PR #18, opened 2020-06-05
Broadcaster creation

aleoli:


internal/advertisement-operator/broadcaster.go
```diff
@@ -68,23 +70,60 @@ func StartBroadcaster(clusterId string, localKubeconfig string, foreignKubeconfi
...
 
-	// get configMaps containing the kubeconfig of the foreign clusters
-	configMaps, err := localClient.CoreV1().ConfigMaps(namespace).List(metav1.ListOptions{})
-	if err != nil {
-		log.Error(err, "Unable to list configMaps")
-		return
-	}
+	if peeringRequestName == "" {
+		// get configMaps containing the kubeconfig of the foreign clusters
+		configMaps, err := localClient.CoreV1().ConfigMaps(namespace).List(metav1.ListOptions{})
```
palexster, line 75: We should use a secret.

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -61,13 +60,43 @@ func (r *ForeignClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
+	if err != nil {
+		r.Log.Error(err, err.Error())
+		return ctrl.Result{}, err
+	}
+	foreignDiscoveryClient, err := v1.NewForConfig(foreignConfig)
+	if err != nil {
+		r.Log.Error(err, err.Error())
+		return ctrl.Result{}, err
+	}
+
```
palexster, line 73: Could you please comment this part of code to make the logic more explicit?

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -94,26 +121,26 @@ func (r *ForeignClusterReconciler) createFederationRequestIfNotExists(clusterID
 
 	localClusterID := r.clusterID.GetClusterID()
 
-	fr, err := foreignDiscoveryClient.FederationRequests().Get(localClusterID, metav1.GetOptions{})
+	pr, err := foreignClient.PeeringRequests().Get(localClusterID, metav1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
```
aleoli, line 126: @palexster I have a problem here. I'm creating peering requests on foreign cluster and I have to create also associated secret. What can I do? Add permission to create secret on foreign cluster or put it inside PeeringRequest and let a mutating webhook to build that secret?

palexster: I would say that the first is ok for the short term.

internal/peering-request-operator/peering-request-controller.go
```diff
@@ -0,0 +1,86 @@
...
+		return ctrl.Result{}, err
+	}
+
+	exists, err := BroadcasterExists(fr, r.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if !exists {
+		r.Log.Info("Deploy Broadcaster")
+		cm, err := r.client.CoreV1().ConfigMaps(r.Namespace).Get("adv-operator-config", metav1.GetOptions{})
```
fraborg, line 68: This hard-coded name can be cumbersome, because it is set in the helm chart. I don't know how we could get the value from it.
However, up to now I think we can keep it this way, you just need to change it since now it is named `liqo-configmap`.

PR #39, opened 2020-06-12
LIQO Installer

palexster:
# Description

This PR introduces a first installer for LIQO.

# How Has This Been Tested?

- [x] With/Without Helm Installed


install.sh
```diff
@@ -1,9 +1,103 @@
...
+    echo "[PRE-INSTALL]: Helm, found version: $HELM_INSTALLED_VERSION"
+    if [ "$HELM_INSTALLED_VERSION" != "$HELM_VERSION" ]; then
+         HELM_INSTALLATION=true
+         echo " [PRE-INSTALL] [HELM] found but having a potential incompatible version... Installing it"
+    fi
+  else
+    echo " [PRE-INSTALL] [HELM]: Helm not found... Installing it"
+    HELM_INSTALLATION=true
+fi
+if [ $HELM_INSTALLATION = true ]; then
```
giorio94, line 52: This looks to me unnecessarily complex. Just go and download your version even if it is already installed. It keeps your script simpler (you also do not need to edit `$PATH`); the probabilities to get your desired version would be anyway very low. Otherwise, insert `helm` as dependency at the beginning (optionally with a minimum version) and fail if it is not found.

For instance, consider that right now you are not using the correct version of helm in case already present, since the download folder is the last in `$PATH`

install.sh
```diff
@@ -1,9 +1,103 @@
...
+    echo " [PRE-INSTALL] [HELM]: Installing Helm $HELM_VERSION"
+    curl --fail -LO $HELM_URL > /dev/null
+    tar zxf helm-v3.2.3-linux-amd64.tar.gz
```
giorio94, line 55: This shouldn't be hardcoded (BTW, it is failing)

Possible alternative:
```bash
curl --fail -L ${HELM_URL} > /dev/null | \
    tar zxf - --directory=${DST} --wildcards '*/helm' --strip 1
```

install.sh
```diff
@@ -1,9 +1,103 @@
...
+fi
+
+if [ ! -d "$FOLDER" ] ; then
+    git clone "$URL" 
+fi
+
+echo "[PRE-INSTALL]: Collecting installation variables. The installer will retrieve installation parameters from your
+ Kubernetes cluster. Fill free to override them, by launching it with those environment variables set in advance."
+if [ -z "$POD_CIDR" ]
```
giorio94, line 64:
```suggestion
if [ -z "$KUBECONFIG" ]
```

install.sh
```diff
@@ -1,9 +1,103 @@
...
+then
+  echo "[PRE-INSTALL]: Kubeconfig variable is set to: $KUBECONFIG"
+else
+  echo "[PRE-INSTALL]: Kubeconfig variable is set to: ~/.kube/config"
+fi
+if [ -z "$POD_CIDR" ]
+then
+    POD_CIDR=$(kubectl cluster-info dump | grep -m 1 cluster-cidr | sed 's/--cluster-cidr=//' | sed 's/[,"]//g' | sed 's/ //g')
```
giorio94, line 72: 1. Maybe here you could directly filter the output in advance to avoid downloading too much data (jsonpath or gotemplate?)
2. You could replace the multiple `sed` with a single `grep` with regex and capture group

install.sh
```diff
@@ -1,9 +1,103 @@
...
+fi
+echo "[PRE-INSTALL]: GATEWAY_PRIVATE_IP is set to: $GATEWAY_PRIVATE_IP"
+
+if [ -z "$GATEWAY_IP" ]
+then
+    GATEWAY_IP=$(kubectl get no -l "liqonet.liqo.io/gateway=true" -o jsonpath='{.items[0].status.addresses[0].address}')
+fi
+echo "[PRE-INSTALL]: GATEWAY_IP is set to: $GATEWAY_IP"
+
+echo "COMMAND: helm install liqo -n liqo ./ --set podCIDR=$POD_CIDR --set serviceCIDR=$SERVICE_CIDR --set gatewayPrivateIP=$GATEWAY_PRIVATE_IP --set gatewayIP=$GATEWAY_IP"
```
giorio94, line 94: This should be a variable, otherwise soon it will be different from the actually executed command

install.sh
```diff
@@ -1,9 +1,103 @@
...
+read -p "Are you sure to run the previous command? (y/N)" -n 1 -r
```
giorio94, line 95:
```suggestion
read -p "Are you sure to run the previous command? (y/N) " -n 1 -r
```

install.sh
```diff
@@ -1,9 +1,103 @@
...
+echo    # (optional) move to a new line
```
giorio94, line 96:
```suggestion
echo
```

install.sh
```diff
@@ -1,9 +1,103 @@
...
+    if [ "$HELM_INSTALLED_VERSION" != "$HELM_VERSION" ]; then
+         HELM_INSTALLATION=true
+         echo " [PRE-INSTALL] [HELM] found but having a potential incompatible version... Installing it"
+    fi
+  else
+    echo " [PRE-INSTALL] [HELM]: Helm not found... Installing it"
+    HELM_INSTALLATION=true
+fi
+if [ $HELM_INSTALLATION = true ]; then
+    echo " [PRE-INSTALL] [HELM]: Installing Helm $HELM_VERSION"
```
giorio94, line 53: As before, I would avoid *Installing*. *Downloading* looks better to me

install.sh
```diff
@@ -1,9 +1,103 @@
...
+
+if [ -z "$SERVICE_CIDR" ]
+then
+    SERVICE_CIDR=$(kubectl cluster-info dump | grep -m 1 service-cluster-ip-range | sed 's/--service-cluster-ip-range=//' | sed 's/[,"]//g' | sed 's/ //g')
+fi
+echo "[PRE-INSTALL]: SERVICE_CIDR is set to: $SERVICE_CIDR"
+
+if [ -z "$GATEWAY_PRIVATE_IP" ]
+then
+    GATEWAY_PRIVATE_IP=192.168.1.1
```
giorio94, line 84: Is this supposed to be hard-coded? If so, I would define it at the beginning of the script

install.sh
```diff
@@ -1,9 +1,103 @@
...
+    HELM_INSTALLATION=true
+fi
+if [ $HELM_INSTALLATION = true ]; then
+    echo " [PRE-INSTALL] [HELM]: Installing Helm $HELM_VERSION"
+    curl --fail -LO $HELM_URL > /dev/null
+    tar zxf helm-v3.2.3-linux-amd64.tar.gz
+fi
+
+if [ ! -d "$FOLDER" ] ; then
+    git clone "$URL"
```
giorio94, line 59: Here, you should consider limiting the depth of the history, to avoid downloading unnecessary data when the repo will grow

install.sh
```diff
@@ -1,9 +1,103 @@
...
+done
+
+set -e
+
+mkdir -p /tmp/liqo
+cd /tmp/liqo || exit 1
+mkdir -p /tmp/liqo
+cd /tmp/liqo || exit 1
+mkdir -p bin/
+cp linux-amd64/helm bin/
```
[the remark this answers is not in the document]
palexster: Fixed, sorry

install.sh
palexster: This will move to a curl release when a first release will be published.

PR #40, opened 2020-06-12
Advertisement modification, broadcaster new logic

fraborg:
# Description

This PR modifies the advertisement logic

- [x] the Broadcaster creates 2 resources on remote cluster: a secret that allows the _foreign cluster_ to create resources on home cluster (at the moment everything) and the `Advertisement` itself
- [x] the `Advertisement` now has an object reference to the previous secret
- [x] when the foreign cluster receives the `Advertisement`, the Controller will be triggered and use this secret to give to the VK the permission it needs

Minor modifications
- slightly modifies the discovery module in order to allow the broadcaster to create `Secrets` on the remote cluster
- removed `ForeignNetowork` field from Adv which is not used
- modifed client creation function to read the kubeconfig from a secret (this affected files in tray-agent, for this reason there is the **ui** label)

# How it has been tested
- [x] manual creation of `PeeringRequest` and `Secret` required for the Broadcaster to run


internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -235,8 +235,8 @@ func (r *ForeignClusterReconciler) createClusterRoleIfNotExists(clusterID string
 				// TODO: set correct access to create advertisements
 				{
 					Verbs:     []string{"get", "list", "create", "delete", "watch"},
-					APIGroups: []string{"protocol.liqo.io"},
-					Resources: []string{"advertisements"},
+					APIGroups: []string{"protocol.liqo.io", ""},
```
fraborg, line 238: @aleoli I added this to allow the broadcaster to create also Secrets: is it correct?

aleoli: Yes

cmd/advertisement-broadcaster/main.go
```diff
@@ -17,12 +19,18 @@ func main() {
...
-	flag.StringVar(&peeringRequestName, "peering-request", "", "Name of PeeringRequest CR where get configurations")
+	flag.StringVar(&peeringRequestName, "peering-request", "", "Name of PeeringRequest CR containing configurations")
 	flag.Parse()
 
+	log := ctrl.Log.WithName("setup")
 	ctrl.SetLogger(zap.New(func(o *zap.Options) {
 		o.Development = true
 	}))
 
+	if peeringRequestName == "" {
```
palexster, line 30: Can we impose the --peering-request as mandatory instead of adding this control?

internal/advertisement-operator/broadcaster.go
```diff
@@ -30,120 +29,82 @@ var (
 	log logr.Logger
 )
 
+type AdvertisementBroadcaster struct {
+	//TODO: transform these functions in methods to avoid all that parameters
```
palexster, line 33: I think we can wait for this todo addressed in this PR. It should not be complex to achieve and I think it would make the code a bit simpler to follow...

internal/advertisement-operator/broadcaster.go
```diff
@@ -188,19 +149,47 @@ func GenerateAdvertisement(wg *sync.WaitGroup, localClient *kubernetes.Clientset
...
 		} else {
-			log.Info("correctly created advertisement on remote cluster " + foreignClusterId)
-			// the watcher must be launched only once
-			go once.Do(func() {
-				scheme := runtime.NewScheme()
-				_ = clientgoscheme.AddToScheme(scheme)
-				_ = protocolv1.AddToScheme(scheme)
-				WatchAdvertisement(localCRDClient, scheme, foreignKubeconfigPath, cm, clusterId, foreignClusterId)
-			})
+			adv := CreateAdvertisement(localClusterId, gatewayIP, gatewayPrivateIP, physicalNodes, virtualNodes, availability, images, limits, kubeconfigSecret)
```
palexster, line 174: some short commentaries may help here.

internal/advertisement-operator/broadcaster.go
fraborg: will be done in the next PR

cmd/advertisement-broadcaster/main.go
fraborg: According to [this](https://stackoverflow.com/questions/31786215/can-command-line-flags-in-go-be-set-to-mandatory#:~:text=The%20flag%20package%20does%20not,halt%20with%20an%20error%20message) it's not possible

PR #41, opened 2020-06-15
Introduce CSR Auto-approver for virtual-kubelet

palexster:
# Description

This commit adds a CSR-approve routine in Advertisement Operator
main function in order to auto-approve CSR coming from the virtual-kubelet.

(to-do): So far, the advertisement operator simply checks if a label is put on the
CSR, in future we will add some more accurate checks.

# How Has This Been Tested?

- [x] Test auto-approve of a virtual-kubelet issued CSR
- [x] Test ignore of other CSRs

### DIFF
--- a/pkg/csrApprover/csr.go
+++ b/pkg/csrApprover/csr.go
@@ -1,0 +1,57 @@
+package csrApprover
+
+import (
+	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8s "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+)
+
+func approveCSR(clientSet k8s.Interface, csr *certificatesv1beta1.CertificateSigningRequest) error {
+	// certificate already added to CSR
+	if csr.Status.Certificate != nil {
+		return nil
+	}
+	// Check if the certificate is already approved but the certificate is still not available
+	for _, b := range csr.Status.Conditions {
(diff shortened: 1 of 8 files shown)


pkg/csrApprover/csr.go
```diff
@@ -0,0 +1,56 @@
...
+	go func() {
+		for event := range watch.ResultChan() {
+			csr, ok := event.Object.(*certificatesv1beta1.CertificateSigningRequest)
+			if !ok {
+				klog.Error("Unable to cast object from watch operation")
+			}
+			// Check if the certificare is already approved
+			err := approveCSR(clientset, csr)
+			if err != nil {
+				klog.Error(err.Error())
```
mLavacca, line 51: You don't necessarily need to use `err.Error()`, that returns the string description of the error. `klog.Error` accepts any kind of interfaces as args (you can use `err` instead of `err.Error()`)

pkg/csrApprover/csr.go
```diff
@@ -0,0 +1,56 @@
...
+
+func WatchCSR(clientset k8s.Interface, label string) {
+	watch, err := clientset.CertificatesV1beta1().CertificateSigningRequests().Watch(metav1.ListOptions{
+		LabelSelector: label,
+	})
+	if err != nil {
+		klog.Error(err.Error())
+	}
+	go func() {
+		for event := range watch.ResultChan() {
```
mLavacca, line 43: Once you finished using the watcher chan, you should close it with `watch.Stop()`

mLavacca:
@dependabot merge

PR #44, opened 2020-06-15
Broadcaster refactoring

fraborg:
# Description
This PR doesn't add functionalities, it's mainly code refactoring

- [x] modify the broadcaster to use methods and avoid many function parameters

Other changes
- modify package functions to use pointers

# How it has been tested
- [x] manual creation of resources given by discovery modules, launching the broadcaster on a cluster and the controller on the other one. All the resources are created correctly

### DIFF
--- a/internal/advertisement-operator/broadcaster.go
+++ b/internal/advertisement-operator/broadcaster.go
@@ -2,8 +2,10 @@ package advertisement_operator
 
 import (
 	"context"
+	"errors"
 	"github.com/liqoTech/liqo/internal/discovery/clients"
 	"github.com/liqoTech/liqo/internal/discovery/kubeconfig"
+	discoveryv1 "github.com/liqoTech/liqo/pkg/discovery/v1"
 	"k8s.io/apimachinery/pkg/fields"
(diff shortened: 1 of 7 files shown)


internal/advertisement-operator/broadcaster.go
```diff
@@ -119,108 +108,135 @@ func GenerateAdvertisement(localClient *kubernetes.Clientset, localCRDClient cli
 		log.Info("created client to remote cluster " + foreignClusterId)
 	}
 
+	//TODO: get the name of the service account
+	sa, err := localClient.CoreV1().ServiceAccounts(pr.Spec.Namespace).Get("broadcaster", metav1.GetOptions{})
```
palexster, line 112: Can we get this by a flag putting broadcaster as default value?

fraborg:
> Do these modifications imply some drawbacks in the virtual-kubelet deployment? The foreign kubeconfig is still taken as before from the vk point of view?

No, the vk is launched as before. It is mainly refactoring of the broadcaster to use methods instead of simple functions

PR #45, opened 2020-06-16
Installer Improvement

palexster:
# Description

This PR improves the installer and the installation documentation by adding:

- [x] Fixing advertisement operator deployment
- [x] Fix Discovery init containers image reference
- [x] Installer Auto-approve for mutating webhook
- [x] Pre-node Labeling check

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,7 +7,7 @@ function cleanup()
 }
 
 set_variable_from_command() {    
-    if [[ ($# -ne 2) ]];
+    if [[ ($# -ne 3) ]];
     then
       echo "Internal Error - Wrong number of parameters"
       exit 1
@@ -16,7 +16,9 @@ set_variable_from_command() {
     DEFAULT_COMMAND=$2
     if [ -z "${!VAR_NAME}" ]
     then
-        result=$(bash -c "${!DEFAULT_COMMAND}") || return $?
+        result=$(bash -c "${!DEFAULT_COMMAND}") || {
+          ret=$?; echo "$3 - Code: $ret"; return $ret; 
+        }
         declare -gx "$VAR_NAME"=$result
     fi
     echo "[PRE-INSTALL]: $VAR_NAME is set to: ${!VAR_NAME}"
@@ -32,6 +34,22 @@ print_help()
    echo "   SERVICE_CIDR: the POD CIDR of your cluster (e.g.; 10.96.0.0/12) . The script will try to detect it, but you can override thisthis by having this variable already set"
    echo "   GATEWAY_PRIVATE_IP: the IP used by the cluster inside the cluster-to-cluster interconnection (e.g.; 192.168.1.1)"
    echo "   GATEWAY_IP: the public IP that will be used by LIQO to establish the interconnection with other clusters"
+}
+
+function wait_and_approve_csr(){
+   max_retry=5
+   retry=0
+   while [ "$retry" -lt "$max_retry" ]; do
+     echo "[INSTALL]: Approving Admission/Mutating Webhook CSRs, $1"
+     if kubectl get csr $1 > /dev/null; then
+       kubectl certificate approve $1 || exit 1
+       break
+     fi
+     echo "[INSTALL]: CSR not found... Retrying..."
+     retry=$((retry+1))
+     sleep 5
+   done
+   return 0
 }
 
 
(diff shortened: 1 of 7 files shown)


install.sh
```diff
@@ -34,6 +45,22 @@ print_help()
    echo "   GATEWAY_IP: the public IP that will be used by LIQO to establish the interconnection with other clusters"
 }
 
+function wait_and_approve_csr(){
+   max_retry=3
+   retry=0
+   while [ "$retry" -lt "$max_retry" ]; do
+     set +e
```
giorio94, line 52: Is this necessary? It seems to me that only `kubectl certificate approve` may fail and you terminate in that case

install.sh
```diff
@@ -89,19 +116,22 @@ else
...
 GATEWAY_IP_COMMAND='kubectl get no -l "liqonet.liqo.io/gateway=true" -o jsonpath="{.items[0].status.addresses[0].address}"'
-set_variable_from_command GATEWAY_IP GATEWAY_IP_COMMAND
+try_set_variable GATEWAY_IP GATEWAY_IP_COMMAND "[ERROR]: You have to assign to a node of the cluster the role of gateway. Label it with 'liqonet.liqo.io/gateway=true': kubectl label yournode liqonet.liqo.io/gateway=true"
 GATEWAY_PRIVATE_IP_COMMAND="echo $DEFAULT_GATEWAY_PRIVATE_IP"
-set_variable_from_command GATEWAY_PRIVATE_IP GATEWAY_PRIVATE_IP_COMMAND
+try_set_variable GATEWAY_PRIVATE_IP GATEWAY_PRIVATE_IP_COMMAND "[ERROR]: Unable to set Gateway Private IP"
+
+TO_EXECUTE="$TMPDIR/bin/helm install liqo -n liqo $TMPDIR/liqo/deployments/liqo_chart --set podCIDR=$POD_CIDR --set serviceCIDR=$SERVICE_CIDR --set gatewayPrivateIP=$GATEWAY_PRIVATE_IP --set gatewayIP=$GATEWAY_IP"
+#Wait for the installation to complete
+helm dependency update $TMPDIR/liqo/deployments/liqo_chart
```
giorio94, line 129: This should point to the downloaded `helm`

install.sh
```diff
@@ -89,19 +116,22 @@ else
...
-set_variable_from_command SERVICE_CIDR SERVICE_CIDR_COMMAND 
+try_set_variable SERVICE_CIDR SERVICE_CIDR_COMMAND "[ERROR]: Unable to find Service CIDR"
 GATEWAY_IP_COMMAND='kubectl get no -l "liqonet.liqo.io/gateway=true" -o jsonpath="{.items[0].status.addresses[0].address}"'
-set_variable_from_command GATEWAY_IP GATEWAY_IP_COMMAND
+try_set_variable GATEWAY_IP GATEWAY_IP_COMMAND "[ERROR]: You have to assign to a node of the cluster the role of gateway. Label it with 'liqonet.liqo.io/gateway=true': kubectl label yournode liqonet.liqo.io/gateway=true"
 GATEWAY_PRIVATE_IP_COMMAND="echo $DEFAULT_GATEWAY_PRIVATE_IP"
-set_variable_from_command GATEWAY_PRIVATE_IP GATEWAY_PRIVATE_IP_COMMAND
+try_set_variable GATEWAY_PRIVATE_IP GATEWAY_PRIVATE_IP_COMMAND "[ERROR]: Unable to set Gateway Private IP"
+
+TO_EXECUTE="$TMPDIR/bin/helm install liqo -n liqo $TMPDIR/liqo/deployments/liqo_chart --set podCIDR=$POD_CIDR --set serviceCIDR=$SERVICE_CIDR --set gatewayPrivateIP=$GATEWAY_PRIVATE_IP --set gatewayIP=$GATEWAY_IP"
```
giorio94, line 127: I would move the command directly below, since the `TO_EXECUTE` variable is no longer of any use

install.sh
palexster: The problem is kubectl get csr which may fail and should not trigger an exit.

install.sh
```diff
@@ -89,19 +110,21 @@ else
...
-set_variable_from_command GATEWAY_IP GATEWAY_IP_COMMAND
+set_variable_from_command GATEWAY_IP GATEWAY_IP_COMMAND "[ERROR]: You have to assign to a node of the cluster the role of gateway. Label it with 'liqonet.liqo.io/gateway=true': kubectl label yournode liqonet.liqo.io/gateway=true"
 GATEWAY_PRIVATE_IP_COMMAND="echo $DEFAULT_GATEWAY_PRIVATE_IP"
-set_variable_from_command GATEWAY_PRIVATE_IP GATEWAY_PRIVATE_IP_COMMAND
+set_variable_from_command GATEWAY_PRIVATE_IP GATEWAY_PRIVATE_IP_COMMAND "[ERROR]: Unable to set Gateway Private IP"
+
+#Wait for the installation to complete
+$TMPDIR/bin/helm dependency update $TMPDIR/liqo/deployments/liqo_chart
+$TMPDIR/bin/helm install liqo -n liqo $TMPDIR/liqo/deployments/liqo_chart --set podCIDR=$POD_CIDR --set serviceCIDR=$SERVICE_CIDR --set gatewayPrivateIP=$GATEWAY_PRIVATE_IP --set gatewayIP=$GATEWAY_IP
+sleep 30
```
giorio94, line 124: Since here you are sleeping for quite a lot, I would say something to the user. Otherwise, he may think the script hanged and terminate it. The same applies also to the `wait_and_approve_csr` function.

install.sh
```diff
@@ -16,7 +16,14 @@ set_variable_from_command() {
     DEFAULT_COMMAND=$2
     if [ -z "${!VAR_NAME}" ]
     then
-        result=$(bash -c "${!DEFAULT_COMMAND}") || return $?
+        set +e
```
giorio94, line 19: A possible alternative approach (if you want to avoid set +/-e) may be:

```
result=$(bash -c "${!DEFAULT_COMMAND}") || {
    ret=$?; echo "..."; return $ret; 
}
```

palexster: nice!

PR #47, opened 2020-06-19
Discovery Improvement

aleoli:
# Description

This PR makes the creation of `ForeignCluster` easier. At cration time has to be provided only the `apiUrl` and the `namespace` where liqo is deployed

This feature will enable:
* easier manual creation
* easier discovery process in LAN
* discovery in WAN (in next PR)

They are different ways to provide the two required params

# How Has This Been Tested?

LAN discovery has been tested with two k3s clusters, note clusters has to be launched with `--anonymous-auth=true` flag


deployments/liqo_chart/subcharts/discovery_chart/templates/unauth-role.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: anonymous-role-peering
+rules:
+  - apiGroups: ["discovery.liqo.io"]
+    resources: ["peeringrequests"]
+    verbs: ["get", "create", "delete", "update"]
```
palexster, line 8: I am a bit scaried about all those privileges. Can't we use just create?

aleoli: At the moment not, because, we need get to check if peering request exsists, update to set the reference to the created secret (with kubeconfig credentials) and delete to delete it when user wants to disjoin clusters

PR #54, opened 2020-06-19
ClusterConfig api first implementation

fraborg:
# Description

This PR introduces the `ClusterConfig` API, through which we can configure some parameters of our cluster, such as how many resources we want to share, how many `Advertisements` we want to accept etc...

- [x] First `ClusterConfig` api definition
```
ClusterConfigSpec{
  ResourceSharingPercentage  int32 
  MaxAcceptableAdvertisement int32
  AutoAccept                 bool 
}
```

- [x] Apply the configuration in Advertisement controller
If the `AutoAccept` flag is set, the controller accepts `Advertisements` until the maximum set in the configuration is reached.
When the configuration changes
- if the maximum has increased a new check is launched over the refused Adv, allowing to accept some of them
- if the maximum has decreased and (with the new maximum) there are too many accepted `Advertisements`, some of them are deleted

- [x] Apply the configuration in Advertisement broadcaster
The amount of resources that the cluster will share is read from the `ClusterConfig`. The same percentage is applied to all created `Advertisements`

## Minor changes
- removed unused field from `Advertisement` status
- slightly modified Adv controller to use methods instead of functions
- minor add in labeler action

# How Has This Been Tested?

- [x] Controller test
- if no `ClusterConfig` is found in the cluster, all Advertisements are refused
- created more `Advertisements` than the `MaxAcceptableAdvertisement` field: the new ones are refused
- updated `ClusterConfig` increasing the `MaxAcceptableAdvertisement` field: if there are some previously refused Adv they are accepted
- updated `ClusterConfig` decreasing the `MaxAcceptableAdvertisement` field: some Adv are deleted until the new maximum is satisfied

- [x] Broadcaster test

- [x] Unit test
- added some unit tests for `CheckAdvertisement()` and `ManageConfigUpdate()` functions, to check the logic that sets Advertisement status to APPROVED or REFUSED

# Known issues
- sometimes at bootstrap the ClusterConfig watcher fails to list Advertisements since the cache is not started yet
- when an Advertisement is manually deleted, the number of accepted Advertisement is not updated
### Proposed solution
Use the standard k8s watcher over `Advertisements` instead of kubebuilder client


internal/advertisement-operator/controller.go
```diff
@@ -93,21 +113,30 @@ func (r *AdvertisementReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // check if the advertisement is interesting and set its status accordingly
-func checkAdvertisement(r *AdvertisementReconciler, ctx context.Context, log logr.Logger,
-	adv *protocolv1.Advertisement) {
-
-	//TODO: implement logic
-	adv.Status.AdvertisementStatus = "ACCEPTED"
-	metav1.SetMetaDataAnnotation(&adv.ObjectMeta, "advertisementStatus", "accepted")
+func (r *AdvertisementReconciler) checkAdvertisement(ctx context.Context, log logr.Logger, adv *protocolv1.Advertisement) {
```
palexster, line 116: Due to the importance of this logic, I think that  a dedicated unit Test is needed.

api/cluster-config/v1/clusterconfig_types.go
```diff
@@ -0,0 +1,57 @@
...
+*/
+
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ClusterConfigSpec defines the desired state of ClusterConfig
+type ClusterConfigSpec struct {
```
palexster, line 23: Question: is it possible to set an maxLimit in this struct or in the crd to say that the max possible value is 100%?

test/unit/advertisement-operator/controller_test.go
```diff
@@ -0,0 +1,126 @@
...
+	// ACCEPT TEST
+	// given a configuration with max 10 Advertisements, create 10 Advertisements and check that they are all accepted
+	for i := 0; i < 10; i++ {
+		adv := createAdvertisement()
+		r.CheckAdvertisement(&adv)
+		assert.Equal(t, "ACCEPTED", adv.Status.AdvertisementStatus)
+	}
+	// check that the Adv counter has been incremented
+	assert.Equal(t, int32(10), r.AcceptedAdvNum)
+
```
palexster, line 53: I would suggest to have one single testCase per Test in order to: (1) foster parallelism and (2) early detection of which testCase is really broken. In this case, one for the ShouldAccept one and one for ShouldRefuse

test/unit/advertisement-operator/controller_test.go
```diff
@@ -0,0 +1,126 @@
...
+	// FALSE TEST
+	// apply again the same configuration
+	// we enter in the false branch of ManageConfigUpdate but nothing should change
+	err, flag = r.ManageConfigUpdate(&config, &advList)
+	assert.Nil(t, err)
+	assert.False(t, flag)
+	assert.Equal(t, config.Spec, r.ClusterConfig)
+	assert.Equal(t, int32(advCount), r.AcceptedAdvNum)
+
+	//TODO: FALSE TEST with config.MaxAcceptableAdvertisement < r.AcceptedAdvNum
```
palexster, line 124: https://book.kubebuilder.io/reference/testing/envtest.html this may be helpful

palexster: A simple example is the following: https://github.com/kubernetes-sigs/kubebuilder/blob/e5fda8fd51e177376e4db6c39028c2b73a7ce66e/docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go

api/cluster-config/v1/clusterconfig_types.go
palexster: https://book.kubebuilder.io/reference/markers/crd-validation.html 
```kubebuilder:validation:Maximum
kubebuilder:validation:Minimum
```

test/unit/advertisement-operator/controller_test.go
fraborg: I'll check it, thanks!

PR #56, opened 2020-06-22
Add new linter and CI checks 

palexster:
# Description

This PRs hardens the amount of linting checks performed on Liqo Code and add a CI check runnable for forked PRs.

It adds:
* [gosec](https://github.com/securego/gosec)
* [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports?tab=doc)
* [prealloc](https://github.com/alexkohler/prealloc)
* [godot](https://github.com/tetafro/godot)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@
 Liqo is a framework to enable dynamic sharing across Kubernetes Clusters. You can run your pods on a remote cluster
 seamlessly, without any modification (Kubernetes or your application). 
 
+#What is Liqo?
+
 Differently from the [Kubernetes Federation](https://github.com/kubernetes-sigs/kubefed) and the
 [Admiralty](https://admiralty.io/) project, Liqo is designed to handle dynamic, temporary
 resource sharing across multi-owner clusters and targets every type of computing resources (e.g. Raspberry PI, 
@@ -57,7 +59,7 @@ Liqo Installer should be capable to look for the cluster parameters required.
 curl https://raw.githubusercontent.com/LiqoTech/liqo/master/install.sh | bash
 ```
 
-#### [K3s](k3s.io)
+#### [K3s](https://k3s.io)
 
 K3s is a minimal Kubernetes cluster which is pretty small and easy to set up. However, it does not store its configuration in the
 way that traditional installers (e.g.; Kubeadm) do. Therefore, it is required to know the configuration you entered for your cluster.
--- a/README.md
+++ b/README.md
@@ -1,17 +1,21 @@
+<p align="center">
+<img alt="Liqo Logo" src="https://doc.liqo.io/images/logo-liqo-blue.svg" />
+</p>
+
+# Liqo
+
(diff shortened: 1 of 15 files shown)


pkg/liqonet/overlay_network.go
```diff
@@ -103,7 +103,7 @@ func Enable_rp_filter() error {
 	for index := range ifaces_list {
 		if ifaces_list[index].Type() == "vxlan" {
 			// Enable loose mode reverse path filtering on the vxlan interface.
-			err = ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+ifaces_list[index].Attrs().Name+"/rp_filter", []byte("2"), 0644)
+			err = ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+ifaces_list[index].Attrs().Name+"/rp_filter", []byte("2"), 0600)
```
palexster, line 106: @alacuku gosec suggested this permission change. Does it sound ok to you?

alacuku: well the default permission of that file are `0644 (-rw-r--r--)` with `user root` and `group root`. what gosec is suggesting is to remove the read permissions of all the users belonging to the` root group` and also to the `rest of the world`. Being it a system file i would maintain the default permission.

AbakusW: 0644 seems to be in fact [the default permission for the root user](https://www.cyberciti.biz/tips/understanding-linux-unix-umask-value-usage.html). Nevertheless also *00 appears to be a common configuration and maybe restricting the privileges a little more as gosec suggested could be a good idea. Do you think it could create troubles?

README.md
```diff
@@ -6,6 +6,8 @@
 Liqo is a framework to enable dynamic sharing across Kubernetes Clusters. You can run your pods on a remote cluster
 seamlessly, without any modification (Kubernetes or your application). 
 
+#What is Liqo?
```
aleoli, line 9:
```suggestion
# What is Liqo?
```

internal/discovery/foreign.go
```diff
@@ -79,7 +79,8 @@ func (discovery *DiscoveryCtrl) UpdateTtl(txts []*TxtData) error {
 		klog.Error(err, err.Error())
 		return err
 	}
-	for _, fc := range fcs.Items {
+	for i := range fcs.Items {
+		fc := fcs.Items[i]
```
aleoli, line 83: Are we sure that is better this kind of for loops?

palexster: No, but gosec told me :)
https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

pkg/liqonet/overlay_network.go
palexster: We discussed with @alacuku, and we agreed that this should not break. I'll test it on a cluster before merging.

PR #64, opened 2020-06-27
Periodic reconciliation of Liqo objects

palexster:
# Description

So far, Liqo controllers are triggered only when a resource is modified.
In this commit, we add a periodic reconciliation to ensure the desired state of Liqo objects.

## Minor changes
- added const variables for Advertisement status

### DIFF
--- a/internal/advertisement-operator/controller.go
+++ b/internal/advertisement-operator/controller.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/klog"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
 )
 
 // AdvertisementReconciler reconciles a Advertisement object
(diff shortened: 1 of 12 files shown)


internal/advertisement-operator/controller.go
```diff
@@ -93,10 +100,10 @@ func (r *AdvertisementReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 			return ctrl.Result{}, err
 		}
 		klog.V(3).Info("Correct creation of virtual kubelet deployment for cluster " + adv.Spec.ClusterId)
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: r.RetryTimeout}, nil
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{RequeueAfter: r.RetryTimeout}, nil
```
fraborg, line 106: when we arrive here, it means everything went well in reconciliation: do we need/want to reque nonetheless?

palexster: I think you are right. This is not necessary if we correctly set up the owner reference on created objects.

aleoli:
Is there a reason why we use sometimes 30 seconds and sometimes 1 minute as default requeue time?

fraborg:
> Is there a reason why we use sometimes 30 seconds and sometimes 1 minute as default requeue time?

I don't think there's a particular reason, if I'm right only the netowrk modules use 30 seconds. I think we can leave that since it's better if networking reconverges rapidly in case of errors, whereas the other operators should not need this.

PR #73, opened 2020-06-30
Cluster Config Improvements and Discovery Config

aleoli:
# Description

A config watcher has been created in an own package. Each component that need it can add an handler function to be notified on new events happened on ClusterConfig

There can be deferent event types:

- Add/Update -> call handler function
- Delete -> try to recreate last config version

Every discovery configuration is dynamic now, and if changed triggers component reload. If you change some setting in mDNS server it is restarted with new configurations


internal/advertisement-operator/config-watcher.go
```diff
@@ -4,91 +4,65 @@ import (
...
-		return err
-	}
-
-	go func() {
-		for event := range watcher.ResultChan() {
-			configuration, ok := event.Object.(*policyv1.ClusterConfig)
-			if !ok {
-				continue
-			}
+func (b *AdvertisementBroadcaster) WatchConfiguration() {
```
aleoli, line 13: @fraborg Do you need to specify kubeconfigPath? Or trying InCluster and, if failed, on ~/.kube/config is enough?

fraborg: I have reused `NewKubeconfig()` function (https://github.com/LiqoTech/liqo/blob/master/pkg/crdClient/v1alpha1/client.go#L29) which receives the kubeconfigPath.
I think it's useful because with that you can pass whichever kubeconfig you want, it not necessarily needs to be located in `.kube/config`

aleoli: ok, so I add this parameter, if not set it uses default path

cmd/advertisement-operator/main.go
```diff
@@ -139,9 +139,7 @@ func main() {
 	}
 	// +kubebuilder:scaffold:builder
 
-	if err = r.WatchConfiguration(localKubeconfig); err != nil {
-		klog.Error(err)
-	}
+	r.WatchConfiguration(localKubeconfig)
```
palexster, line 142: Why is this necessary?

config/cluster-config/crd/bases/policy.liqo.io_clusterconfigs.yaml
```diff
@@ -36,19 +36,58 @@ spec:
...
+                domain:
+                  type: string
+                enableAdvertisement:
+                  type: boolean
+                enableDiscovery:
+                  type: boolean
+                name:
+                  type: string
+                port:
+                  type: integer
```
palexster, line 68: max, min values possible

deployments/liqo_chart/crds/policy.liqo.io_clusterconfigs.yaml
```diff
@@ -36,29 +36,68 @@ spec:
...
+                  type: boolean
+                domain:
+                  type: string
+                enableAdvertisement:
+                  type: boolean
+                enableDiscovery:
+                  type: boolean
+                name:
+                  type: string
+                port:
```
palexster, line 67: as before, the domain can be constrained to 1-65355

internal/advertisement-operator/config-watcher.go
```diff
@@ -4,91 +4,65 @@ import (
...
 	protocolv1 "github.com/liqoTech/liqo/api/advertisement-operator/v1"
 	policyv1 "github.com/liqoTech/liqo/api/cluster-config/v1"
+	"github.com/liqoTech/liqo/pkg/clusterConfig"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/klog"
 )
 
-func (b *AdvertisementBroadcaster) WatchConfiguration(kubeconfigPath string) error {
-	configClient, err := policyv1.CreateClusterConfigClient(kubeconfigPath)
```
palexster, line 13: Why is it necessary to change this and do not exit if watch is broken?

aleoli: The change allow us to have one watcher function and pass it only a function to handle our specific need in the component

For the exit I can add a logging and a os exit instead of return

palexster: Ok, got it. I think that we can consider a standard policy is: if it is unable to get the configuration, the process will fail.

aleoli: ok, and I think it's implemented [here](https://github.com/LiqoTech/liqo/blob/aol/cluster-config-improvement/pkg/clusterConfig/watchConfig.go#L33)

pkg/clusterConfig/watchConfig.go
```diff
@@ -0,0 +1,61 @@
...
+		klog.Error(err, err.Error())
+		os.Exit(1)
+	}
+
+	for event := range watcher.ResultChan() {
+		handler(event)
+	}
+}
+
+func NewConfig(kubeconfigPath string) (*rest.Config, error) {
```
mLavacca, line 42: Why did you replicate this function from the client library? We have a shared library for this purpose

internal/advertisement-operator/config-watcher.go
```diff
@@ -4,91 +4,65 @@ import (
...
-			case watch.Added, watch.Modified:
-				b.ClusterConfig.ResourceSharingPercentage = configuration.Spec.ResourceSharingPercentage
-			case watch.Deleted:
-				// TODO: set default config?
-			}
+		switch event.Type {
+		case watch.Added, watch.Modified:
+			b.ClusterConfig.ResourceSharingPercentage = configuration.Spec.AdvertisementConfig.ResourceSharingPercentage
+		case watch.Deleted:
+			// TODO: set default config?
```
mLavacca, line 24: I think we should take this decision before merging it

internal/advertisement-operator/config-watcher.go
```diff
@@ -4,91 +4,65 @@ import (
...
+			}
+			if updateFlag {
+				for _, adv := range advList.Items {
+					r.UpdateAdvertisement(&adv)
 				}
-			case watch.Deleted:
-				// TODO: set default config?
 			}
+		case watch.Deleted:
+			// TODO: set default config?
```
mLavacca, line 55: as above

api/cluster-config/v1/clusterconfig_types.go
```diff
@@ -57,4 +84,12 @@ type ClusterConfigList struct {
 
 func init() {
 	SchemeBuilder.Register(&ClusterConfig{}, &ClusterConfigList{})
+
+	if err := AddToScheme(scheme.Scheme); err != nil {
+		panic(err)
+	}
+	v1alpha1.AddToRegistry("clusterconfigs", &ClusterConfig{}, &ClusterConfigList{}, nil, schema.GroupResource{
```
fraborg, line 91: This is already done [here](https://github.com/LiqoTech/liqo/blob/master/api/cluster-config/v1/clusterConfigClient.go#L27). Do we need to replicate?

aleoli: I think that is better to do in init function because is automatically executed, can I delete the other one?

fraborg: Fine for me

palexster:
Just a small general comment: Could you please squash some of your commits and have a more compelling description of the content?

PR #76, opened 2020-07-01
Advertisement management test

fraborg:
# Description

This PR adds some tests for functions related to Advertisement management

- [x] `pkg/advertisement-operator` folder
At the moment we test ownerReference creation functions, Virtual-Kubelet deployment creation and Pod CreateOrUpdate(). The other resources (Deployment, Secret etc...) are not tested because it would require a lot of boilerplate code which maybe can be avoided.

In the next PR some tests for `internal/advertisement-operator` folder will be added, they have not been implemented yet because I would like to modify remote-watcher logic before.

### DIFF
--- a/test/unit/advertisement-operator/creation_test.go
+++ b/test/unit/advertisement-operator/creation_test.go
@@ -1,0 +1,178 @@
+package advertisement_operator
+
+import (
+	"context"
+	v1 "github.com/liqoTech/liqo/api/advertisement-operator/v1"
+	advertisement_operator "github.com/liqoTech/liqo/pkg/advertisement-operator"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


test/unit/advertisement-operator/creation_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// test adv creation
+	err := advertisement_operator.CreateOrUpdate(c, context.Background(), adv)
+	assert.Nil(t, err)
+
+	// creation requires some time to be effective
+	// wait 1 second to assure the resource is retrieved by Get
+	time.Sleep(1 * time.Second)
+
+	// test adv update
```
palexster, line 180: Can you please wipe the following lines if they are commented, in order to keep the code clean.

PR #77, opened 2020-07-01
Vk ep reflection filtering

mLavacca:
This PR introduces a check that verifies whether the ep event has to be replicated remotely. This check is based on checking if the addresses of the foreign ep are aligned with the local one. In addition to that, The e2e ep reflection test has been slightly improved in its robustness.

### DIFF
--- a/internal/kubernetes/endpoints.go
+++ b/internal/kubernetes/endpoints.go
@@ -25,6 +25,10 @@ func (p *KubernetesProvider) manageRemoteEpEvent(event watch.Event) error {
 		return err
 	}
 
+	if !hasToBeUpdated(endpoints.Subsets, foreignEps.Subsets) {
+		return nil
+	}
+
 	foreignEps.Subsets = p.updateEndpoints(endpoints.Subsets, foreignEps.Subsets)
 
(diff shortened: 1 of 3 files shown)


internal/kubernetes/endpoints.go
```diff
@@ -113,3 +121,21 @@ func (p *KubernetesProvider) updateEndpoints(eps, foreignEps []corev1.EndpointSu
...
+
+func hasToBeUpdated(home, foreign []corev1.EndpointSubset) bool {
+	if len(home) != len(foreign) {
+		return true
+	}
+	for i := 0; i < len(home); i++ {
+		if len(home[i].Addresses) > len(foreign[i].Addresses) {
+			return true
+		}
+		for j := 0; j != len(home[i].Addresses); j++ {
```
fraborg, line 133: just a curiosity: is there a reason you used `!=` instead of `<` ? (it's the same)

mLavacca: I got confused, the symbols `!=` and `>` in 130 and 133 had to be swapped. Thank you!

PR #81, opened 2020-07-02
e2e configmap reflection test

mLavacca:
# Configmap reflection e2e test
This PR implements an e2e test for configmaps reflection. A new namespaceNattingTable is created (hence, a namespace is remotely reflected), then some CRUD operations are performed on the configmaps (create, update, delete).
After each CRUD operation set, the coherency of the foreign cache status with the local one is asserted.

### DIFF
--- a/internal/kubernetes/configMap.go
+++ b/internal/kubernetes/configMap.go
@@ -80,10 +80,11 @@ func (p *KubernetesProvider) updateConfigMap(cm *corev1.ConfigMap, namespace str
 		return err
 	}
 
-	cm.SetNamespace(namespace)
-	cm.SetResourceVersion(cmOld.ResourceVersion)
-	cm.SetUID(cmOld.UID)
-	_, err = p.foreignClient.Client().CoreV1().ConfigMaps(namespace).Update(cm)
+	cm2 := cm.DeepCopy()
+	cm2.SetNamespace(namespace)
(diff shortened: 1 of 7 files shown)


internal/kubernetes/configMap.go
```diff
@@ -80,10 +80,11 @@ func (p *KubernetesProvider) updateConfigMap(cm *corev1.ConfigMap, namespace str
 		return err
 	}
 
-	cm.SetNamespace(namespace)
-	cm.SetResourceVersion(cmOld.ResourceVersion)
-	cm.SetUID(cmOld.UID)
-	_, err = p.foreignClient.Client().CoreV1().ConfigMaps(namespace).Update(cm)
+	cm2 := cm.DeepCopy()
```
fraborg, line 83: why did you copied the cm?

mLavacca: Because cm is a pointer to a cm stored in memory. In the tests the configmaps are stored in memory, therefore the manipulation of the object pointed by the pointer implies the modification of the original stored object. As result, if I don't copy the whole object, all the modifications of the object are applied to the original element, which is stored in the local cache, leading to an inconsistency between the local and foreign caches. This problem exists only in the testing phase.

fraborg: ok thanks

PR #95, opened 2020-07-07
Node update from advertisement test

mLavacca:
This PR introduces:
* implementation of a fake `crdClient` for testing purpose
* virtual node update from advertisement test


palexster:
Hi @mLavacca! This PR is really nice! Thanks for this! I think that you could explain a bit better the refactoring you did in the advertisement client and in particular the new interface for the client.

PR #103, opened 2020-07-11
Possible schema for the documentation of individual components

frisso:
This PR adds a template for per-component documentation.


palexster:
Hi @frisso! Thanks for your contribution! A minor comment, I think it may be less confusing to have this template in a different directory w.r.t. the actual documentation (e.g.; docs/templates/ instead of docs/design/)

PR #106, opened 2020-07-14
Update Installer and add E2E tests

palexster:
# Description

This PR adds the version customization via environment variable for the installer.
This can be used to test a specific liqo, version by setting the corresponding suffix
and version.

# How Has This Been Tested?

- [x] E2E Testing

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -77,6 +77,12 @@ HELM_ARCHIVE=helm-${HELM_VERSION}-linux-amd64.tar.gz
 HELM_URL=https://get.helm.sh/$HELM_ARCHIVE
 DEFAULT_GATEWAY_PRIVATE_IP=192.168.1.1
 NAMESPACE_DEFAULT="liqo"
+# The following variable are used a default value to select the images when installing LIQO.
+# When installing a non released version:
+# - Export LIQO_SUFFIX="-ci" if the commit is not on the master branch
+# - Export LIQO_VERSION to the id of your commit
+LIQO_VERSION_DEFAULT="latest"
+LIQO_SUFFIX_DEFAULT=""
 
 # Necessary Commands
 commands="curl kubectl"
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -117,12 +128,16 @@ GATEWAY_PRIVATE_IP_COMMAND="echo $DEFAULT_GATEWAY_PRIVATE_IP"
...
+LIQO_SUFFIX_COMMAND="echo $LIQO_SUFFIX_DEFAULT"
+set_variable_from_command LIQO_SUFFIX LIQO_SUFFIX_COMMAND "[ERROR]: Error setting the Liqo suffix... "
+LIQO_VERSION_COMMAND="echo $LIQO_VERSION_DEFAULT"
+set_variable_from_command LIQO_VERSION LIQO_VERSION_COMMAND "[ERROR]: Error setting the Liqo version... "
+
 
 #Wait for the installation to complete
-echo $NAMESPACE
-kubectl create ns $NAMESPACE
+#kkubectl create ns $NAMESPACE
```
giorio94, line 138:
```suggestion

```

install.sh
```diff
@@ -96,7 +98,16 @@ mkdir -p $TMPDIR/bin/
...
-git clone "$URL" $TMPDIR/liqo --depth 1
+if [ "$LIQO_SUFFIX" == "-ci" ]; then
+  git clone "$URL" $TMPDIR/liqo
+  cd  $TMPDIR/liqo
+  git pull origin
+  git checkout "$LIQO_VERSION" > /dev/null 2> /dev/null
+  cd -
+else
+  git clone "$URL" $TMPDIR/liqo --depth 1
+fi
```
giorio94, line 109: A couple of comments here (if I got your goal correctly)
1. I would replace this block with something like `git clone "${URL} "${TMPDIR}/liqo" --branch "${LIQO_VERSION}"  --depth 1`, that is much simpler to understand (and it is a oneliner)
2. The variable LIQO_VERSION may not be defined here (you are setting it at line 134), so it may be better to move this down

install.sh
```diff
@@ -77,6 +77,8 @@ HELM_ARCHIVE=helm-${HELM_VERSION}-linux-amd64.tar.gz
 HELM_URL=https://get.helm.sh/$HELM_ARCHIVE
 DEFAULT_GATEWAY_PRIVATE_IP=192.168.1.1
 NAMESPACE_DEFAULT="liqo"
+LIQO_VERSION_DEFAULT="latest"
```
giorio94, line 80: Should the default be *latest* or *master*?

install.sh
```diff
@@ -117,12 +128,16 @@ GATEWAY_PRIVATE_IP_COMMAND="echo $DEFAULT_GATEWAY_PRIVATE_IP"
...
+set_variable_from_command LIQO_VERSION LIQO_VERSION_COMMAND "[ERROR]: Error setting the Liqo version... "
+
 
 #Wait for the installation to complete
-echo $NAMESPACE
-kubectl create ns $NAMESPACE
+#kkubectl create ns $NAMESPACE
 $TMPDIR/bin/helm dependency update $TMPDIR/liqo/deployments/liqo_chart
-$TMPDIR/bin/helm install liqo -n liqo $TMPDIR/liqo/deployments/liqo_chart --set podCIDR=$POD_CIDR --set serviceCIDR=$SERVICE_CIDR --set gatewayPrivateIP=$GATEWAY_PRIVATE_IP --set gatewayIP=$GATEWAY_IP
+$TMPDIR/bin/helm template liqo -n liqo $TMPDIR/liqo/deployments/liqo_chart --set podCIDR=$POD_CIDR --set serviceCIDR=$SERVICE_CIDR --set gatewayPrivateIP=$GATEWAY_PRIVATE_IP --set gatewayIP=$GATEWAY_IP --set global.suffix="$LIQO_SUFFIX" --set global.version="$LIQO_VERSION"
```
giorio94, line 140: I would break this in multiple lines with `\`, since it is becoming too long and requires scrolling

install.sh
palexster: It temporarily latest, it will be replaced by a specific version when the first one will be available.

PR #108, opened 2020-07-14
Unjoin

aleoli:
# Description

This PR adds first step of unjoin process. When you finish to use remote resources you can set `join` flag on that cluster to delete the `PeeringRequest` (-> broadcaster) and the `Advertisement` (-> vk, node, ...)

For other side we can delete received `PeeringRequest` and created `Advertisement` when we want to stop resource sharing. Adv related FC will be triggered by OwnerReference and updates its status

Another new feature is `DiscoveryType` in `ForeignCluster` that let us to know how the specific resource has been discovered. Possible values are `LAN`, `WAN` and `Manual`. For LAN discovered FC is set a TTL, if this cluster is no more visible on mDNS discovery this counter is decreased. When TTL reach 0 value, FC is deleted

# How Has This Been Tested?

- [x] new tests on TTL feature written
- [x] new tests on disable `join` flag and `Advertisement` delete written
- [x] due to test environment garbage collector tests on triggering event on Adv delete can't be tested, I've done some manual tests on my k3s installation

### DIFF
--- a/api/discovery/v1/foreigncluster_methods.go
+++ b/api/discovery/v1/foreigncluster_methods.go
@@ -1,12 +1,15 @@
 package v1
 
 import (
+	protocolv1 "github.com/liqoTech/liqo/api/advertisement-operator/v1"
+	"github.com/liqoTech/liqo/pkg/crdClient"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"k8s.io/klog"
 )
 
 func (fc *ForeignCluster) GetConfig(client kubernetes.Interface) (*rest.Config, error) {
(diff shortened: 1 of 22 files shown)


palexster:
Hi @aleoli, thanks for this PR! Is this documented somewhere in the repository documentation?

aleoli:
No, I'm going to add it in documentation PR

api/discovery/v1/foreigncluster_methods.go
```diff
@@ -84,3 +87,32 @@ func (fc *ForeignCluster) LoadForeignCA(localClient kubernetes.Interface, localN
...
+func (fc *ForeignCluster) SetAdvertisement(adv *protocolv1.Advertisement, discoveryClient *crdClient.CRDClient) error {
+	if fc.Status.Advertisement == nil {
+		// Advertisement has not been set in ForeignCluster yet
+		fc.Status.Advertisement = &v1.ObjectReference{
+			Kind:       "Advertisement",
+			Name:       adv.Name,
+			UID:        adv.UID,
+			APIVersion: "protocol.liqo.io/v1",
+		}
+		_, err := discoveryClient.Resource("foreignclusters").Update(fc.Name, fc, metav1.UpdateOptions{})
```
fraborg, line 100: I'm not sure, but since you are modifyng the resource status I think here you need to use `UpdateStatus()` method

internal/discovery/foreign-cluster-operator/setup.go
```diff
@@ -42,10 +43,17 @@ func StartOperator(mgr *manager.Manager, namespace string, requeueAfter time.Dur
 		os.Exit(1)
 	}
 
+	advClient, err := protocolv1.CreateAdvertisementClient(filepath.Join(os.Getenv("HOME"), ".kube", "config"), nil)
```
fraborg, line 46: maybe it's remained from debug
if you need an in-cluster client you should leave the kubeconfigPath to ""
if you want to use a specified kubeconfigPath I would pass it as parameter rather than hard-code it

api/discovery/v1/foreigncluster_methods.go
aleoli: Yes, I think that it would be slightly more efficient

PR #119, opened 2020-07-16
Add checks on negative advertised resources

fraborg:
# Description

This PR adds some checks on the resources announced in the Advertisement
- If the computed resources by the broadcaster are negative, set them to 0
- If the advertisement operator receives an Advertisement with negative resources, refuse it


palexster:
Hi @fraborg ! Thanks for this! One question: can we enforce a strong type checking on advertisement CRD in order to prevent at the source wrong modifications of the CRD?

fraborg:
> Hi @fraborg ! Thanks for this! One question: can we enforce a strong type checking on advertisement CRD in order to prevent at the source wrong modifications of the CRD?

I think the only possibility is to modify  [this](https://github.com/LiqoTech/liqo/blob/master/config/advertisement-operator/crd/bases/protocol.liqo.io_advertisements.yaml#L220) regex. 
The problem is that CRD is auto-generated when we compile the struct definition, and I think we cannot apply annotations over the struct (because we use predefined struct types)

PR #124, opened 2020-07-20
Advertisement deletion for unjoin

fraborg:
# Description
This PR adds the second part of unjoin process, when a cluster doesn't want to share his resources anymore.

1. Added `enableBroadcaster` flag in `ClusterConfig`
1. When the `enableBroadcaster` flag is set to `false`, the home cluster notifies to the foreign he wants to stop sharing by setting the advertisement status to `DELETING`.
2. The virtual kubelet on the foreign cluster is triggered: 
    1. it deletes all the offloaded resources
    2. it deletes the Advertisement, causing the deletion of the virtual-kubelet deployment and virtual node through ownerReference.

## Other features
- the `ClusterRole` created by discovery modules has been updated to allow operations over Advertisement status
- the remote watcher is closed when the adv is deleted
- added periodic garbage collection in advertisement controller to remove expired adv
- added `createClusterConfigClient` function to have a simple way to create a client for `ClusterConfig` resource

# How Has This Been Tested?

- [x] tested with kind
- [x] ClusterConfig unit tests
  - [x] when we modify the SharingPercentage the advertisement is updated with the new quantity of shared resources
  - [x] when we disable Advertisement the Advertisement is deleted on foreign cluster


internal/kubernetes/reflectionController.go
```diff
@@ -112,6 +112,18 @@ func (p *KubernetesProvider) controlLoop() {
 
 func (p *KubernetesProvider) cleanupNamespace(ns string) error {
 
+	pods, err := p.foreignClient.Client().CoreV1().Pods(ns).List(metav1.ListOptions{})
```
mLavacca, line 115: You can use provider's foreign caches instead of performing this API server request

internal/kubernetes/reflectionController.go
```diff
@@ -120,7 +132,7 @@ func (p *KubernetesProvider) cleanupNamespace(ns string) error {
 	for _, svc := range svcs.Items {
 		err = p.foreignClient.Client().CoreV1().Services(ns).Delete(svc.Name, &metav1.DeleteOptions{})
 		if err != nil {
-			klog.Error(err, "cannot delete remote service")
+			klog.Error(err, "cannot delete remote service %v", svc.Name)
```
mLavacca, line 135: as above

internal/kubernetes/reflectionController.go
```diff
@@ -112,6 +112,18 @@ func (p *KubernetesProvider) controlLoop() {
...
 
+	pods, err := p.foreignClient.Client().CoreV1().Pods(ns).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, po := range pods.Items {
+		err = p.foreignClient.Client().CoreV1().Pods(ns).Delete(po.Name, &metav1.DeleteOptions{})
+		if err != nil {
+			klog.Error(err, "cannot delete remote pod %v", po.Name)
```
mLavacca, line 123: In order to use the escape feature, you should use `klog.errorf` instead of `klog.error`

internal/kubernetes/virtualNodeupdater.go
```diff
@@ -115,3 +126,25 @@ func (p *KubernetesProvider) updateFromAdv(adv advv1.Advertisement) error {
 
 	return p.nodeController.UpdateNodeFromOutside(false, no)
 }
+
+func (p *KubernetesProvider) deleteAdv(adv *advv1.Advertisement) error {
+	// delete all reflected resources in reflected namespaces
+	for ns := range p.reflectedNamespaces.ns {
```
mLavacca, line 132: Instead of performing this manual operation, we could exploit [this](https://github.com/LiqoTech/liqo/blob/7e35c77582552d2847c0a8804727270499349ccb/internal/kubernetes/namespace.go#L169) function, that is triggered for every update of the `NamespaceNattingTable`. It is enough to delete the whole `NattingTable` from the `NamespaceNattingTable` and update the corresponding k8s resource. This operation ends up triggering the mechanism that cleans all the reflected namespaces.

mLavacca: However, this approach implies that we must be sure that all the reflection routines have been stopped before deleting the advertisement. This behavior can be achieved by polling to [this](https://github.com/LiqoTech/liqo/blob/7e35c77582552d2847c0a8804727270499349ccb/internal/kubernetes/reflectionController.go#L320) function for each namespace

internal/kubernetes/reflectionController.go
```diff
@@ -112,6 +112,18 @@ func (p *KubernetesProvider) controlLoop() {
 
 func (p *KubernetesProvider) cleanupNamespace(ns string) error {
 
+	pods, err := p.foreignClient.Client().CoreV1().Pods(ns).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, po := range pods.Items {
+		err = p.foreignClient.Client().CoreV1().Pods(ns).Delete(po.Name, &metav1.DeleteOptions{})
```
mLavacca, line 121: Before deleting all the pods, I think we should set the virtual node as `notReady`, otherwise, if we delete the foreign pod, the vk, will notify this event and will delete also the local copy, leading to the `replicaController` trigger that re-creates the pod, scheduling it again to the virtual node.

mLavacca: This comment is related to this line, even though the node status setting should be performed while detecting the advertisement deletion [here](https://github.com/LiqoTech/liqo/blob/4b29ac14aa1f67de0ded0a248a2f5983218eac3a/internal/kubernetes/virtualNodeupdater.go#L59).

internal/kubernetes/virtualNodeupdater.go
```diff
@@ -55,6 +56,16 @@ func (p *KubernetesProvider) ReconcileNodeFromAdv(event watch.Event) {
 		return
 	}
 
+	if adv.Status.AdvertisementStatus == advertisement_operator.AdvertisementDeleting {
+		klog.Infof("advertisement %v is going to be deleted... delete all offloaded resources", adv.Name)
+		if err := p.deleteAdv(adv); err != nil {
+			klog.Infof("something went wrong during advertisement deletion - %v", err)
```
palexster, line 62: Retry?

internal/advertisement-operator/broadcaster.go
```diff
@@ -325,6 +326,29 @@ func (b *AdvertisementBroadcaster) SendAdvertisementToForeignCluster(advToCreate
...
+		adv := obj.(*protocolv1.Advertisement)
+		adv.Status.AdvertisementStatus = AdvertisementDeleting
+		_, err = b.RemoteClient.Resource("advertisements").UpdateStatus(adv.Name, adv, metav1.UpdateOptions{})
+		if err != nil {
+			klog.Error("Unable to update Advertisement " + adv.Name)
+			return err
+		}
+	}
+	// delete the peering request to delete the broadcaster
+	if err := b.DiscoveryClient.Resource("peeringrequests").Delete(b.PeeringRequestName, metav1.DeleteOptions{}); err != nil {
```
aleoli, line 345: Await for advertisement to be deleted to avoid PeeringRequest recreation

If ForeignCluster operator has an update between PR delete and Adv delete, it is aware of that process and it will create new PR

fraborg: thanks!

internal/kubernetes/reflectionController.go
fraborg: thanks!

internal/kubernetes/virtualNodeupdater.go
```diff
@@ -55,6 +56,34 @@ func (p *KubernetesProvider) ReconcileNodeFromAdv(event watch.Event) {
...
+			klog.Infof("advertisement %v is going to be deleted... set node status not ready", adv.Name)
+			no, err := p.nodeUpdateClient.Client().CoreV1().Nodes().Get(p.nodeName, metav1.GetOptions{})
+			if err != nil {
+				klog.Error(err)
+				continue
+			}
+			for _, condition := range no.Status.Conditions {
+				if condition.Type == v1.NodeReady {
+					no.Status.Conditions[0].Status = v1.ConditionFalse
+					_, err = p.nodeUpdateClient.Client().CoreV1().Nodes().UpdateStatus(no)
```
mLavacca, line 70: Instead of performing this manual operation, you can update the node status by using the provider's nodeController with [this](https://github.com/LiqoTech/liqo/blob/e0433b9b79574feef7ec9494d65bef7d80127e8a/internal/kubernetes/virtualNodeupdater.go#L145) method. By following this approach you avoid the concurrency problem.

internal/advertisement-operator/config-watcher.go
```diff
@@ -4,15 +4,40 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
+	"time"
 )
 
-func (b *AdvertisementBroadcaster) WatchConfiguration(kubeconfigPath string) {
+func (b *AdvertisementBroadcaster) WatchConfiguration(kubeconfigPath string, client *crdClient.CRDClient) {
 	go clusterConfig.WatchConfiguration(func(configuration *policyv1.ClusterConfig) {
-		if configuration.Spec.AdvertisementConfig.ResourceSharingPercentage != b.ClusterConfig.ResourceSharingPercentage {
+		if !configuration.Spec.DiscoveryConfig.EnableAdvertisement {
```
aleoli, line 16: This config is to enable mDNS advertisement. I suggest to add another flag

I will change this name to EnableLANAdvertisement to avoid misunderstandings

internal/advertisement-operator/config-watcher.go
```diff
@@ -4,15 +4,40 @@ import (
...
 			klog.V(3).Info("ClusterConfig changed")
-			b.ClusterConfig.ResourceSharingPercentage = configuration.Spec.AdvertisementConfig.ResourceSharingPercentage
+			klog.Info("Stopping sharing resources with cluster " + b.ForeignClusterId)
+			err := b.NotifyAdvertisementDeletion()
```
aleoli, line 19: If advertisement does not exist in remote cluster this is a never ending process

If we don't find our Adv we can delete local PeeringRequest even if an error is returned

internal/advertisement-operator/config-watcher.go
fraborg: so at the moment we don't have a flag to enable/disable advertisement (i.e. the creation/deletion of the broadcaster)?

internal/kubernetes/virtualNodeupdater.go
fraborg: thanks a lot!

PR #131, opened 2020-07-22
Add custom resolution for testing purpose

aleoli:
# Description

This PR allows DNS discovery process to use CNAME record as the last step

So, it's possible to use both IP or server name to contact remote API Server

This is useful for testing purpose

### DIFF
--- a/internal/discovery/search-domain-operator/wan.go
+++ b/internal/discovery/search-domain-operator/wan.go
@@ -74,11 +74,25 @@ func ResolveWan(c *dns.Client, dnsAddr string, ptr *dns.PTR) (*discovery.TxtData
 		return nil, err
 	}
 	if len(in.Answer) == 0 {
-		klog.Error("A record is not set for " + srv.Target)
-		return nil, errors.New("A record is not set for " + srv.Target)
+		// no A record is set, let's try with CNAME
+		msg = GetDnsMsg(srv.Target, dns.TypeCNAME)
+		in, _, err = c.Exchange(msg, dnsAddr)
+		if err != nil {
+			klog.Error(err, err.Error())
+			return nil, err
+		}
+
+		if len(in.Answer) == 0 {
+			klog.Error("no A record or CNAME record is set for " + srv.Target)
+			return nil, errors.New("no A record or CNAME record is set for " + srv.Target)
+		}
+
+		cname := in.Answer[0].(*dns.CNAME)
+		return discovery.Decode(cname.Target, strconv.Itoa(int(srv.Port)), txt)
+	} else {
+		a := in.Answer[0].(*dns.A)
+		return discovery.Decode(a.A.String(), strconv.Itoa(int(srv.Port)), txt)
 	}
-	a := in.Answer[0].(*dns.A)
-	return discovery.Decode(a.A.String(), strconv.Itoa(int(srv.Port)), txt)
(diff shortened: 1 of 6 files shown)


frisso:
@aleoli Not sure why we have to proceed in resolving A/CNAME records. Should't we simply connect to the given host, once we get the result from the SRV record? The name resolution would then left to the socket interface, hence simplifying the code.

aleoli:
Because API Server certificate may not be valid for SRV record name

For example, my local installation has a valid certificate for alternative names:
* localhost
* 127.0.0.1
* 192.168.2.215

This certificate is automatically managed by k8s. If we address the server with another name https client will return an error

frisso:
> Because API Server certificate may not be valid for SRV record name
> 
> For example, my local installation has a valid certificate for alternative names:
> 
> * localhost
> * 127.0.0.1
> * 192.168.2.215
> 
> This certificate is automatically managed by k8s. If we address the server with another name https client will return an error

This looks to me a server problem.
I consider very strange that the user has to take care of DNS resolution by hand. The socket interface solved this problem 20 years ago, exactly to avoid programmers to take care of this. So, I would simplify the code by removing the A/CNAME part.
@palexster what do you think (we can also discuss by voice this afternoon)

aleoli:
So the suggestion is to remove A/CNAME resolution and add alternative name in certifiate, it makes sense 😅

palexster:
> > Because API Server certificate may not be valid for SRV record name
> > For example, my local installation has a valid certificate for alternative names:
> > 
> > * localhost
> > * 127.0.0.1
> > * 192.168.2.215
> > 
> > This certificate is automatically managed by k8s. If we address the server with another name https client will return an error
> 
> This looks to me a server problem.
> I consider very strange that the user has to take care of DNS resolution by hand. The socket interface solved this problem 20 years ago, exactly to avoid programmers to take care of this. So, I would simplify the code by removing the A/CNAME part.
> @palexster what do you think (we can also discuss by voice this afternoon)?

I agree with you @frisso. For my perspective, having the manual resolution make possible to easily test the feature without having to setup a whole dns server. For the certificate perspective, WAN discovery is thought for well established clusters, which are normally disposed of appropriate certificates for their clusters.

frisso:
> So the suggestion is to remove A/CNAME resolution and add alternative name in certifiate, it makes sense 😅

Yes, exactly.

internal/discovery/search-domain-operator/wan.go
```diff
@@ -66,19 +66,38 @@ func ResolveWan(c *dns.Client, dnsAddr string, ptr *dns.PTR) (*discovery.TxtData
...
-	msg = GetDnsMsg(srv.Target, dns.TypeA)
-	in, _, err = c.Exchange(msg, dnsAddr)
-	if err != nil {
-		klog.Error(err, err.Error())
-		return nil, err
-	}
-	if len(in.Answer) == 0 {
-		klog.Error("A record is not set for " + srv.Target)
-		return nil, errors.New("A record is not set for " + srv.Target)
+	if test {
```
palexster, line 69: It is ok to me, but I would move this logic outside this function in order to not pollute the default behavior with this code which is useful for testing only.

aleoli: ok

PR #134, opened 2020-07-24
Trust CA

aleoli:
# Description

At the moment, the foreign cluster's root CA is downloaded at first request leading to the impossibility to authenticate the remote cluster.

With this PR can trust CAs different from the default cluster CA (the one used by pods) and use the ones signed by default trusted root CAs. We can set the local cluster to not provide its CA at the first request but require that its CA is already installed on the remote cluster.

We have three ways to get the remote cluster CA:
1. Read from public readable secret containing its CA (we cannot verify that the remote cluster is who it claims to be) [not authenticated]
2. Using public root CA [authenticated]
3. Adding trusted root CA [authenticated]

## Trusted Foreign Clusters

If a remote cluster requires authentication, `ForeignCluster` CR will be created with the `requireCa` flag enabled in its Spec. If the remote cluster certificate is signed by a "default" root CA we are ok, else if not we have to add its root CA (provided Out-Of-Band) in our `trusted-ca-certificates` ConfigMap

Example:
```bash
kubectl edit cm trusted-ca-certificates
```
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: trusted-ca-certificates
data:
  remote: |
    -----BEGIN CERTIFICATE-----
    MIIBVjCB/qADAgECAgEAMAoGCCqGSM49BAMCMCMxITAfBgNVBAMMGGszcy1zZXJ2
    ...
    APKY9n4CRdSWSQ==
    -----END CERTIFICATE-----
```

When we will try the peering, https client will check that API server certificate was issued by one of trusted CAs

__NOTE:__ when this ConfigMap is updated, the discovery component will trigger a restart to reload new CA configurations


api/cluster-config/v1/clusterconfig_types.go
```diff
@@ -63,6 +63,10 @@ type DiscoveryConfig struct {
 	// --- DNS ---
 
 	DnsServer string `json:"dnsServer"`
+
+	// --- CA ---
+
+	RequireCa bool `json:"requireCa"`
```
palexster, line 69: RequireCa is a bit hard to understand, I would name AllowUntrustedCA

scripts/discovery/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# update CA certificates
+update-ca-certificates 2>/dev/null
```
palexster, line 4: Can we do this at Docker image build time?

deployments/liqo_chart/crds/foreign-cluster-crd.yaml
```diff
@@ -52,6 +52,9 @@ spec:
                 - LAN
                 - WAN
                 - Manual
+            requireCa:
+              type: boolean
+              description: Requires CA installation
```
palexster, line 57: A bit hard to understand...

scripts/discovery/entrypoint.sh
aleoli: no, we need user provided CAs in ConfigMap mounted in `/usr/local/share/ca-certificates`

api/cluster-config/v1/clusterconfig_types.go
```diff
@@ -58,11 +58,16 @@ type DiscoveryConfig struct {
 	EnableDiscovery     bool `json:"enableDiscovery"`
 	EnableAdvertisement bool `json:"enableAdvertisement"`
 
-	AutoJoin bool `json:"autojoin"`
+	AutoJoin          bool `json:"autojoin"`
+	AutoJoinUntrusted bool `json:"autojoinUntrusted"`
```
alacuku, line 62: This new fields are set as required. Did you check if we need to update the various tests that uses the `clusterconfig.policy.liqo.io` CRD?

aleoli: No, tests have not to be updated, they are not tested yet because they are not testable locally (at least it is not easy), there will be added some e2e test to check trusted/untrusted peering

PR #144, opened 2020-08-03
Add ForeignCluster on PeeringRequest accepted

aleoli:
# Description

When a `PeeringRequest` is accepted, a new `ForeignCluster` is created/updated according to it. In this way, we can use `ForeignCluster`s as a catalog where we can check the status for every connection, both incoming and outgoing

## Changes

New DiscoveryType is added: "IncomingPeering", in ForeignCLuster Status joined flag has been splitter in `OutgoingJoin` and `IncomingJoin` flags

When we create new PeeringRequest we can specify if we want that remote cluster will join us back settings `ReverseJoin` to `RequireReverseJoin`, otherwise `ForeignCluster` will be created on the remote cluster but it will not try to join us

## Fixes

A `cluster-id` label is added to each `ForeignCluster` to retrieve them querying on this parameter even if they are been created manually (so ClusterId and name can be different)


api/discovery/v1/foreigncluster_types.go
```diff
@@ -52,11 +53,12 @@ type ForeignClusterStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	Joined             bool                `json:"joined"`
-	PeeringRequestName string              `json:"peering-request-name"`
-	CaDataRef          *v1.ObjectReference `json:"caDataRef,omitempty"`
-	Advertisement      *v1.ObjectReference `json:"advertisement,omitempty"`
-	Ttl                int                 `json:"ttl,omitempty"`
+	Joined                   bool                `json:"joined"`
```
palexster, line 56: Proposal: Maybe we can split between the incoming and the outgoing part to avoid confusion between parameters affecting one or the other direction? Does it seem useful?

aleoli: This field will become an enum, it will be "None, Incoming, Outgoing, Bidirectional"

api/discovery/v1/foreigncluster_types.go
```diff
@@ -53,11 +54,13 @@ type ForeignClusterStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	Joined             bool                `json:"joined"`
-	PeeringRequestName string              `json:"peering-request-name,omitempty"`
-	CaDataRef          *v1.ObjectReference `json:"caDataRef,omitempty"`
-	Advertisement      *v1.ObjectReference `json:"advertisement,omitempty"`
-	Ttl                int                 `json:"ttl,omitempty"`
+	OutgoingJoin             bool                `json:"outgoingJoin"`
```
palexster, line 57: Proposal: can we have two distinct sections for outgoing and incoming? This way we can clearly understand the status of peering in each direction, and the overall parameters...

internal/peering-request-operator/foreign-cluster.go
```diff
@@ -0,0 +1,158 @@
...
+				Kind:       pr.Kind,
+				Name:       pr.Name,
+				UID:        pr.UID,
+				APIVersion: pr.APIVersion,
+			},
+		},
+	}
+
+	if r.ForeignConfig != nil {
+		fc.Spec.ApiUrl = strings.Replace(fc.Spec.ApiUrl, "https://", "http://", -1)
```
palexster, line 120: Why do we need this replace? Can't we assume that the ApiURL is https?

internal/peering-request-operator/foreign-cluster.go
```diff
@@ -0,0 +1,158 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+	"k8s.io/utils/pointer"
+	"strings"
+)
+
+func (r *PeeringRequestReconciler) UpdateForeignCluster(pr *v1.PeeringRequest) error {
+	tmp, err := r.crdClient.Resource("foreignclusters").Get(pr.Spec.ClusterID, metav1.GetOptions{})
```
palexster, line 16: Does we require both get operations? I think we have to make a clear design choice (e.g. get by label) since we have control on how the ForeignCluster are created or am I missing something?

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -245,6 +284,21 @@ func (r *ForeignClusterReconciler) checkJoined(fc *discoveryv1.ForeignCluster, f
 	return fc, nil
 }
 
+func (r *ForeignClusterReconciler) getOriginClusterSets() discoveryv1.OriginClusterSets {
```
palexster, line 287: I think it could be nice to have a comment on this method

internal/peering-request-operator/foreign-cluster.go
aleoli: I'm sorry, I forgot to add a comment. r.ForeignConfig is only set in test environment (where we have only one ForeignConfig) and in this environment API server is only available in http, but field is automatically created with https, because yes, we can assume that ApiUrl is https

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -245,6 +284,21 @@ func (r *ForeignClusterReconciler) checkJoined(fc *discoveryv1.ForeignCluster, f
...
+	if r.ForeignConfig != nil {
+		// testing
```
palexster, line 289: As a general behavior, I would suggest to avoid adding testing branches when not strictly needed. In this case, for example, does the r.ForeignConfig != nil make sense only for testing?

internal/peering-request-operator/foreign-cluster.go
aleoli: In 90% of times (every time that is not manually create) we can use Get function that is faster, if get fails we can try to get it by label

If it seems too much confusing I can use label only

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
aleoli: r.ForeignConfig != nil is true only in testing environment, because of slightly different behaviour. In this case I have not ClsuterConfig running, so I can't get policy on trusting CAs

internal/peering-request-operator/foreign-cluster.go
palexster: Which are the case which are not accessible using a get by name?

aleoli: Only if someone creates it manually

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
palexster: The problem is: let's suppose that there is a bug in the operator and the r.ForeignConfig is not Nil as expected when you are running this code outside testing, this would result in a error which we are treating as a testCase. I would suggest to refactor this and treat this unexpected parameter as an unexpected behavior.

internal/peering-request-operator/foreign-cluster.go
palexster: We should avoid at most to add code for testing purposes, and put it in fake classes.

internal/peering-request-operator/foreign-cluster.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+}
+
+func (r *PeeringRequestReconciler) createForeignCluster(pr *v1.PeeringRequest) (*v1.ForeignCluster, error) {
+	var cnf *rest.Config
+	var err error
+
+	cnf, err = pr.GetConfig(r.crdClient.Client())
+	if err != nil {
+		if strings.Contains(err.Error(), "yaml") && r.Testing {
```
mLavacca, line 89: this condition isn't robust at all; what if the error message changes?  Cannot we implement a more robust way to detect the error type?

internal/advertisement-operator/controller.go
```diff
@@ -140,12 +140,26 @@ func (r *AdvertisementReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // set Advertisement reference in related ForeignCluster
 func (r *AdvertisementReconciler) UpdateForeignCluster(adv *protocolv1.Advertisement) (error, bool) {
-	tmp, err := r.DiscoveryClient.Resource("foreignclusters").Get(adv.Spec.ClusterId, metav1.GetOptions{})
+	tmp, err := r.DiscoveryClient.Resource("foreignclusters").List(metav1.ListOptions{
```
fraborg, line 143: why do you use `List` if you're interested only on 1 ForeignCluster object?

internal/peering-request-operator/foreign-cluster.go
```diff
@@ -0,0 +1,129 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+	"k8s.io/utils/pointer"
+	"strings"
+)
+
+func (r *PeeringRequestReconciler) UpdateForeignCluster(pr *v1.PeeringRequest) error {
+	tmp, err := r.crdClient.Resource("foreignclusters").List(metav1.ListOptions{
```
fraborg, line 15: as above, is it necessary to use `List` and then select the first one instead of directly use `Get`?

internal/peering-request-operator/foreign-cluster.go
```diff
@@ -0,0 +1,129 @@
...
+		},
+		Spec: v1.ForeignClusterSpec{
+			ClusterID:        pr.Spec.ClusterID,
+			Namespace:        pr.Spec.Namespace,
+			Join:             false,
+			ApiUrl:           cnf.Host,
+			DiscoveryType:    v1.IncomingPeeringDiscovery,
+			AllowUntrustedCA: pr.Spec.OriginClusterSets.AllowUntrustedCA,
+		},
+		Status: v1.ForeignClusterStatus{
```
fraborg, line 87: if I remember well, when you create an object the status is dropped so this should be ignored

internal/advertisement-operator/controller.go
aleoli: Because I can use LabelSelector. When a foreign cluster is created is added a label indicating its clusterId, so I can retrieve it by this label that is in our control. If someone creates a ForeignClsuter manually it's possible that the name and the clusterId are different

internal/peering-request-operator/foreign-cluster.go
aleoli: As above

internal/peering-request-operator/foreign-cluster.go
aleoli: No, it is working. I tested it in my k3s installation

fraborg: Good to know!

docs/pages/Architecture/Discovery and Peering/discovery-component.md
```diff
@@ -91,4 +93,8 @@ When we no longer need foreign resources we can disable `join` flag to trigger p
 1. `PeeringRequest` delete
     * that triggers `Broadcaster` termination
 2. `Advertisement` delete
-    * that triggers `VirtualKubelet` termination
\ No newline at end of file
+    * that triggers `VirtualKubelet` termination
+
+### Witch fields are managed by who?
```
fraborg, line 98:
```suggestion
### Which fields are managed by who?
```

PR #148, opened 2020-08-06
[tray-agent]: LiqoDash integration

AbakusW:
# Description

This PR adds the Liqo Agent feature "launch LiqoDash" allowing the user to open in the default browser the [Liqodash](https://github.com/LiqoTech/dashboard) application and easily access the required access token directly from the clipboard.

### DIFF
--- a/internal/tray-agent/agent/client/agent-client.go
+++ b/internal/tray-agent/agent/client/agent-client.go
@@ -8,11 +8,20 @@ import (
 	"github.com/liqoTech/liqo/pkg/crdClient"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	"os"
 	"path/filepath"
 	"sync"
+)
+
+const (
+	//EnvLiqoKConfig defines the env var containing the path of the kubeconfig file of the
+	//cluster associated to Liqo Agent.
+	EnvLiqoKConfig = "LIQO_KCONFIG"
 )
 
 //AgentController singleton.
@@ -44,11 +53,17 @@ func DestroyMockedAgentController() {
 
 //AgentController is the data structure that manages Tray Agent interaction with the cluster.
 type AgentController struct {
-	client    *crdClient.CRDClient
-	valid     bool
-	connected bool
-	advCache  *AdvertisementCache
-	mocked    bool
+	kubeClient kubernetes.Interface
+	client     *crdClient.CRDClient
+	valid      bool
+	connected  bool
+	advCache   *AdvertisementCache
+	mocked     bool
+}
+
+//KubeClient returns a Kubernetes client to interact with the cluster.
+func (ctrl *AgentController) KubeClient() kubernetes.Interface {
+	return ctrl.kubeClient
 }
 
 //Mocked returns if the AgentController is mocked (true).
@@ -63,6 +78,9 @@ func (ctrl *AgentController) Client() *crdClient.CRDClient {
 
 //Connected returns if the controller client is actually connected to the cluster.
 func (ctrl *AgentController) Connected() bool {
+	if mockedController {
+		return true
+	}
 	return ctrl.connected
 }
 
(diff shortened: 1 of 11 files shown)


internal/tray-agent/agent/client/dashboard-client.go
```diff
@@ -0,0 +1,186 @@
...
+	"errors"
+	"fmt"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"os"
+	"strings"
+)
+
+const (
```
palexster, line 14: Are those default values or are they hardcoded values? Configuration should be taken from external environment/configuration files.

internal/tray-agent/agent/client/dashboard-client.go
```diff
@@ -0,0 +1,186 @@
...
+		return err
+	}
+	if err = os.Unsetenv(EnvLiqoDashPort); err != nil {
+		return err
+	}
+	//preliminary check to verify the LiqoDash pod is running
+	var dashPodL *corev1.PodList
+	dashPodL, err = ctrl.kubeClient.CoreV1().Pods(liqoDashboardNamespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "app=" + liqoDashboardServiceName,
+		FieldSelector: fields.OneTermEqualSelector("status.phase", "Running").String(),
```
palexster, line 50: Suggestion: May the check be hardened looking if the pod is ready to serve traffic and not only running?

internal/tray-agent/agent/client/dashboard-client.go
```diff
@@ -0,0 +1,186 @@
...
+	})
+	if err != nil {
+		return errors.New("cluster connection not available")
+	}
+	if len(dashPodL.Items) < 1 {
+		return errors.New("the LiqoDash is not currently available")
+	}
+	/*-----------------------------------------------------------------------------------
+	CASE 1: check the presence of an ingress controller for the LiqoDash
```
palexster, line 59: Are you looking for an ingress or an ingress controller?

internal/tray-agent/agent/client/dashboard-client.go
AbakusW: Yes they are default values. Since the aim of the button action is to launch the Dashboard, I tried to reduce the processing time specifically switching for each resource from a two-step procedure (List w/ listOptions + find) to a single step (this does not apply obviously to pods). I assumed that enforcing a specific name just for the LiqoDash resources was an acceptable trade-off to gain a quicker response.

Do you think a List solution is more suitable?

internal/tray-agent/agent/client/dashboard-client.go
AbakusW: wrong comment, my bad. It is an Ingress.

internal/tray-agent/agent/client/agent-client.go
```diff
@@ -189,3 +209,26 @@ func AcquireKubeconfig() bool {
...
+	}
+	kubeconfig, ok := os.LookupEnv(EnvLiqoKConfig)
+	if !ok || kubeconfig == "" {
+		return nil, errors.New("no kubeconfig provided")
+	}
+	var (
+		cfg *rest.Config
+		err error
+	)
+	cfg, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
```
fraborg, line 229: why do you use the var block and don't use `cfg, err := ...` ?

internal/tray-agent/agent/client/agent-client.go
```diff
@@ -44,11 +53,17 @@ func DestroyMockedAgentController() {
 
 //AgentController is the data structure that manages Tray Agent interaction with the cluster.
 type AgentController struct {
-	client    *crdClient.CRDClient
-	valid     bool
-	connected bool
-	advCache  *AdvertisementCache
-	mocked    bool
+	kubeClient kubernetes.Interface
```
fraborg, line 56: it seems to me that this `kubeClient` is connected to the same cluster as `client` field: can't you reuse this one?

palexster (approved):
/approve

internal/tray-agent/agent/client/agent-client.go
AbakusW: yes you're right, it has no particular meaning and maybe comes from an unfixed copy&paste or a previous version.

internal/tray-agent/agent/client/agent-client.go
AbakusW: The inner structure of AgentController needs to be updated in future versions in order to manage also the clients of other CRDs. I preferred separating the two clients at least for the moment. In the future I could consider to use the inner client of the ClusterConfig CRD.

fraborg (approved):

PR #150, opened 2020-08-07
Simplify LIQO installattion procedure

palexster:
# Description

This commit add the possibility to add auto-labeling of gateway node. If no node has been already labelled ***liqonet.liqo.io/gateway***, the last node in kubectl get no is selected as gateway node.

# How Has This Been Tested?

This commit has been tested with the E2E pipeline.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -6,7 +6,7 @@ function cleanup()
    set +e
 }
 
-set_variable_from_command() {    
+function set_variable_from_command() {
     if [[ ($# -ne 3) ]];
     then
       echo "Internal Error - Wrong number of parameters"
@@ -24,7 +24,7 @@ set_variable_from_command() {
     echo "[PRE-INSTALL]: $VAR_NAME is set to: ${!VAR_NAME}"
 }
 
-print_help()
+function print_help()
 {
    echo "Arguments:"
    echo "   --help: print this help"
@@ -52,6 +52,15 @@ function wait_and_approve_csr(){
    return 0
 }
 
+function set_gateway_node() {
+   test=$(kubectl get no -l "liqonet.liqo.io/gateway=true" 2> /dev/null | wc -l)
+   if [ $test == 0 ]; then
+      node=$(kubectl get no -o jsonpath="{.items[-1].metadata.name}")
+      kubectl label no $node liqonet.liqo.io/gateway=true > /dev/null
+   fi
+   address=$(kubectl get no -l "liqonet.liqo.io/gateway=true" -o jsonpath="{.items[0].status.addresses[0].address}")
+   echo "$address"
+}
 
 if [[ ($# -eq 1 && $1 == '--help') ]];
 then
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -52,6 +52,15 @@ function wait_and_approve_csr(){
...
 }
 
+function set_gateway_node() {
+   test=$(kubectl get no -l "liqonet.liqo.io/gateway=true" 2> /dev/null | wc -l)
+   if [ $test == 0 ]; then
+      node=$(kubectl get no -o jsonpath="{.items[-1].metadata.name}")
+      kubectl label no $node liqonet.liqo.io/gateway=true > /dev/null
+      address=$(kubectl get no -l "liqonet.liqo.io/gateway=true" -o jsonpath="{.items[0].status.addresses[0].address}")
+   fi
+   echo "$address"
```
giorio94, line 62:
```suggestion
   return "$address"
```

Looks better to me

install.sh
```diff
@@ -52,6 +52,15 @@ function wait_and_approve_csr(){
    return 0
 }
 
+function set_gateway_node() {
+   test=$(kubectl get no -l "liqonet.liqo.io/gateway=true" 2> /dev/null | wc -l)
+   if [ $test == 0 ]; then
+      node=$(kubectl get no -o jsonpath="{.items[-1].metadata.name}")
+      kubectl label no $node liqonet.liqo.io/gateway=true > /dev/null
+      address=$(kubectl get no -l "liqonet.liqo.io/gateway=true" -o jsonpath="{.items[0].status.addresses[0].address}")
+   fi
```
giorio94, line 61:
```suggestion
      kubectl label no $node liqonet.liqo.io/gateway=true > /dev/null
   fi
   address=$(kubectl get no -l "liqonet.liqo.io/gateway=true" -o jsonpath="{.items[0].status.addresses[0].address}")
```

Otherwise the address variable gets no value if the node was already labeled

install.sh
```diff
@@ -121,12 +130,18 @@ else
   echo "[PRE-INSTALL]: Kubeconfig variable is not set. Kubectl will use: ~/.kube/config"
 fi
 
+
+echo "[PRE-INSTALL]: Setting Gateway IP"
+GATEWAY_IP=$(set_gateway_node)
```
giorio94, line 135:
```suggestion
GATEWAY_IP=set_gateway_node
```

install.sh
```diff
@@ -121,12 +130,18 @@ else
...
+
+
 POD_CIDR_COMMAND='kubectl cluster-info dump | grep -m 1 -Po "(?<=--cluster-cidr=)[0-9.\/]+"'
 set_variable_from_command POD_CIDR POD_CIDR_COMMAND "[ERROR]: Unable to find POD_CIDR"
 SERVICE_CIDR_COMMAND='kubectl cluster-info dump | grep -m 1 -Po "(?<=--service-cluster-ip-range=)[0-9.\/]+"'
 set_variable_from_command SERVICE_CIDR SERVICE_CIDR_COMMAND "[ERROR]: Unable to find Service CIDR"
-GATEWAY_IP_COMMAND='kubectl get no -l "liqonet.liqo.io/gateway=true" -o jsonpath="{.items[0].status.addresses[0].address}"'
-set_variable_from_command GATEWAY_IP GATEWAY_IP_COMMAND "[ERROR]: You have to assign to a node of the cluster the role of gateway. Label it with 'liqonet.liqo.io/gateway=true': kubectl label yournode liqonet.liqo.io/gateway=true"
+#GATEWAY_IP_COMMAND='echo $GATEWAY_IP'
+#set_variable_from_command GATEWAY_IP GATEWAY_IP_COMMAND "[ERROR]: You have to assign to a node of the cluster the role of gateway. Label it with 'liqonet.liqo.io/gateway=true': kubectl label yournode liqonet.liqo.io/gateway=true"
```
giorio94, line 144: Should these lines be deleted?

install.sh
palexster: I think that return expects a value between 0-255 and it will not accept an IPv4 address as argument.

install.sh
palexster: Are you sure of this? I got this output:
```
[PRE-INSTALL]: Setting Gateway IP
[PRE-INSTALL]: GATEWAY_IP is set to: set_gateway_node
[PRE-INSTALL]: POD_CIDR is set to: 10.244.0.0/16
```

install.sh
giorio94: Sorry, my fault. Wrote the comment without actually trying it

install.sh
giorio94: Same as above. No not consider this

Issue #21, opened 2020-06-08
[Epic] Discovery and Peering Fundamentals


palexster:
## Improvements and known issues
- [x] discovery-broadcaster integration (Ref. #18 ): When a new for that `foreign cluster` is discovered, a new broadcaster deployment for that cluster should be started.
The broadcaster needs to be modified in the way it takes the foreign kubeconfig (not from a ConfigMap anymore)
- [x] WAN: Discovery process has to be extended to use DNS protocol
- [x] Retrieve public kubeconfig: Actually public kubeconfig is stored in ConfigMap served by an nginx Deployment, we have to give public access throw API Server
- [x] Advertisement acceptance: At the moment, all Advertisements are automatically accepted. We need a logic (to integrate with systray?) which triggers a notify every time we receive a new Advertisement and allows the user to accept/decline the CR
- [x] Policies: Discuss the implementation of outgoing/incoming policies (customize outgoing Adv, accept incoming Adv...) (Ref. #176)
- [x] A ForeignCluster should be added as reference when receiveing a PeeringCluster, if not previously discovered.
- [x] ForeignCluster resource should provide the status of Ingoing and Outgoing peers of a specific ForeignCluster (Ref. #171)
- [x] So far, the cluster which send an Advertisement has no visibility of the status of remote cluster. We should introduce a feedback loop in ForeignCluster status to notice the provider of advertisement statuses.
- [x] Advertisement complete lifecycle: Delete an Advertisement if the timeToLive has expired.

PR #152, opened 2020-08-07
Added Advertisement status in PeeringRequest

fraborg:
# Description
This PR adds in the `PeeringRequest` object the information about Advertisement acceptance by the foreign cluster.
When the `Advertisement` is created on the foreign cluster, the broadcaster starts watching it: when the foreign cluster accepts/refuses the Advertisement (by setting the field `Status.AdvertisementStatus`), this information is copied in the local `PeeringRequest` object, allowing the home cluster to display it.

## Other changes
- added `AdvertisementStatus` field in PeeringRequest status
- added utility function to create a crdClient for PeeringRequest resources (following the same pattern used for other CRD)
- `WatchAdvertisement` function is now a method of the Broadcaster

# How Has This Been Tested?

- [x] unit test on the new feature:
1. create a `PeeringRequest` on home cluster
2. create an `Advertisement` on foreign cluster
3. set adv status to **ACCEPTED**
4. check the `PeeringRequest` status is set accordingly


palexster:
Ref. #21

Issue #151, opened 2020-08-07
[Improvement] Network configuration support for Unidirectional Peering


alacuku:
# Network Remodeling

## Current Network Model

The network configuration is exchanged between two peering clusters using the `advertisement.protocol.liqo.io` CRD. It was fine before the discovery protocol was implemented. In some cases the `advertisement.protocol.liqo.io` CRD is not symmetrically exchanged between two peering clusters, having a state where one of the two clusters is missing the network configuration of the other cluster. The following pictures depicts the three different states:
![JoinFlow](https://user-images.githubusercontent.com/17413176/89640584-2a13b600-d8b0-11ea-8d2e-c794f5890a3c.png)

### Known Limitations

* Both clusters should exchange their advertisements to successfully establish a network interconnection

## Enhancement Proposal

A possible solution could be to separate the network parameters from the advertisement protocol, and exchange them using a different CRD having in such manner symmetric sharing of the network configuration between two peering clusters.

### Improvements

* Clusters can establish their interconnection without having to exchange their advertisements
* Early detection of network incompatibility

### New Architecture

![DispatcherDetails](https://user-images.githubusercontent.com/17413176/89640496-00f32580-d8b0-11ea-9fe2-c88e471544b9.png)

1. The network configuration of the local cluster is saved in a local CRD (TEP1->2) containing also the clusterID of the remote cluster to whom is destined;
2. This CRD is replicated to the remote cluster by the dispatcher, running in the local cluster who knows how to interact with the remote cluster, and discriminates multiple peering clusters by their clusterID
3. In the remote cluster this CRD (TEP1->2) is processed and its status updated with the NAT information.
4. The Dispatcher (cluster 1) reflects this change in the local resource (TEP1->2)

At this point the same steps are performed by the peering cluster and in Cluster 1 we have two CRDs: TEP1->2, TEP2->1. The first one contains the information of the local cluster sent to the peering cluster in the `spec` section and in the `status` we have the NATing information given by the remote cluster (Cluster2). The second one in its `spec` has the network parameters of the Cluster2. Combining the `status` of TEP1->2 and the `spec` of TEP2->1 we have all the needed information to establish a connection with the remote Cluster2

5. Combining the `status` of TEP1->2 and the `spec` of TEP2->1 we have all the needed information to establish a connection with the remote Cluster2

[...]

PR #162, opened 2020-08-13
Dispatcher

alacuku:
# Description

This is the first version of CRD Dispatcher. It replicates a CRD from the local cluster on remotes clusters. 
The resources to be replicated are identified by their GVR and given to the operator through the `clusterconfig.policy.liqo.io` CRD.
As far as now the Operator does not manages watchers to monitor the resources replicated on the remote clusters. It only monitors the local resources and replicates the following data:

- Metadata: name, namespace and labels
- Spec: all the fields on the spec section
- Status: all the fields on the status section

# How Has This Been Tested?
Unit tests covers the more then 74% of the statements
The remaining statements will be covered by Integration tests with the next PRs.

# Related Issues
#151


scripts/liqonet/Makefile
```diff
@@ -42,30 +42,22 @@ deploy: manifests
...
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) paths="./api/liqonet/v1alpha1" rbac:roleName=manager-role crd:crdVersions=v1beta1 crd:trivialVersions=false output:crd:artifacts:config=config/liqonet/crd/bases
-
-	#rm config/liqonet.liqo.io_networkconfigs.yaml
+	$(CONTROLLER_GEN) $(CRD_OPTIONS)  paths="../../api/liqonet/..." rbac:roleName=manager-role output:crd:artifacts:config=../../config/liqonet/crd/bases output:rbac:artifacts:config=../../config/liqonet/rbac
 
 # Run go fmt against code
 fmt:
 	go fmt ./...
 
+lint:
```
palexster, line 51: I think we can suppose that golangci-lint is in the path...

internal/dispatcher/dispatcher-operator.go
```diff
@@ -0,0 +1,440 @@
...
+	UnregisteredResources []string
+	//for each cluster we save the tuples (registeredResource, chan)
+	RunningWatchers map[string]chan bool
+	Started         bool
+}
+
+func (d *DispatcherReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+
+	d.StartWatchers()
+	d.StopWatchers()
```
palexster, line 43: defer?

internal/dispatcher/dispatcher-operator.go
```diff
@@ -0,0 +1,440 @@
...
+			}
+		}
+	}
+	//if we come here it means that we have to create the resource on the remote cluster
+	spec, b, err := unstructured.NestedMap(obj.Object, "spec")
+	if !b || err != nil {
+		klog.Errorf("an error occurred while processing the 'spec' of the resource %s: %v ", name, err)
+		return err
+	}
+	remRes := &unstructured.Unstructured{
```
palexster, line 159: Do we need some annotations to say that the resource have been reflected? From the following resource, it is not clear to me how to detect which resources have been reflected and which are "native"...

palexster (approved):
/approve

PR #169, opened 2020-08-17
Keepalive mechanism

fraborg:
# Description
In this PR a first implementation of the keepalive mechanism is presented.
There are 3 possibilities:
- **METHOD 1** : create a client to the foreign cluster and try to list the pods every 20 seconds
This approach is the simplest one, but it requires the permissions to list the pods on a given namespace on the foreign cluster.
- **METHOD 2** : create a `CronJob` which executes a `curl -k <url>` every minute
This approach has the limitation given by the Cron format, whose minimum timestamp is 1 minute. Therefore, the convergence time would be 3 times greater.
- **METHOD 3** : perform a `curl -k <url>` every 20 seconds. 
`<url>` is the `healthz` endpoint of the foreign cluster. 
`healthz` endpoint is meant to reply _200 OK_ if the cluster is up and running
Very similar to method 1, the problem with this approach is the `-k` option which is insecure

After 3 failed attempts, the foreign cluster is supposed to be down and its advertisement is deleted.

In the code the first approach is used, the other 2 have been left commented.

## Other changes
- updated `VkReference` field with a description and optional fields

# How Has This Been Tested?

Tested with 2 Kind clusters

### DIFF
--- a/internal/advertisement-operator/controller.go
+++ b/internal/advertisement-operator/controller.go
@@ -141,12 +141,10 @@ func (r *AdvertisementReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 			// if everything works well, the check is an infinite loop
 			// therefore, if an error is returned, the foreign cluster is not reachable anymore
 			if err != nil {
-				// the foreign cluster is down
+				// the foreign cluster is down: set adv status to trigger the unjoin
 				klog.Error(err)
-				//TODO: decide which action to do (delete the adv/peering request...
-				// if we delete the adv we need to set a bool variable for tests
-				err = r.Delete(ctx, &adv, &client.DeleteOptions{})
-				if err != nil {
+				adv.Status.AdvertisementStatus = AdvertisementDeleting
+				if err := r.Status().Update(context.Background(), &adv); err != nil {
 					klog.Error(err)
 				}
 			}
@@ -340,7 +338,6 @@ func (r *AdvertisementReconciler) checkClusterStatus(adv protocolv1.Advertisemen
 		return err
 	}
 
-	// METHOD 1 : kubectl get pods (on foreign cluster)
 	remoteClient, err := protocolv1.CreateAdvertisementClient("", remoteKubeconfig)
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


internal/advertisement-operator/controller.go
```diff
@@ -302,8 +326,129 @@ func (r *AdvertisementReconciler) cleanOldAdvertisements() {
...
+		} else {
+			retry = 0
+		}
+		if retry == 3 {
+			return err
+		}
+		time.Sleep(20 * time.Second)
+	}
+
+	// METHOD 2 : CRONJOB
```
palexster, line 362: Deadcode should be put in a dedicated commit if we want to keep it and then removed.

internal/advertisement-operator/controller.go
```diff
@@ -302,8 +326,129 @@ func (r *AdvertisementReconciler) cleanOldAdvertisements() {
...
+	//			},
+	//		},
+	//	},
+	//}
+	//_, err = r.AdvClient.Client().BatchV1beta1().CronJobs(r.KubeletNamespace).Create(context.Background(), cronJob, metav1.CreateOptions{})
+	//if err != nil {
+	//	return err
+	//}
+
+	// METHOD 3 : curl -k <foreign_cluster_url>
```
palexster, line 399: The same as before

internal/advertisement-operator/controller.go
```diff
@@ -302,8 +326,129 @@ func (r *AdvertisementReconciler) cleanOldAdvertisements() {
...
+	}
+	retry := 0
+	for {
+		_, err = remoteClient.Client().CoreV1().Pods("default").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			retry++
+		} else {
+			retry = 0
+		}
+		if retry == 3 {
```
palexster, line 356: Suggestion: The threshold and sleep amount should be parametrized in clusterConfig, in order to let the user tweak the delay and the threshold.

internal/advertisement-operator/controller.go
```diff
@@ -127,6 +134,23 @@ func (r *AdvertisementReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+		// start the keepalive check for the new cluster
+		r.checkRemoteCluster[adv.Spec.ClusterId] = new(sync.Once)
+		go r.checkRemoteCluster[adv.Spec.ClusterId].Do(func() {
+			err := r.checkClusterStatus(adv)
+			// if everything works well, the check is an infinite loop
+			// therefore, if an error is returned, the foreign cluster is not reachable anymore
+			if err != nil {
+				// the foreign cluster is down
+				klog.Error(err)
+				//TODO: decide which action to do (delete the adv/peering request...
```
palexster, line 146: I suggest to have two different behaviors, similarly as done by the kube-controller manager:
-  ```--node-monitor-grace-period duration```    Default: 40s  Amount  of time which we allow running Node to be unresponsive before marking  it unhealthy. Must be N times more than kubelet's  nodeStatusUpdateFrequency, where N means number of retries allowed for  kubelet to post node status. 
- ```--pod-eviction-timeout duration```     Default: 5m0s | The grace period for deleting pods on failed nodes.

When the first threshold is met --> the virtual node becomes unhealty
When the second threshold is met --> the unjoin is performed

palexster (approved):
/approve

PR #173, opened 2020-08-18
Removing the GatewayPrivateIP parameter from the network module

alacuku:
# Description
This PR removes the GatewayPrivateIP parameter from the network module. The Linux kernel does not strictly require a IP address for the tunnel interface in order to route the traffic through it. By removing this parameter we greatly simplify the configuration of a LIQO cluster.
Now the traffic from hosts to remote pods is NATed in two different ways:
- the local cluster have not been NATed by the remote one: we take the first IP address of the podCIDR of the cluster
- the local cluster have be NATed by the remote one: we take the first IP address from the new subnet

The parameter have been removed from the install script and from the helm chart as well.

The route-operator gets the podCIDR parameters from the` clusterconfig.policy.liqo.io` CRD.

# How Has This Been Tested?
The unit tests of the network module have been updated to not consider the iptables rules and routes involving the GatewayPrivateIP parameter.

End to End tests with two k8s-clusters:
- without podCIDR conflicts, no NAT required
- with podCIDR conflicts, NAT configured

Both pod to pod traffic and node to pod traffic have been tested.

### DIFF
--- a/api/cluster-config/v1/clusterconfig_types.go
+++ b/api/cluster-config/v1/clusterconfig_types.go
@@ -77,9 +77,9 @@ type DiscoveryConfig struct {
 
 type LiqonetConfig struct {
 	//contains a list of reserved subnets in CIDR notation used by the k8s cluster like the podCIDR and ClusterCIDR
-	ReservedSubnets  []string               `json:"reservedSubnets"`
-	GatewayPrivateIP string                 `json:"gatewayPrivateIP"`
-	VxlanNetConfig   liqonet.VxlanNetConfig `json:"vxlanNetConfig,omitempty"`
+	ReservedSubnets []string               `json:"reservedSubnets"`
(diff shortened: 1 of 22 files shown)


palexster:
Hi @alacuku ! Thanks for this! Could you please update the install documentation and the installer script accordingly to this modification?

frisso:
> Hi @alacuku ! Thanks for this! Could you please update the install documentation and the installer script accordingly to this modification?

The install documentation does not mention the gateway already.
The script, instead, needs to be updated.

alacuku:
> Hi @alacuku ! Thanks for this! Could you please update the install documentation and the installer script accordingly to this modification?

done, and updated the helm chart as well.

fraborg (approved):

palexster (approved):
/approve

PR #174, opened 2020-08-18
Fix names of vk resources

fraborg:
# Description

This PR modifies the names of resources related to virtual kubelet in order to be more consistent and allow an easier navigation among them.
Every resource related to virtual-kubelet is now named `vk-<clusterID>`


palexster:
What about liqo + clusterID?

frisso (approved):

frisso:
> What about liqo + clusterID? Instead of vk... As an admin/user, I would rather recognize that it is part of LIQO, since an administrator may not know the internal details about the virtual-kubelet...

Completely agreed.
For the rest, LGTM.

Issue #176, opened 2020-08-19
[Feature] In-going Advertisement Policing


palexster:
**Is your feature request related to a problem? Please describe.**
So far, in-going advertisement are automatically accepted. It would be more effective to have advertisement acceptance configuration, letting the user to accept or refuse the incoming advertisement.

Policies:

- [x] ClusterConfig (Incoming advertisement): (Ref. #181)
    - Autoaccept
    - Manual
    - Auto-deny
- [x] ClusterConfig (Outgoing advertisement) (Done)
    - % cluster (Equal for everybody)
    - Auto-advertise / No-Advertise
    - Auto-join
- [x] Documentation (Configure Liqo)
    - How to set configuration
    - How to accept/deny

palexster:
Ref. #21

PR #181, opened 2020-08-20
Advertisement accept policies

fraborg:
# Description
This PR modifies the `ClusterConfig` CRD to define a policy for the acceptance of Advertisements.
The AdvertisementConfig has been split into 2 main fields:
- **OutgoingConfig** (related to the broadcaster)
```
ResourceSharingPercentage int32
EnableBroadcaster         bool
```
- **IngoingConfig** (related to adv operator)
```
MaxAcceptableAdvertisement int32
AcceptPolicy AcceptPolicy
```

`AcceptPolicy` is an enum which can have the following values
- **_AutoAcceptWithinMaximum_** means all the Advertisement received will be accepted until the MaxAcceptableAdvertisement limit is reached
- **_ManualAccept_** means every Advertisement received will need a manual accept/refuse, which can be done by updating its status (this is not implemented yet)

`AutoAcceptAll `can be achieved by setting MaxAcceptableAdvertisement to infinite
`AutoRefuseAll` can be achieved by setting MaxAcceptableAdvertisement to 0
Further policies can be easily added

The general approach towards config updates has changed too:
- **old** approach : if newMax < oldMax, delete some advertisements until the newMax is respected
- **new** approach : "apply the config from now on": advertisements already accepted are not deleted if the config changes

## Other fixes
- fixed adv operator and peering-request operator deployments in helm chart (the parameters for the broadcaster and virtual-kubelet images were missing)

# How Has This Been Tested?

- [x] moved tests from `test/controller.go` to `test/config-watcher.go` adding Advertisement creation on a test cluster


palexster:
Fix. #176
Fix. #21

palexster (approved):
/approve

mLavacca (approved):

api/cluster-config/v1/clusterconfig_types.go
```diff
@@ -25,26 +25,61 @@ import (
...
+	//EnableBroadcaster flag allows you to enable/disable the broadcasting of your Advertisement to the foreign clusters
+	EnableBroadcaster bool `json:"enableBroadcaster"`
+}
+
+// AcceptPolicy defines the policy to accept/refuse an Advertisement
+type AcceptPolicy string
+
+const (
+	// AutoAcceptWithinMaximum means all the Advertisement received will be accepted until the MaxAcceptableAdvertisement limit is reached
+	// AutoAcceptAll can be achieved by setting MaxAcceptableAdvertisement to infinite
```
frisso, line 64: @fraborg Sorry for coming so late but... what's the value for "infinite" here?

api/cluster-config/v1/clusterconfig_types.go
```diff
@@ -25,26 +25,61 @@ import (
...
+	// AutoRefuseAll can be achieved by setting MaxAcceptableAdvertisement to 0
+	AutoAcceptWithinMaximum AcceptPolicy = "AutoAcceptWithinMaximum"
```
frisso, line 66: @fraborg I would rename it AutoAcceptWithinMaximum ==> MaxAutoAccept

api/cluster-config/v1/clusterconfig_types.go
fraborg: the maximum value of int32 is 2147483647, but we can set the MaxAcceptableAdvertisement range from 0 to a value we decide, which can be our "infinite"

frisso: I would write this in the code (or, better, the definition of the structure), so that users know what they can configure.
Btw, not sure if we're telling users if this number refers to the max number of *contemporary* advertisements accepted, or the total number over time. Just in case, make sure this is clearly explained.

PR #184, opened 2020-08-21
Add dispatcher permission to PeeringRequest related identity

aleoli:
# Description

With this PR ForeignClusterOperator will add the dispatcher ClusterRole to created ServiceAccount assigned to each PeeringRequest created on a remote cluster

For each ServiceAccount a new ClusterRoleBinding is created to the same ClusterRole. This ClusterRole is managed and updated by the cluster config watcher, so when the config is updated also this role will be

### DIFF
--- a/internal/discovery/discovery-config.go
+++ b/internal/discovery/discovery-config.go
@@ -1,9 +1,13 @@
 package discovery
 
 import (
+	"context"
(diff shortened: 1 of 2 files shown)


alacuku (approved):
/lgtm

palexster (approved):
/approve

Issue #153, opened 2020-08-07
[Feature] Uninstall Script


palexster:
**Is your feature request related to a problem? Please describe.**

So far, it is very complex to correctly uninstall all the liqo resources.

**Describe the solution you'd like**

Similarly to install.sh script, we should have a uninstall.sh script to safely remove all liqo objects.

**Describe alternatives you've considered**

Liqo uninstall is uncomfortable and requires to have helminstalled and liqo repository downloaded.

Steps:

- [x] Implement Uninstall function (Ref. #185)
- [x] Linting Shellcheck (Ref. #229)
- [x] Sleep deletion + Auto-crt approval (Ref. #232)
- [ ] Default mode: Install the latest "stable" release
- [x] Checkout tagged releases

PR #185, opened 2020-08-21
Uninstall script

aleoli:
# Description

This PR adds an uninstall script that clean all resources created by Liqo

Ref. #153

### DIFF
--- a/docs/pages/User/GettingStarted/uninstall.md
+++ b/docs/pages/User/GettingStarted/uninstall.md
@@ -1,0 +1,26 @@
+---
+title: Uninstall Liqo
+weight: 4
+---
+
+## Uninstall steps
+
+This procedure uninstalls Liqo on your cluster.
+
+```bash
+curl https://raw.githubusercontent.com/LiqoTech/liqo/master/install.sh | bash -s -- --uninstall
+```
+
+_NOTE:_ all LIQO resources (i.e. CRDs) will not be purged, so you will not lose your discovered clusters. If you want to delete these resources after uninstallation, call the same script with `--deleteCrd` flag.
+
+### Purge LIQO data
+
+If you want that the installed CRD will be completely purged, add `--deleteCrd` flag to the script.
+
+```bash
+curl https://raw.githubusercontent.com/LiqoTech/liqo/master/install.sh | bash -s -- --uninstall --deleteCrd
+```
+
+### What happens to my deployed applications?
+
+During uninstall script your cluster does a de-peering from each peered cluster, hence giving up to the foreign used resources. Hence, your offloaded applications will be rescheduled on your local cluster: you will see them running locally in a few minutes.
--- a/install.sh
+++ b/install.sh
@@ -4,6 +4,9 @@ set -e
 function cleanup()
 {
    set +e
+   if [ -n "$TMPDIR" ]; then
+     rm -rf "$TMPDIR"
+   fi
 }
 
 function set_variable_from_command() {
@@ -29,10 +32,13 @@ function print_help()
    echo "Arguments:"
    echo "   --help: print this help"
    echo "   --test: source file only for testing"
+   echo "   -u, --uninstall: trigger uninstall workflow to remove LIQO from your cluster"
+   echo "   --deleteCrd: remove all CRDs installed by LIQO"
    echo "This script is designed to install LIQO on your cluster. This script is configurable via environment variables:"
    echo "   POD_CIDR: the POD CIDR of your cluster (e.g.; 10.0.0.0/16). The script will try to detect it, but you can override this by having this variable already set"
-   echo "   SERVICE_CIDR: the POD CIDR of your cluster (e.g.; 10.96.0.0/12) . The script will try to detect it, but you can override thisthis by having this variable already set"
+   echo "   SERVICE_CIDR: the POD CIDR of your cluster (e.g.; 10.96.0.0/12) . The script will try to detect it, but you can override this by having this variable already set"
    echo "   GATEWAY_IP: the public IP that will be used by LIQO to establish the interconnection with other clusters"
+   echo "   NAMESPACE: the namespace where LIQO control plane resources will be created"
 }
 
 function wait_and_approve_csr(){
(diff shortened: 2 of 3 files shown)


docs/pages/User/GettingStarted/install.md
```diff
@@ -55,3 +55,19 @@ Therefore you may need to repeat the above procedure on another cluster in order
 
 Once you have two clusters ready, you can start the peering procedure, which is presented in the [next step](../peer).
 
+## Uninstall
```
frisso, line 58: @palexster I would suggest to create a new, dedicated page for the uninstall. Your opinion?

docs/pages/User/GettingStarted/install.md
```diff
@@ -55,3 +55,19 @@ Therefore you may need to repeat the above procedure on another cluster in order
...
+```
+
+### Delete CRDs
+
+If you want that the installed CRD will be completely purged, export `DELETE_CRD` environment variable before to call the script.
+
+```bash
+export DELETE_CRD=1
+curl https://raw.githubusercontent.com/LiqoTech/liqo/master/uninstall.sh | bash
+```
```
frisso, line 73: I would mention what happens about the testing applications that you installed on your cluster: are those still available, are automatically removed, or what?

docs/pages/User/GettingStarted/install.md
palexster: Totally agree, otherwise it would break the tutorial "flow".

uninstall.sh
```diff
@@ -0,0 +1,116 @@
...
+sleep 30
+kubectl delete ns $NAMESPACE
+
+kubectl delete MutatingWebhookConfiguration mutatepodtoleration
+kubectl delete ValidatingWebhookConfiguration peering-request-operator
+
+if [ -n "$DELETE_CRD" ]; then
+  echo "delete CRD"
+  clone_repo
+  kubectl delete -f "$TMPDIR"/liqo/deployments/liqo_chart/crds
```
giorio94, line 112: Same as above

uninstall.sh
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function print_help() {
+   echo "Arguments:"
+   echo "   --help: print this help"
+   echo "   --test: source file only for testing"
+}
+
+function unjoin() {
+  kubectl get clusterconfig configuration -o yaml | sed "s/enableBroadcaster: true/enableBroadcaster: false/g" | kubectl replace -f -
```
giorio94, line 33: Two comments here:
1. I would improve the error handling (i.e. is the `clusterconfig` guaranteed to exist named that way in any possible situation?)
2. I would replace this with `kubectl patch` if possible

uninstall.sh
```diff
@@ -0,0 +1,116 @@
...
+  sleep 30
+
+  for id in $(kubectl get foreignclusters -o jsonpath="{.items[*].metadata.name}"); do
+    kubectl get foreignclusters $id -o yaml | sed "s/join: true/join: false/g" | kubectl replace -f -
```
giorio94, line 37: Same as above

uninstall.sh
```diff
@@ -0,0 +1,116 @@
...
+
+function print_help() {
+   echo "Arguments:"
+   echo "   --help: print this help"
+   echo "   --test: source file only for testing"
+}
+
+function unjoin() {
+  kubectl get clusterconfig configuration -o yaml | sed "s/enableBroadcaster: true/enableBroadcaster: false/g" | kubectl replace -f -
+  sleep 30
```
giorio94, line 34: What is this sleep for?

giorio94 (commented):
Thanks @aleoli for this PR.

I left you some comments inline. In addition, I would make sure the entire script works also in case of unexpected situations (what would it happen if one of the delete commands failed - e.g. since one of the resources does not exist or was not installed correctly).

As a more general comment, there is a fair amount of code replicated from the install script, which may lead to inconsistencies in case of future modifications. However, I have not yet found an ideal solution. Possible alternatives that came to my mind:
* Keep the current solution
* Merge the two scripts in only one controlled by a parameter (I believe this one may be the best)
* Move the common parts to a common script sourced by the others (but needs to be downloaded, since these scripts are used as one-liners)
@palexster WDYT?

uninstall.sh
```diff
@@ -0,0 +1,116 @@
...
+echo "[PRE-UNINSTALL]: Unpeering clusters"
+unjoin
+
+$TMPDIR/bin/helm uninstall liqo -n $NAMESPACE
+echo "[UNINSTALL]: Uninstalling LIQO on your cluster..."
+sleep 30
+kubectl delete ns $NAMESPACE
+
+kubectl delete MutatingWebhookConfiguration mutatepodtoleration
+kubectl delete ValidatingWebhookConfiguration peering-request-operator
```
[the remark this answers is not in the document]
aleoli: No, because they are created by init containers and helm does not know their existence

uninstall.sh
aleoli: Helm handles only crd installation, but nor update neither delete of them

uninstall.sh
aleoli: Liqo discovery components need maximum 30 seconds to reconcile status across clusters, to have a safe delete of created components/applications

install.sh
```diff
@@ -153,13 +153,13 @@ set_variable_from_command LIQO_VERSION LIQO_VERSION_COMMAND "[ERROR]: Error sett
 #Wait for the installation to complete
 kubectl create ns $NAMESPACE
 $TMPDIR/bin/helm dependency update $TMPDIR/liqo/deployments/liqo_chart
-$TMPDIR/bin/helm install liqo -n liqo $TMPDIR/liqo/deployments/liqo_chart --set podCIDR=$POD_CIDR --set serviceCIDR=$SERVICE_CIDR \
+$TMPDIR/bin/helm install liqo -n $NAMESPACE $TMPDIR/liqo/deployments/liqo_chart --set podCIDR=$POD_CIDR --set serviceCIDR=$SERVICE_CIDR \
 --set gatewayPrivateIP=$GATEWAY_PRIVATE_IP --set gatewayIP=$GATEWAY_IP --set global.suffix="$LIQO_SUFFIX" --set global.version="$LIQO_VERSION"
 echo "[INSTALL]: Installing LIQO on your cluster..."
```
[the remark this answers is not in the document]
palexster: Yeah, it is a short sleep to avoid stressing the API Server before containers are created.

uninstall.sh
palexster: This is a quite optimistic assumption, since some component may be stuck and not able to properly reconcile its state. May be we can check that the broadcaster have been really shutoff?

uninstall.sh
```diff
@@ -0,0 +1,116 @@
...
+done
+
+TMPDIR=$(mktemp -d)
+mkdir -p $TMPDIR/bin/
+echo "[PRE-UNINSTALL] [HELM] Checking HELM installation..."
+echo "[PRE-UNINSTALL] [HELM]: Downloading Helm $HELM_VERSION"
+curl --fail -L ${HELM_URL} | tar zxf - --directory="$TMPDIR/bin/" --wildcards '*/helm' --strip 1
+
+NAMESPACE_COMMAND="echo $NAMESPACE_DEFAULT"
+set_variable_from_command NAMESPACE NAMESPACE_COMMAND "[ERROR]: Error while creating the namespace... "
```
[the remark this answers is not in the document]
palexster: Nice.

palexster:
> Thanks @aleoli for this PR.
> 
> I left you some comments inline. In addition, I would make sure the entire script works also in case of unexpected situations (what would it happen if one of the delete commands failed - e.g. since one of the resources does not exist or was not installed correctly).
> 
> As a more general comment, there is a fair amount of code replicated from the install script, which may lead to inconsistencies in case of future modifications. However, I have not yet found an ideal solution. Possible alternatives that came to my mind:
> 
>     * Keep the current solution
> 
>     * Merge the two scripts in only one controlled by a parameter (I believe this one may be the best)
> 
>     * Move the common parts to a common script sourced by the others (but needs to be downloaded, since these scripts are used as one-liners)
>       @palexster WDYT?

I also think that the second option is the best.

docs/pages/User/GettingStarted/uninstall.md
```diff
@@ -0,0 +1,27 @@
...
+title: Uninstall Liqo
+weight: 4
+---
+
+## Uninstall steps
+
+This procedure uninstalls Liqo on your cluster.
+
+```bash
+export UNINSTALL=1
```
frisso, line 11: I feel "export" variables are a pain and I suggest to avoid them.
I would prefer to rename the script and call it with an additional parameter:
install.sh ==> liqo-setup.sh
==> liqo-setup.sh install (to install everything)
==> liqo-setup.sh remove (to remove everything)
@palexster Your opinion?

palexster: Concerning the uninstall script, I totally agree since we are asking for a complete different behaviour that should be made explicit as much as possible. We can still preserve the one-liner and no-clone policy by using an approach like [that](https://stackoverflow.com/questions/4642915/passing-parameters-to-bash-when-executing-a-script-fetched-by-curl). I would suggest something like:

```
curl https://raw.githubusercontent.com/LiqoTech/liqo/master/liqo-setup.sh | bash -s -- unistall
```

For this PR, I suggest to make install.sh with an "option" interface. Due to the implications of renaming to CI and documentation, I would rather do this in a dedicated back-to-back PR.

docs/pages/User/GettingStarted/uninstall.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Uninstall steps
+
+This procedure uninstalls Liqo on your cluster.
+
+```bash
+curl https://raw.githubusercontent.com/LiqoTech/liqo/master/install.sh | bash -s -- --uninstall
+```
+
+### Delete CRDs
```
frisso, line 14: Can we replace "CRD" with something simpler? Not sure who will test Liqo has full understanding of what CRD means.

docs/pages/User/GettingStarted/uninstall.md
```diff
@@ -0,0 +1,24 @@
...
+
+If you want that the installed CRD will be completely purged, add `--deleteCrd` flag to the script.
+
+```bash
+curl https://raw.githubusercontent.com/LiqoTech/liqo/master/install.sh | bash -s -- --uninstall --deleteCrd
+```
+
+### What happens to my deployed applications?
+
+During uninstall script your cluster does an unpeering from each peered cluster, at this point your offloaded applications will be rescheduled on your local cluster, and, after uninstallation, you will see them running there.
```
frisso, line 24:
```suggestion
During uninstall script your cluster does a de-peering from each peered cluster, hence giving up to the foreign used resources. Hence, your offloaded applications will be rescheduled on your local cluster: you will see them running locally in a few minutes.
```

docs/pages/User/GettingStarted/uninstall.md
aleoli: What about something like "Delete all stored resources"?

palexster: Maybe something like "Purging data"?

frisso: "Purge LIQO data" looks definitely simpler.
However, wondering why this has to be chosen by the user. I assume, by default, that if I want to uninstall LIQO, data should be removed. I don't see reason for keeping Liqo data alive.

palexster (approved):

giorio94 (approved):
/approve

palexster (approved):

PR #190, opened 2020-08-25
Fix LAN discovery

### LINKED ISSUE #194: LAN discovery does not create new ForeignClusters
**Describe the bug**
When the two peering hosts have at least one overlapping IP, mDNS packets are received but ForeignClusters are not created

**To Reproduce**
1. Add an interface with the same IP on both peering hosts (in different networks)
2. Start LAN discovery
3. Using your favorite network capture tool check that mDNS packets are coming
4. Check that no new ForeignCluster has been created

**Expected behavior**
A new ForeignCluster should appear

aleoli:
# Description

This PR fixes a problem in LAN discovery when hosts have some overlapping IPs. This PR changes the logic of how a cluster is considered local or remote. Previously if there was at least an equal IP on both, they were considered the same. Now if they have at least a different IP they are considered different hosts

Fixes #194

### DIFF
--- a/internal/discovery/register.go
+++ b/internal/discovery/register.go
@@ -10,7 +10,12 @@ import (
 
 func (discovery *DiscoveryCtrl) Register() {
 	if discovery.Config.EnableAdvertisement {
-		txt, err := discovery.GetTxtData().Encode()
+		txtData, err := discovery.GetTxtData()
+		if err != nil {
+			klog.Error(err)
+			return
+		}
+		txt, err := txtData.Encode()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


palexster:
hi @aleoli! thank you for this! Could you please fill an issue or reference it if already existing to track this bug and its solution?

aleoli:
> hi @aleoli! thank you for this! Could you please fill an issue or reference it if already existing to track this bug and its solution?

Updated

palexster (approved):
/approve

fraborg (approved):

PR #211, opened 2020-08-31
Scheduling node operator fix

mLavacca:
# Description

This PR introduces a fix to a problem caused by the scope change (namespaced to clustered) of the Advertisement CRD. The new scope of the Advertisement is taken into account and SchedulingNode resources that coherently reflect the nodes of the cluster (both the physical and the virtual ones) are created.


internal/schedulingNodeOperator/schedulingNodeFromNode.go
```diff
@@ -75,18 +74,10 @@ func (r *SchedulingNodeReconciler) createSchedulingNode(ctx context.Context, nod
 func (r *SchedulingNodeReconciler) setNeighborsFromAdv(sn *v1alpha1.SchedulingNode, ctx context.Context, node corev1.Node) error {
 	var adv advv1.Advertisement
 
-	namespacedNodeName := types.NamespacedName{
-		Name:      strings.Join([]string{"advertisement", node.Annotations["cluster-id"]}, "-"),
-		Namespace: "default",
-	}
+	advname := strings.Replace(node.Labels["kubernetes.io/hostname"], "vk-", "", 1)
```
fraborg, line 77: I don't know if this affects the operator, but the names of the virtual nodes have been changed to `liqo-<clusterID>`

fraborg (approved):

internal/schedulingNodeOperator/schedulingNodeFromNode.go
mLavacca: It may be, what is the naming for the advertisement?

palexster (approved):

build/scheduling-node-operator/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.13 as builder
```
fraborg, line 1: in the other Dockerfiles we updated to 1.14

mLavacca: updated, thanks!

PR #213, opened 2020-08-31
Update dashboard subchart and installer

nappozord:
# Description

This PR updates the installer of Liqo in regard of the last changes on the dashboard.
The parameter apiServerURL of the helm chart (as well as the installer variable that set it) has been deleted because no longer needed.


palexster (approved):

alacuku (approved):

deployments/liqo_chart/subcharts/liqodash_chart/values.yaml
```diff
@@ -4,7 +4,7 @@
 
 image:
   repository: liqo/dashboard
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
frisso, line 7: Is there any reason we do always "pull"?

nappozord: I guess not, it was just some testing leftover. Thank you for pointing that out!

PR #214, opened 2020-09-01
[tray-agent]: introduction of CRDController for the AgentController

AbakusW:
# Description

This PR updates and standardizes the way *Liqo Agent* manages connections with k8s clusters, with the introduction of a **CRDController** for each CRD.

Each CRDController both manages the client operations for a specific CRD and a cache, reducing the requests to the APIserver and providing a simple way to handle events.

# How Has This Been Tested?

Unit tests


deployments/liqo_chart/values.yaml
```diff
@@ -88,6 +88,9 @@ liqodash_chart:
     repository: "liqo/dashboard"
     pullpolicy: "Always"
   enabled: true
+  service: "liqo-dashboard"
```
fraborg, line 91: since you put these values for the clusterConfig, I would modify also the liqodash subchart (e.g. in the ServiceAccount template the name to use this value instead of the hard-coded one we now have, th same for the other values)

internal/tray-agent/agent/client/agent-config.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	confL := objL.(*clusterConfig.ClusterConfigList)
+	if len(confL.Items) < 1 {
+		return nil, errors.New("no ClusterConfig is present")
+	}
+	return &(confL.Items[0]), nil
+}
+
+//createClusterConfig creates a mocked ClusterConfig CR.
+func createClusterConfig() (*clusterConfig.ClusterConfig, error) {
```
fraborg, line 76: I would put this in test files

fraborg (commented):
I haven't understood why you needed a crdController and couldn't use the normal crdClient, could you explain me a bit? Thank you 😄

internal/tray-agent/agent/client/agent-config.go
fraborg: not essential, it would require many modifications

deployments/liqo_chart/values.yaml
```diff
@@ -99,3 +103,5 @@ global:
   configmapName: "liqo-configmap"
   suffix: ""
   version: ""
+  liqodash:
```
fraborg, line 106: I think this is not needed

AbakusW: yes this line had to be removed with the last change. I'll do it now.

fraborg (approved):

palexster (approved):

PR #218, opened 2020-09-01
Add support for network interconnection with uniderectional peering

alacuku:
# Description

The network parameters are now exchanged between two peering clusters using the `networkconfig` CRD. The tunnelEndpointCreator-operator monitors the presence of `peeringRequests` and `advertisements` CRDs. When one of them is created a new instance of `networkconfig` CRD is created for the peering cluster. When a local cluster has already a `networkconfig` created for a remote peering cluster and receives a `networkconfig` from the same remote peering cluster then
the operator creates a `tunnelendpoint` CR. At this point all the network parameters to set up a point to point connection with the remote peering cluster are available.

## VirtualKubelet update
The virtual kukelet has been modified in order to take the podCIDR from the `TunnelEndpoint` resource and not from the `Advertisement.`
The virtualNodeUpdater now watches both Advertisement and TunnelEndpoint.
When the virtual node needs to be updated, we check if the podCIDR (coming from the `TunnelEndpoint`) has already been set: if not, the node status is set to `NetworkUnavailable`. When both the podCIDR and the resources from Advertisement are set, the node status becomes `Ready`.

# How Has This Been Tested?

Integration tests to check the:

- the correct processing of a received `networkconfig` instance
- the correct creation of a `tunnelEndpoint` when we have the local and remote `networkconfig` instances
- the correct update of a `tunnelEndpoint` when an update is done on the local and remote `network` instances

Fix. #151


api/config/v1alpha1/clusterconfig_types.go
```diff
@@ -114,6 +114,7 @@ type LiqonetConfig struct {
 	//contains a list of reserved subnets in CIDR notation used by the k8s cluster like the podCIDR and ClusterCIDR
 	ReservedSubnets []string               `json:"reservedSubnets"`
```
frisso, line 115: @alacuku Please add the proper description of the members of all the resources, otherwise we may not be able to understand what these are about.

frisso:
Shouldn't we get rid of "liqonet.liqo.io" and rename it as "net.liqo.io"?

alacuku:
> Shouldn't we get rid of "liqonet.liqo.io" and rename it as "net.liqo.io"?

the refactoring of the API has be done on the PR #221. once it is merged a rebase will apply those changes to this PR.

palexster:
Ref. #151

test/unit/liqonet/env_test.go
```diff
@@ -54,12 +52,16 @@ func TestMain(m *testing.M) {
...
 	}
-	adv := getAdv()
-	err = tunEndpointCreator.Create(ctx, adv)
+	err = setupTunnelEndpointCreatorOperator()
 	if err != nil {
-		klog.Error(err, err.Error())
-		os.Exit(-2)
+		os.Exit(-1)
 	}
+	/*	adv := getAdv()
```
fraborg, line 59: is this needed?

fraborg (approved):

internal/liqonet/tunnelEndpointCreator-operator.go
```diff
@@ -32,26 +40,49 @@ import (
...
+	Mutex              sync.Mutex
+	IsConfigured       bool
+	Configured         chan bool
+	AdvWatcher         chan bool
+	PReqWatcher        chan bool
+	RunningWatchers    bool
+	RetryTimeout       time.Duration
 }
 
 // +kubebuilder:rbac:groups=sharing.liqo.io,resources=advertisements,verbs=get;list;watch;create;update;patch;delete
```
fraborg, line 88: I think this is not needed anymore

cmd/liqonet/main.go
```diff
@@ -49,14 +53,25 @@ var (
...
+	}
 )
 
 func init() {
 	_ = clientgoscheme.AddToScheme(scheme)
 
 	_ = netv1alpha1.AddToScheme(scheme)
 
-	_ = advtypes.AddToScheme(scheme)
+	_ = netv1alpha1.AddToScheme(scheme)
```
fraborg, line 74: this can be deleted, an addition from a rebase I think

test/unit/liqonet/env_test.go
```diff
@@ -54,12 +51,16 @@ func TestMain(m *testing.M) {
...
+		err = tunEndpointCreator.Create(ctx, adv)
```
palexster: Remove commented code

internal/liqonet/tunnelEndpointCreator-config.go
```diff
@@ -41,6 +41,19 @@ func (r *TunnelEndpointCreator) GetConfiguration(config *configv1alpha1.ClusterC
 	return reservedSubnets, nil
 }
 
+func (r *TunnelEndpointCreator) SetNetParameters(config *configv1alpha1.ClusterConfig) {
+	podCIDR := config.Spec.LiqonetConfig.PodCIDR
+	serviceCIDR := config.Spec.LiqonetConfig.ServiceCIDR
+	if r.PodCIDR != podCIDR {
```
palexster, line 47: Why do you need those ifs? For logging?

palexster (approved):
/approve

PR #219, opened 2020-09-02
Fix init-vk permissions

fraborg:
# Description
This PR fixes the permissions on the virtual-kubelet init container script (which caused the e2e test to fail): we had no permission to execute, therefore the init container of the virtual-kubelet failed.
The permissions on `scripts/virtual-kubelet/script.sh` have been changed from 644 to 755
The code has also been formatted according to gofmt.


mLavacca (commented):
Hi @fraborg, it looks like there no changes in `scripts/virtual-kubelet/script.sh`, even if the file is in the "Files change" list.

fraborg:
> Hi @fraborg, it looks like there are no changes in `scripts/virtual-kubelet/script.sh`, even if the file is in the "Files change" list.

Yeah, that's exactly the problem we had in the script and I hadn't notice: the permissions where 644, so no permission to execute the script. I changed them to 755

mLavacca:
> > Hi @fraborg, it looks like there are no changes in `scripts/virtual-kubelet/script.sh`, even if the file is in the "Files change" list.
> 
> Yeah, that's exactly the problem we had in the script and I hadn't notice: the permissions where 644, so no permission to execute the script. I changed them to 755.
> Thanks, I changed the description that wasn't very clear

I don't think that changing the permission in your localhost and committing the file, the modification is reflected in the repo. Shouldn't we set the correct permission in the proper [dockerfile](https://github.com/LiqoTech/liqo/blob/master/build/init-vkubelet/Dockerfile)?

fraborg:
> I don't think that changing the permission in your localhost and committing the file, the modification is reflected in the repo. Shouldn't we set the correct permission in the proper dockerfile?

Actually I have no idea, but now the e2e test works fine

internal/advertisement-operator/controller.go
```diff
@@ -18,8 +18,8 @@ package advertisement_operator
 import (
 	"context"
 	goerrors "errors"
-	discoveryv1alpha1 "github.com/liqoTech/liqo/api/discovery/v1alpha1"
 	configv1alpha1 "github.com/liqoTech/liqo/api/config/v1alpha1"
+	discoveryv1alpha1 "github.com/liqoTech/liqo/api/discovery/v1alpha1"
```
aleoli, line 22: why ci linter did tell it to me when I did this change?

fraborg: I think when you merged your PR that import was "policyv1" and this occurred when we modified the ClusterConfig version

aleoli (approved):

mLavacca (approved):
/approve

PR #222, opened 2020-09-03
VirtualKubelet network update

fraborg:
# Description
In this PR the virtual kukelet has been modified in order to take the podCIDR from the `TunnelEndpoint` resource and not from the `Advertisement.`
The virtualNodeUpdater now watches both Advertisement and TunnelEndpoint.
When the virtual node needs to be updated, we check if the podCIDR (coming from the `TunnelEndpoint`) has already been set: if not, the node status is set to `NetworkUnavailable`. When both the podCIDR and the resources from Advertisement are set, the node status becomes `Ready`.

# Other changes
- removed all Network info from Advertisement

# How Has This Been Tested?

- [x] Test with kind creating manually an Advertisement and TunnelEndpoint anch checking the node status


internal/kubernetes/virtualNodeupdater.go
```diff
@@ -89,45 +104,45 @@ func (p *KubernetesProvider) ReconcileNodeFromAdv(event watch.Event) {
...
-// Initialization of the Virtual kubelet, that implies:
-func (p *KubernetesProvider) initVirtualKubelet(adv advtypes.Advertisement) error {
-	klog.Info("vk initializing")
+func (p *KubernetesProvider) ReconcileNodeFromTep(event watch.Event) {
+	tep, ok := event.Object.(*nettypes.TunnelEndpoint)
+	if !ok {
+		klog.Fatal("error in casting tunnelEndpoint")
+	}
+	if event.Type == watch.Deleted {
+		klog.Infof("tunnelEndpoint %v deleted", tep.Name)
```
palexster, line 119: If there is no TEP, should the node be unreacheable?

fraborg: good point, thank you!

PR #223, opened 2020-09-03
CRDReplicator Enhancement

alacuku:
# Description
The "update" functions used to update a resource now can recover from "Conflicting" when trying to perform an update on the status or spec fields.

Fixing a minor bug which caused the operator to run even if the clusterID configmap is not found

Fixing race conditions on the creation and replication of the local resources on the remote peering clusters. With this fix, given a remote peering cluster, for each type of resource there is a local and remote watcher. The local watcher is created after that a connection to the API server of the remote peering cluster exists.  

# How Has This Been Tested?

- added a test where the spec and status fields of a resource are modified in order to check that the conflicting errors are handled properly

- updated the unit tests on the functions that start and stop the watchers for a given resource.

### DIFF
--- a/cmd/crdReplicator/main.go
+++ b/cmd/crdReplicator/main.go
@@ -49,6 +49,7 @@ func main() {
 	clusterID, err := util.GetClusterID(k8sClient, clusterIDConfMap, namespaceName)
 	if err != nil {
 		klog.Errorf("an error occurred while retrieving the clusterID: %s", err)
+		os.Exit(-1)
 	} else {
 		klog.Infof("setting local clusterID to: %s", clusterID)
 	}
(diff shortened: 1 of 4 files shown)


frisso:
Thanks, Aldo!
Minor question @palexster about the naming: is it really a "CRD" replicator, or "object" replicator? Service replicator? Resource replicator?

palexster (approved):

palexster:
> Thanks, Aldo!
> Minor question @palexster about the naming: is it really a "CRD" replicator, or "object" replicator? Service replicator? Resource replicator?

The replicator is designed to replicate CR for a configurable set of CRD. So far, we assume that CRD are already installed on remote cluster, in the future we would enable CRD replication.

I think it is out of the scope of this PR but we can discuss to rename it future to a more generic ObjectReplicator or ResourceReplicator. I personally would avoid service which has a different meaning in K8s.

frisso:
> 
> The replicator is designed to replicate CR for a configurable set of CRD. So far, we assume that CRD are already installed on remote cluster, in the future we would enable CRD replication.
> 
> I think it is out of the scope of this PR but we can discuss to rename it future to a more generic ObjectReplicator or ResourceReplicator. I personally would avoid service which has a different meaning in K8s.

Perfect. Let's keep the current name for the time being.
We can consider to change it when the code will be used for a broader scope in the future. Thanks!

fraborg (approved):

Issue #230, opened 2020-09-07
[Feature] User documentation review


giorio94:
**Is your feature request related to a problem? Please describe.**

Some comments, suggestions and issues which came to my mind when reading the Liqo user documentation (still WIP):

- [ ] I would move *Liqo in brief* before *Getting started* (at least, it seems more natural to me);
- [ ] The *Peer to a foreign cluster* does not explain how to peer with a foreign cluster (to me, this is a big missing point before being able to try liqo)
- [ ] In the discovery section, it is not clear to me the *Trust Remote Cluster* section and its subsections.

palexster:
Ref. #179

PR #237, opened 2020-09-09
Add ingress section to dashboard documentation

nappozord:
# Description

This PR adds an ingress a new section of the dashboard documentation, under _Accessing LiqoDash_, that explains how to set up an Ingress for the dashboard.

Also, fixes a minor bug in the View CRD, where the object **layout** would not update. This was because said parameter was left unspecified. 
ref. #230

### DIFF
--- a/docs/pages/User/dashboard.md
+++ b/docs/pages/User/dashboard.md
@@ -17,7 +17,7 @@ LiqoDash is installed by default during the installation of Liqo and it is bound
 
 ## Accessing LiqoDash
 
-### Node Port
+### NodePort
 In order to access the dashboard you need to first get the port on which LiqoDash is exposed, which can be done with the following command:
 ```
 kubectl -n liqo describe service liqo-dashboard | grep NodePort
@@ -30,8 +30,8 @@ NodePort:      https  32421/TCP
 In this case, the dashboard has been exposed to the port ``32421``
 Now, you can access LiqoDash using your master node IP and the service port you just found: ``https://<MASTER_IP>:<LIQODASH_PORT>``
 
-**NOTE: to get your master node IP, you can run ``kubectl get nodes -o wide | grep master``, and take the
-``INTERNAL-IP``**
+**NOTE:** to get your master node IP, you can run ``kubectl get nodes -o wide | grep master``, and take the
+``INTERNAL-IP``
 
 ### Port-Forward
 A simple way to access the dashboard with a simpler URL than the one specified above is to use ``kubectl port-forward``.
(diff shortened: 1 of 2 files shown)


docs/pages/User/dashboard.md
```diff
@@ -44,6 +44,44 @@ https://localhost:6443
...
+installing Liqo, for example:
+```
+export DASHBOARD_INGRESS=example.liqodash.com
+```
+Once installed, you can see that an Ingress resource is created, accessible through:
+```
+kubectl get -n liqo ingress
+```
+Along with a secret that you will have to edit, providing a certificate and key. You can do it
+with the following command:
```
giorio94, line 60: AFAIK, the secret is automatically generated with a self-signed certificate. Hence, no actions are required it this is the behavior desired. Conversely, to have a valid certificate it is either possible to edit that secret (something along the lines of `kubectl replace secret tls ${CERT_NAME} --key ${KEY_FILE} --cert ${CERT_FILE}`) or use ad-hoc tools such as `cert-manager`, which nonetheless require to add an annotation to the ingress. In the latter case, I am not sure what would it happen after an upgrade of Liqo (is the additional annotation deleted by helm?)

docs/pages/User/dashboard.md
```diff
@@ -53,6 +91,9 @@ The first time you access LiqoDash, it is likely that your browser will show you
 that you are trying to access a website that uses a self-signed certificate, and thus it is not trusted
 by default. Simply accept the risk and continue to access the dashboard.
 
+If you are using Ingress, you can either use a certificate provided by a public trusted Certificate Authority (like 
+[Let's encrypt](https://letsencrypt.org/)) or a self-signed one (you can use [OpenSSL](https://www.openssl.org/))
```
giorio94, line 95:
```suggestion
If you are exposing the LiqoDash through an Ingress, you can either use a certificate provided by a public trusted Certificate Authority (like 
[Let's encrypt](https://letsencrypt.org/)) or a self-signed one.
```

giorio94 (commented):
@nappozord I left you a couple of suggestions. As a general comment, I would add a brief introduction after `Accessing LiqoDash` about why there are different alternatives to access the dashboard and when each one should be used.

docs/pages/User/dashboard.md
palexster: My 2 cents: IIRC I think that the secret is not automatically generated. If nginx does not find the secret, it automatically switch to the default one.

giorio94: @palexster Thanks, you are definitely right. I just checked and this is also the current behavior of nginx. However, I feel there is still no need to manually create the secret unless a custom certificate is to be provided. What do you think?

palexster: Totally agree.

palexster (approved):

giorio94 (approved):
/lgtm

PR #238, opened 2020-09-09
Endpoint IP translation

mLavacca:
# Description

This PR fixes a problem in endpoint reflection between clusters having the
same podCIDR. The home IP addresses must be translated from home to foreign
before being remotely reflected.


palexster (approved):

giorio94 (approved):
/lgtm

fraborg (approved):

Issue #246, opened 2020-09-11
The uninstaller hangs forever while deleting the Liqo CRDs


giorio94:
**Describe the bug**

When Liqo is uninstalled using the provided script (with --deleteCrd enabled), the procedure hangs forever. In particular, the CRDs `networkconfigs.net.liqo.io` and `tunnelendpoints.net.liqo.io` fail to be deleted since the corresponding resources are associated to finalizers that do not longer exist (since all pods have already been deleted in the previous step).

**To Reproduce**
1. Install Liqo
2. Establish a peering
3. Uninstall Liqo (with `--deleteCrd`)
4. The deletion process never terminates

**Expected behavior**
The uninstallation should be completed correctly

PR #247, opened 2020-09-11
peering process improvements

aleoli:
# Description

This PR improves the stability of the peering/unjoin process

## Peering/Unjoin
* previously, in some situation the advertisement was deleted instead of setting the delete flag, this was leading to incorrect delete of the virtualkubelet related resources
* now in PeeringRequest status is available the reference to the created broadcaster
  * a PeeringRequest is now owner and controller of broadcaster deployment, so if it is deleted it is immediately recreated
* during PeeringRequest creation the secret name is generated by API Server, avoiding name collisions
  * it's now possible to peer n clusters (tested with 3 kind clusters), we will see n-1 new nodes in each one

## Uninstall - Ref #246
Some networking resources are no more deleted with unjoin after the CrdReplicator addition, these resources are now deleted in the uninstall script before Liqo operators uninstall

### Problem
If we delete all CRD before uninstall, during uninstallation is still possible that a remote CrdReplicator is sending us new replicas with its finalizers, at the moment there is no way to avoid it

### DIFF
--- a/api/discovery/v1alpha1/peeringrequest_types.go
+++ b/api/discovery/v1alpha1/peeringrequest_types.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	advtypes "github.com/liqotech/liqo/api/sharing/v1alpha1"
 	"github.com/liqotech/liqo/pkg/crdClient"
+	object_references "github.com/liqotech/liqo/pkg/object-references"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -53,7 +54,8 @@ type OriginClusterSets struct {
 
 // PeeringRequestStatus defines the observed state of PeeringRequest
 type PeeringRequestStatus struct {
-	AdvertisementStatus advtypes.AdvPhase `json:"advertisementStatus,omitempty"`
+	BroadcasterRef      *object_references.DeploymentReference `json:"broadcasterRef,omitempty"`
+	AdvertisementStatus advtypes.AdvPhase                      `json:"advertisementStatus,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/install.sh
+++ b/install.sh
@@ -63,11 +63,26 @@ function clone_repo() {
   fi
 }
 
+function check_unjoined() {
+  tmp=$(kubectl get foreignclusters -o jsonpath="{.items[*].status.incoming.joined} {.items[*].status.outgoing.joined}")
+  for f in ${tmp}; do
+    if [ "$f" == "true" ]; then
+      # they are not completely unjoined
+      return 0
+    fi
+  done
+  return 1
+}
+
 function unjoin() {
+  for id in $(kubectl get foreignclusters -o jsonpath="{.items[*].metadata.name}"); do
+    kubectl patch foreignclusters "$id" -p '{"spec":{"join":false}}' --type 'merge'
+  done
+
   kubectl patch clusterconfig configuration -p '{"spec":{"advertisementConfig":{"outgoingConfig":{"enableBroadcaster":false}}}}' --type 'merge'
 
   retry=10
-  while [ "$(kubectl get pods -n "$NAMESPACE" | grep -c broadcaster)" -gt 0 ]; do
+  while check_unjoined; do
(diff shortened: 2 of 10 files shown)


api/discovery/v1alpha1/peeringrequest_types.go
```diff
@@ -53,7 +53,8 @@ type OriginClusterSets struct {
 
 // PeeringRequestStatus defines the observed state of PeeringRequest
 type PeeringRequestStatus struct {
-	AdvertisementStatus advtypes.AdvPhase `json:"advertisementStatus,omitempty"`
+	BroadcasterRef      *v1.ObjectReference `json:"broadcasterRef,omitempty"`
```
fraborg, line 56: you can use `DeploymentReference` struct ([here](https://github.com/liqotech/liqo/blob/master/pkg/object-references/types.go)) as suggested in the new version of kubernetes, which discourages the use of `ObjectReference`

install.sh
```diff
@@ -64,26 +64,39 @@ function clone_repo() {
 }
 
 function unjoin() {
+  for id in $(kubectl get foreignclusters -o jsonpath="{.items[*].metadata.name}"); do
+    kubectl patch foreignclusters "$id" -p '{"spec":{"join":false}}' --type 'merge'
+  done
+
+  for id in $(kubectl get advertisements -o jsonpath="{.items[*].metadata.name}"); do
+    kubectl patch advertisements "$id" -p '{"status":{"advertisementStatus":"Deleting"}}' --type 'merge'
+  done
```
palexster, line 73: My 2 cents: This second patch should not be necessary, since the first one should imply the status change of advertisements

install.sh
```diff
@@ -64,26 +64,39 @@ function clone_repo() {
...
-  for id in $(kubectl get foreignclusters -o jsonpath="{.items[*].metadata.name}"); do
-    kubectl patch foreignclusters "$id" -p '{"spec":{"join":false}}' --type 'merge'
-  done
 }
 
 function deleteCrd() {
   echo "delete CRD"
   clone_repo
+
+  # there are only resources that are been replicated during unjoin process, finalizers are useless
```
palexster, line 92: AFAIK, replicated resources should not have finalizers. Correct @alacuku ?

alacuku: A replicated resource in a remote cluster may have finalizers. For example, if an operator that processes a replicated resource need to do some clean-up after the deletion of the resource than it may set the finalizers for this purpose.
If your question is if we replicate also the finalizers then the answer is: no we do not.

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -409,53 +416,62 @@ func (r *ForeignClusterReconciler) createPeeringRequestIfNotExists(clusterID str
...
-							APIVersion: "discovery.liqo.io/v1alpha1",
-							Kind:       "PeeringRequest",
-							Name:       pr.Name,
-							UID:        pr.UID,
-						},
+		}
+		secret := &apiv1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				// generate name will lead to different names every time, avoiding name collisions
+				GenerateName: "pr-" + r.clusterID.GetClusterID() + "-",
```
palexster, line 423: strings.join may be more efficient

internal/peering-request-operator/peering-request-controller.go
```diff
@@ -75,19 +77,27 @@ func (r *PeeringRequestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
 	}
 
-	exists, err := r.BroadcasterExists(pr)
-	if err != nil {
-		klog.Error(err, err.Error())
-		return ctrl.Result{RequeueAfter: r.retryTimeout}, err
+	exists := pr.Status.BroadcasterRef != nil
+	if exists {
+		// check if it really exists
+		exists, err = r.BroadcasterExists(pr)
```
palexster, line 83: If the deployment does not exist, do we erase the reference somewhere?

aleoli: No, the only two ways that the deployment does not exist are:
* the PeeringRequest is just been created, so the reference is not set
* someone has deleted the deployment manually, so we recreate it immediately and we overwrite the reference

palexster (approved):

fraborg (approved):

PR #249, opened 2020-09-11
Add watcher recreation in VirtualKubelet

### LINKED ISSUE #250: Virtual Kubelet crashes after failing casting TunnelEndpoint and Advertisement
# Description
Sometimes the watchers in the VirtualKubelet fail to cast the triggering event into a TunnelEndpoint or Advertisement object.

# Proposed solution
When a casting error occurs, stop the watcher and recreate it; doing so, the watcher should restart working properly.

fraborg:
# Description
This PR adds the recreation of the watchers over `Advertisement` and `TunnelEndpoint` in the VirtualKubelet when a casting error occurs.
It also adds tests on network updates in the VirtualKubelet.

Fixes #250 

## Other changes
Fix ownerReference in the virtual node: it pointed to `liqo-<clusterID>` deployment, whereas the virtualKubelet deployment is named `virtual-kubelet-<clusterID>`

# How Has This Been Tested?

- [x] Tests on VirtualKubelet network update after the creation of a `TunnelEndpoint` with no remapping
- [x] Test VirtualKubelet network update with remapping

### DIFF
--- a/cmd/virtual-kubelet/internal/commands/root/root.go
+++ b/cmd/virtual-kubelet/internal/commands/root/root.go
@@ -16,6 +16,7 @@ package root
 
 import (
 	"context"
+	"github.com/liqotech/liqo/pkg"
 	"k8s.io/klog"
 	"os"
 	"path"
@@ -152,7 +153,7 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
 		leaseClient = client.CoordinationV1beta1().Leases(corev1.NamespaceNodeLease)
 	}
 
-	deployName := strings.Join([]string{"liqo", c.ClusterId}, "-")
+	deployName := strings.Join([]string{pkg.VirtualKubeletPrefix, c.ClusterId}, "")
 	refs := createOwnerReference(client, deployName, c.KubeletNamespace)
 
 	var nodeRunner *node.NodeController
--- a/internal/kubernetes/test/nodeTestCases.go
+++ b/internal/kubernetes/test/nodeTestCases.go
@@ -11,6 +11,50 @@ var (
 	Cpu2, _    = resource.ParseQuantity("2")
 	Memory1, _ = resource.ParseQuantity("1000")
 	Memory2, _ = resource.ParseQuantity("2000")
+	Condition1 = []corev1.NodeCondition{
+		{
+			Type:   corev1.NodeReady,
+			Status: corev1.ConditionFalse,
+		},
+		{
+			Type:   corev1.NodeMemoryPressure,
+			Status: corev1.ConditionFalse,
+		},
+		{
+			Type:   corev1.NodeDiskPressure,
+			Status: corev1.ConditionFalse,
+		},
+		{
+			Type:   corev1.NodePIDPressure,
+			Status: corev1.ConditionFalse,
+		},
+		{
+			Type:   corev1.NodeNetworkUnavailable,
+			Status: corev1.ConditionTrue,
+		},
+	}
+	Condition2 = []corev1.NodeCondition{
+		{
+			Type:   corev1.NodeReady,
+			Status: corev1.ConditionTrue,
+		},
+		{
+			Type:   corev1.NodeMemoryPressure,
+			Status: corev1.ConditionFalse,
+		},
+		{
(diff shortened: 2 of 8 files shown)


alacuku (approved):

internal/kubernetes/virtualNodeupdater.go
```diff
@@ -42,9 +44,31 @@ func (p *KubernetesProvider) StartNodeUpdater(nodeRunner *node.NodeController) (
 		for {
 			select {
 			case ev := <-advWatcher.ResultChan():
-				p.ReconcileNodeFromAdv(ev)
+				err = p.ReconcileNodeFromAdv(ev)
+				if err != nil {
+					klog.Error(err)
+					advWatcher.Stop()
```
mLavacca, line 50: Did you verify that this `Stop()` statement doesn't crash? Because I think that it could cause panic caused by a "closing of already closed channel" error. In that case I would add a `recover()` statement ([example](https://play.golang.org/p/LqfFbAV0cg1))

internal/kubernetes/test/nodeTestCases.go
```diff
@@ -11,6 +11,50 @@ var (
 	Cpu2, _    = resource.ParseQuantity("2")
 	Memory1, _ = resource.ParseQuantity("1000")
 	Memory2, _ = resource.ParseQuantity("2000")
+	Condition1 = []corev1.NodeCondition{
```
mLavacca, line 14: Since we address this variable also in virtual kubelet code, I would move it to another file (not testing)

https://github.com/liqotech/liqo/blob/ce42ff0d6b739d034dbccbd23c086746d83a557e/internal/kubernetes/virtualNodeupdater.go#L193

cmd/virtual-kubelet/internal/commands/root/root.go
```diff
@@ -152,7 +153,7 @@ func runRootCommand(ctx context.Context, s *provider.Store, c Opts) error {
 		leaseClient = client.CoordinationV1beta1().Leases(corev1.NamespaceNodeLease)
 	}
 
-	deployName := strings.Join([]string{"liqo", c.ClusterId}, "-")
+	deployName := strings.Join([]string{advop.VirtualKubeletPrefix, c.ClusterId}, "")
```
mLavacca, line 156: I think that the advertisement-operator internals is the wrong place in which to put this const. Can we move it to a better place? (Maybe a shared pkg file)

mLavacca (changes requested):

internal/kubernetes/virtualNodeupdater.go
fraborg: Good point, thank you! I think it would be a very strange behaviour, since we are in the switch case where the channel received an event (so it was open) and before the `Stop()` statement nobody closes it. 
However I think that if that happens we don't have a way to recover the error and could only log a fatal error; do you have a possible recovery?

internal/kubernetes/test/nodeTestCases.go
fraborg: I agree with you, I reused that variable just to avoid some lines of code but it's better not to use test files in the normal code

internal/kubernetes/virtualNodeupdater.go
mLavacca: Since we are re-opening the channel due to its close from the API server, I think that it is very likely to happen. maybe we could implement a function that re-open the chan and call it whenever the recover routine is triggered (just an idea based on the example that I linked in the comment above).

fraborg: I have written a recover function but while debugging I found out the `Stop()` function already checks if the channel is closed ([doc](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/watch/streamwatcher.go#L81)), so I think we don't need the recover; however, if it can be useful I commit it ( it's very simple, it siply restart the nodeUpdater)

mLavacca: If `Stop()` already handles this error, I'm perfectly fine with your solution!

mLavacca (approved):

PR #251, opened 2020-09-11
Fix TunnelEndpointCreator broken watchers on CRD

alacuku:
# Description

The operator used watchers to monitor the creation of `peeringrequests.discovery.liqo.io` and `advertisements.sharing.liqo.io` and eventually create a `networkconfigs.net.liqo.io` to send to the peering cluster. Watchers may miss events (e.g. network problems), to avoid this scenario Informers are used instead of the more primitive watcher construct.

Fixes #(issue)

# How Has This Been Tested?
Old tests are still valid, changed the way to do things but the outcome is the same.

### DIFF
--- a/cmd/liqonet/main.go
+++ b/cmd/liqonet/main.go
@@ -20,19 +20,18 @@ import (
 	"flag"
 	"github.com/coreos/go-iptables/iptables"
 	clusterConfig "github.com/liqotech/liqo/api/config/v1alpha1"
-	discoveryv1alpha1 "github.com/liqotech/liqo/api/discovery/v1alpha1"
 	netv1alpha1 "github.com/liqotech/liqo/api/net/v1alpha1"
-	advtypes "github.com/liqotech/liqo/api/sharing/v1alpha1"
 	"github.com/liqotech/liqo/internal/liqonet"
 	"github.com/liqotech/liqo/pkg/liqonet"
 	"github.com/vishvananda/netlink"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"net"
 	"os"
@@ -52,17 +51,6 @@ var (
 		DeviceName: "liqonet",
 		Port:       "4789", //IANA assigned
 		Vni:        "200",
-	}
-
-	peeringReqGVR = schema.GroupVersionResource{
-		Group:    discoveryv1alpha1.GroupVersion.Group,
-		Version:  discoveryv1alpha1.GroupVersion.Version,
-		Resource: "peeringrequests",
-	}
-	advGVR = schema.GroupVersionResource{
-		Group:    advtypes.GroupVersion.Group,
-		Version:  advtypes.GroupVersion.Version,
-		Resource: "advertisements",
 	}
 )
 
@@ -148,7 +136,7 @@ func main() {
 			setupLog.Error(err, "unable to initialize iptables: %v. check if the ipatable are present in the system", err)
 		}
 
-		r := &controllers.RouteController{
+		r := &liqonetOperators.RouteController{
 			Client:                             mgr.GetClient(),
 			Log:                                ctrl.Log.WithName("route-operator"),
 			Scheme:                             mgr.GetScheme(),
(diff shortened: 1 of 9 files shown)


cmd/liqonet/main.go
```diff
@@ -236,18 +230,25 @@ func main() {
...
-		go r.Watcher(r.DynClient, peeringReqGVR, r.PeeringRequestHandler, r.PReqWatcher)
-		go r.Watcher(r.DynClient, advGVR, r.AdvertisementHandler, r.AdvWatcher)
+		//starting the watchers
+		go r.Watcher(r.DynFactory, liqonetOperators.AdvGVR, cache.ResourceEventHandlerFuncs{
+			AddFunc:    r.AdvertisementHandlerAdd,
+			UpdateFunc: r.AdvertisementHandlerUpdate,
+		}, r.AdvStartWatcher, r.AdvStopWatcher)
+		go r.Watcher(r.DynFactory, liqonetOperators.PeeringReqGVR, cache.ResourceEventHandlerFuncs{
+			AddFunc:    r.PeeringRequestHandlerAdd,
+			UpdateFunc: r.PeeringRequestHandlerUpdate,
```
palexster, line 240: Don't we need to react also to delete events?

palexster (approved):

internal/liqonet/tunnelEndpointCreator-operator.go
```diff
@@ -444,26 +467,46 @@ func (r *TunnelEndpointCreator) CreateTunnelEndpoint(param networkParam) error {
...
 	adv := &advtypes.Advertisement{}
-	err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, adv)
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(objUnstruct.Object, adv)
 	if err != nil {
-		klog.Errorf("an error occurred while converting resource %s of type %s to typed object: %s", obj.GetName(), obj.GetKind(), err)
+		klog.Errorf("an error occurred while converting resource %s of type %s to typed object: %s", objUnstruct.GetName(), objUnstruct.GetKind(), err)
 		return
 	}
+	klog.Infof("processing %s", adv.Name)
 	_ = r.createNetConfig(adv.Spec.ClusterId)
```
fraborg, line 483: why do you return an error if it's not used?

internal/liqonet/tunnelEndpointCreator-operator.go
```diff
@@ -444,26 +467,46 @@ func (r *TunnelEndpointCreator) CreateTunnelEndpoint(param networkParam) error {
...
-		klog.Errorf("an error occurred while converting resource %s of type %s to typed object: %s", obj.GetName(), obj.GetKind(), err)
+		klog.Errorf("an error occurred while converting resource %s of type %s to typed object: %s", objUnstruct.GetName(), objUnstruct.GetKind(), err)
 		return
 	}
+	klog.Infof("processing %s", peeringReq.Name)
 	_ = r.createNetConfig(peeringReq.Spec.ClusterID)
 }
 
+func (r *TunnelEndpointCreator) PeeringRequestHandlerUpdate(oldObj interface{}, newObj interface{}) {
+	r.PeeringRequestHandlerAdd(newObj)
```
fraborg, line 507: same as Advertisement handler

internal/liqonet/tunnelEndpointCreator-operator.go
```diff
@@ -444,26 +467,46 @@ func (r *TunnelEndpointCreator) CreateTunnelEndpoint(param networkParam) error {
...
+		klog.Errorf("an error occurred while converting resource %s of type %s to typed object: %s", objUnstruct.GetName(), objUnstruct.GetKind(), err)
 		return
 	}
+	klog.Infof("processing %s", adv.Name)
 	_ = r.createNetConfig(adv.Spec.ClusterId)
 }
 
-func (r *TunnelEndpointCreator) PeeringRequestHandler(obj *unstructured.Unstructured) {
+func (r *TunnelEndpointCreator) AdvertisementHandlerUpdate(oldObj interface{}, newObj interface{}) {
+	r.AdvertisementHandlerAdd(newObj)
```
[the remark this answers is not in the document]
alacuku: what if the `networkconfigs.net.liqo.io` is deleted? or if the `add` event is lost? how would you manage those situations listening only for `add` events? remember that after the re-sync period the informer returns an `update` event.

internal/liqonet/tunnelEndpointCreator-operator.go
alacuku: it's a generic function, that checks, logs, and return the error if one is verified.

internal/liqonet/tunnelEndpointCreator-operator.go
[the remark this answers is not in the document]
fraborg: ok, I didn't know the deletion policy on `NetworkConfig`, thanks!

internal/liqonet/tunnelEndpointCreator-operator.go
palexster: And why don't we check the return value?

alacuku: in the go routine running that code we can not recover from errors. We have to options:

1. panik, exit the program
2. log the error and continue

We are interacting with the api-server, errors could be permissions or network related. The former could be solved by the user, and logging is enough, the latter is transitory we log, and retry later.

cmd/liqonet/main.go
alacuku: this is a little bit complicated. Just because a peeringrequests or and advertisement is deleted  we should not tear down the network connection between the two clusters. Im thinking how to integrate the network connection in the "unjoin" flow.

fraborg (approved):

PR #253, opened 2020-09-12
crdReplicator: replace watchers with informers

alacuku:
# Description

Using informers instead of watchers. Informers  permit to have a re-sync period on the watched resource.

# How Has This Been Tested?

Unit and integration tests

### DIFF
--- a/internal/crdReplicator/crdReplicator-operator.go
+++ b/internal/crdReplicator/crdReplicator-operator.go
@@ -10,10 +10,11 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/client-go/util/retry"
@@ -21,6 +22,7 @@ import (
 	"reflect"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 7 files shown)


internal/crdReplicator/crdReplicator-operator.go
```diff
@@ -230,56 +215,47 @@ func (d *CRDReplicatorReconciler) RemoteResourceModifiedHandler(obj *unstructure
...
-			err := d.UpdateStatus(client, gvr, localObj, clusterID, remoteStatus)
-			return err
-		}
-		return nil
-	})
-	if retryErr != nil {
-		klog.Error(retryErr)
+	//we reflect on the local resource only the changes of the status of the remote one
+	//we do not reflect the changes to labels, annotations or spec
+	//TODO:support labels and annotations
```
palexster, line 221: My 2 cents: I would avoid to only have todos in the code but also create issues

palexster (approved):

test/unit/crdReplicator/env_test.go
```diff
@@ -130,6 +132,11 @@ func setupEnv() {
 		peeringClustersManagers[peeringClusterID] = manager
 		dynClient := dynamic.NewForConfigOrDie(manager.GetConfig())
 		peeringClustersDynClients[peeringClusterID] = dynClient
+		dynFac := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynClient, 30*time.Second, metav1.NamespaceAll, func(options *metav1.ListOptions) {
```
palexster, line 135: I would put the amount of seconds as a variable outside this call.

internal/crdReplicator/env_test.go
```diff
@@ -53,6 +57,12 @@ func setupEnv() {
 		panic(err)
 	}
 	dynClient = dynamic.NewForConfigOrDie(configLocal)
+	dynFac = dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynClient, 30*time.Second, metav1.NamespaceAll, func(options *metav1.ListOptions) {
```
mLavacca, line 60: I think it can be useful to parameterize the `resyncPeriod`. Maybe we could use a flag to enable the customization of such parameter?

test/unit/crdReplicator/crdReplicator-operator_test.go
```diff
@@ -33,17 +34,28 @@ var (
 func setupDispatcherOperator() error {
 	var err error
 	localDynClient := dynamic.NewForConfigOrDie(k8sManagerLocal.GetConfig())
+	localDynFac := dynamicinformer.NewFilteredDynamicSharedInformerFactory(localDynClient, crdReplicator.ResyncPeriod, metav1.NamespaceAll, func(options *metav1.ListOptions) {
+		//we want to watch only the resources that should be replicated on a remote cluster
+		klog.Infof("printing labels: %s", options.LabelSelector)
+		if options.LabelSelector == "" {
+			options.LabelSelector = crdReplicator.LocalLabelSelector + "=true"
+		} else {
+			options.LabelSelector = options.LabelSelector + "," + crdReplicator.LocalLabelSelector + "=" + "true"
```
mLavacca, line 43: In order to make the string building more efficient, we could use either `strings.Builder` or `strings.Join()`

alacuku: thanks for the suggestion. Using `strings.Join()` instead of the `+` operator.

internal/crdReplicator/env_test.go
alacuku: it was already parametrized, so it's now used where needed. The `resyncPeriod` value is used to periodically check the watched resource in case if any of the events was last due to network problems or timeouts. I don't think it could be of great use for the user to be able to set it.

mLavacca (approved):

PR #255, opened 2020-09-14
Fix DNS Discovery

aleoli:
# Description

Now DNS discovery uses the default pod's DNS server (i.e. the one that is indicated in `/etc/resolv.conf`)

## Changes
* The DNS server is no more needed in ClusterConfig
* A new finalizer has been added on ForeignCluster to avoid accidentally (or not) delete of peered ones, they will be unpeered and then deleted


giorio94 (approved):
As far as I can understand (I do not know the details of the implementation), looks good to me.

palexster (approved):
/lgtm

PR #260, opened 2020-09-15
Untrusted CA accepted by default with DNS Discovered clusters

aleoli:
# Description

The default value of allow untrusted ca is now true

Exposed CA certificate is no more deleted when cluster config trusted mode changes


api/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -37,27 +37,35 @@ const (
...
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	Outgoing Outgoing `json:"outgoing,omitempty"`
-	Incoming Incoming `json:"incoming,omitempty"`
-	Ttl      int      `json:"ttl,omitempty"`
+	Outgoing Outgoing  `json:"outgoing,omitempty"`
+	Incoming Incoming  `json:"incoming,omitempty"`
+	Ttl      int       `json:"ttl,omitempty"`
+	Trusted  TrustMode `json:"trusted"`
```
fraborg, line 68: I would add the enum [annotation](https://book.kubebuilder.io/reference/markers/crd-validation.html) so that if we automatically generate the crd it reflects the struct definition

api/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -37,27 +37,35 @@ const (
 	IncomingPeeringDiscovery DiscoveryType = "IncomingPeering"
 )
 
+type TrustMode string
+
+const (
+	TrustUnknown   TrustMode = ""
```
fraborg, line 43: Just a personal preference: I would name these constants differently (for example  `TrustModeUnknown` or `ModeUnknown` or simply `Unknown`, depending on how you name the respective field in the struct) so that when you compare it is clearer the meaning.

fraborg (approved):

deployments/liqo_chart/crds/foreign-cluster-crd.yaml
```diff
@@ -108,4 +104,11 @@ spec:
             ttl:
               type: integer
               description: If discoveryType is LAN and this counter reach 0 value, this FC will be removed
+            trustMode:
+              type: string
+              description: Indicates if this remote cluster is trusted or not
+              enum:
+                - ""
```
palexster, line 111: Why do we need an empty value?

test/unit/discovery/discovery_test.go
```diff
@@ -327,7 +327,9 @@ func testBidirectionalJoin(t *testing.T) {
 	// wait reconciliation
 	time.Sleep(1 * time.Second)
 
-	tmp, err := serverCluster.client.Resource("foreignclusters").List(metav1.ListOptions{})
+	tmp, err := serverCluster.client.Resource("foreignclusters").List(metav1.ListOptions{
+		LabelSelector: "discovery-type=" + string(v1alpha1.IncomingPeeringDiscovery),
```
palexster, line 331: String.Join is preferable

test/unit/discovery/env_test.go
```diff
@@ -50,7 +50,7 @@ func getClientCluster() *Cluster {
 		cluster.client,
 		cluster.advClient,
 		cluster.clusterId,
-		1*time.Minute,
+		10*time.Minute,
```
palexster, line 53: Why 10 minutes?

palexster (approved):

PR #262, opened 2020-09-15
Establishing the network connection based on the join/unjoin flow

alacuku:
# Description
The network connection is now handled based on the status of the foreign cluster. If at least one of the incoming or outgoing connections is set to "joined = true" then the connection is established. When they are both false then the operator tears down the network connection.

Ref. #246

# How Has This Been Tested?
Given a foreign cluster with the following status:
1. {outgoingJoined: true, incomingJoined: false}
2. {outgoingJoined: false, incomingJoined: true}
3. {outgoingJoined: true, incomingJoined: true}
4. {outgoingJoined: false, incomingJoined: false}

we check that for the cases 1, 2 and 3 the networkconfigs.net.liqo.io is created and removed when the foreigncluster.discovery.liqo.io is removed.
For the fourth case we check that the networkconfigs.net.liqo.io is not created at all.


mLavacca (approved):
/lgtm

palexster (approved):

PR #266, opened 2020-09-16
 Fixing route-operator not processing tunnelendpoints.net.liqo.io resources on restart

### LINKED ISSUE #264: Upon node reboot, liqo routes are not re-created
**Describe the bug**
Upon the reboot of a node in the cluster, routes for external traffic are not re-created in Liqo.

**To Reproduce**
1. Install Liqo
2. Check the routes
3. Reboot a node
4. Check the routes again

**Expected behavior**
Routes should be re-created after a node become unavailable.

alacuku:
# Description
This PR fixes the route-operator not processing `tunnelendpoints.net.liqo.io` resources when restarted.
When the operator proccessed a `tunnelendpoints.net.liqo.io` instance it added a label to the resource to say that it has been processed. Each time the reconcile function of the operator is called upon an update event of the resource it checks if the label is present if yes it would do nothing. This label prevents also the operator to reconcile the state of a resource according to the `resyncPeriod`. The `resyncPeriod` is really important to check that the routes and iptables rules have not been modified by an external actor.
This check was introduced to sync the various operators that handled a `tunnelendpoints.net.liqo.io` resource.  
Fixes #264 

# How Has This Been Tested?

Restarting the route-operator pod and verifying that:

1. liqo-routes are not present while the route-operator is not running
2. liqo-routes are installed after the restart

### DIFF
--- a/internal/liqonet/route-operator.go
+++ b/internal/liqonet/route-operator.go
@@ -23,7 +23,6 @@ import (
 	liqonetOperator "github.com/liqotech/liqo/pkg/liqonet"
 	"github.com/vishvananda/netlink"
 	k8sApiErrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"net"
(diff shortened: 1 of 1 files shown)


palexster (approved):

palexster:
/lgtm
My 2 cents: Can you please explain what you did in this PR and what was the problem?

palexster (approved):

mLavacca (approved):

PR #268, opened 2020-09-16
Bug fixing and enhancement network operators

alacuku:
# Description
This PR brings bugfixing and some refactoring  of the network operators

## TunnelEndpointCreator
The `tunnelendpoints.net.liqo.io` resource is created now in one step, specifying the spec and status field at once. It used to first create the resource with the spec field, wait for the creation and then update the status field. This process is now delegated to the client-go library.
A newly created `tunnelendpoint` resource has the status field named `phase` set to _Ready_. This is needed for the tunnel and route operators. Otherwise the resource may be consumed before that some status fields are set.

## Route-Operator

- Correcting some misleading information in log messages. 
- Fixing a bug causing the LIQONET-POSTROUTING chain not to be flushed when the operator pod is removed
- The reconcile function is idempotent with a resync period of 30 seconds where it checks that routing table and iptables rules are correctly configured.
- When iptables rules and routes are added or removed `events` are sent to the resource been reconciled 
tep_events
This way if an error occurs the user knows which instance/node to check for misconfiguration. If an error occurs while removing the external resources it is reported through an event on the related `tunnelendpoint` resource and waits for the user to solve the conflicts.

## Tunnel-Operator
- Fixing the bug which caused the operator not to reconcile the `tunnelendpoints` resources if restarted.
- Periodically ensures the existence of the tunnel network interface for the peering clusters
- The creation of the tunnel network interface is now idempotent: if it exists a check if the configuration is the same is done and it is recreated only when the existing one and the wanted one differs.
- Instal/uninstall of network interfaces is reported by means of events to the related `tunnelendpoint` resource. 

Fixes #(issue)

# How Has This Been Tested?

Existing unit and integration test  have been updated to use the status field `phase` with new value which for now could be set to _Ready_ only.

### DIFF
--- a/internal/liqonet/route-operator.go
+++ b/internal/liqonet/route-operator.go
@@ -25,6 +25,7 @@ import (
 	k8sApiErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/record"
 	"net"
 	"os"
 	"os/signal"
@@ -54,7 +55,7 @@ type RouteController struct {
 	Log            logr.Logger
 	Scheme         *runtime.Scheme
 	clientset      kubernetes.Clientset
-	RouteOperator  bool
+	Recorder       record.EventRecorder
 	NodeName       string
 	ClientSet      *kubernetes.Clientset
 	RemoteVTEPs    []string
@@ -96,7 +97,14 @@ func (r *RouteController) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		r.Log.Error(err, "unable to fetch endpoint")
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
-
+	//we can process a tunnelendpoint resource only if the tunnel interface has been created and configured
+	//this network interface is managed by the tunnel operator who sets this field when process the same tunnelendpoint
+	//resource and creates the network interface.
+	//if it has not been created yet than we return and wait for the resource to be processed by the tunnel operator.
+	if endpoint.Status.TunnelIFaceName == "" {
+		log.Info("the tunnel network interface is not ready")
+		return ctrl.Result{RequeueAfter: r.RetryTimeout}, nil
+	}
 	// examine DeletionTimestamp to determine if object is under deletion
 	if endpoint.ObjectMeta.DeletionTimestamp.IsZero() {
 		if !liqonetOperator.ContainsString(endpoint.ObjectMeta.Finalizers, routeOperatorFinalizer) {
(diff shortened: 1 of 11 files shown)


internal/liqonet/route-operator.go
```diff
@@ -196,7 +219,7 @@ func (r *RouteController) createAndInsertIPTablesChains() error {
 	} else {
 		log.Info("created", "chain", LiqonetPreroutingChain, "in table", NatTable)
 	}
-	r.IPTablesChains[LiqonetPostroutingChain] = liqonetOperator.IPTableChain{
+	r.IPTablesChains[LiqonetPreroutingChain] = liqonetOperator.IPTableChain{
```
palexster, line 222: Out of curiosity: Why do we moved to PreRoutingChain?

alacuku: it was a little bug where we saved the _LiqonetPreroutingChain_ as the _LiqonetPostroutingChain_ in the _IPTablesChains_ table. Looking at line 217 it is creating the _LiqonetPreroutingChain_.

palexster (approved):

internal/liqonet/route-operator.go
```diff
@@ -120,13 +128,18 @@ func (r *RouteController) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 	} else {
 		//the object is being deleted
+		//if we encounter an error while removing iptables rules or the routes than we record an
+		//event on the resource to notify the user
+		//the finalizer is not removed
 		if liqonetOperator.ContainsString(endpoint.Finalizers, routeOperatorFinalizer) {
 			if err := r.deleteIPTablesRulespecForRemoteCluster(&endpoint); err != nil {
 				r.Log.Error(err, "error while deleting rulespec from iptables")
+				r.Recorder.Event(&endpoint, "Error", "", err.Error())
```
fraborg, line 137: just a minor thing: can we use "Error" as event type? in the documentation it says only "Normal" and "Warning"

internal/liqonet/route-operator_test.go
```diff
@@ -81,7 +80,7 @@ func TestCreateAndInsertIPTablesChains(t *testing.T) {
 	for i := 3; i >= 0; i-- {
 		err := r.createAndInsertIPTablesChains()
 		assert.Nil(t, err, "error should be nil")
-		assert.Equal(t, 3, len(r.IPTablesChains), "there should be three new chains")
+		assert.Equal(t, 4, len(r.IPTablesChains), "there should be three new chains")
```
fraborg, line 83:
```suggestion
		assert.Equal(t, 4, len(r.IPTablesChains), "there should be 4 new chains")
```

alacuku: thanks.

internal/liqonet/route-operator.go
alacuku: you are right. corrected.

fraborg (approved):

PR #269, opened 2020-09-18
Associate a Cluster Name to Home and Foreign Clusters

aleoli:
# Description

This PR adds a "nice" name for our cluster. This can be set in the `ClusterConfig` CR

The remote clusters that will discover us will found our name in ForeignCluster in `Spec.ClusterIdentity.ClusterName`

If the `ClusterName` is updated in the ClusterConfig, the `ForeignCluster` on the already peered remote clusters is not updated, but the future clusters will discover it with the new name.

### DIFF
--- a/api/discovery/v1alpha1/foreigncluster_types.go
+++ b/api/discovery/v1alpha1/foreigncluster_types.go
@@ -42,12 +42,17 @@ type ForeignClusterSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	ClusterID        string        `json:"clusterID"`
-	Namespace        string        `json:"namespace"`
-	Join             bool          `json:"join"`
-	ApiUrl           string        `json:"apiUrl"`
-	DiscoveryType    DiscoveryType `json:"discoveryType"`
-	AllowUntrustedCA bool          `json:"allowUntrustedCA"`
+	ClusterIdentity  ClusterIdentity `json:"clusterIdentity"`
+	Namespace        string          `json:"namespace"`
+	Join             bool            `json:"join"`
+	ApiUrl           string          `json:"apiUrl"`
+	DiscoveryType    DiscoveryType   `json:"discoveryType"`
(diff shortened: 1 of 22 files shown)


api/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -42,12 +42,17 @@ type ForeignClusterSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	ClusterID        string        `json:"clusterID"`
-	Namespace        string        `json:"namespace"`
-	Join             bool          `json:"join"`
-	ApiUrl           string        `json:"apiUrl"`
-	DiscoveryType    DiscoveryType `json:"discoveryType"`
-	AllowUntrustedCA bool          `json:"allowUntrustedCA"`
+	ClusterIdentity  *ClusterIdentity `json:"clusterIdentity"`
```
frisso, line 45: Why this change? 
`ClusterIdentity`  looks weird; `ClusterID` looks much more common (if it is an ID, of course).

api/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -42,12 +42,17 @@ type ForeignClusterSpec struct {
...
+	Namespace        string           `json:"namespace"`
+	Join             bool             `json:"join"`
```
frisso, line 47: What does this mean?
`Join` looks meaningless to me. `AutoJoin`, maybe?
@palexster ?

aleoli: This is the flag to set to start the peering process, that can be set or because of an `AutoJoin` in Config or manually (or by dashboard) by the user

palexster: I totally agree that "Join" does not explicit well the status of the peering, maybe something like "Peered". I suggest to handle this in a dedicated PR, since I feel it requires a wider refactoring which involve many components.

palexster (approved):

nappozord (approved):

PR #279, opened 2020-09-21
Move ClusterConfig watch to shared informer

palexster:
# Description

This PR changes the way ClusterConfig is kept updated in watchConfig pkg moving to a shared informer.

# Testing

This PR was tested with Unit and E2E tests.

### DIFF
--- a/pkg/clusterConfig/watchConfig.go
+++ b/pkg/clusterConfig/watchConfig.go
@@ -5,12 +5,14 @@ import (
 	"github.com/liqotech/liqo/pkg/crdClient"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 	"os"
+	"time"
 )
 
 func WatchConfiguration(handler func(*configv1alpha1.ClusterConfig), client *crdClient.CRDClient, kubeconfigPath string) {
+	var rsyncPeriod = 1 * time.Second
 	if client == nil {
 		config, err := crdClient.NewKubeconfig(kubeconfigPath, &configv1alpha1.GroupVersion)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/clusterConfig/watchConfig.go
```diff
@@ -25,30 +27,42 @@ func WatchConfiguration(handler func(*configv1alpha1.ClusterConfig), client *crd
...
+			configuration, ok := newObj.(*configv1alpha1.ClusterConfig)
+			if !ok {
+				os.Exit(1)
+			}
 			handler(configuration)
-		case watch.Deleted:
-			klog.Error("please, do not delete ClusterConfigs")
+		},
+		DeleteFunc: func(config interface{}) {
+			klog.Error("please,obj do not delete ClusterConfigs")
```
fraborg, line 48:
```suggestion
			klog.Error("please, do not delete ClusterConfigs")
```

pkg/clusterConfig/watchConfig.go
```diff
@@ -25,30 +27,42 @@ func WatchConfiguration(handler func(*configv1alpha1.ClusterConfig), client *crd
...
-			continue
-		}
-		klog.V(3).Info("ClusterConfig changed")
-		switch event.Type {
-		case watch.Added, watch.Modified:
+	ehf := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			configuration, ok := obj.(*configv1alpha1.ClusterConfig)
+			if !ok {
+				os.Exit(1)
```
fraborg, line 36: I would log something before exiting

pkg/clusterConfig/watchConfig.go
```diff
@@ -5,12 +5,14 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 	"os"
+	"time"
 )
 
-func WatchConfiguration(handler func(*configv1alpha1.ClusterConfig), client *crdClient.CRDClient, kubeconfigPath string) {
+func WatchConfiguration(handler func(*configv1alpha1.ClusterConfig), client *crdClient.CRDClient, kubeconfigPath string) chan struct{} {
```
fraborg, line 14: it seems to me that the returned chan is never used, and in the function you only return nil at the end

test/unit/advertisement-operator/config-watcher_test.go
```diff
@@ -127,6 +162,34 @@ func testDisableBroadcaster(t *testing.T) {
 	assert.Equal(t, advtypes.AdvertisementDeleting, adv2.Status.AdvertisementStatus)
 }
 
+func waitEventOnClusterConfig(client *crdClient.CRDClient, resourcetype string, name string) error {
```
fraborg, line 165: since this is also used for Advertisement I would simply rename it to "waitEvent" or something similar

fraborg (approved):

AbakusW (approved):

PR #285, opened 2020-09-24
Migrate advertisement remote-watcher to informer

### LINKED ISSUE #290: Secret recreation error
**Describe the bug**
When the broadcaster tries to recreate the `Secret` for the VirtualKubelet, an error occurs due to `ResouceVersion` field being set.

```
E0925 10:11:22.708028       1 broadcaster.go:344] Unable to create secret vk-kubeconfig-secret-433f86df-2734-49e4-9dd7-9f6fd364b88f on remote cluster 23ca2126-3b7c-4a35-8186-78c4c76e5f36; error: resourceVersion should not be set on objects to be created
E0925 10:11:22.708051       1 broadcaster.go:176] resourceVersion should not be set on objects to be created Error while sending Secret for virtual-kubelet to cluster 23ca2126-3b7c-4a35-8186-78c4c76e5f36
```

**To Reproduce**
1. Peer 2 clusters
2. Delete the `Advertisement` on cluster1 (or directly delete the secret `vk-kubeconfig-secret-<clusterID>`
3. On cluster2, `kubectl logs -n liqo broadcaster-<clusterID>`

fraborg:
# Description
In this PR the remote-watcher in the broadcaster has been upgraded to use an informer.

## Other changes
- Fix #290 bug in broadcaster that prevented `Secret` recreation

### DIFF
--- a/internal/advertisement-operator/remote-watcher.go
+++ b/internal/advertisement-operator/remote-watcher.go
@@ -3,46 +3,63 @@ package advertisementOperator
 import (
 	discoveryv1alpha1 "github.com/liqotech/liqo/api/discovery/v1alpha1"
 	advtypes "github.com/liqotech/liqo/api/sharing/v1alpha1"
+	"github.com/liqotech/liqo/pkg/crdClient"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
+	"os"
+	"time"
 )
 
 func (b *AdvertisementBroadcaster) WatchAdvertisement(homeAdvName string) {
+	var err error
+	resyncPeriod := 1 * time.Minute
 
 	klog.Info("starting remote advertisement watcher")
-	watcher, err := b.RemoteClient.Resource("advertisements").Watch(metav1.ListOptions{
+	// events are triggered only by modifications on the Advertisement created by the broadcaster on the remote cluster
+	// homeClusterAdv is the Advertisement created by home cluster on foreign cluster -> stored remotely
+	lo := metav1.ListOptions{
 		FieldSelector: "metadata.name=" + homeAdvName,
-		Watch:         true,
-	})
+	}
+	ehf := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			b.handlerFunc(obj)
+		},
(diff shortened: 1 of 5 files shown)


internal/advertisement-operator/remote-watcher.go
```diff
@@ -3,46 +3,63 @@ package advertisementOperator
...
-	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
+	"os"
+	"time"
 )
 
 func (b *AdvertisementBroadcaster) WatchAdvertisement(homeAdvName string) {
+	var err error
+	resyncPeriod := 1 * time.Second
```
aleoli, line 16: Do we need a so short resync period on the remote cluster?

fraborg: I don't think so, I didn't think about it.. thanks!

fraborg: actually I think we could also set it to 0, so that it is only triggered by Watch event and we don't annoy the remote cluster with periodic List (since this watcher should be triggered very few times). what do you think @palexster?

palexster: My 2 cents: I would force a resync in order to recover from deadlocks or missed events ( If I understand the scenario correctly). For example, 60 seconds should be ok.

internal/advertisement-operator/remote-watcher.go
```diff
@@ -3,46 +3,63 @@ package advertisementOperator
...
-		case watch.Deleted:
-			klog.Info("Adv " + homeAdvName + " has been deleted")
-			watcher.Stop()
+func (b *AdvertisementBroadcaster) handlerFunc(obj interface{}) {
+	homeClusterAdv, ok := obj.(*advtypes.Advertisement)
+	if !ok {
+		klog.Error("Error casting Advertisement")
+		os.Exit(1)
+	}
+	if homeClusterAdv.Status.AdvertisementStatus != "" {
```
palexster, line 57: My 2 cents, not necessarily involving this PR: Can we avoid the nil status of resources? I find it a bit hard to follow. I imagine the status "" is mapped to a "New". Am I right?

palexster (approved):

internal/advertisement-operator/remote-watcher.go
fraborg: Yes, with defaulting it should be quite simple

aleoli (approved):

PR #289, opened 2020-09-25
Uninstall E2E

aleoli:
# Description

This PR adds E2E tests for Liqo uninstall.

1. Liqo is uninstalled on the cluster1 using the script (2 minutes of timeout)
   * the timeout will make tests to fail if some finalizers were not correctly removed and resources can not be deleted
2. Runs e2e tests to assert pods are still running on the cluster2 and that there are no more virtual nodes
3. Liqo is uninstalled on the cluster1 using the script (2 minutes of timeout)

### DIFF
--- a/test/e2e/unjoin_e2e/unjoin_test.go
+++ b/test/e2e/unjoin_e2e/unjoin_test.go
@@ -1,0 +1,39 @@
+package unjoin_e2e
+
+import (
+	context2 "context"
+	"github.com/liqotech/liqo/test/e2e/util"
+	"gotest.tools/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+	"testing"
+)
+
+func TestUnjoin(t *testing.T) {
+	context := util.GetTester()
+	NoPods(context.Client1, context.Namespace, t, "cluster1")
+
+	NoJoined(context.Client2, t, "cluster2")
+	util.ArePodsUp(context.Client2, context.Namespace, t, "cluster2")
+}
+
+func NoPods(clientset *kubernetes.Clientset, namespace string, t *testing.T, clustername string) {
+	pods, err := clientset.CoreV1().Pods(namespace).List(context2.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Error(err)
+		t.Fail()
+	}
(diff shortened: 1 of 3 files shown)


AbakusW (approved):

test/e2e/util/util.go
```diff
@@ -89,13 +91,13 @@ func createTester() *Tester {
 	}
 	clusterID1, err := getClusterID(clientset1, namespace)
 	if err != nil {
-		klog.Errorf("an error occurred while getting cluster-id configmap %s", err)
-		os.Exit(1)
+		klog.Warningf("an error occurred while getting cluster-id configmap %s", err)
+		clusterID1 = ""
```
fraborg, line 95: why don't you return an error? can we work properly without clusterID?

test/e2e/unjoin_e2e/unjoin_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"gotest.tools/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+	"testing"
+)
+
+func TestUnjoin(t *testing.T) {
+	context := util.GetTester()
+	NoPods(context.Client1, context.Namespace, t, "cluster1")
```
fraborg, line 15: I think I'm not seeing it: where do you begin unjoin process?

test/e2e/util/util.go
aleoli: yes, after uninstall I have to access standard Kubernetes resource (as pods and nodes), but cluster-id doesn't exist anymore

test/e2e/unjoin_e2e/unjoin_test.go
aleoli: It is in liqops actions, these tests are run after the uninstall script

fraborg: ok I found it

test/e2e/util/util.go
fraborg: ok I get it

fraborg (approved):

palexster:
Hi @aleoli, thanks for this! Maybe in a next PR: Can we add more assertions to assess the state of the cluster? For example, the existence of CRDs and other complementary resources...

aleoli:
Yes, sure, these are very high-level tests

Issue #305, opened 2020-10-01
Impossible to install a dashboard version different from latest


giorio94:
**Describe the bug**
Whatever `LIQO_VERSION` is selected, the dashboard is always installed using the `latest` tag. This is inconsistent with the rest of the components and raises the usual concerns regarding the usage of the latest tag (e.g. a restart may cause a new version to be used).

**To Reproduce**
1. Instal liqo: `curl -sL https://raw.githubusercontent.com/liqotech/liqo/master/install.sh | LIQO_VERSION=... bash`
2. Check the dashboard container version: `kubectl get po liqo-dashboard-7977f68bc4-ml4sg -o yaml | grep ' image:'` 

**Expected behavior**
Since the dashboard is located on a different repository, the commit sha will never be the same of the other components. However, I believe that the tagged versions should be kept aligned between the two repo to avoid confusions, while for installations from master the latest commit tag should be retrieved and used as for the other components.

PR #308, opened 2020-10-01
Dashboard version install

nappozord:
# Description
This PR fixes the dashboard not installing with the version given during installation.
- If a version is specified, the dashboard and liqo will be installed with the same version.
- If a commit is specified, the dashboard will be installed with the master tag image, while liqo will be installed with the commit specified, as always.
- If nothing is specified, the master images will be installed for every component.

ref. #305

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -22,7 +22,7 @@ set -o pipefail  # Fail if one of the piped commands fails
 # Environment variables:
 #
 #   - LIQO_VERSION
-#     the version of Liqo to install. It can be a released version, a commit SHA or 'latest'.
+#     the version of Liqo to install. It can be a released version, a commit SHA or 'master'.
 #     Defaults to the latest released version.
 #
 #   - LIQO_NAMESPACE
@@ -48,6 +48,8 @@ EXIT_FAILURE=1
 
 LIQO_REPO="liqotech/liqo"
 LIQO_CHARTS_PATH="deployments/liqo_chart"
+
+LIQO_DASHBOARD_REPO="liqotech/dashboard"
 
 LIQO_NAMESPACE_DEFAULT="liqo"
 CLUSTER_NAME_DEFAULT=$(printf "LiqoCluster%04d" $(( RANDOM%10000 )) )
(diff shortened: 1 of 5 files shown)


AbakusW (approved):

install.sh
```diff
@@ -22,9 +22,13 @@ set -o pipefail  # Fail if one of the piped commands fails
...
 #
 #   - LIQO_VERSION
-#     the version of Liqo to install. It can be a released version, a commit SHA or 'latest'.
+#     the version of Liqo to install. It can be a released version, a commit SHA or 'master'.
 #     Defaults to the latest released version.
 #
+#	- LIQO_DASHBOARD_VERSION
+#	  the version of the Liqo Dashboard to install. It can be a released version, a commit SHA or 'master'.
+#	  Defaults to the latest released version.
+#
```
giorio94, line 31: These lines should be dropped, since this is not a user-supplied environment variable

giorio94 (commented):
Thanks @nappozord 

Besides the minor comment inline, I would integrate the `setup_liqo_version` tests to verify that the new variable is configured correctly.

install.sh
nappozord: Yes, sorry I forgot to get rid of them

test/installer/install-test.sh
```diff
@@ -182,94 +182,136 @@ setup() {
...
+	# get_repo_tags returns no tags, hence should fallback to master
 	unset LIQO_VERSION
 	setup_liqo_version
 	assert_equal "${LIQO_VERSION}" "master"
 	assert_equal "${LIQO_IMAGE_VERSION}" "f2de258b07d8b507b461f55f87e17f1bb619f926"
+	assert_equal "${LIQO_DASHBOARD_IMAGE_VERSION}" "a3de258b07d8b507b461f55f87e17f1bb619f926"
 	assert [ -z "${LIQO_SUFFIX:-}" ]
 }
 
+@test "setup_liqo_version fails if the dashboard repo do not have the same tag as the liqo repo" {
```
giorio94, line 299:
```suggestion
@test "setup_liqo_version fails if the dashboard repo does not have the same tag as the liqo repo" {
```

test/installer/install-test.sh
```diff
@@ -182,94 +182,136 @@ setup() {
 
 @test "setup_liqo_version correctly configures the environment if a commit is specified" {
 	LIQO_VERSION="f2de258b07d8b507b461f55f87e17f1bb619f926"
+	function get_repo_master_commit() { 
+		if [[ "$1" = "liqotech/dashboard" ]]; then
+			echo "a3de258b07d8b507b461f55f87e17f1bb619f926";
+		else
+			echo "f2de258b07d8b507b461f55f87e17f1bb619f926";
```
giorio94, line 189: Here, I would use a different SHA from `LIQO_VERSION`, just too avoid false positives.

test/installer/install-test.sh
```diff
@@ -182,94 +182,136 @@ setup() {
...
+	function get_repo_tags() { 
+		if [[ "$1" = "liqotech/dashboard" ]]; then
+			echo "";
+		else
+			printf "v1.0.0\nv0.1.0-alpha";
+		fi
+	}
+	declare -f get_repo_tags
+
+	# get_repo_tags returns a list of tags, hence the latest should be used (assumed to be sorted)
```
giorio94, line 309:
```suggestion

```

giorio94 (approved):
/lgtm

Just a couple of very minor comments inline...

giorio94:
Fixes #305

PR #313, opened 2020-10-02
Discovery Type Priority

### LINKED ISSUE #307: The SearchDomain autojoin=true flag is not respected if an unidirectional peering already exists
**Describe the bug**
In the context of DNS discovery, when an unidirectional peering has already been established, the creation of a `SearchDomain` for the opposite direction does not trigger the autojoin process (although autojoin=true). Instead, it is necessary to manually set the join property in the corresponding `foreigncluster` resource (which is otherwise set to false).

**To Reproduce**
1. Install liqo on two clusters (A and B)
2. Create a `SearchDomain` resource in A pointing to B (with autojoin=true) and observe that the virtual kubelet is correctly created
3. Create a `SearchDomain` resource in B pointing to A (with autojoin=true) and observe that the virtual kubelet is NOT created
4. Check the corresponding `foreigncluster` resource in B and observe that the join property is set to false

**Expected behavior**
Both clusters should correctly perform the autojoin

aleoli:
# Description

This PR adds a priority in Discover Types.

If a cluster was already discovered with `IncomingPeering` type (and so not automatically joined) and it is discovered again in another way (`WAN`, `LAN` or `Manual`) the discovery type is overwritten.

If the new Discovery Type is `WAN` the join flag is set accordingly to the value set in the `SearchDomain`.

Fixes #307 

# How Has This Been Tested?

- [x] Local K3s deployment

### DIFF
--- a/apis/discovery/v1alpha1/foreigncluster_methods.go
+++ b/apis/discovery/v1alpha1/foreigncluster_methods.go
@@ -156,3 +156,10 @@ func (fc *ForeignCluster) DeleteAdvertisement(advClient *crdClient.CRDClient) er
 	}
 	return nil
 }
+
+// if we discovered a cluster with IncomingPeering we can upgrade this discovery
+// when we found it also in other way, for example inserting a SearchDomain or
+// adding it manually
+func (fc *ForeignCluster) HasHigherPriority(discoveryType DiscoveryType) bool {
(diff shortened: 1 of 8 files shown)


palexster (approved):

giorio94 (approved):
/lgtm

PR #323, opened 2020-10-07
[WIP] Tweak Advertisement Operator resources

palexster:
# Description

This PR reduces the amount of privileges provided to the advertisement-operator service account.
In addition, it also sets requests-limits for virtual-kubelet inside the advertisement operator code

# Tests

- [ ] E2E

### DIFF
--- a/deployments/liqo/subcharts/advertisementOperator/templates/adv-deploy.yaml
+++ b/deployments/liqo/subcharts/advertisementOperator/templates/adv-deploy.yaml
@@ -9,6 +9,121 @@ metadata:
     app: liqo.io
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: liqo-advertisement-operator-crd-role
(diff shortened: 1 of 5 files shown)


deployments/liqo/subcharts/advertisementOperator/templates/adv-deploy.yaml
```diff
@@ -9,6 +9,98 @@ metadata:
     app: liqo.io
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: advertisement-operator-cluster-role
```
giorio94, line 14: Just a comment on the fly. What about something along the lines of `liqo-advertisement-operator` (or `liqo.io:advertisement-operator`) as a name? I feel it would simplify the management as all the liqo cluster roles (and cluster rolebindings if modified accordingly) would be grouped together. Looking at our cluster, it also seems to be the standard practice of most other projects.

Issue #303, opened 2020-09-30
[Feature] Reduce the rights of Liqo Components


palexster:
So far many liqo components are relying on a "cluster admin" ClusterRole, which is not necessary every time. We should move to better tailored clusterRoles for each component.

PR #326, opened 2020-10-08
Discovery Roles

aleoli:
# Description

This PR reduces the amount of privileges provided to the discovery service account.

Ref. #303 

# How Has This Been Tested?

- [x] E2E
- [x] Locally on kind

### DIFF
--- a/deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
+++ b/deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
@@ -9,6 +9,162 @@ metadata:
     app: liqo.io
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: liqo-discovery
+rules:
+  - apiGroups:
+      - discovery.liqo.io
+    resources:
+      - foreignclusters
+      - searchdomains
+    verbs:
+      - get
+      - list
+      - patch
+      - update
+      - watch
+      - create
+      - delete
+  - apiGroups:
+      - discovery.liqo.io
+    resources:
+      - peeringrequests
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - config.liqo.io
+    resources:
+      - clusterconfigs
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+  - apiGroups:
+      - net.liqo.io
+    resources:
+      - networkconfigs
+      - tunnelendpoints
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - sharing.liqo.io
+    resources:
+      - advertisements
+    verbs:
+      - get
+      - list
+      - watch
+      - delete
+  - apiGroups:
+      - ""
+    resources:
+      - nodes
+    verbs:
+      - list
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - clusterroles
+    verbs:
+      - get
+      - create
+      - update
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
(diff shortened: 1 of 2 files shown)


palexster (approved):

deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
```diff
@@ -9,6 +9,148 @@ metadata:
     app: liqo.io
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: liqo-discovery-role
```
giorio94, line 14: My opinion is that `-role` in the name is a bit wierd (no one seems doing that). Just an opinion though.

deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
```diff
@@ -9,6 +9,148 @@ metadata:
...
+      - create
+
+  # to satisfy ClusterRoles creation
+  - apiGroups:
+      - ""
+      - sharing.liqo.io
+    resources:
+      - advertisements
+      - advertisements/status
+      - secrets
```
mLavacca, line 93: Do we need to give all of these cluster-scoped privileges for the secrets?

deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
```diff
@@ -9,6 +9,148 @@ metadata:
...
+      - get
+      - create
+      - update
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - clusterrolebindings
+    verbs:
+      - get
+      - create
```
giorio94, line 84: Just curiosity, why do you need this? You are effectively becoming `cluster-admin`

deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
```diff
@@ -9,6 +9,148 @@ metadata:
...
+    resources:
+      - networkconfigs
+      - networkconfigs/status
+    verbs:
+      - "*"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: liqo-discovery-role-namespaced
```
giorio94, line 112: 1. Could this be a `Role` instead of a `ClusterRole`?
2. Same comment as above concerning `-role` within the name

deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
aleoli: Probably not, but these are the privileges needed to create the role for the remote broadcaster and the crdReplicator. That, again, probably does not need these privileges, but it is a bigger thing to do...

@palexster what do you suggest, doing it in this pr or open an issue and to do it in the future?

deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
aleoli: The role of the crdReplicator is created dynamically when a peering occurs, this is required to do that

deployments/liqo/subcharts/discoveryOperator/templates/discovery-deploy.yaml
palexster: I would suggest to open a dedicated issue and handle this in the very next iteration. This PR has still the merit to explicit permissions required by discovery operator, which are way too extended for the moment

mLavacca (approved):

PR #329, opened 2020-10-09
Network operators supporting peering sessions with more than two clusters

alacuku:
# Description

The resources of type `networkconfigs.net.liqo.io` used to exchange the network parameters are now identified by using `labels` and not their name. The name is generated automatically at creation time. Before the clusterID of a remote cluster was used to name a `networkconfig` resource causing conflicts when multiple clusters were sending their parameters to a remote cluster.

The VPN tunnels are now named based on the last 5 characters of the related `tunnelendpoints.net.liqo.io` resource.

# How Has This Been Tested?

The unit tests have been updated to use labels instead of fixed names for the resources involved.


palexster:
Hi @alacuku! Thanks for this PR. Could you please fill/reference an issue/feature this PR belongs to?

palexster (approved):

mlavacca (approved):

PR #330, opened 2020-10-12
Improve pod translation in Liqo virtual kubelet

aleoli:
# Description

This PR adds some improvements to the virtual node and the virtual kubelet.

Virtual Node:
* Generic management of allocatables and capacities, in the same way before we managed CPU and RAM, with sharing percentage
* Generic custom label aggregation from cluster nodes to the remote virtual node
  * We can define the label to care about in the `ClusterConfig` with a `LabelPolicy`
  * This label will be added to the remote virtual node according to this policy
  * New policies are easy to add

Virtual Kubelet translation:
* Some fields are added in the pod translation
  * initContainers
  * imagePullSecrets
  * restartPolicy
  * terminationGracePeriodSeconds
  * securityContext
  * hostname
  * nodeSelector

# How Has This Been Tested?

- [x] Locally on kind
- [x] On real k8s clusters
- [x] Label Policies have unit tests


apis/config/v1alpha1/clusterconfig_types.go
```diff
@@ -84,6 +87,15 @@ type AdvOperatorConfig struct {
 	AcceptPolicy AcceptPolicy `json:"acceptPolicy"`
 }
 
+type LabelPolicy struct {
+	// Label Key
```
palexster, line 91: Maybe we can add some more text to describe what this new section is about.

internal/advertisement-operator/broadcaster.go
```diff
@@ -242,33 +244,35 @@ func (b *AdvertisementBroadcaster) CreateAdvertisement(virtualNodes *corev1.Node
...
+		return nil, nil, nil, nil, nil, nil, err
 	}
 	reqs, limits := GetAllPodsResources(nodeNonTerminatedPodsList)
 	// compute resources to be announced to the other cluster
 	availability, images = ComputeAnnouncedResources(physicalNodes, reqs, int64(b.ClusterConfig.AdvertisementConfig.OutgoingConfig.ResourceSharingPercentage))
 
-	return physicalNodes, virtualNodes, availability, limits, images, nil
+	labels = GetLabels(physicalNodes, b.ClusterConfig.AdvertisementConfig.LabelPolicies)
+
+	return physicalNodes, virtualNodes, availability, limits, images, labels, nil
```
palexster, line 275: Can we create a struct to encapsulate those parameters?

palexster (approved):

mlavacca (approved):

PR #336, opened 2020-10-16
Docker images CI suffix

mlavacca:
# Description

This commit applies the CI-suffix even for the images related to master

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -47,15 +47,16 @@ jobs:
         uses: actions/checkout@v2
         with:
           ref: "${{ env.commit_ref }}"
-      - name: Push ${{ matrix.component }} image on repo (Master)
+
+      - name: Push ${{ matrix.component }} image (unversioned)
         uses: elgohr/Publish-Docker-Github-Action@master
         with:
-          name: liqo/${{ matrix.component }}
+          name: liqo/${{ matrix.component }}-ci
           username: ${{ secrets.DOCKER_USERNAME }}
           dockerfile: build/${{ matrix.component }}/Dockerfile
           password: ${{ secrets.DOCKER_PASSWORD }}
           tags: "latest,${{ env.commit_ref }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Push ${{ matrix.component }} image on repo (Master)
         uses: elgohr/Publish-Docker-Github-Action@master
         with:
-          name: liqo/${{ matrix.component }}
+          name: liqo/${{ matrix.component }}-ci
```
palexster, line 53: May we fusion this part with the development part?

install.sh
```diff
@@ -239,6 +239,7 @@ function setup_liqo_version() {
 		warn "[PRE-FLIGHT] [DOWNLOAD]" "An unreleased version of Liqo is going to be downloaded"
 		LIQO_IMAGE_VERSION=$(get_repo_master_commit ${LIQO_REPO}) ||
 			fatal "[PRE-FLIGHT] [DOWNLOAD]" "Failed to retrieve the latest commit of the master branch"
+    LIQO_SUFFIX="-ci"
```
giorio94, line 242: Please, fix the indentation. The installer uses tabs for the indentation since spaces do not work in the `cat <<EOF` blocks

.github/workflows/go.yml
giorio94: Totally agree. As a future step it may be also possible to merge all of them and use a preliminary step to set the tag as required

giorio94 (approved):
/lgtm

Just a couple of minor comments

.github/workflows/go.yml
mlavacca: Yep, I grouped the two steps together

mlavacca: Yep, I grouped the two steps

palexster (approved):

PR #338, opened 2020-10-20
New iconset for Liqo Agent

AbakusW:
# Description

This PR adds a new iconset for the Liqo Agent according to Liqo official logo and style.

- The files of the **github.com/liqotech/liqo/internal/tray-agent/icon** package are no longer part of the repository.

  - They are generated automatically in the *golint* and *integration* pipelines when they are needed during testing and building.

  - They can be generated manually by executing locally the provided script **make_icon.sh** *from the project root directory*. The iconset is ignored by Git.

- Minor changes to code and test are made to maintain consistency with the new icons.


internal/tray-agent/icon/README.md
```diff
@@ -1,22 +1,21 @@
...
-[systray](https://github.com/getlantern/systray) package exploited 
-by **Indicator**.
+images into **Golang []byte**. This is the format required by the 
+[systray](https://github.com/getlantern/systray) package the **Indicator** exploits.
 
 > ```bash
 >$GOPATH/bin/2goarray <IcoVarName> icon < myimage.png > <IcoVarName>.go
 > ```
 
-- a simple script ```make_icon.sh``` is already provided to easy the
```
palexster, line 16: Can we generate icons at buildtime, using the make_icon.sh instead of committing all this .go files?

palexster (approved):

internal/tray-agent/icon/README.md
mlavacca: I totally agree, the fewer files we have to handle, the better.

AbakusW: IMHO it would not be a great improvement.

The main reason is that the .go files are generated starting from the .png files of the icons, and so at that point we would have to upload them.

Secondly, since they are .go files of a Go package, implementing the automatic generation on the pipeline would imply that to make building, formatting, testing work locally, one should have its local copy of the go package to copy in the repo folder at every checkout etc (and having a new rule in the .gitignore of course).

Am I missing a simpler solution?

(for what it's worth, I tried my best to achieve the best resolution/size ratio for each variable)

palexster (approved):

mlavacca (commented):
Thanks for this PR @AbakusW, it looks good.  I have only a final comment: can we move the  `png`  files under `/assets`? I give you a link to the standard repository template: https://github.com/golang-standards/project-layout#assets

alacuku (approved):
/lgtm

PR #351, opened 2020-10-27
Acquire Liqodash configuration from Clusterconfig CR

AbakusW:
# Description

This PR introduces the *AgentConfig* field inside the ClusterConfig CRD. The Liqo Agent can now load and store configuration data in the ClusterConfig CR.

Currently this section is used to read the required parameters to interact with the LiqoDash. This mechanism may partially change with the future removal of the LiqoDash subchart.

# How Has This Been Tested?

Unit tests


internal/tray-agent/agent/client/agent-client.go
```diff
@@ -190,6 +194,8 @@ func GetAgentController() *AgentController {
 				if agentCtrl.ConnectionTest() {
 					if err = agentCtrl.StartCaches(); err == nil {
 						agentCtrl.connected = true
+						//init configuration data
+						agentCtrl.acquireClusterConfiguration()
```
aleoli, line 198: So, if I understand correctly, we are not able to change these configurations while the tray-agent is running, is it correct?

AbakusW: This method only manages startup configuration the Agent uses to acquire config params (like the dashboard resources) and in future releases also its last configuration (e.g. level of notification). 
As soon as I Implement the update function for the ClusterConfig cache, we will be able to also have a runtime update.

aleoli (approved):

mlavacca (approved):

PR #356, opened 2020-10-29
service account secret reflection

aleoli:
# Description

This PR adds the reflection of ServiceAccount Secrets, we have to change the type to Opaque and to set a label with ServiceAccount name to easily retrieve them after


palexster (approved):

palexster:
Hi @aleoli, thanks for this. May be is it worth to add this scenario to the existing test at https://github.com/liqotech/liqo/blob/master/test/unit/reflection/secrets_test.go ?

pkg/virtualKubelet/apiReflection/reflectors/outgoing/secrets.go
```diff
@@ -139,6 +139,13 @@ func (r *SecretsReflector) PreAdd(obj interface{}) interface{} {
 	}
 	secretRemote.Labels[apimgmt.LiqoLabelKey] = apimgmt.LiqoLabelValue
 
+	if secretRemote.Type == corev1.SecretTypeServiceAccountToken {
```
palexster, line 142: I think a comment to explain this if will be useful for future extensions.

palexster (approved):

pkg/virtualKubelet/apiReflection/reflectors/outgoing/secrets.go
```diff
@@ -200,14 +212,11 @@ func (r *SecretsReflector) PreDelete(obj interface{}) interface{} {
 }
 
 func (r *SecretsReflector) isAllowed(obj interface{}) bool {
-	sec, ok := obj.(*corev1.Secret)
+	_, ok := obj.(*corev1.Secret)
```
mlavacca, line 215: Since we do not need the blacklisting mechanism anymore, you can delete this function. You should also delete the field  assignation [here](https://github.com/liqotech/liqo/blob/f5cf34fc99c77ff19d7cf0a9e37c615f0003a319/pkg/virtualKubelet/apiReflection/reflectors/outgoing/secrets.go#L24)

pkg/virtualKubelet/apiReflection/reflectors/outgoing/secrets.go
```diff
@@ -205,10 +224,9 @@ func (r *SecretsReflector) isAllowed(obj interface{}) bool {
 		klog.Error("cannot convert obj to secret")
 		return false
 	}
-	if sec.Type == corev1.SecretTypeServiceAccountToken {
-		return false
-	}
-	return true
+	// if this annotation is set, this secret will not be reflected to the remote cluster
+	_, ok = sec.Annotations["liqo.io/not-reflect"]
```
mlavacca, line 228: can you assert also the annotation value? maybe "true" (just a proposal)?

aleoli: ok

mlavacca (approved):

PR #357, opened 2020-11-02
mount service account secret

aleoli:
# Description

Mount reflected ServiceAccount secrets in the offloaded pods, in this way when an offloaded pod contact the home API server is able to authenticate itself with the ServiceAccount associated in the home cluster

# How Has This Been Tested?

- [x] Specific tests added to assert that secret volumes are correctly mounted in pods

### DIFF
--- a/pkg/virtualKubelet/translation/mutation.go
+++ b/pkg/virtualKubelet/translation/mutation.go
@@ -3,8 +3,10 @@ package translation
 import (
 	"fmt"
 	v1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/kubernetes/pod.go
```diff
@@ -42,6 +42,14 @@ func (p *KubernetesProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 	}
 
 	podTranslated := translation.H2FTranslate(pod, nattedNS)
+	remoteServices, err := p.apiController.ListMirroringObjects(apimgmgt.Secrets, podTranslated.Namespace)
```
mlavacca, line 45: I guess that the variable name is `remoteSecrets`

pkg/virtualKubelet/translation/mutation.go
```diff
@@ -180,3 +187,52 @@ func ChangePodIp(newPodCidr string, oldPodIp string) (newPodIp string) {
...
 }
+
+func TranslateSA(destPod *v1.Pod, origPod *v1.Pod, remoteSecrets []interface{}) (*v1.Pod, error) {
+	var secret *v1.Secret = nil
+	for _, obj := range remoteSecrets {
+		s, ok := obj.(*v1.Secret)
+		if !ok {
+			continue
+		}
+		if sa, ok := s.Labels["kubernetes.io/service-account.name"]; ok && sa == origPod.Spec.ServiceAccountName {
```
mlavacca, line 198: before checking the label, check also that `type==kubernetes.io/service-account-token`

pkg/virtualKubelet/translation/mutation.go
```diff
@@ -180,3 +187,52 @@ func ChangePodIp(newPodCidr string, oldPodIp string) (newPodIp string) {
...
+			secret = s
+			break
+		}
+	}
+	if secret == nil {
+		return nil, kerrors.NewServiceUnavailable("service account token not found")
```
mlavacca, line 204: I do not know if this error suits our case. I would use either `notFound` or a custom error

palexster (approved):

palexster:
Closing due to major refactoring.

PR #362, opened 2020-11-05
kubernetes service env var translation

aleoli:
# Description

Insert the IP of the home API server in the environment variables of the offloaded pods. In this way, the offloaded pods are able to contact our home cluster in the same way if they were scheduled locally

# How Has This Been Tested?

- [x] new unit test added


palexster:
Closing due to major refactoring of the VK.

PR #366, opened 2020-11-09
move service env var translation

aleoli:
# Description

This PR affects the injection of environment variables inside the virtual kubelet when offloading a Pod.

In particular, it moves the service environment variable translation inside the Kubernetes provider.

This avoids conflicts in the creation of the natted namespace.


palexster (dismissed):

pkg/virtualKubelet/storage/test/mockCacheReader.go
```diff
@@ -0,0 +1,62 @@
+package test
```
mlavacca, line 1: we already have a mocked cache manager under `/pkg/virtualKubelet/storage/test/mockCacheManager.go`

mlavacca (changes requested):

pkg/virtualKubelet/apiReflection/reflectors/incoming/replicaSets.go
```diff
@@ -83,7 +83,7 @@ func (r *ReplicaSetsIncomingReflector) preDelete(obj interface{}) interface{} {
 		return nil
 	}
 
-	po, err := r.GetCacheManager().GetHomeNamespacedObject(apimgmt.Pods, homeNamespace, foreignReplicaSet.Name)
+	po, err := r.GetCacheManager().GetHomeNamespacedObject(apimgmt.Pods, homeNamespace, podName)
```
aleoli, line 86: @mlavacca In order to use a unique mock cache, I had to change a little the maps in the mock struct. After that, the tests on this function were no more passing. I found that probably this was wrong, can you confirm it?

PS: if you confirm I will squash these commits

pkg/virtualKubelet/translation/serviceEnv/serviceEnv.go
```diff
@@ -0,0 +1,120 @@
...
+	for i := range services {
+		// We always want to add environment variables for master kubernetes service
+		// from the default namespace, even if enableServiceLinks is false.
+		// We also add environment variables for other services in the same
+		// namespace, if enableServiceLinks is true.
+
+		tmp := services[i]
+		service, ok := tmp.(*v1.Service)
+		if !ok {
+			continue
```
mlavacca, line 77: Can you add a comment with high verbosity (>=4)? For having visibility on this fault (very unlikely to me)

pkg/virtualKubelet/translation/serviceEnv/serviceEnv.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		serviceName := service.Name
+
+		if service.Namespace == metav1.NamespaceDefault && masterServices.Has(serviceName) {
+			// TODO: we have to set env vars for our home kubernetes service
```
mlavacca, line 82: I am not sure that we should set this environment variable for now (I would remove this if statement with its TODO and maybe we can keep track of this in an issue). In order to allow the remote pods communicating with the home API server in this way, we also should reflect that namespace and nat its address, but for now, it is blacklisted (i.e., not reflected [here](https://github.com/liqotech/liqo/blob/734d7544da18c8af58d9aae44df37167965b80a0/pkg/virtualKubelet/apiReflection/reflectors/outgoing/blacklists.go#L14))

pkg/virtualKubelet/translation/serviceEnv/serviceEnv.go
```diff
@@ -0,0 +1,120 @@
...
+	return m, nil
+}
+
+func addService(serviceMap *map[string]*v1.Service, cacheManager storage.CacheManagerReader, namespace string, name string, checkNamespace bool) error {
+	tmp, err := cacheManager.GetForeignNamespacedObject(apimgmgt.Services, namespace, name)
+	if err != nil {
+		return err
+	}
+	if tmp == nil {
+		return nil
```
mlavacca, line 108: Can you also add here a log with high verbosity for having visibility on this issue?

pkg/virtualKubelet/translation/serviceEnv/serviceEnv.go
```diff
@@ -0,0 +1,120 @@
...
+			})
+		}
+	}
+	return container.DeepCopy()
+}
+
+func getServiceEnvVarMap(ns string, enableServiceLinks bool, remoteNs string, cacheManager storage.CacheManagerReader) (map[string]string, error) {
+	var (
+		serviceMap = make(map[string]*v1.Service)
+		m          = make(map[string]string)
```
mlavacca, line 58: can you use a more verbose name (maybe envVars or something similar)?

pkg/virtualKubelet/translation/serviceEnv/serviceEnv.go
```diff
@@ -0,0 +1,120 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
+	"k8s.io/kubernetes/pkg/kubelet/envvars"
+)
+
+var masterServices = sets.NewString("kubernetes")
+
+func TranslateServiceEnvVariables(pod *v1.Pod, localNS string, nattedNS string, cacheManager storage.CacheManagerReader) (*v1.Pod, error) {
+	resPod := pod.DeepCopy()
```
mlavacca, line 16: This `deepcopy` is not needed, we already work on a brand-new pod obtained from [here](https://github.com/liqotech/liqo/blob/734d7544da18c8af58d9aae44df37167965b80a0/pkg/virtualKubelet/translation/mutation.go#L105)

pkg/virtualKubelet/translation/serviceEnv/serviceEnv.go
aleoli: the Kubernetes service translation is already done in another pr [here](https://github.com/liqotech/liqo/pull/362/files), and it is done not with a service but by providing the IP address of the home API server

mlavacca: ok, then I suppose we can remove this condition related to the Kubernetes default service, am I right?

mlavacca (approved):
/lgtm

palexster (approved):

PR #369, opened 2020-11-10
auth service

aleoli:
# Description

This pr is a first iteration over the Auth Service

## What's new
* A new HTTP service that, given the clusterID and a secret token creates a new ServiceAccount
* This new ServiceAccount has permissions on
  * Secrets in liqo namespace:
    * Create, update, delete, get
  * Peering requests:
    * Create, update, delete, get
* A secret informer creates a new secret token if no one exists

Ref. issue #371

# How Has This Been Tested?

Tests will be added in a very next future iteration

At the moment with a simple curl command, we can check that the service is working as expected

### DIFF
--- a/internal/auth-service/auth-service.go
+++ b/internal/auth-service/auth-service.go
@@ -1,0 +1,117 @@
+package auth_service
+
+import (
+	"context"
+	"github.com/julienschmidt/httprouter"
+	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
+	"github.com/liqotech/liqo/pkg/crdClient"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+	"net/http"
+	"strings"
+	"time"
+)
+
+type AuthServiceCtrl struct {
+	namespace      string
+	clientset      kubernetes.Interface
+	saInformer     cache.SharedIndexInformer
+	nodeInformer   cache.SharedIndexInformer
+	secretInformer cache.SharedIndexInformer
+}
+
+func NewAuthServiceCtrl(namespace string, kubeconfigPath string, resyncTime time.Duration) (*AuthServiceCtrl, error) {
+	config, err := crdClient.NewKubeconfig(kubeconfigPath, &discoveryv1alpha1.GroupVersion)
+	if err != nil {
+		return nil, err
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(clientset, resyncTime, informers.WithNamespace(namespace))
+
+	saInformer := informerFactory.Core().V1().ServiceAccounts().Informer()
+	saInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
+
+	nodeInformer := informerFactory.Core().V1().Nodes().Informer()
+	nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
+
+	secretInformer := informerFactory.Core().V1().Secrets().Informer()
+	secretInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
+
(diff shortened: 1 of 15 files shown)


AbakusW (approved):

internal/auth-service/auth-service.go
```diff
@@ -0,0 +1,116 @@
...
+func (authService *AuthServiceCtrl) Start() error {
+	if err := authService.configureToken(); err != nil {
+		return err
+	}
+
+	router := httprouter.New()
+
+	router.POST("/role", authService.role)
+
+	err := http.ListenAndServe(":5000", router)
```
alacuku, line 69: same as above for the `listeningPort`

internal/auth-service/auth-token.go
```diff
@@ -0,0 +1,87 @@
...
+	v1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog"
+	"strings"
+)
+
+func (authService *AuthServiceCtrl) getToken() (string, error) {
+	obj, exists, err := authService.secretInformer.GetStore().GetByKey(strings.Join([]string{authService.namespace, "auth-token"}, "/"))
```
alacuku, line 17: i would save the _auth-token_ key value into a `const` variable.

internal/auth-service/auth-token.go
```diff
@@ -0,0 +1,87 @@
...
+		}, "auth-token")
+		klog.Error(err)
+		return "", err
+	}
+
+	return authService.getTokenFromSecret(secret.DeepCopy())
+}
+
+func (authService *AuthServiceCtrl) createToken() error {
+	_, exists, _ := authService.secretInformer.GetStore().GetByKey(strings.Join([]string{authService.namespace, "auth-token"}, "/"))
```
alacuku, line 44: same as above

internal/auth-service/clusterRole.go
```diff
@@ -0,0 +1,38 @@
...
+					APIVersion: "v1",
+					Kind:       "ServiceAccount",
+					Name:       sa.Name,
+					UID:        sa.UID,
+				},
+			},
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{"discovery.liqo.io"},
```
alacuku, line 25: instead of `discovery.liqo.io` i would use the groupversion variable of the discovery api. something like this `	discoveryv1alpha1.GroupVersion.Group`

internal/auth-service/clusterRole.go
```diff
@@ -0,0 +1,38 @@
...
+				Resources: []string{"peeringrequests"},
+				Verbs:     []string{"create"},
+			},
+			{
+				APIGroups:     []string{"discovery.liqo.io"},
```
alacuku, line 30: same here

internal/auth-service/roleBinding.go
```diff
@@ -0,0 +1,37 @@
...
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      sa.Name,
+				Namespace: sa.Namespace,
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
```
alacuku, line 31: you could use: `rbacv1.SchemeGroupVersion.Group`

internal/auth-service/auth-service.go
```diff
@@ -0,0 +1,116 @@
...
+
+	saInformer := informerFactory.Core().V1().ServiceAccounts().Informer()
+	saInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
+
+	nodeInformer := informerFactory.Core().V1().Nodes().Informer()
+	nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
+
+	secretInformer := informerFactory.Core().V1().Secrets().Informer()
+	secretInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
+
```
alacuku, line 47: i think i'm missing something here. you have three informers but only for the **Secrets** informer you do register valid _EvendHandlerFunctions_. Which is the function of the remaining informers?

aleoli: The other two are used for caching purposes:
* the ServiceAccount one in `serviceAccont.go` at line 60 (this one, I admit, it's a bit useless and I can remove it)
* the Node one in `kubeconfig.go` at line 24

alacuku (approved):
/lgtm

PR #373, opened 2020-11-12
Refactor Build Pipeline

palexster:
# Description

This PR refactors the build pipeline simplifying its configuration and deleting legacy code:

- [x] Fix Doc pipeline bug
- [x] Fix Fork Build


.github/workflows/integration.yml
```diff
@@ -25,78 +64,43 @@ jobs:
...
+        uses: docker/build-push-action@v2
         with:
-          name: liqo/${{ matrix.component }}-ci
-          dockerfile: build/${{ matrix.component }}/Dockerfile
-          push: false
-        if: github.ref != 'refs/heads/master' && github.event.pull_request.head.repo.full_name != 'liqotech/liqo' &&
-            !startsWith(github.ref, 'refs/tags/v')
+          tags: |
+            liqo/${{ matrix.component }}:latest
+            liqo/${{ matrix.component }}:${{ needs.configure.outputs.commit_ref }}
```
giorio94, line 89: Looks like `repo-suffix` is not used.

.github/workflows/integration.yml
```diff
@@ -106,18 +110,21 @@ jobs:
          run: echo "branch_owner=`git log -1 --format='%ae' $commit_ref^!`"
        - name: "Install python dependencies"
          run: pip install requests
-       - name: "Set branch name"
-         run: echo "branch_name=`echo ${GITHUB_REF#refs/*/}`" >> $GITHUB_ENV
```
giorio94, line 110: If this part is deleted, then the steps before could be deleted as well

.github/workflows/integration.yml
```diff
@@ -213,29 +210,37 @@ jobs:
         tar -zxvf kubebuilder_${version}_linux_${arch}.tar.gz
         mv kubebuilder_${version}_linux_${arch} kubebuilder && sudo mv kubebuilder /usr/local/
         export PATH=$PATH:/usr/local/kubebuilder/bin
```
giorio94, line 212: This line could be dropped as (if I am not wrong) it is not persisted to next step

.github/workflows/integration.yml
palexster: Nice catch!

.github/workflows/integration.yml
```diff
@@ -25,99 +64,61 @@ jobs:
...
+      - name: Build (and Publish) ${{ matrix.component }} image
+        uses: docker/build-push-action@v2
         with:
-          name: liqo/${{ matrix.component }}-ci
-          dockerfile: build/${{ matrix.component }}/Dockerfile
-          push: false
-        if: github.ref != 'refs/heads/master' && github.event.pull_request.head.repo.full_name != 'liqotech/liqo' &&
-            !startsWith(github.ref, 'refs/tags/v')
+          tags: |
+            liqo/${{ matrix.component }}:latest
```
giorio94, line 88: Probably the suffix should be added also here

giorio94 (approved):
/lgtm

Just one minor comment inline

.github/workflows/integration.yml
```diff
@@ -25,99 +64,61 @@ jobs:
         - init-pod-mutator
         - pod-mutator
         - peering-request-webhook-init
-        - crdReplicator
+        - crdreplicator
```
mlavacca, line 67: can we use the same naming convention? I think that kebab-case fits better than flat-case, just an opinion, though.

mlavacca (approved):

PR #374, opened 2020-11-12
Southbound driver for vpn backends

alacuku:
# Description
### Liqo-Gateway
The tunnel-operator has been renamed to liqo-gateway. It handles the VPN connections to remote peers through a southbound driver, an interface which defines a set of methods. The solution is flexible and separates the underlying VPN technology from the operator's business core logic. Based on the information of the tunnelendpoints.net.liqo.io resource the correct implementation of the driver is called to establish the VPN connection hence  for different peering clusters different VPN backends can be used. A first implementation of the drives is done for the WireGuard VPN.
Further more the vpn network interface does not live on the host's network namespace where the pod runs but in the pod's network namespace. Not being bound anymore to a specific host, and not running in the host's network, nodePorts/loadBalancers services are used to expose the VPN endpoint to the remote peering clusters. Here are the features introduced for the liqo-gateway by this PR:

- southbound driver for VPN technologies;
- support for WireGuard VPN, a fast and secure solution;
- the network interfaces, and all the needed configuration for liqo network live in the pod's namespace, isolated from the host's network namespace;
- support for geographical peering by exposing the VPN endpoint through a loadBalancer service.

### Liqo-Route
The route-operator has been renamed to liqo-route. The operator has been simplified since the NAT configuration has been moved to liqo-gateway. The VxLAN approach used before to connect the nodes to the gateway has been changed since it presented limitations with the new architecture where the gateway has been moved to a separate namespace. Each host establishes a VPN connection to the the liqo-gateway in order to route network traffic to remote services. Other than configure the overlay network the operator handles the routing rules on each host for the remote peers.


mlavacca (approved):
LGTM

palexster (approved):

Issue #371, opened 2020-11-11
[Feature] Auth Service


aleoli:
Provide an Auth Service to allow users to authenticate before to create a new PeeringRequest

This can enable integration with those k8s services where the system:anonymous user is not available

### Features
- [x] HTTP service - ref #369 
- [x] secret token - ref #369 
- [x] the ForeignCluster wait for the token before to join the remote cluster - ref #398 
- HTTP service discovery
  - [x] LAN - ref #377
  - [x] WAN - ref #413
  - [x] Manual
- [x] ids endpoint -> to get ClusterID, ClusterName and LiqoGuestNamespace - ref #409 
- [x] remove liqo_api mDNS service - ref #409 
- [ ] ~~delete role (?)~~
- [x] expose the HTTP service (Ingress and/or NodePort) - ref #385

PR #377, opened 2020-11-16
discover auth service

aleoli:
# Description

Before to contact the remote auth service we have to discover it. With this pr we will be able to know the IP and the port of this service

## What's new

A new service is added to mDNS discovery, the two services (API server host+port and Auth service host+port) are linked by the same instance id. When the mDNS client discover both of them it is able to create the new ForeignCluster CR

To be able to match services by instance name, this pr adds a new cache map:
* key: string
* value:
  * auth service data
  * API server data

Ref. issue #371


AbakusW (approved):

mlavacca (approved):
/lgtm

PR #381, opened 2020-11-18
fix provider failed

aleoli:
# Description

This pr fixes a problem with the new VirtualKubelet's cache manager. If the manager was not completely ready, some pods were in ProviderFailed status, now the reflector is able to detect the temporary problem and to retry the pod creation


internal/virtualKubelet/node/pod.go
```diff
@@ -127,6 +127,17 @@ func (pc *PodController) handleProviderError(ctx context.Context, span trace.Spa
...
 
+	case metav1.StatusReasonServiceUnavailable, metav1.StatusReasonAlreadyExists:
+		// if the pod creation/update process ends with one of these errors we have not to set the pod in the provider
+		// failed status
+		// StatusReasonAlreadyExists: if the cache is not aligned we can have double pod creation, if that happens
+		// simply ignore second creation
+		// StatusReasonServiceUnavailable: it can be generated by 3 different things, with both it will retried very soon
+		// 1. the cache for the remote cluster is not started yet
+		// 2. no secret referencing the same ServiceAccount of the pod is available in the remote cluster, probably
+		//    because the secret reflection is not completed yet
```
AbakusW, line 138: I think it should be "...by 2 different things...it will be retried" ?

AbakusW (approved):

palexster (approved):

Issue #380, opened 2020-11-18
[Feature] Pod resiliency


mlavacca:
Every time a pod is scheduled in the foreign cluster, there are two different copies of the pod:
* home pod: keeps track of the pod spec and status, it is only a placeholder for having visibility on the offloaded pod
* foreign pod: actual pod which is running.

If the foreign pod dies or is deleted for any reason, it is up to the local cluster to get triggered by that status change and to react accordingly, by re-creating a new local pod that, if scheduled in the remote cluster, will be then offloaded remotely according to the previous paradigm.

We want to provide resiliency to the remote pod, hence, if a pod has been offloaded remotely and the actual remote pod is deleted, then the re-creation of the remote pod will be handled by the foreign cluster.
This behavior would introduce resiliency to the connectivity fault and a self-healing mechanism in the pods' management.

To introduce this feature we have to:
- [x] create, update and delete remote `ReplicaSets` instead of pods ref #383 
- [x] create a new incoming reflector that handles the `ReplicaSets` ref #363 
- [x] provide a mechanism for handling the coexistence of multiple foreign pods when a foreign pod is deleted ref #402 
- [x] handle the deletion of a foreign `Replicaset`, triggering the delete of a home `Pod` ref #411

PR #383, opened 2020-11-19
Replicaset integration in virtual kubelet provider

mlavacca:
# Description

This PR heavily changes the Liqo provider of the virtual kubelet intending to implement the management of foreign `replicasets` instead of `pods`. The CRUD operations until now "pod-oriented" have to be migrated toward an approach "replica-set" oriented, by mapping the operations as follows:

* `CreatePod` has to create a new remote `replicaset`
* `UpdatePod` has to update a remote `replicaset`
* `DeletePod` has to delete a remote `replicaset`

The `replicasets` have one replica each that aims at mapping home pods one by one with foreign pods. The mapping of each home pod with a foreign `pod` is implemented by a label in the replicaset spec `podTemplateSpec` field.

Address #380


pkg/virtualKubelet/forge/apiForger.go
```diff
@@ -0,0 +1,74 @@
...
+	return nil, errors.New("unknown received api")
+}
+
+func ForeignToHome(foreignObj, homeObj runtime.Object, reflectionType string) (runtime.Object, error) {
+	switch foreignObj.(type) {
+	case *corev1.Pod:
+		return forger.podForeignToHome(foreignObj, homeObj, reflectionType)
+	}
+
+	return nil, errors.New("unknown received api")
```
palexster, line 29: This error message is a bit hard to follow.

pkg/virtualKubelet/forge/apiForger.go
```diff
@@ -0,0 +1,74 @@
...
+		return forger.configmapHomeToForeign(homeObj.(*corev1.ConfigMap), foreignObj.(*corev1.ConfigMap))
+	case *discoveryv1beta1.EndpointSlice:
+		return forger.endpointsliceHomeToForeign(homeObj.(*discoveryv1beta1.EndpointSlice), foreignObj.(*discoveryv1beta1.EndpointSlice))
+	case *corev1.Pod:
+		return forger.podHomeToForeign(homeObj, foreignObj, reflectionType)
+	case *corev1.Service:
+		return forger.serviceHomeToForeign(homeObj.(*corev1.Service), foreignObj.(*corev1.Service))
+	}
+
+	return nil, errors.New("unknown received api")
```
palexster, line 44: This error message is a bit "implicit".

pkg/virtualKubelet/forge/pods.go
```diff
@@ -0,0 +1,201 @@
...
+		for _, v := range volumes {
+			if vm.Name == v.Name {
+				volumeMounts = append(volumeMounts, vm)
+			}
+		}
+	}
+	return volumeMounts
+}
+
+func ChangePodIp(newPodCidr string, oldPodIp string) (newPodIp string) {
```
palexster, line 157: I would add a comment to explain this method, due to its importance.

test/e2e/util/util.go
```diff
@@ -124,10 +126,20 @@ func WaitForPodToBeReady(client *kubernetes.Clientset, waitSeconds time.Duration
...
 
-func PodWatcher(client *kubernetes.Clientset, clusterID, namespace, podName string, podReady, stopCh chan bool) {
-	watcher, err := client.CoreV1().Pods(namespace).Watch(context.TODO(), metav1.ListOptions{
-		FieldSelector: fields.OneTermEqualSelector(api.ObjectNameField, podName).String(),
-	})
+func PodWatcher(location string, client *kubernetes.Clientset, clusterID, namespace, podName string, podReady, stopCh chan bool) {
+	var watcher watch.Interface
+	var err error
+
+	if location == "home" {
```
palexster, line 133: I think that a comment there is needed to understand properly the different test-cases.

palexster (approved):

pkg/virtualKubelet/provider/pods.go
```diff
@@ -23,122 +24,119 @@ import (
...
 
-	podTranslated := translation.H2FTranslate(pod, nattedNS)
+	foreignReplicaset := forge.ReplicasetFromPod(foreignPod.(*corev1.Pod))
 
-	apiController, err := p.GetApiController()
-	if err != nil {
-		return err
-	}
-	podTranslated, err = serviceEnv.TranslateServiceEnvVariables(podTranslated, pod.Namespace, nattedNS, apiController.CacheManager())
+	_, err = p.foreignClient.AppsV1().ReplicaSets(foreignReplicaset.Namespace).Create(context.TODO(), foreignReplicaset, metav1.CreateOptions{})
```
aleoli, line 46: Probably we are missing service environment translations

pkg/virtualKubelet/forge/pods.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func forgeVolumes(volumesIn []corev1.Volume) []corev1.Volume {
+	volumesOut := make([]corev1.Volume, 0)
+	for _, v := range volumesIn {
+		if v.ConfigMap != nil || v.EmptyDir != nil || v.DownwardAPI != nil {
+			volumesOut = append(volumesOut, v)
+		}
+		// copy all volumes of type Secret except for the default token
+		if v.Secret != nil && !strings.Contains(v.Secret.SecretName, "default-token") {
```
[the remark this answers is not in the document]
mlavacca: I guess this is the heritage from our missing support to service account secrets. I will remove that condition in a new PR and I will check the correctness of the mechanism, if you agree

pkg/virtualKubelet/provider/pods.go
mlavacca: Nice catch, we were going to miss them.  I introduced them again. Anyway, I have some doubts about that function: what happens if, let's say one over ten services hasn't been replicated yet? The translated pod will have all the service environment variables but one?

pkg/virtualKubelet/forge/pods.go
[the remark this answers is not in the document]
aleoli: Yes, fine for me

pkg/virtualKubelet/provider/pods.go
aleoli: Yes, good point, but I think that is the same for local deployment. If the pod comes before that a service is available it can't have it in env var but has to discover it with DNS

I think that in all the cases you have to be sure that the service was available a lot of time before the creation of the pod

mlavacca: Yes, I suppose there is the same issue even in local deployments.

aleoli (approved):

PR #398, opened 2020-11-30
Introduce dedicated authentication service

aleoli:
# Description

This pr changes the way to get the remote role, from the system:anonymous role to the one provided by the AuthService

### Changes

* Adds in the cluster config a new section about auth settings, where is possible to enable/disable auth with an empty token
* In the ForeignCluster status is now present a field to indicate the phase of the "getting role" process
* CaData secret is no more needed

### 2 auth methods

1. with an empty token (the default)
    * when the remote cluster is discovered we try to get a role sending an empty token, the join is automatic, and (for the user) there is no change from the previous peer workflow
2. with a token (you have to disable empty token in the ClsuterConfig > authConfig > allowEmptyToken)
    * this workflow is simple as getting the required data from the remote cluster
```bash
# get the token from the remote cluster

token=$(kubectl get secret -n liqo auth-token --kubeconfig "$K2" -o jsonpath="{.data.token}" | base64 -d)
echo -e "Token:\t\t$token"

# get remote cluster id

clusterId=$(kubectl get configmap -n liqo cluster-id --kubeconfig "$K2" -o jsonpath="{.data.cluster-id}")
echo -e "ClusterId:\t$clusterId"
```

    * and paste them to the local cluster

```bash
# create local secret

secret_name="remote-token"

kubectl create secret generic "$secret_name" \
  -n liqo \
  --kubeconfig "$K1" \
  --from-literal=token="$token"

# label it

kubectl label secret "$secret_name" \
  -n liqo \
  --kubeconfig "$K1" \
  cluster-id="$clusterId" \
  liqo-auth-token=""

# patch foreign cluster CR

kubectl patch foreignclusters "$clusterId" \
  --patch '{"status":{"authStatus":"Pending"}}' \
  --type 'merge' \
  --kubeconfig "$K1"
```

Ref. issue #371 

## Note

The tests on discovery will be completely rewritten in a next pr


apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -45,6 +45,15 @@ const (
...
+
+const (
+	AuthStatusPending      AuthStatus = "Pending"
+	AuthStatusAccepted     AuthStatus = "Accepted"
+	AuthStatusRefused      AuthStatus = "Refused"
+	AuthStatusEmptyRefused AuthStatus = "EmptyRefused"
+)
+
 const (
 	LastUpdateAnnotation string = "LastUpdate"
```
mlavacca, line 58:
```suggestion
	LastUpdateAnnotation string = "discovery.liqo.io/LastUpdate"
```

deployments/liqo/templates/clusterconfig.yaml
```diff
@@ -30,8 +30,10 @@ spec:
     enableDiscovery: true
     name: MyLiqo
     port: 6443
-    service: _liqo._tcp
+    service: _liqo_api._tcp
     ttl: 90
+  authConfig:
+    allowEmptyToken: true
```
mlavacca, line 36: Why are we setting this flag to true?

internal/auth-service/auth-service-config.go
```diff
@@ -0,0 +1,34 @@
+package auth_service
+
+import (
+	configv1alpha1 "github.com/liqotech/liqo/apis/config/v1alpha1"
+	"github.com/liqotech/liqo/pkg/clusterConfig"
+)
+
+func (authService *AuthServiceCtrl) GetAuthServiceConfig(kubeconfigPath string) error {
+	waitFirst := make(chan bool)
```
mlavacca, line 9: Since we are using this chan only once we could use a `chan struct{}`. In order to Communicate that the event has happened, you can close it. In this case, line 19 wouldn't be needed anymore.

internal/auth-service/auth-service-config.go
```diff
@@ -0,0 +1,34 @@
...
+	<-waitFirst
+	close(waitFirst)
+
+	return nil
+}
+
+func (authService *AuthServiceCtrl) handleConfiguration(config configv1alpha1.AuthConfig) {
+	authService.configMutex.Lock()
+	defer authService.configMutex.Unlock()
+	authService.config = config
```
mlavacca, line 27: I think you should use `deepcopy`, because of the pointers in the AuthConfig struct.

internal/auth-service/auth-service-config.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func (authService *AuthServiceCtrl) GetConfig() configv1alpha1.AuthConfig {
+	authService.configMutex.RLock()
+	defer authService.configMutex.RUnlock()
+	return authService.config
```
mlavacca, line 33: The same as above

internal/discovery/authData.go
```diff
@@ -27,6 +36,13 @@ func (authData *AuthData) IsComplete() bool {
 	return authData.address != "" && authData.port > 0
 }
 
+func (authData *AuthData) GetUrl() string {
+	if authData.isTest {
```
mlavacca, line 40: We should avoid distinguishing between production and testing code.

internal/discovery/foreign-cluster-operator/auth.go
```diff
@@ -0,0 +1,204 @@
...
+//
+// first of all, if the status is pending we can try to get a role with an empty token, if the remote cluster allows it
+// the status will become accepted
+//
+// if our status is EmptyRefused, this means that the remote cluster refused out request with the empty token, so we will
+// wait to have a token to ask for the role again
+//
+// while we are waiting for that secret this function will return no error, but an empty client
+func (r *ForeignClusterReconciler) getRemoteClient(fc *discoveryv1alpha1.ForeignCluster, gv *schema.GroupVersion) (*crdClient.CRDClient, error) {
+	if strings.HasPrefix(fc.Spec.AuthUrl, "fake://") {
```
mlavacca, line 38: The same as above, try using interfaces

internal/discovery/foreign-cluster-operator/auth.go
```diff
@@ -0,0 +1,204 @@
...
+		kubeconfigStr, err := r.askRemoteIdentity(fc)
+		if err != nil {
+			klog.Error(err)
+			return nil, err
+		}
+		if kubeconfigStr == "" {
+			return nil, nil
+		}
+		_, err = kubeconfig.CreateSecret(r.crdClient.Client(), r.Namespace, kubeconfigStr, map[string]string{
+			"cluster-id":           fc.Spec.ClusterIdentity.ClusterID,
```
mlavacca, line 77: Please parameterize these labels and use a liqo prefix (e.g., `liqo.io/cluster-id`, `liqo.io/remote-identity`)

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -473,7 +437,70 @@ func (r *ForeignClusterReconciler) checkJoined(fc *discoveryv1alpha1.ForeignClus
...
+				Resource: string(apiv1.ResourceServices),
+			}, "auth-service")
+			klog.Error(err)
+			return "", err
+		}
+
+		port = fmt.Sprintf("%v", svc.Spec.Ports[0].NodePort)
+	}
+
+	_, ok := os.LookupEnv("TEST")
```
mlavacca, line 499: Same as above regarding testing

mlavacca (changes requested):

deployments/liqo/templates/clusterconfig.yaml
aleoli: We set it to true in order to have the same "automatic peer" behavior. With an empty token allowed the clusters will join automatically as before the AuthService

We can disable it to not allow automatic join, but each time the user has to do manual actions to provide the token, this choice makes simpler the first user experience

internal/auth-service/auth-service-config.go
aleoli: I don't think so, because AuthConfig is a struct passed by value with only a boolean as filed at the moment

But I'm adding the DeepCopy for possible future additions in this struct

deployments/liqo/templates/clusterconfig.yaml
mlavacca: Gotcha, thanks!

internal/auth-service/auth-service-config.go
mlavacca: You are right, I did not check the struct :grimacing: . Anyway, if we plan to add more fields, a `deepcopy` would be better (I already faced the lacking of `deepcopy` bug and it was very hard to troubleshoot)

AbakusW (changes requested):

AbakusW (approved):

mlavacca (approved):

PR #401, opened 2020-12-01
LiqoAgent dynamic lists discovered peers

AbakusW:
# Description

This PR provides the *Liqo Agent* the ability to show a dynamic list of the available peers known by the home cluster.

* Each entry currently shows 
  * the **ClusterName** of the *peer* cluster, i.e. a user-defined parameter to characterize its own Liqo instance. It changes accordingly to variations to this info made by the peer.
  * a nested "status" element showing the *peer* **ClusterID** which is the unique identifier, useful to avoid impersonation. 

# Test

In the *Indicator* package the **EventTester** component of the *GuiProvider* has been added in order to better test the changes performed by single callbacks of the Listeners.

### DIFF
--- a/internal/tray-agent/agent/client/crdClients.go
+++ b/internal/tray-agent/agent/client/crdClients.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"errors"
+	"fmt"
 	"github.com/liqotech/liqo/pkg/crdClient"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
@@ -16,6 +17,8 @@ const (
 	CRClusterConfig CustomResource = "clusterconfigs"
 	//CRAdvertisement is the resource id for the Advertisement CRD.
 	CRAdvertisement CustomResource = "advertisements"
+	//CRForeignCluster is the resource id for the ForeignCluster CRD.
+	CRForeignCluster CustomResource = "foreignclusters"
 )
 
 //customResources contains all the registered CustomResource managed by the AgentController.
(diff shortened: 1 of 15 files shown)


palexster (approved):

internal/tray-agent/agent/client/crdClients.go
```diff
@@ -35,28 +39,29 @@ type crdManager struct {
...
 		return errors.New("no kubeconfig provided")
 	}
 	//creation of each single CRDController
 	var err error
 	var crdCtrl *CRDController
-	// - CLUSTERCONFIG
-	crdCtrl, err = createClusterConfigController(kubeconfig)
-	if err != nil {
-		return errors.New("connection error on clusterconfigs client creation")
+	for _, cr := range customResources {
```
mlavacca, line 51: Why are you using this loop that iterates over a slice and switches on the elements? Can't you simply call the functions one after the other?

AbakusW: I thought was a good idea to reuse the common code for each 'creation', like (60-63), instead of always rewriting the same checks.

mlavacca: I think it is a bit confusing, it seems we are adding complexity where the workflow is pretty simple.

AbakusW: I re-checked the code: I moved to the iteration mechanism also to use the same scheme of the tests. 

The *customResources* variable, similarly to notifyChannelNames, is a slice containing all the CRDs currently managed by the AgentController and the *TestAgentControllerComponentsReadiness* iterates on that, checking the respective clients are active. I thought that using the same approach in this init phase could have compensated a mechanism maybe a little convoluted.

If you think it's still worth reverting to sequential calls, I can do it without problems

mlavacca: yes please, try using the clearest and simplest approach if there is no complexity.

mlavacca (approved):

palexster (approved):

PR #402, opened 2020-12-04
Foreign pod deletion transitory state

mlavacca:
# Description

Whenever a foreign pod is deleted in the foreign cluster, a new pod is created by the foreign `replicaset`. This situation leads to a transitory state in which there are two foreign pods mapped to the same local one:
* old marked for deletion pod in `terminating` status
* new pod in `pending`, `containerCreating`, or `running` status

If we simply reflect the status of the foreign pod to the local one, there are fluctuations between the old pod and the new one. To face this issue, when a pod is marked for deletion, we blacklist that pod in the incoming reflector, such that the old pod will not be handled anymore by the reflector, except for the deletion routine, that will remove the object from the blacklist map, freeing the occupied space, and allowing the mechanism to be efficient and sustainable at scale.

Address #380


pkg/virtualKubelet/apiReflection/reflectors/incoming/pods.go
```diff
@@ -149,3 +162,21 @@ func (r *PodsIncomingReflector) CleanupNamespace(namespace string) {
 		}
 	}
 }
+
+func (r *PodsIncomingReflector) isAllowed(ctx context.Context, obj interface{}) bool {
```
palexster, line 166: I think that a comment to this method may help to understand this part.

pkg/virtualKubelet/apiReflection/reflectors/blacklists.go
```diff
@@ -0,0 +1,17 @@
+package reflectors
+
+import (
+	apimgmt "github.com/liqotech/liqo/pkg/virtualKubelet/apiReflection"
+)
+
+type blackList map[string]struct{}
+
```
palexster, line 8: I think a comment to this function would help to understand how the blacklist works.

pkg/virtualKubelet/apiReflection/reflectors/incoming/pods.go
```diff
@@ -111,8 +121,11 @@ func (r *PodsIncomingReflector) preAddUpdate(foreignPod *corev1.Pod) *corev1.Pod
 	return homePod.(*corev1.Pod)
 }
 
-// TODO: add here custom unavailable status
-func (r *PodsIncomingReflector) PreDelete(_ interface{}) interface{} {
+func (r *PodsIncomingReflector) PreDelete(obj interface{}) interface{} {
```
palexster, line 124: A short comment to specify what is done here with the blacklist may be useful.

palexster (approved):
/lgtm

AbakusW (approved):

PR #406, opened 2020-12-07
Local config file for Liqo Agent

AbakusW:
# Description

This PR introduces for Liqo Agent to load and store some configuration data on a *.yaml* config file on the local filesystem, inside the *liqo* directory where other resources are already present (e.g. the desktop icons).

This feature brings some benefits:

- [x] The Agent can acquire some parameters that would be otherwise inaccessible, e.g. the indication of a specific kubeconfig filepath directly injected by the Liqo installer in the config file.

- [x] Persistency of the information: the config file is stille accessible after a Turn-Off of the Agent, allowing a stateful configuration

- [x] easy configuration and fields expansion thanks to yaml format

- further developments
  - Switching between clusters and ApiServers can be easier

  - users profiles

##### Note

Notifications setting will be moved in the config file in a further PR.

# Tests

unit tests and local testing

### DIFF
--- a/internal/tray-agent/agent/client/agent-client.go
+++ b/internal/tray-agent/agent/client/agent-client.go
@@ -19,6 +19,8 @@ const (
 	//EnvLiqoKConfig defines the env var containing the path of the kubeconfig file of the
 	//cluster associated to Liqo Agent.
 	EnvLiqoKConfig = "LIQO_KCONFIG"
+	//EnvLiqoPath defines the env var containing the path of the root directory of the Liqo Agent on the local file system.
+	EnvLiqoPath = "LIQO_PATH"
 )
 
 //AgentController singleton.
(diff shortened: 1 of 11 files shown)


internal/tray-agent/agent/client/agent-client.go
```diff
@@ -104,16 +106,19 @@ func (ctrl *AgentController) StopCaches() {
 	}
 }
 
-/*acquireKubeconfig sets the LIQO_KCONFIG env variable.
+/*acquireKubeconfig sets the EnvLiqoKConfig env variable.
```
palexster, line 109: Environment variable should be uppercase and snake case

AbakusW: yes, you're right. In this case *EnvLiqoKConfig* is the name of the Golang const that contains the actual environment variable name. I made this change in the comment so that when developers display this documentation in a IDE or in the [golang doc site](https://pkg.go.dev/), this information automatically contains a link to the constant itself.

AbakusW: if you think this could be misleading, I could rephrase as *"sets the env variable named after EnvLiqoKConfig"* or something like that

mlavacca (approved):

palexster (approved):

PR #409, opened 2020-12-10
Add ClusterIdentity endpoint to authentication service

aleoli:
# Description

This PR adds an HTTP endpoint where to get the foreign ClusterIdentity (ClusterID and ClusterName), in order to be removed from the mDNS/DNS TXT packets

Ref. issue #371

### DIFF
--- a/internal/auth-service/idsHttpHandler.go
+++ b/internal/auth-service/idsHttpHandler.go
@@ -1,0 +1,40 @@
+package auth_service
+
+import (
+	"encoding/json"
+	"github.com/julienschmidt/httprouter"
+	"github.com/liqotech/liqo/pkg/auth"
+	"k8s.io/klog"
+	"net/http"
+)
+
+// this HTTP handler returns home cluster information to the foreign clusters that are asking for them
+// it returns a JSON encoded ClusterInfo struct with the following fields:
+// - clusterID		-> the id of the home cluster
+// - clusterName	-> the custom name for the home cluster (to be displayed in GUIs)
+// - guestNamespace	-> the namespace where to create secrets and resources to be shared with the home cluster
+func (authService *AuthServiceCtrl) ids(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	idsResponse := authService.getIdsResponse()
+
+	res, err := json.Marshal(idsResponse)
+	if err != nil {
+		klog.Error(err)
+		authService.handleError(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	if _, err = w.Write(res); err != nil {
+		klog.Error(err)
+		return
+	}
+}
+
+func (authService *AuthServiceCtrl) getIdsResponse() *auth.ClusterInfo {
+	conf := authService.GetDiscoveryConfig()
+	return &auth.ClusterInfo{
+		ClusterID:      authService.clusterId.GetClusterID(),
+		ClusterName:    conf.ClusterName,
+		GuestNamespace: auth.LiqoGuestNamespace,
+	}
+}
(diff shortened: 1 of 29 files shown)


internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -473,7 +437,70 @@ func (r *ForeignClusterReconciler) checkJoined(fc *discoveryv1alpha1.ForeignClus
 // check if the error is due to a TLS certificate signed by unknown authority
 func isUnknownAuthority(err error) bool {
 	var err509 x509.UnknownAuthorityError
-	return goerrors.As(err, &err509)
+	var err509Hostname x509.HostnameError
+	return goerrors.As(err, &err509) || goerrors.As(err, &err509Hostname)
+}
+
+func (r *ForeignClusterReconciler) getHomeAuthUrl() (string, error) {
+	address, _ := os.LookupEnv("APISERVER")
```
palexster, line 445: Can we use the default kubernetes service environment variable?

aleoli: no, this variable has an internal IP of the service, but here we need an address that can be reachable from the foreign cluster

internal/discovery/foreign-cluster-operator/auth.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/liqotech/liqo/pkg/auth"
 	"github.com/liqotech/liqo/pkg/crdClient"
 	"github.com/liqotech/liqo/pkg/discovery"
+	discoveryPkg "github.com/liqotech/liqo/pkg/discovery"
```
mlavacca, line 14: I think `somethingPkg` is pretty strange as an import name (it is obviously a package that you are importing).

internal/auth-service/idsHttpHandler.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"encoding/json"
+	"github.com/julienschmidt/httprouter"
+	"github.com/liqotech/liqo/pkg/auth"
+	"k8s.io/klog"
+	"net/http"
+)
+
+func (authService *AuthServiceCtrl) ids(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
mlavacca, line 11: can you add a comment about what this function serves to the client?

pkg/auth/const.go
```diff
@@ -0,0 +1,5 @@
+package auth
+
+const (
+	LiqoGuestNamespace = "liqo" // TODO: move to "liqo-public"
```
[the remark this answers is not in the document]
aleoli: Yes, this namespace will be changed when we will move all foreign-created secrets in the liqo-public namespace, there some work to do. For example, we have to add this namespace in the installer/uninstaller

internal/discovery/foreign-cluster-operator/auth.go
aleoli: I removed where possible, there is a name collision in some packages

mlavacca: do these collisions derive from importing both `internal/discovery` and `pkg/discovery` in the same files?

pkg/auth/const.go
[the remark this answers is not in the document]
mlavacca: :+1:

internal/discovery/foreign-cluster-operator/auth.go
aleoli: yes, and because in the discovery internals the object that owns the functions has also "discovery" name

mlavacca (approved):

palexster (approved):

PR #412, opened 2020-12-14
protection for concurrent operations on Indicator components

AbakusW:
# Description

This PR adds protections for concurrent operations on:
* MenuNode attributes
* resources of the Indicator that handle graphic operations, e.g. change of the tray icon, change of the tray label and the launch of desktop notifications

### DIFF
--- a/internal/tray-agent/app-indicator/indicator.go
+++ b/internal/tray-agent/app-indicator/indicator.go
@@ -29,6 +29,16 @@ const (
 	IconLiqoNil
 )
 
+//graphicResource defines a graphic interaction handled by the Indicator.
+type graphicResource int
+
+//graphicResource currently handled by the Indicator.
+const (
+	resourceIcon graphicResource = iota
+	resourceLabel
+	resourceDesktop
+)
+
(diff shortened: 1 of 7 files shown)


internal/tray-agent/app-indicator/indicator.go
```diff
@@ -247,12 +264,18 @@ func (i *Indicator) SetMenuTitle(title string) {
...
+	gr.RLock()
+	defer gr.RUnlock()
 	return i.icon
 }
 
 //SetIcon sets the Indicator tray icon. If 'ico' is not a valid argument or ico == IconLiqoNil,
 //SetIcon does nothing.
 func (i *Indicator) SetIcon(ico Icon) {
+	gr := i.graphicResource[resourceIcon]
+	gr.Lock()
```
palexster, line 277: Can we lock just the status write operation instead the whole function body?

AbakusW: yes you are right!

mlavacca (approved):

palexster (approved):

PR #413, opened 2020-12-14
Move the DNS/WAN discovery to use the new AuthService

aleoli:
# Description

This PR changes the way to do WAN discovery with DNS

Now, we have to discover the AuthService to get foreign cluster information. SRV records are enow pointing to the ingress/service that is exposing the AuthService on the foreign cluster

The registration of the home cluster is simpler than before, no TXT record is needed because those data are retrieved from the new service as for LAN discovery

ref. issue #371


test/unit/discovery/dns_test.go
```diff
@@ -26,7 +24,7 @@ func TestDns(t *testing.T) {
 // ------
 // tests if is able to get txtData from local DNS server
 func testDns(t *testing.T) {
-	targetTxts := []*discovery.TxtData{
+	/*targetTxts := []*discovery.TxtData{
```
AbakusW, line 27: Can these tests be directly removed since they refer to the old type? (to respect the "no commented code" policy)

test/unit/discovery/mdns_test.go
```diff
@@ -13,12 +12,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/klog"
 	"strconv"
-	"strings"
 	"testing"
 	"time"
 )
 
-var txtData discovery.TxtData
+//var txtData discovery.TxtData
```
AbakusW, line 19: other commented tests

test/unit/discovery/dns_test.go
aleoli: Tests will be completely rewritten in the next pr, this is a reminder for me of which kind of test we were doing

AbakusW (approved):

internal/discovery/search-domain-operator/search-domain-controller.go
```diff
@@ -49,68 +48,15 @@ func (r *SearchDomainReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
-	fcs := r.DiscoveryCtrl.UpdateForeignWAN(txts, sd)
-	if len(fcs) > 0 {
-		// new FCs added, so update the list
-		AddToList(sd, ForeignClustersToObjectReferences(fcs))
-		update = true
-	}
-
-	// check deletion, ForeignClusters that are no more in DNS list will be deleted
-	toDelete, err := r.CheckForDeletion(sd, txts)
+	authData, err := Wan(r.DnsAddress, sd.Spec.Domain)
```
palexster, line 51: Wan as function name is a bit hard to understand...

internal/discovery/search-domain-operator/wan.go
```diff
@@ -6,12 +6,11 @@ import (
 	"github.com/miekg/dns"
 	"k8s.io/klog"
 	"net"
-	"strconv"
 	"time"
 )
 
-func Wan(dnsAddr string, name string) ([]*discovery.TxtData, error) {
-	txtData := []*discovery.TxtData{}
+func Wan(dnsAddr string, name string) ([]*discovery.AuthData, error) {
```
palexster, line 12: Missing comment to describe the behavior of this method.

internal/discovery/foreign.go
```diff
@@ -132,11 +140,13 @@ func (discovery *DiscoveryCtrl) createForeign(data *discoveryData, trustMode dis
 				UID:        sd.UID,
 			},
 		}
+		if fc.Labels == nil {
+			fc.Labels = map[string]string{}
+		}
+		fc.Labels["searchdomain"] = sd.Name
```
mlavacca, line 146: can you parameterize this label key? (I would suggest `discovery.liqo.io/searchdomain`)

internal/discovery/foreign.go
```diff
@@ -38,26 +38,34 @@ func (discovery *DiscoveryCtrl) UpdateForeignLAN(data *discoveryData, trustMode
 	}
 }
 
-func (discovery *DiscoveryCtrl) UpdateForeignWAN(data []*TxtData, sd *v1alpha1.SearchDomain) []*v1alpha1.ForeignCluster {
+func (discovery *DiscoveryCtrl) UpdateForeignWAN(data []*AuthData, sd *v1alpha1.SearchDomain) []*v1alpha1.ForeignCluster {
```
mlavacca, line 41: could you provide a brief comment that describes what this function is intended to do?

pkg/discovery/const.go
```diff
@@ -4,6 +4,7 @@ const (
 	ClusterIdLabel      = "liqo.io/cluster-id"
 	AuthTokenLabel      = "liqo.io/auth-token"
 	RemoteIdentityLabel = "liqo.io/remote-identity"
+	DiscoveryTypeLabel  = "liqo.io/discovery-type"
```
mlavacca, line 7: what do you think about adding the `discovery` prefix to all these labels?

aleoli: Do you mean something like "discovery.liqo.io/..."?

mlavacca: yep

mlavacca: if you think it is appropriate for these labels of course

mlavacca (approved):

PR #414, opened 2020-12-14
Home pods finalizer

mlavacca:
# Description

This PR addresses the following concurrency problem in the `pods` management inside the virtual-kubelet: if a pod is created and suddenly deleted (e.g., quick scale up and down), a strange situation can happen:
* the home pod has been bound to the virtualKubelet
* the foreign replicaset hasn't been created yet (the create event is enqueued but not handled yet)
* the home pod is deleted
* the delete event processing of the home Pod fails while trying to delete the foreign replicaset (not existing yet)
* the create event of the home pod is then handled by the vk, leading to an inconsistency: the home pod has been deleted, the foreign replicaset has just been created.

In order to avoid this situation, the vk before creating the remote replicaset adds a finalizer to the home pod. If this operation succeeds (i.e., the home Pod still exists, if the deletion timestamp is set, finalizers addition is not allowed by the apiserver), the foreign replicaset is created. The deletion of the home pod is then allowed only by the incoming reflector of the virtual kubelet that whenever a replicaset is deleted patches the corresponding home pod, removing the finalizer and allowing its deletion.

### DIFF
--- a/pkg/virtualKubelet/provider/pods_test.go
+++ b/pkg/virtualKubelet/provider/pods_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"flag"
-	"github.com/liqotech/liqo/internal/virtualKubelet/node"
 	apimgmt "github.com/liqotech/liqo/pkg/virtualKubelet/apiReflection"
 	test2 "github.com/liqotech/liqo/pkg/virtualKubelet/apiReflection/controller/test"
 	"github.com/liqotech/liqo/pkg/virtualKubelet/forge"
@@ -24,7 +23,7 @@ import (
 
 var _ = Describe("Pods", func() {
 	var (
-		provider              node.PodLifecycleHandler
+		provider              *LiqoProvider
 		namespaceMapper       namespacesMapping.MapperController
 		namespaceNattingTable *test.MockNamespaceMapper
 		foreignClient         kubernetes.Interface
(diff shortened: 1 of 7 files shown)


pkg/virtualKubelet/provider/pods_test.go
```diff
@@ -60,18 +59,21 @@ var _ = Describe("Pods", func() {
...
 
-			It("create pod", func() {
-				err := provider.CreatePod(context.TODO(), pod)
-				Expect(err).NotTo(HaveOccurred())
-				rs, err := foreignClient.AppsV1().ReplicaSets("homeNamespace-natted").Get(context.TODO(), "testObject", metav1.GetOptions{})
-				Expect(err).NotTo(HaveOccurred())
-				Expect(rs.Name).To(Equal(pod.Name))
-				Expect(rs.Namespace).To(Equal("homeNamespace-natted"))
-			})
+			/*
```
palexster, line 65: Should we remove this test or change the clients to make this test pass?

pkg/virtualKubelet/provider/provider.go
```diff
@@ -26,7 +26,7 @@ type LiqoProvider struct { // nolint:golint]
 
 	advClient     *crdClient.CRDClient
 	tunEndClient  *crdClient.CRDClient
-	homeClient    *crdClient.CRDClient
+	nntClient     *crdClient.CRDClient
```
palexster, line 29: nntClient stands for?

pkg/virtualKubelet/provider/pods_test.go
mlavacca: To make this test pass we need to use a standard Kubernetes client instead of the CRDClient (because it is mockable). This job will be done in the next slight refactoring regarding also the NodeController. I commented this test because it will be useful (and unchanged) when we will change the client.

pkg/virtualKubelet/provider/provider.go
mlavacca: NamespaceNattingTableClient, too long :grinning:

mlavacca: anyway, all the three local clients will be replaced by a single dynamic client ASAP

pkg/virtualKubelet/provider/pods_test.go
palexster: It is ok for me if you plan to do so in a very next PR, otherwise I think we should drop dead-code, git will take care of the code for us.

palexster (approved):

AbakusW (approved):

PR #415, opened 2020-12-16
Test on Managed Kubernetes

aleoli:
# Description

This PR includes some small changes needed in the Helm Chart and in the Installer to make Liqo work on managed Kubernetes clusters

Users can now set some environment variables to set the API server hostname and the AuthService hostname. If the AuthService hostname is set this service is exposed as ingress, and only as a nodeport

# How Has This Been Tested?

- [x] on AKS managed clusters with Kubernetes version 1.19.1, setting the following variables
  * POD_CIDR=10.244.0.0/16
  * SERVICE_CIDR=10.0.0.0/16
  * LIQO_INGRESS_CLASS=addon-http-application-routing
  * LIQO_APISERVER_ADDR=test-liqo-1-dns-4b5279e4.hcp.westeurope.azmk8s.io
  * LIQO_AUTHSERVER_ADDR=auth.319075fecaae46988083.westeurope.aksapp.io
- [x] locally on kind to test that the old installation was still working

### DIFF
--- a/deployments/liqo/subcharts/authService/templates/auth-service-deploy.yaml
+++ b/deployments/liqo/subcharts/authService/templates/auth-service-deploy.yaml
@@ -140,14 +140,16 @@ spec:
         run: auth-service
     spec:
       serviceAccountName: auth-service-sa
+      {{ if not .Values.ingress.enable }}
       initContainers:
         - name: cert-creator
-          image: nginx
+          image: nginx:1.19
           volumeMounts:
             - mountPath: '/certs'
               name: certs
           command: [ "/bin/sh" ]
           args: [ "-c", 'openssl req -x509 -subj "/C=IT/ST=Turin/O=Liqo" -nodes -days 365 -newkey rsa:4096 -keyout /certs/key.pem -out /certs/cert.pem' ]
+      {{ end }}
       containers:
         - image: {{ .Values.image.repository }}{{ .Values.global.suffix | default .Values.suffix }}:{{ .Values.global.version | default .Values.version }}
           name: auth-service
@@ -159,19 +161,24 @@ spec:
           - "--resyncSeconds"
           - "30"
           - "--listeningPort"
+          {{ if .Values.ingress.enable }}
           - "5000"
+          {{ else }}
+          - "443"
+          - "--useTls"
+          {{ end }}
           env:
             - name: POD_NAMESPACE
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
-            {{ if .Values.apiServerIp }}
+            {{ if .Values.global.apiServer.ip }}
             - name: APISERVER
-              value: {{ .Values.apiServerIp }}
+              value: "{{ .Values.global.apiServer.ip }}"
             {{ end }}
-            {{ if .Values.apiServerPort }}
+            {{ if .Values.global.apiServer.port }}
             - name: APISERVER_PORT
-              value: {{ .Values.apiServerPort }}
+              value: "{{ .Values.global.apiServer.port }}"
             {{ end }}
           resources:
             limits:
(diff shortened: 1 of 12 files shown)


deployments/liqo/subcharts/authService/templates/auth-service-deploy.yaml
```diff
@@ -201,5 +208,39 @@ spec:
...
+      {{ if .Values.ingress.enable }}
+      port: 5000
       targetPort: 5000
+      {{ else }}
+      port: 443
+      targetPort: 443
+      {{ end }}
+---
+{{ if .Values.ingress.enable }}
+apiVersion: networking.k8s.io/v1beta1
```
palexster, line 220: v1beta1 is deprecated, can we use v1 ingresses?

internal/peering-request-operator/broadcaster.go
```diff
@@ -41,6 +42,25 @@ func GetBroadcasterDeployment(request *discoveryv1alpha1.PeeringRequest, nameSA
 		remoteSA,
 	}
 
+	env := []v1.EnvVar{
```
palexster, line 45: Could you please add some comments here to discuss why those variables have been added?

alacuku (approved):
/lgtm

deployments/liqo/subcharts/authService/templates/auth-service-deploy.yaml
```diff
@@ -195,11 +202,50 @@ metadata:
...
+                port:
+                  number: 5000
+            path: /
+            pathType: Prefix
+      {{ if (.Values.ingress.host) }}
+      host: {{ .Values.ingress.host }}
+      {{ end }}
+  tls:
+    - hosts:
+      - {{ .Values.ingress.host }}
```
palexster, line 250: host is not protected by the same if present in http field.

aleoli: thanks, now it is protected

palexster (approved):

install.sh
```diff
@@ -501,6 +518,24 @@ function configure_installation_variables() {
 				"Please, manually specify it with 'export SERVICE_CIDR=...' before executing again this script"
 	fi
 	info "[INSTALL] [CONFIGURE]" "Service CIDR: ${SERVICE_CIDR}"
+
+	# Set variables for kubernetes ingress
+	if [ -z "${LIQO_INGRESS_CLASS:-}" ] || [ -z "${LIQO_AUTHSERVER_ADDR:-}" ]; then
```
giorio94, line 523: I would avoid forcing to specify an ingress class. In case the parameter is not present, you could just omit the annotation and it would use the default ingress controller (e.g. in our cluster on-premise)

deployments/liqo/subcharts/authService/templates/auth-service-deploy.yaml
```diff
@@ -140,6 +140,7 @@ spec:
         run: auth-service
     spec:
       serviceAccountName: auth-service-sa
+      {{ if not .Values.ingress.enable }}
       initContainers:
         - name: cert-creator
           image: nginx
```
giorio94, line 146: I know this is not part of this PR, but I would avoid using an image without a tag, as it forces the download and increases the start-up time

deployments/liqo/subcharts/authService/templates/auth-service-deploy.yaml
```diff
@@ -159,19 +161,24 @@ spec:
           - "--resyncSeconds"
           - "30"
           - "--listeningPort"
+          {{ if .Values.ingress.enable }}
           - "5000"
+          {{ else }}
+          - "443"
```
giorio94, line 167: Is it necessary to use two different ports? I feel it increases the complexity (e.g. in the service definition).

giorio94 (approved):
Just a couple of minor comments about the installer

deployments/liqo/subcharts/authService/templates/auth-service-deploy.yaml
aleoli: No, it's not necessary, but I think that it's better because if we expose the service with a non-TLS backend on port 443 some ingress controller may intend it as a TLS backend [(traefik docs)](https://doc.traefik.io/traefik/v1.7/configuration/backends/kubernetes/#tls-communication-between-traefik-and-backend-pods)

Vice versa, exposing a TLS service on a non 443 port can be strange and hard to understand

Note: we expose the pod port with TLS if we are exposing it with a NodePort, without TLS if we are using ingresses

giorio94: Ok, got it

PR #420, opened 2020-12-18
Status Menunode synchronization with peerings information

AbakusW:
# Description

This PR operates a refactoring on the *Status* MenuNode (which shows some metrics on the tray menu) and introduces the possibility for the Agent to manage the status of active peerings (read only).

* STATUS MenuNode
  * code refactoring for a cleaner and more efficient workflow
  * added an explicit *peers* counter to aggregate all data inside the Status
  * the tray agent *label* (small text beside the tray icon) is now used to dynamically display the number of active peerings (with a distinction between **incoming** and **outgoing** ones)

* Peerings management (read only)
  * The AgentController now handles also events on the *ForeignCluster* CRD regarding the establishment and disconnection of peerings
  * The Indicator listens to these two kinds of events and reacts by:
    * notifying the event (acording to current notification settings)
    * updating STATUS data
    * signalling the event inside the "*peers* list" for the corresponding peer; currently a ✔️ is used to indicate whether a peering in a given direction is active
    * the function handlers regarding the Advertisements have been removed, since the updates about peerings are now acquired from the ForeignCluster

### NOTE: 

The peerings feature implemented in this PR is an early version. In a very next PR there will be a code refactoring for the **NotifyChannel**s and Indicator **Listener**s, thus enabling a more efficient and compact handling of these events.

# How Has This Been Tested?

Since this is not a definitive version and a relevant portion of code will be soon updated, the tests only cover the STATUS refactoring. New tests will be added in the updating PR. Some live tests have been conducted locally.

### DIFF
--- a/internal/tray-agent/agent/client/advertisementController.go
+++ b/internal/tray-agent/agent/client/advertisementController.go
@@ -19,33 +19,6 @@ func createAdvertisementController(kubeconfig string) (*CRDController, error) {
 	return controller, nil
 }
 
-//advertisementAddFunc is the ADD event handler for the Advertisement CRDController.
-func advertisementAddFunc(obj interface{}) {
-	newAdv := obj.(*advertisementApi.Advertisement)
-	if newAdv.Status.AdvertisementStatus == advertisementApi.AdvertisementAccepted {
-		agentCtrl.NotifyChannel(ChanAdvAccepted) <- newAdv.Name
-	} else {
-		agentCtrl.NotifyChannel(ChanAdvNew) <- newAdv.Name
-	}
-}
-
-//advertisementUpdateFunc is the UPDATE event handler for the Advertisement CRDController.
-func advertisementUpdateFunc(oldObj interface{}, newObj interface{}) {
-	oldAdv := oldObj.(*advertisementApi.Advertisement)
-	newAdv := newObj.(*advertisementApi.Advertisement)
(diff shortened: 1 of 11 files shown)


mlavacca (approved):

internal/tray-agent/agent/client/advertisementController.go
```diff
@@ -8,11 +8,7 @@ import (
 
 //createAdvertisementController creates a new CRDController for the Liqo Advertisement CRD.
 func createAdvertisementController(kubeconfig string) (*CRDController, error) {
-	controller := &CRDController{
-		addFunc:    advertisementAddFunc,
-		updateFunc: advertisementUpdateFunc,
-		deleteFunc: advertisementDeleteFunc,
-	}
```
aleoli, line 15: why did you remove this advertisement function?

AbakusW: In this new version, the main point of access is the foreign cluster. It's the ForeignClusterController's cache that monitors with its handlers the changes on a FC, like the 'joined' field for the peerings.

I considered at least for this moment to remove the handling of advertisements events, since a notification would now be useless and there is no associated behavior to these changes.

Moreover, in the next PR I'll bring some refactoring to the Listener mechanism, and so there would be additional useless code to refactor

aleoli (approved):

Issue #367, opened 2020-11-09
Virtual kubelet to-do list


mlavacca:
## Description

The virtual kubelet has to be enriched with some additional modules and to see its code refactored in some parts.

### New features
* implement metrics API (Ref. #426)
* add the vk deployment as owner reference for the `NamespaceNattingTable`
* implement `podUpdate`

## refactoring

* `virtualNodeUpdater` better integration
* old logs package deletion
* old and useless flags deletion
* `server.go` file refactoring
* deletion of all unused tracing functionalities

palexster:
I would add the following items:
```:
- [ ] kubectl cp
- [ ] kubectl port-forward
```

PR #426, opened 2020-12-21
Virtual kubelet metrics API

mlavacca:
# Description

This PR introduces the metrics API implementation in the virtual kubelet. The virtual kubelet exposes a new endpoint for retrieving the stats of the offloaded pods running in the remote cluster.

* the `vk` iterates over all the namespaces currently natted
* all the pods in each natted namespace in the remote cluster having a specific label (a reference to the offloading virtual kubelet) are listed from the metrics API server
* the data structure used by the kubelet is built upon the remote data and is retrieved to the caller.


palexster:
ref. #367

internal/virtualKubelet/node/api/server.go
```diff
@@ -47,8 +50,22 @@ func PodHandler(p PodHandlerConfig, debug bool) http.Handler {
...
 	r.HandleFunc("/containerLogs/{namespace}/{pod}/{container}", HandleContainerLogs(p.GetContainerLogs)).Methods("GET")
-	r.HandleFunc("/exec/{namespace}/{pod}/{container}", HandleContainerExec(p.RunInContainer)).Methods("POST")
+	r.HandleFunc(
+		"/exec/{namespace}/{pod}/{container}",
+		HandleContainerExec(
+			p.RunInContainer,
+		),
+	).Methods("POST", "GET")
+
+	if p.GetStatsSummary != nil {
```
palexster, line 63: Why do we have this if? We are checking if the remote cluster has metrics enabled?

mlavacca: This is for being sure the metrics handler has been set. Actually, this if is always true in our implementation, I can remove it.

palexster (approved):

alacuku (approved):

PR #441, opened 2021-01-05
Discovery Test

aleoli:
# Description

This PR completely rewrites unit tests for the discovery component.

The new tests are written with the Ginkgo framework and the old one has been deleted.

A new interface `ConfigProvider` is used to mask discovery configuration making possible to mock it in a testing environment. Also the `ClusterID` has changed type from struct to interface for similar purposes.


internal/discovery/discovery_test.go
```diff
@@ -0,0 +1,860 @@
...
+
+	Describe("AuthData", func() {
+
+		var (
+			entry zeroconf.ServiceEntry
+		)
+
+		BeforeEach(func() {
+			entry = zeroconf.ServiceEntry{
+				HostName: "www.polito.it",
```
palexster, line 45: A fake name is preferable

AbakusW (approved):

AbakusW (approved):

mlavacca (approved):

PR #449, opened 2021-01-18
Update install and helm chart for dashboard

nappozord:
# Description

Updated helm chart and installer so that LiqoDash and Liqo will be now installed with two separate helm charts (anyway, they will be both installed in the same namespace).
To specify a version of the dashboard, export the variable `LIQO_DASHBOARD_VERSION`. The default is the latest tagged version.

Also uninstalling Liqo will uninstall the dashboard and related resources.


giorio94 (approved):
/lgtm

palexster (approved):

PR #457, opened 2021-01-28
Helm Chart Refactoring

palexster:
# Description

This PR refactors the code of the Liqo helm chart (Ref. #460):

- [x] Remove existing subcharts
- [x] Splits templates into different files to simplify troubleshooting

# How Has This Been Tested?

* [x] E2E Testing


deployments/liqo/templates/auth-service-deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+          - "--namespace"
+          - "$(POD_NAMESPACE)"
+          - "--resyncSeconds"
+          - "29"
+          - "--listeningPort"
+          {{ if .Values.authService.ingress.enable }}
+          - "4999"
+          {{ else }}
+          - "442"
+          - "--useTls"
```
aleoli, line 46: Is there any reason why these 3 numbers have been decreased by 1? (the same for the ingress and the service)

deployments/liqo/templates/discovery-deploy.yaml
```diff
@@ -0,0 +1,67 @@
...
+      containers:
+        - image: {{ .Values.discoveryOperator.image.repository }}{{ .Values.global.suffix | default .Values.suffix }}:{{ .Values.global.version | default .Values.version }}
+          name: discovery
+          imagePullPolicy: {{ .Values.pullPolicy }}
+          command: ["/usr/bin/entrypoint.sh", "/usr/bin/discovery"]
+          args:
+          - "--namespace"
+          - "$(POD_NAMESPACE)"
+          - "--requeueAfter"
+          - "29"
```
aleoli, line 30: same as before

deployments/liqo/templates/auth-service-deployment.yaml
palexster: very strange

deployments/liqo/templates/auth-service-service.yaml
```diff
@@ -0,0 +1,26 @@
...
+    run: auth-service
+  ports:
+    - name: https
+      protocol: TCP
+      {{ if .Values.authService.ingress.enable }}
+      port: 4999
+      targetPort: 4999
+      {{ else }}
+      port: 442
+      targetPort: 442
```
giorio94, line 25: Shouldn't those be 5000 and 443 (at least those are the ports specified in the deployment). Also in the ingress the port is wrong

AbakusW (approved):

deployments/liqo/values.yaml
```diff
@@ -120,3 +119,26 @@ global:
...
+    repository: "liqo/virtual-kubelet"
+    pullPolicy: "IfNotPresent"
+
+initVk:
+  image:
+    repository: "liqo/init-vkubelet"
+    pullPolicy: "IfNotPresent"
+
+suffix: ""
+version: "latest"
```
aleoli, line 144: These two fields are already set at the beginning of the file, and there are also inside the global object

```yaml
suffix: ""
version: "latest"
...
global:
  suffix: ""
  version: ""
...
suffix: ""
version: "latest
```

I think that we can delete some of them

aleoli (approved):

PR #458, opened 2021-01-28
Fix Advertisements Counter

### LINKED ISSUE #459: Advertisement operator does not update the number of advertisements already accepted.
When a peering is destroyed the number of currently accepted advertisements is not decreased, leading to a maximum of 5 peering with the default configurations.

aleoli:
# Description

This pr fixes the counter of currently accepted advertisements. Currently, this counter is increased but never decreased, with these changes the operator intercepts the delete event and, if the advertisement was in the accepted state, decreases the counter.

Fixes #459 

# How Has This Been Tested?

- [x] in a kind cluster trying to peer and unpeer for 6 times

### DIFF
--- a/internal/advertisement-operator/controller.go
+++ b/internal/advertisement-operator/controller.go
@@ -95,6 +95,12 @@ func (r *AdvertisementReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 		}
 	}
 
+	// if the Advertisement is deleting and it was in accepted state, we have to decrease the counter
+	if r.isDeleting(&adv) && adv.Status.AdvertisementStatus == advtypes.AdvertisementAccepted {
+		r.AcceptedAdvNum--
+		klog.Infof("Currently accepted Advertisements: %v", r.AcceptedAdvNum)
+	}
+
 	// we do that on Advertisement creation
 	err, update := r.UpdateForeignCluster(&adv)
 	if err != nil {
@@ -161,6 +167,11 @@ func (r *AdvertisementReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&advtypes.Advertisement{}).
 		Complete(r)
+}
+
+// checks that the Advertisement is being deleted
+func (r *AdvertisementReconciler) isDeleting(adv *advtypes.Advertisement) bool {
+	return !adv.DeletionTimestamp.IsZero()
 }
 
 // set Advertisement reference in related ForeignCluster


palexster (approved):

internal/advertisement-operator/controller.go
```diff
@@ -160,6 +162,20 @@ func (r *AdvertisementReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 func (r *AdvertisementReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&advtypes.Advertisement{}).
+		WithEventFilter(predicate.Funcs{
```
alacuku, line 165: The event filters, as the name suggests, should be used to filter the events that could trigger the reconcile logic of the operator. Here we are putting the operational logic of the operator inside a predicate function. A viable solution would be to check the `adv.ObjectMeta.DeletionTimestamp`, done inside the reconcile logic, to determine if the object is being deleted. The `.DeletionTimestamp` is meaningful only when the resource has at least a finalizer, in our case the `advertisement` resource meets the condition, hence no need to add a new finalizer.

alacuku (changes requested):

internal/advertisement-operator/controller.go
aleoli: Ok, but they are handler passed to the internal reconciler watcher, it is not doing anything on the adv resource, but is only updating an internal operator counter (that it is an internal state of operator, another thing that we should change). In my opinion, putting it in the event filter makes the reconcile function cleaner.

But if according to you it's better to move in inside the reconcile it's not a problem

palexster: My 2 cents: I also think that this code should be put in the reconcile function. It is a bit strange to see the eventFilter used to modify the internal state of the operator instead of real filtering.

internal/advertisement-operator/controller.go
```diff
@@ -163,6 +169,11 @@ func (r *AdvertisementReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+// checks that the Advertisement is being deleted
+func (r *AdvertisementReconciler) isDeleting(adv *advtypes.Advertisement) bool {
+	return adv.DeletionTimestamp != nil
```
alacuku, line 174: The `DeletionTimestamp` field could be set to `zero`, to be sure we check that case we could use `adv.DeletionTimestamp.IsZero()` method, which returns true if the value is nil or time is zero.

alacuku (approved):
/lgtm

PR #467, opened 2021-02-02
HelmChart generic helper functions

alacuku:
# Description

In order to template labels, annotations, and names for the deployments, cluster roles, service account some helper functions have been written. At the same time the cluster roles for the liqo operators are automatically generated by the `Makefile` using the kubebuilder annotations. This files are saved under the `files` directory in the helmchart and automatically included to the template files.
At install time annotations and labels can be added for each liqo component, and at the same time the same can be done for the k8s `services`.
The following components has been updated using the new helper functions:

- liqo-advertisement
- liqo-route
- liqo-gateway


deployments/liqo/templates/advertisement-deployment.yaml
```diff
@@ -4,29 +4,24 @@ kind: Deployment
...
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       labels:
-        run: advertisement-operator
+        {{- include "advertisement.selectorLabels" . | nindent 8 }}
     spec:
-      affinity:
-        nodeAffinity:
```
palexster, line 21: AFAIK This should not be removed

deployments/liqo/templates/_helpers.tpl
```diff
@@ -0,0 +1,105 @@
...
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "liqo.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "liqo.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
```
giorio94, line 51: I suppose this could be parametrized to take the component name (and optionally the module name) as argument, hence avoiding `advertisement.selectorLabels` and so on. Additionally, I would stick to standard k8s labels instead of custom ones (https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/):

I would propose something like:
```
app.kubernetes.io/name: <component-name> (e.g. advertisement)
app.kubernetes.io/component: <module-name> (e.g. sharing-protocol)
app.kubernetes.io/part-of: liqo
```

The definition of the common labels above should also be adapted with the parameters.

deployments/liqo/templates/_helpers.tpl
```diff
@@ -0,0 +1,105 @@
...
+
+{{/*
+Common labels
+*/}}
+{{- define "liqo.labels" -}}
+helm.sh/chart: {{ include "liqo.chart" . }}
+{{ include "liqo.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
```
giorio94, line 40: Would define an helper `liqo.version` and use that for consistency. Additionally, it may also be a good idea to use the `required` construct to throw an error if neither `.Values.version` nor `.Chart.AppVersion`  is set, to avoid subtle errors (I didn't in CrownLabs and I deployed the chart multiple times without a valid tag for this reason). BTW, the AppVersion should be probably dropped from the chart and configured with `helm package`.

```
{{/*
The version of the application to be deployed
*/}}
{{- define "liqo.version" -}}
{{- .Values.version | default .Chart.AppVersion }}
{{- end }}
{{- end }}
```

deployments/liqo/templates/advertisement-rbac.yaml
```diff
@@ -2,24 +2,24 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: advertisement-operator
+  name: {{ include "generic.service-account" (dict "prefix" (include "advertisement.name" .)) }}
   namespace: {{ .Release.Namespace }}
   labels:
-    k8s-app: advertisement-operator
-    app: liqo.io
+    {{- include "advertisement.selectorLabels" . | nindent 4 }}
```
giorio94, line 8: The global labels should also be present (the same for all components).

deployments/liqo/templates/advertisement-rbac.yaml
```diff
@@ -2,24 +2,24 @@
...
+    name: {{ include "generic.service-account" (dict "prefix" (include "advertisement.name" .)) }}
     namespace: {{ .Release.Namespace }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: cluster-admin
\ No newline at end of file
+  name: {{ include "advertisement.name" . }}
+
+{{ (.Files.Get (include "generic.rbac-filename" (dict "prefix" (include "advertisement.name" .)))) }}
```
giorio94, line 25: The best to me would be to explicitly define the metadata (to obtain consistency with names and labels), and only include the rules, but I do not know if can be done easiliy (it may be possible to just crop the first lines of the files, but that relies on the structure to be immutable).

giorio94 (commented):
@alacuku Seems definitely a good starting point to me! I left a few comments inline with some suggestions from my point of view.

palexster (approved):

mlavacca (approved):

PR #506, opened 2021-02-25
Enable Cluster Authentication by Default

aleoli:
# Description

This pr enables by default the inter-cluster authentication when Liqo is installed with helm. This makes it possible to have a minimal authentication layer before allowing an external cluster to schedule jobs in the local one.

This feature is not enabled we using the one-line installer, but it can be enabled by editing the cluster config CR [https://doc.liqo.io/user/post-install/authentication/](https://doc.liqo.io/user/post-install/authentication/)


deployments/liqo/values.yaml
```diff
@@ -137,7 +137,7 @@ auth:
     port: ""
     class: ""
   config:
-    allowEmptyToken: true
+    allowEmptyToken: false
```
mlavacca, line 140: Could you please add a brief description to be rendered in the helm documentation about what concerning the empty token?

palexster (approved):

mlavacca (approved):

palexster (approved):

PR #515, opened 2021-03-03
BugFix: crdReplicator forget keys of remove objects

alacuku:
# Description

Fixing a problem that causes the `reconcile` function to enter an infinite loop of logging when a `foreignclusters.discovery.liqo.io` is removed.


mlavacca (approved):
LGTM

palexster (approved):

PR #527, opened 2021-03-13
Add initial ARM support

palexster:
# Description

This PR introduces first support for ARM64 architecture.


.github/workflows/integration.yml
```diff
@@ -5,7 +5,7 @@ on:
       - 'v*'
     branches:
       - master
-  pull_request_target:
+  pull_request:
```
giorio94, line 8: This change has to be dropped

build/advertisement-broadcaster/Dockerfile
```diff
@@ -3,7 +3,7 @@ ENV PATH /go/bin:/usr/local/go/bin:$PATH
 ENV GOPATH /go
 COPY . /go/src/github.com/liqotech/liqo
 WORKDIR /go/src/github.com/liqotech/liqo
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./cmd/advertisement-broadcaster/
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=$(go env GOARCH) go build ./cmd/advertisement-broadcaster/
 RUN cp advertisement-broadcaster /usr/bin/advertisement-broadcaster
 
 FROM scratch
```
giorio94, line 9: For the future, it may be worth to uniform the base image for the different controllers. Probably, alpine or busybox are good compromises as they allow some basic debugging with no particular effort

build/liqo-controller-manager/Dockerfile
```diff
@@ -1,6 +1,13 @@
 FROM alpine:3.9.5
 RUN apk add --update curl bash openssl
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.17.0/bin/linux/amd64/kubectl && chmod +x ./kubectl && cp kubectl /usr/bin/kubectl
+RUN <<EOR
+  case "`uname -i`" in
+  x86_64|amd64)
+    curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.17.0/bin/linux/amd64/kubectl && chmod +x ./kubectl && cp kubectl /usr/bin/kubectl
```
giorio94, line 6: Could it be worth to upgrade the kubectl version? In any case, a fixed version could lead to problems in the future.
Additionally, it could be better to just set a variable and then run the command, to avoid code duplication which can easily become inconsistent.

giorio94 (approved):
/lgtm

alacuku (approved):

giorio94 (approved):

PR #531, opened 2021-03-16
Introduce retry logic into "get ClusterID" function 

### LINKED ISSUE #521: [Feature] Wait for ClusterID without failing
Many components make use of the `GetClusterID` function. The `ClusterID` is a configmap created by the discovery component. Until it is created, other components, such as the CRDReplicator exit with error because the fetch ends up with a NotFound error.
The `ClusterId` function should use a retry mechanism (or something similar) to allow the callers to not crash in this race condition.

GetClusterID function:
https://github.com/liqotech/liqo/blob/756e9c1cf297a7370c1f151d45745719a0c63b28/pkg/liqonet/utils.go#L109

lomluca:
# Description

This PR make the `GetClusterID` more reliable in the situation when the ConfigMap it tries to read from has not been created yet.

I reused the retry primitive included in the k8s `client-go` library, in order to be more consistent with the rest of the project. The function wait 30 seconds for 10 times. We could hard-code an opinionated configuration (current implementation) or make configurable by the caller (if useful, maybe not). 

Fixes #521 

## Plan

- [x] Implementation
- [x] Local build
- [x] Preliminary tests on development k8s cluster
- [x] Write test cases

### DIFF
--- a/cmd/crd-replicator/main.go
+++ b/cmd/crd-replicator/main.go
@@ -2,6 +2,9 @@ package main
 
 import (
 	"flag"
+	"os"
+	"time"
+
 	configv1alpha1 "github.com/liqotech/liqo/apis/config/v1alpha1"
 	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
 	"github.com/liqotech/liqo/internal/crdReplicator"
@@ -9,12 +12,12 @@ import (
 	"github.com/liqotech/liqo/pkg/mapperUtils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/klog/v2"
-	"os"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
@@ -50,7 +53,16 @@ func main() {
 		klog.Errorf("namespace env variable not set, please set it in manifest file of the operator")
 		os.Exit(-1)
 	}
-	clusterID, err := util.GetClusterID(k8sClient, clusterIDConfMap, namespaceName)
+
+	// 7 attempts with 30 seconds sleep between one another
+	// for a total of 3 minutes
+	backoff := wait.Backoff{
+		Steps:    7,
+		Duration: 30 * time.Second,
+		Factor:   1.0,
+		Jitter:   0,
+	}
+	clusterID, err := util.GetClusterID(k8sClient, clusterIDConfMap, namespaceName, backoff)
 	if err != nil {
 		klog.Errorf("an error occurred while retrieving the clusterID: %s", err)
 		os.Exit(-1)
--- a/pkg/liqonet/utils.go
+++ b/pkg/liqonet/utils.go
@@ -3,18 +3,20 @@ package liqonet
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
-	"net"
-	"os"
-	"syscall"
-
 	"github.com/liqotech/liqo/internal/utils/errdefs"
 	"github.com/vishvananda/netlink"
 	"golang.org/x/tools/go/ssa/interp/testdata/src/errors"
+	"io/ioutil"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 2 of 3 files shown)


lomluca:
I added three test cases to cover every possible situation. Only the last one is significant actually, the first twos are reported only for completeness (and be aware that the first one will make the testing process 3 minutes longer): we could keep only the last one.

palexster (approved):

palexster:
Hi @lomluca, can you please squash your commits?

pkg/liqonet/utils_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/liqotech/liqo/pkg/liqonet"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Liqonet", func() {
+
+	ns := "default"
```
mlavacca, line 16: According to the Kingko paradigm, the variables should be first declared and then instantiated in a beforeEach block. This approach is required to be sure to have the variables always clean and not affected by previous test cases.

lomluca:
Ok, I followed your feedback :) Hope it's ok now, let me know if there are any other change you would do. If not, I'm going to squash everything.

cmd/crd-replicator/main.go
```diff
@@ -48,7 +51,16 @@ func main() {
 		klog.Errorf("namespace env variable not set, please set it in manifest file of the operator")
 		os.Exit(-1)
 	}
-	clusterID, err := util.GetClusterID(k8sClient, clusterIDConfMap, namespaceName)
+
+	// 7 attempts with 30 seconds sleep between one another
+	// for a total of 3 minutes
+	backoff := wait.Backoff{
+		Steps:    3,
+		Duration: 15 * time.Second,
```
mlavacca, line 59: The comment says 30 seconds but the backoff has 15 seconds. I would use 30 secs

pkg/liqonet/utils.go
```diff
@@ -117,13 +121,27 @@ func VerifyNoOverlap(subnets map[string]*net.IPNet, newNet *net.IPNet) bool {
...
+	var cm *corev1.ConfigMap
+	var err error
+
+	notFound := func(error) bool {
+		return k8serrors.IsNotFound(err)
 	}
+
+	klog.Info("Getting ClusterID from ConfigMap...")
+	retryErr := retry.OnError(backoff, notFound, func() error {
+		klog.V(4).Info("Trying to get ConfigMap resource.")
```
mlavacca, line 135: I would move this log message in the notFound handler by logging something like "error while getting ClusterID configmap, retrying..."

pkg/liqonet/utils_test.go
```diff
@@ -0,0 +1,124 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      clusterIDKey,
+					Namespace: ns,
+				},
+				Data: configMapData,
+			}
+
+			// 3 attempts with 15 seconds sleep between one another
+			// for a total of 30 seconds
+			backoff = wait.Backoff{
```
mlavacca, line 48: Since we just want to test the correctness of the function, I would use a backoff time way lower than 3*15 seconds.

pkg/liqonet/utils_test.go
```diff
@@ -0,0 +1,124 @@
...
+				clusterID, err := liqonet.GetClusterID(clientset, clusterIDKey, "default", backoff)
+				end := time.Now()
+				Expect(err).NotTo(HaveOccurred())
+
+				minimum_sleep := 2 * backoff.Duration
+				Expect(end).Should(BeTemporally(">=", start.Add(minimum_sleep)))
+				Expect(clusterID).To(Equal(clusterIDValue))
+
+				// clean
+				err = clientset.CoreV1().ConfigMaps(ns).Delete(
```
mlavacca, line 114: Since this code is duplicated, I would embed the 2nd and the 3rd contexts in a Describe function with an AfterEach statement that cleans up the created configmap

EDIT: If you'd use the fake clientset instead of the controller-runtime test environment, the explicit delete is not needed at all, because the client is instantiated for each test case.

pkg/liqonet/liqonet_suite_test.go
```diff
@@ -0,0 +1,38 @@
+package liqonet_test
```
mlavacca, line 1: Do we really need the controller-runtime environment to perform this simple test? I think a [FakeClientSet](https://pkg.go.dev/k8s.io/client-go/kubernetes/fake#NewSimpleClientset) would fit better without the necessity to set up all this environment

lomluca: I took into consideration the fake client, but as far as I understood you need to declare the objects at the beginning, so the third test is not possible. Do you agree or am I missing something?

cmd/crd-replicator/main.go
lomluca: yes, copy&paste mistake I will fix

pkg/liqonet/liqonet_suite_test.go
mlavacca: You can instantiate a client and then create resources wherever you want, there are no constraints.

mlavacca (approved):

PR #532, opened 2021-03-17
Add Logging in Discovery and Authentication Components

aleoli:
# Description

This pr add some loggings in the Discovery and Authentication Components


palexster (approved):

mlavacca (commented):
Thanks for this PR @aleoli. Just a dumb question: why all those square brackets in the log messages? Why not "" instead?

aleoli:
No particular reason, it's only to have a clearly different part in the logging, in this way is clear that this section is not part of the logging message but it's more like a "context"

mlavacca (approved):

PR #539, opened 2021-03-22
Avoid Start-Time Throttling

aleoli:
# Description

This pr adds a different Mapper implementation to reduce the number of calls during components startup

A helper function has been added to add the resources belonging to the Liqo groups and the core/v1, apps/v1, and rbac/v1 groups into the new Mapper

### DIFF
--- a/cmd/advertisement-operator/main.go
+++ b/cmd/advertisement-operator/main.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
 	"github.com/liqotech/liqo/pkg/crdClient"
+	"github.com/liqotech/liqo/pkg/mapperUtils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
@@ -84,6 +85,7 @@ func main() {
 	}
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		MapperProvider:     mapperUtils.LiqoMapperProvider(scheme),
 		Scheme:             scheme,
 		MetricsBindAddress: "0",
 		LeaderElection:     enableLeaderElection,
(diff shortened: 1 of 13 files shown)


cmd/advertisement-operator/main.go
```diff
@@ -84,6 +89,18 @@ func main() {
 	}
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		MapperProvider: func(c *rest.Config) (meta.RESTMapper, error) {
```
alacuku, line 92: i've seen that this function is repeated for each operator of Liqo. I would suggest to create a wrapper function in `mapperUtils.go` which takes `*scheme` object (being it different for each operator) and returns a `func(c *rest.Config) (meta.RESTMapper, error)`. More or less something like this: 
```
type liqoMapper func(c *rest.Config) (meta.RESTMapper, error)

func LiqoDefaultMapperProvider (scheme *runtime.Scheme) (liqoMapper) {
	mapper = meta.NewDefaultRESTMapper(scheme.PrioritizedVersionsAllGroups())

	return func(c *rest.Config) (meta.RESTMapper, error) {
		dClient, err := discovery.NewDiscoveryClientForConfig(c)
		if err != nil {
			klog.Error(err)
			return nil, err
		}
		if err = AddDefaults(dClient, mapper); err != nil {
			klog.Error(err)
			return nil, err
		}
		return mapper, nil
	}
}
```

alacuku (approved):

pkg/mapperUtils/mapper.go
```diff
@@ -0,0 +1,107 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+)
+
+type LiqoMapper func(c *rest.Config) (meta.RESTMapper, error)
+
+// get the default liqo mapper
+func DefaultMapperProvider(scheme *runtime.Scheme, additionalGroupVersions ...schema.GroupVersion) LiqoMapper {
```
mlavacca, line 24:
```suggestion
func LiqoMapperProvider(scheme *runtime.Scheme, additionalGroupVersions ...schema.GroupVersion) LiqoMapper {
```

mlavacca (approved):

PR #542, opened 2021-03-26
Namespace-controller

Andreagit97:
# Description

Namespace-controller watches namespace's labels, creating and deleting corresponding entries in NamespaceMaps.

Fixes #(issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B


pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	klog.Flush()
+	return ctrl.Result{}, nil
+}
+
+// This function is useful only if we decide to accept the renaming of namespaces in "mapping.liqo.io" label.
+// If we don't want to change the name of namespaces, in function reconcile when we check that the name in the NNT is
+// different from the new name, with client we change the value of the new inserted label
+func mappingLabelUpdate(old map[string]string, new map[string]string) bool {
```
mlavacca, line 331: please, change the naming. `new` can collide with the builtin function name

cmd/advertisement-operator/main.go
```diff
@@ -153,6 +154,16 @@ func main() {
...
 	}
+
+	r2 := &namectrl.NamespaceReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+	}
+
+	if err = r2.SetupWithManager(mgr); err != nil {
+		klog.Error(err)
+		os.Exit(1)
```
mlavacca, line 165: You can use `klog.Fatal` instead of `klog.Error` and `os.Exit` jointly

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+
+			for _, node := range nodes.Items {
+				i := 0
+				found := false
+				dim := len(node.Labels)
+				for key := range node.Labels {
+					i++
+					// I have to check only the offloading.liqo.io/ labels
+					// virtual nodes will always have these offloading.liqo.io/ labels
+					if len(key) > 18 && key[0:19] == offloadingPrefixLabel {
```
mlavacca, line 282: You can use the `strings.StartWith` instead of this hardcoded check

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+
+			klog.Infof("---------------------- Watch for virtual-nodes labels")
+
+			// 2.b Iterate on all nodes labels (next step only virtual nodes), if the namespace has all the requested
+			// labels, is necessary to create the remote namespace on the remote cluster associated with the virtual
+			// node
+
+			nodes := &v1.NodeList{}
+			if err := r.List(context.Background(), nodes, client.MatchingLabels{"type": "virtual-node"}); err != nil {
+				//if err := r.List(context.Background(), nodes); err != nil {
```
mlavacca: Can we remove this line if useless?

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+				found := false
+				dim := len(node.Labels)
+				for key := range node.Labels {
+					i++
+					// I have to check only the offloading.liqo.io/ labels
+					// virtual nodes will always have these offloading.liqo.io/ labels
+					if len(key) > 18 && key[0:19] == offloadingPrefixLabel {
+						if _, ok := labels[key]; !ok {
+							break
+						} else {
```
mlavacca, line 286: AFAIU this else condition is useless (`found=true` is always hit)

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+							found = true // found will be no more necessary with only virtual nodes
+						}
+					}
+
+					if i == dim && found {
+						klog.Infof("---------------------- Create namespace for that remote cluter")
+						// TODO : remove from "erase" this virtual node
```
mlavacca: This TODO should be implemented now?

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+			klog.Error(err, " unable to fetch Namespace for some reasons")
+			return ctrl.Result{}, err
+		}
+	}
+
+	labels := namespace.Labels
+
+	// TODO : var "erase" will become a map/vector with some references to all virtual nodes on which, remote namespace
+	//        mustn't be present
+	erase := true
```
mlavacca: Can we use another more meaningful name? Something like `removeMappings`?

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+		if errors.IsNotFound(err) {
+			klog.Infof("---------------------- I have to delete all entries from NNT, if present")
+			return ctrl.Result{}, nil
+		} else {
+			klog.Error(err, " unable to fetch Namespace for some reasons")
+			return ctrl.Result{}, err
+		}
+	}
+
+	labels := namespace.Labels
```
mlavacca: Instead of using this brand-new var, referring to `namespace.Labels` every time would increase code readability

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+		}
+
+	}
+
+	if erase {
+		klog.Infof("---------------------- Delete all unnecessary mapping in NNT")
+	}
+
+	klog.Flush()
```
mlavacca: Why this Flush?

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
```diff
@@ -0,0 +1,172 @@
...
+			newLabels := e.MetaNew.GetLabels()
+			return ((len(oldLabels) != len(newLabels)) && (mappingLabelPresence(oldLabels) || mappingLabelPresence(newLabels))) || mappingLabelUpdate(oldLabels, newLabels)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false // i want to use only finalizer
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			return mappingLabelPresence(e.Meta.GetLabels())
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
```
mlavacca, line 369: Instead of setting it explicitly, I wouldn't set any handler for `GenericFunction`

Andreagit97: so is better to not filter this kind of `GenericEvent` ?

mlavacca: yep, I think so

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
Andreagit97: If we consider that virtual nodes are already created with right labels this check is useless, otherwise we could have some problems without this check (remote namespaces could be created on virtual nodes that don't have any `offloading.liqo.io/` label...). Anyway considering the first case i have removed `found` variable

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
Andreagit97: Changed in `newLabels`, thanks.

cmd/advertisement-operator/main.go
Andreagit97: Thank you, i have changed it.

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
Andreagit97: Thank you, i have changed it.

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
Andreagit97: Yes i have removed it, thank you

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
Andreagit97: Now in this version i have implemented it.

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
Andreagit97: Yes i have used `removeMappings` as you suggested, thank you.

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
Andreagit97: Thanks for the advice.

pkg/liqo-controller-manager/namespace-controller/namespace_controller.go
Andreagit97: Now is useless, i have removed it, thank you.

Andreagit97:
i will split a little bit this PR

PR #553, opened 2021-04-06
Add Pre-delete hook

palexster:
# Description

This PR adds a pre-delete hook to correctly remove all Liqo peerings before Liqo uninstall.


.github/workflows/integration.yml
```diff
@@ -5,7 +5,7 @@ on:
       - 'v*'
     branches:
       - master
-  pull_request_target:
+  pull_request:
```
giorio94, line 8: I suppose this should be reverted

.github/workflows/integration.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       id: configure
       run: |
         # The ref of the commit to checkout (do not use the merge commit if pull request)
-        if [ "${{ github.event_name }}" == "pull_request_target" ]; then
+        if [ "${{ github.event_name }}" == "pull_request" ]; then
```
giorio94, line 33: Same here

apis/config/v1alpha1/groupversion_info.go
```diff
@@ -29,6 +29,12 @@ var (
 
 	GroupResource = schema.GroupResource{Group: GroupVersion.Group, Resource: "clusterconfigs"}
 
+	GroupVersionResource = schema.GroupVersionResource{
```
giorio94, line 32:
```suggestion
	ClusterConfigGroupVersionResource = schema.GroupVersionResource{
```

I suppose

cmd/uninstaller/main.go
```diff
@@ -0,0 +1,70 @@
...
+	"path/filepath"
+)
+
+func main() {
+	var kubeconfigPath string
+	var runAs string
+	var config *rest.Config
+	klog.Info("Loading client config")
+	flag.StringVar(&kubeconfigPath, "kubeconfigPath", filepath.Join(os.Getenv("HOME"), ".kube", "config"), "For debug purpose, set path to local kubeconfig")
+	flag.StringVar(&runAs, "run-as", "pre-delete", "")
```
giorio94, line 21: Suppose this should be dropped

cmd/uninstaller/main.go
```diff
@@ -0,0 +1,70 @@
...
+	// Trigger unjoin clusters
+	err = uninstaller.UnjoinClusters(client)
+	if err != nil {
+		klog.Errorf("Unable to unjoin clusters: %s", err)
+		os.Exit(1)
+	}
+	klog.Info("Cluster Unjoin Correctly Performed")
+
+	if err = uninstaller.DisableBroadcasting(client); err != nil {
+		klog.Errorf("Unable to unjoin clusters: %s", err)
```
giorio94, line 43: This error should probably be different

deployments/liqo/templates/pre-delete-job.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: {{ include "liqo.prefixedName" $predelete }}
+  labels:
+    {{- include "liqo.labels" $predelete| nindent 4 }}
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
```
giorio94, line 14:
```suggestion
    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```

deployments/liqo/templates/pre-delete-rbac.yaml
```diff
@@ -0,0 +1,55 @@
+{{- $predelete := (merge (dict "name" "pre-delete" "module" "pre-delete") .) -}}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
```
giorio94, line 5: I think those should also have the hook annotations, to be created only at deletion time

pkg/uninstaller/resources.go
```diff
@@ -0,0 +1,200 @@
...
+			labelSelector: metav1.LabelSelector{},
+		},
+		{
+			gvr:           v1alpha1.NetworkConfigGroupVersionResource,
+			labelSelector: metav1.LabelSelector{},
+		},
+		{
+			gvr: podGVR,
+			labelSelector: metav1.LabelSelector{
+				MatchLabels: map[string]string{"app": "virtual-kubelet"},
```
giorio94, line 50: Those labels should probably not be hardcoded, otherwise issues will arise as soon as they are changed

giorio94 (approved):
/lgtm

Some nits inline

apis/config/v1alpha1/groupversion_info.go
```diff
@@ -27,7 +27,13 @@ var (
 	// GroupVersion is group version used to register these objects
 	GroupVersion = schema.GroupVersion{Group: "config.liqo.io", Version: "v1alpha1"}
 
-	GroupResource = schema.GroupResource{Group: GroupVersion.Group, Resource: "clusterconfigs"}
+	ClusterConfigGroupResource = schema.GroupResource{Group: GroupVersion.Group, Resource: "clusterconfigs"}
+
+	ClusterConfigGroupVersionResource = schema.GroupVersionResource{
+		Group:    GroupVersion.Group,
+		Resource: "clusterconfigs",
+		Version:  "v1alpha1",
```
alacuku, line 35:
```suggestion
		Version:  GroupVersion.Version,
```

pkg/uninstaller/resources.go
```diff
@@ -0,0 +1,200 @@
...
+	labelSelector metav1.LabelSelector
+}
+
+type resultType struct {
+	Resource toCheckDeleted
+	Success  bool
+}
+
+var (
+	podGVR = schema.GroupVersionResource{
```
alacuku, line 32:
```suggestion
	podGVR = v1.SchemeGroupVersion.WithResource("pods")
```

pkg/uninstaller/resources.go
```diff
@@ -0,0 +1,200 @@
...
+	if labelString, err = generateLabelString(toCheck.labelSelector); err != nil {
+		return false, err
+	}
+	options := metav1.ListOptions{
+		LabelSelector: labelString,
+	}
+
+	objectList, err = client.Resource(toCheck.gvr).Namespace("").List(context.TODO(), options)
+
+	if apierrors.IsNotFound(err) {
```
alacuku, line 122: i think we do not need this check. List will return an UnstructuredList with the field Items having length 0 if no objects have been found.

pkg/advertisement-operator/forge.go
```diff
@@ -0,0 +1,160 @@
...
+			Name:      "virtual-kubelet-crt",
+			MountPath: "/etc/virtual-kubelet/certs",
+		},
+	}
+
+	return []v1.Container{
+		{
+			Name:            "virtual-kubelet",
+			Image:           vkImage,
+			ImagePullPolicy: v1.PullAlways,
```
mlavacca, line 118: Do we really want the Always policy? Isn't it too impactful on the performance?

pkg/uninstaller/resources.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+)
+
+func UnjoinClusters(client dynamic.Interface) error {
+	r1 := client.Resource(discoveryV1alpha1.ForeignClusterGroupVersionResource)
+	t, err := r1.Namespace("").List(context.TODO(), metav1.ListOptions{TypeMeta: metav1.TypeMeta{}})
+	if err != nil {
+		return err
+	}
+	klog.V(8).Info("Getting ForeignClusters list")
```
mlavacca, line 66: I would use a verbosity lower than 8, which is very high. I think that if you set 8, finding what you need without grepping can be very difficult. Moreover, consider that the program is fairly simple, we don't have too many prints.

pkg/uninstaller/resources.go
```diff
@@ -0,0 +1,198 @@
...
+			close(quit)
+		case <-ticker.C:
+			value, wError = CheckReconciliation(client, toCheck)
+			if value {
+				res.Success = true
+				close(quit)
+			} else if wError != nil {
+				close(quit)
+			}
+			printLabels, _ := generateLabelString(toCheck.labelSelector)
```
mlavacca, line 181: Since CheckReconciliation cannot return (true, err), this else is useless. Besides, if `err != nil`, I would log the error at least.

pkg/advertisement-operator/forge.go
palexster: I just copied what was already there, but I totally agree.

build/uninstaller/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ENV PATH /go/bin:/usr/local/go/bin:$PATH
+ENV GOPATH /go
+COPY . /go/src/github.com/liqotech/liqo
+WORKDIR /go/src/github.com/liqotech/liqo
+
+ARG BUILD_TAGS=""
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=$(go env GOARCH) go build ./cmd/uninstaller/
+RUN cp uninstaller /usr/bin/uninstaller
+
+FROM alpine:3.13.2
```
mlavacca, line 11: Specifying only 3.13 ensures that you are up to date. 3.13.2 is already out of date

cmd/advertisement-operator/main.go
```diff
@@ -155,12 +153,17 @@ func main() {
 	}
 	// +kubebuilder:scaffold:builder
 
-	r.WatchConfiguration(localKubeconfig, nil)
+	go r.WatchConfiguration(localKubeconfig, nil)
```
mlavacca, line 156: can you add this goroutine to the waiting group?

deployments/liqo/pre-delete.yaml
```diff
@@ -0,0 +1,41 @@
...
+      labels:
+        app.kubernetes.io/name: "pre-delete"
+        app.kubernetes.io/instance: "RELEASE-NAME-pre-delete"
+        app.kubernetes.io/component: "discovery"
+        app.kubernetes.io/part-of: "liqo"
+        helm.sh/chart: "liqo-0.2.1"
+        app.kubernetes.io/version: "0.2.1"
+        app.kubernetes.io/managed-by: "Helm"
+    spec:
+      restartPolicy: Never
```
mlavacca, line 33: If the pod fails, it is never restarted. Do we want this behavior?

internal/advertisement-operator/controller.go
```diff
@@ -342,37 +340,39 @@ func (r *AdvertisementReconciler) createVirtualKubelet(ctx context.Context, adv
...
 func (r *AdvertisementReconciler) recordEvent(msg string, eventType string, eventReason string, adv *advtypes.Advertisement) {
 	klog.Info(msg)
 	r.EventsRecorder.Event(adv, eventType, eventReason, msg)
 }
 
-func (r *AdvertisementReconciler) cleanOldAdvertisements() {
+func (r *AdvertisementReconciler) CleanOldAdvertisements(quit chan struct{}, wg *sync.WaitGroup) {
+	defer wg.Done()
 	var advList advtypes.AdvertisementList
+	ticker := time.NewTicker(10 * time.Minute)
```
mlavacca, line 353: can we parameterize 10 minutes?

mlavacca (approved):

deployments/liqo/pre-delete.yaml
palexster: Yes, it is one-shot and I think it is better to avoid restart in order to simplify troubleshooting.

build/uninstaller/Dockerfile
palexster: I personally prefer to have a slightly outdated version but know precisely which image we are using and have all Liqo components aligned to the same one.

PR #556, opened 2021-04-07
Support for additional network pools

davidefalcone1:
Currently Liqo uses a fixed set of network ('pools') used to map a cluster network into another one, if necessary. 
The set of network pools is: [10.0.0.0/8, 192.168.0.0/16, 172.16.0.0/12]

This PR allows the user to specify additional network pools to be added to the default set. These new pools can be specified by means of the clusterconfig resource, where a new field, called 'AdditionalPools' has just been added.


internal/liqonet/tunnelEndpointCreator/tunnelEndpointCreator-config.go
```diff
@@ -71,30 +81,91 @@ func (tec *TunnelEndpointCreator) UpdateConfiguration(reservedSubnets []string)
...
 			if err := tec.IPManager.AcquireReservedSubnet(subnet); err != nil {
+				klog.Error(err)
+			}
+			tec.addReservedSubnet(subnet)
+		}
+	}
+	return nil
+}
+
+func (tec *TunnelEndpointCreator) UpdatePools(additionalPools []string) error {
```
alacuku, line 102: this function does the same as the UpdateReservedSubnets. could we move the logic outside these functions, having it in one place. We call it to check if the configuration in the clusterConfig resource has changed and then return what subnets that have been added or removed. Then the UpdateReservedSubnets and UpdatePools update accordingly the configuration inside the `TunnelEndpointCreator` instance.

davidefalcone1: Good tip

alacuku (approved):

pkg/liqonet/ipam.go
```diff
@@ -323,3 +339,95 @@ func (liqoIPAM *IPAM) FreeSubnetPerCluster(clusterID string) error {
 	klog.Infof("Network assigned to cluster %s has just been freed", clusterID)
 	return nil
 }
+
```
palexster, line 342: Since those methods are exported, I would add a short comment describing their purpose

davidefalcone1: Done!

palexster (approved):

PR #557, opened 2021-04-07
Reserve a subnet equal to network pool

davidefalcone1:
This PR adds the capability to reserve a network which is equal to a network pool.

### DIFF
--- a/pkg/liqonet/ipam.go
+++ b/pkg/liqonet/ipam.go
@@ -2,7 +2,7 @@ package liqonet
 
 import (
 	"fmt"
-	"strconv"
+	"net"
 	"strings"
 
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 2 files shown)


pkg/liqonet/ipam.go
```diff
@@ -89,8 +89,20 @@ func (liqoIPAM *IPAM) AcquireReservedSubnet(reservedNetwork string) error {
 		return err
 	}
 	if ok && reservedNetwork == pool {
-		klog.Errorf("Network %s is equal to a pool", reservedNetwork)
-		return fmt.Errorf("network %s is equal to a pool so it cannot be reserved", reservedNetwork)
+		klog.Infof("Network %s is equal to a pool, acquiring first half..", reservedNetwork)
+		mask := getMask(reservedNetwork)
```
palexster, line 93: Can you please briefly comment on this logic?

davidefalcone1: Sure!

palexster (approved):

pkg/liqonet/ipam.go
```diff
@@ -251,7 +266,7 @@ func (liqoIPAM *IPAM) mapNetwork(network string) (string, error) {
 
 /* Helper function to get a mask from a net.IPNet */
 func getMask(network string) uint8 {
-	stringMask := network[len(network)-2:]
+	stringMask := strings.Split(network, "/")[1]
```
alacuku, line 269: can we avoid manipulating strings in order to get the mask length? i would use the "net" package to achieve the same result. something like
```
	_, network, err := net.ParseCIDR(network)
	if err != nil{
		return err
	}
	ones := network.Mask.Size())
	return ones

davidefalcone1: Good idea

alacuku (approved):

PR #559, opened 2021-04-10
Fixed changePodIp func

davidefalcone1:
# Description

Function changePodIp forges a new IP taking the first 2 octets from the newPodCidr and the last 2 from the original Pod Ip. Since Liqo does support networks with random masks, this function needs to be modified: thanks to this PR, the function now takes a number of bits equal to the mask length from the newPodCidr and 32-mask bits from the old Pod IP.

### DIFF
--- a/pkg/virtualKubelet/forge/pods.go
+++ b/pkg/virtualKubelet/forge/pods.go
@@ -2,6 +2,9 @@ package forge
 
 import (
 	"fmt"
+	"net"
+	"strings"
+
 	"github.com/liqotech/liqo/internal/liqonet/tunnelEndpointCreator"
 	"github.com/liqotech/liqo/pkg/virtualKubelet"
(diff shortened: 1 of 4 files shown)


palexster (approved):
/lgtm

pkg/virtualKubelet/forge/pods.go
```diff
@@ -183,30 +188,38 @@ func filterVolumeMounts(volumes []corev1.Volume, volumeMountsIn []corev1.VolumeM
...
-			newPodIpBuilder.WriteString(".")
-		}
+		return oldPodIp, nil
 	}
-	for i, s := range oldPodIpTokenized {
-		if i > 1 && i < 4 {
-			newPodIpBuilder.WriteString(s)
-			newPodIpBuilder.WriteString(".")
-		}
+	// Parse newPodCidr
```
palexster, line 198: This logic may deserve to be in a separated function

alacuku (approved):

PR #567, opened 2021-04-15
ExternalCidr for remote clusters

davidefalcone1:
# Description

This PR introduces support for ExternalCidr. 

ExternalCidr is used to represent the network used by the local cluster for service endpoints of the remote cluster. Traffic toward this network is forwarded to the remote cluster. Thanks to this network we avoid possible conflicts between IPs of service endpoints reflected in the remote cluster and IPs of pods of the local cluster.

At startup, the IPAM module choose randomly the ExternalCIDR of the local cluster.

The remote cluster receives the network conf with the localExternalCIDR(Spec->ExternalCIDR). The remote cluster, as it does for the PodCIDR, checks for conflicts and eventually remaps the network, setting Status->ExternalCIDRNAT in the network config. If no conflicts are found, Status->ExternalCIDRNAT is "None".

A set of new fields in the tunnelendpoint crd has been added as well.

The Ipam module reserves an ExternalCIDR when the cluster receives a networkconfig from a remote cluster and stores (without reserving) the ExternalCIDR used in the remote cluster and stored in the Status of the local networkconfig.


apis/net/v1alpha1/tunnel_endpoint_types.go
```diff
@@ -43,11 +45,14 @@ type TunnelEndpointStatus struct {
...
 	// Important: Run "make" to regenerate code after modifying this file\
 	Phase                 string     `json:"phase,omitempty"` //two phases: New, Processed
-	LocalRemappedPodCIDR  string     `json:"localRemappedPodCIDR,omitempty"`
+	LocalNATPodCIDR       string     `json:"localNATPodCIDR,omitempty"`
 	LocalPodCIDR          string     `json:"localPodCIDR,omitempty"`
-	RemoteRemappedPodCIDR string     `json:"remoteRemappedPodCIDR,omitempty"`
-	OutgoingNAT           bool       `json:"outgoingNAT,omitempty"` // if true, the local podCIDR has been remapped by the remote cluster
-	IncomingNAT           bool       `json:"incomingNAT,omitempty"` // if true, the remote podCIDR has been remapped by the local cluster
+	RemoteNATPodCIDR      string     `json:"remoteNATPodCIDR,omitempty"`
+	LocalExternalCIDR     string     `json:"localExternalCIDR,omitempty"`     // External CIDR for remote cluster.
```
palexster, line 51: Can you please add comments to be included in the CRD definition as a description field?

davidefalcone1: Sure!

liqo-cluster-config.yaml
```diff
@@ -0,0 +1,10 @@
+kind: Cluster
```
palexster, line 1: I think this file should be dropped.

pkg/liqonet/utils.go
```diff
@@ -132,13 +132,10 @@ func EnableIPForwarding() error {
 }
 
 /* Helper function to get a mask from a CIDR */
-func GetMask(network string) (uint8, error) {
-	_, net, err := net.ParseCIDR(network)
-	if err != nil {
-		return 0, err
-	}
+func GetMask(network string) uint8 {
```
palexster, line 135: Do you introduce the err control somewhere else?

liqo-cluster-config.yaml
davidefalcone1: Right!

pkg/liqonet/utils.go
davidefalcone1: This function receives always a network cidr already validated.

alacuku (approved):

palexster (approved):

PR #575, opened 2021-04-19
Fix - Get Node Address

aleoli:
# Description

This pr fixes a problem retrieving the node IP that could cause the node not to become ready in multi peering scenarios

# How Has This Been Tested?

- [x] Locally on KinD

### DIFF
--- a/internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
+++ b/internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
@@ -34,6 +34,7 @@ import (
 	"github.com/liqotech/liqo/pkg/crdClient"
 	discoveryPkg "github.com/liqotech/liqo/pkg/discovery"
 	"github.com/liqotech/liqo/pkg/kubeconfig"
+	liqoControllerManager "github.com/liqotech/liqo/pkg/liqo-controller-manager"
 	apiv1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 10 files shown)


internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -508,8 +509,25 @@ func (r *ForeignClusterReconciler) getAddress() (string, error) {
 		}
 	}
 
+	// only physical nodes
```
palexster, line 512: Can you comment more to explain why this extra filter is necessary?

palexster (approved):

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -508,8 +509,29 @@ func (r *ForeignClusterReconciler) getAddress() (string, error) {
...
+	//
+	// we need to get an address from a physical node, if we have established peerings in the past with other clusters,
+	// we may have some virtual nodes in our cluster. Since their IPs will not be reachable from other clusters, we cannot use them
+	// as address for a local NodePort Service
+	labelSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      "type",
+				Operator: metav1.LabelSelectorOpNotIn,
+				Values:   []string{"virtual-node"},
```
giorio94, line 522: Just a minor comment, it would probably better to refer to constants instead of using hard-coded values, to avoid issues if the key or the value are changed in the future

giorio94 (approved):

Issue #573, opened 2021-04-19
Authentication and PeeringProcess Refactoring


aleoli:
# Description

This issue tracks the migration from the ServiceAccount-based authentication to the new Certificate-based authentication, with the usage of namespaced resources

# Steps

### General
- [x] define cluster roles in the helm chart - #580
  - Network Resources
  - Advertisement
  - Resource Request
- [ ] namespaced resources
  - all the Liqo resources, but not the ForeignCluster

### ForeignCluster Operator (Authentication)
- [x] "system" namespace creation - #574 #588
- [x] certificate creation - #578 #588
- [x] store remote system namespace name in the ForeignCluster - #588
- [ ] switch to the new auth endpoint (after auth 4) - #588
- [ ] certificate rotation (after CrdR 2)
  - set shorter time
  -  create a new certificate when deadline is near
  -  recreate all the clients pointing to the remote cluster

### Authentication Service
- [x] create a new HTTP endpoint - #583 
- [x] CRS approval - #578 #583
- [x] "system" namespace creation - #574 #583
- [x] create role binding (resource request and network resources) in the "system" namespace - #574 #583
- [ ] allow only certificates with a short time to live (after FC Op Auth 5)

### ForeignCluster Operator (Peering)
- [x] create role binding (advertisement and network resources) in the "system" namespace - #574 #588
- [x] create local ResourceRequest in the “system” namespace - #588
- [ ] enable cr replication (after CrdR 1) - #663 
- [ ] delete the cluster scoped PeeringRequest (when the broadcaster is able to handle the new resource)

### CrdReplicator
- [x] allow different source and destination namespaces - #632
- [x] enable the replication of CRDs in different peering phases - #649 #663 
- [ ] allow client recreation when the client identity is updated

### ResourceOffer Operator (Peering / Node Creation)
- [x] add the new operator - #637 
- [x] filter only replicated resources - #637 
- [x] set the phase accordingly - #637 
- [x] add vk cluster role - #646  
- [x] create virtual node - #653 
  - role binding
  - create deployment
  - manage lifecycle

PR #580, opened 2021-04-22
Remote Peering RBAC

aleoli:
# Description

This pr adds the ClsuterRoles required during the peering with a remote cluster, these ClusterRoles will be bound during the peering with RoleBindings to the Tenant Control Namespace.

In particular, 3 roles has been defined:
* `liqo-remote-peering-basic` that will be activated when the namespace is created with limited permissions
* `liqo-remote-peering-incoming` that will be activated when an incoming peering has been accepted, after the reception of a ResourceRequest
* `liqo-remote-peering-outgoing` that will be activated when we want to establish an outgoing peering, so when we create a ResourceRequest in a remote cluster

Ref. #573


palexster (approved):

deployments/liqo/templates/liqo-remote-peering-rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+# when the Pods will be offloaded to the local cluster
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "liqo.prefixedName" $peeringIncoming }}
+{{ .Files.Get (include "liqo.cluster-role-filename" (dict "prefix" ( include "liqo.prefixedName" $peeringIncoming))) }}
+
+---
+# to be enabled when we send a ResourceRequest,
+# this ClusterRole has the perissions required to a remote cluster to manage an incoming peering (outgoing for the local cluster),
```
giorio94, line 25:
```suggestion
# this ClusterRole has the permissions required to a remote cluster to manage an incoming peering (outgoing for the local cluster),
```

giorio94 (approved):
/lgtm

PR #583, opened 2021-04-22
Certificate Identity Http Endpoint

aleoli:
# Description

This pr adds an HTTP endpoint to the authentication service to be used to issue certificate identities.

When the request is accepted, it creates a new TenantControlNamespace (if not present). Then it issues the certificate and sends to the remote client the name of the namespace, the certificate, the API server URL, and the ApiServer CA. These values will be used to create a new remote client.

Ref #573

# How Has This Been Tested?

- [x] add new unit tests
- [x] deploy locally on KinD and try to contact it with curl


pkg/identityManager/signingRequest.go
```diff
@@ -39,7 +99,7 @@ func (certManager *certificateIdentityManager) ApproveSigningRequest(signingRequ
 			Groups: []string{
 				"system:authenticated",
 			},
-			Request: signingRequest,
+			Request: signingBytes,
```
palexster, line 102: Should we add also a signer here to be compliant with v1?

pkg/identityManager/test/csrApprover.go
```diff
@@ -0,0 +1,43 @@
+package test
```
palexster, line 1: Maybe the name test is not really clear here

palexster (approved):

alacuku (approved):

Issue #572, opened 2021-04-18
[Feature] Support for deployments spanning across selected clusters.


Andreagit97:
# Description

So far when Users deploy apps using Liqo, they do not have the opportunity to choose which remote clusters use to offload their workload. This new feature allows you to select specif remote clusters according to defined requirements by simply creating a custom resource called NamespaceOffloading.

# Required Steps

- [x]  **STEP 1**: Introduce a new `NamespaceMap` CRD that will replace `NamespaceNattingTable` resource. #569 

- [x]  **STEP  2**: 
     - Implement a new Virtual-node Controller that spawns `NamespaceMaps` corresponding to just created `VirtualNodes`. This controller takes also care of `NamespaceMaps`' and `VirtualNodes`' deletion lifecycle. #561 
     - Implement a new Namespace Controller that watches `Namespace`'s labels and decides if a `Namespace` has to be offloaded onto a remote cluster or not. #571 

- [x]  **STEP 3**:
    - Implement a new NamespaceMap Controller that watches `NamespaceMap`'s Spec to understand which `Namespaces` have to offloaded onto the remote cluster, and updates `NamespaceMap`'Status with the current state of the just created remote `Namespaces`. #587
    - `NamespaceMap`'s refactor due to new features introduced by NamespaceMap Controller. #585
 
- [x]  **STEP 4**: Possibility to specify, at installation time, cluster's labels; these labels must be inserted in `ClusterConfig` CRD with all other information about the cluster.   #597

- [x]  **STEP 5**: Cluster's labels (of the previous step) must be imported into `ResourceOffer` CRD, so that VirtualKubelet could expose them on the corresponding `VirtualNodes`. #720 

- [x]  **STEP 6**: Introduce a new `NamespaceOffloading` CRD that allows the user to specify, with desired granularity, on which cluster his workload must be offloaded. #617

- [x]  **STEP 7**: Refactor of Namespace controller in order to support `NamespaceOffloading` management. #639 

- [x]  **STEP 8**: New features for NamespaceMap Controller: #643 
    
- When there is a problem in remote namespace creation onto a remote cluster, the associated `NamespaceMap` is updated with a Refused phase for that namespace (`Status`->`CurrentMapping`->` RemoteStatus`->` Phase`=Refused). The NamespaceMap Controller will try to create that Namespace until the problem will be solved with exponential backoff.

- The controller updates the remote namespace condition in the `NamespaceOffloading` resource associated with the just created remote namespace.

[...]

PR #587, opened 2021-04-26
Add NamespaceMap-controller to Liqo's controllers

Andreagit97:
# Description

This new controller monitors NamespaceMap's Spec in order to create and remove remote Namespaces. Besides it updates NamespaceMap's status with the current status of the just created remote Namespace.

# Fixes

**STEP 3** #572

### DIFF
--- a/pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller.go
+++ b/pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller.go
@@ -1,0 +1,98 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package namespaceMap_controller
+
+import (
+	"context"
+	mapsv1alpha1 "github.com/liqotech/liqo/apis/virtualKubelet/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller.go
```diff
@@ -0,0 +1,98 @@
...
+			return ctrl.Result{}, err
+		}
+		klog.Infof("Finalizer correctly added on NamespaceMap '%s'", namespaceMap.GetName())
+	}
+
+	if err := r.manageRemoteNamespaces(namespaceMap); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if len(namespaceMap.Status.CurrentMapping) == 0 {
```
mlavacca, line 65: Why are you removing the finalizer if the status has no current mappings? Shouldn't you check the deletionTimestamp?

mlavacca: I saw the update handler in the eventFilter, I got it.

mlavacca (approved):

pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller.go
Andreagit97: This finalizer is only used to ensure that the NamespaceMap is not deleted until there are remote Namespaces onto the corresponding remote clusters. The deletion logic of NamespaceMap is managed by Virtual-Node Controller, it will check deletion timestamp... I don't know if it is clear. There are 2 finalizers on NamespaceMap one setted by VirtualNode Controller and this one. If the NamespaceMap is ready for deletion and the virtualNode's finalizer is already been removed, we couldn't delete really the NamespaceMap until the are remote Namespaces, this is the scope of this new finalizer.

Andreagit97: Great, thank you for the review

palexster (approved):

PR #588, opened 2021-04-29
Update the ForeignCluster Operator with new Authentication

aleoli:
# Description

This pr adds the new authentication service client in the ForeignCluster Operator.

When a new ForeignCluster comes, it creates a new TenantControlNamespace (if not present). Then it creates a new private key and a new certificate signing request to be sent to the remote cluster.

Ref #573

# How Has This Been Tested?

- [x] add new unit test
- [x] locally on KinD


internal/discovery/foreign-cluster-operator/auth.go
```diff
@@ -161,52 +187,103 @@ func (r *ForeignClusterReconciler) getAuthToken(fc *discoveryv1alpha1.ForeignClu
...
-	jsonRequest, err := json.Marshal(roleRequest)
+	return sendIdentityRequest(&roleRequest, fc)
+}
+
+// validateIdentity sends an HTTP request to validate the identity for the remote cluster (Certificate).
+func (r *ForeignClusterReconciler) validateIdentity(fc *discoveryv1alpha1.ForeignCluster) error {
+	remoteClusterID := fc.Spec.ClusterIdentity.ClusterID
+	token := r.getAuthToken(fc)
+
+	_, err := r.identityManager.CreateIdentity(remoteClusterID)
```
palexster, line 206: What if the identity already exists?

aleoli: * If the identity already exists in the local cluster, this function will not be even invoked. This will only be invoked if there is not a secret containing the issued certificate.

* If the identity already exists in the remote cluster, it will return a forbidden error, because of the impossibility to validate a second identity for the same clusterID. If we accept the creation of a second identity with the same clusterID, when a cluster A is peered with a cluster B, and a cluster C knows the ID of cluster A, it will be able to forge an identity and to act as A taking control of its resources.

palexster: Clear. Thanks!

palexster (approved):

alacuku (approved):

PR #589, opened 2021-05-01
Support for MappingLabel renaming and defaulting

Andreagit97:
# Description

This PR adds the possibility to have a default value for mappingLabel if the user does not specify it. In case of MappingLabel renaming during offloading, an Annotation notifies user about the impossibility to perform this action. From now on liqo/pkg/utils will be the common place for controllers' utilities.

### DIFF
--- a/pkg/liqo-controller-manager/namespace-controller/manage_namespace_labels.go
+++ b/pkg/liqo-controller-manager/namespace-controller/manage_namespace_labels.go
@@ -1,7 +1,9 @@
 package namespace_controller
 
 import (
-	const_ctrl "github.com/liqotech/liqo/pkg/liqo-controller-manager"
+	"context"
+	"fmt"
+	liqocontrollerutils "github.com/liqotech/liqo/pkg/utils"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/util/slice"
@@ -10,12 +12,35 @@ import (
 	"strings"
 )
 
+// This function sets remote Namespace name to default value "localName-clusterID" , and adds an Annotation to notify user
+func (r *NamespaceReconciler) checkMappingLabelDefaulting(n *corev1.Namespace) error {
+	clustedID, err := liqocontrollerutils.GetClusterID(r.Client)
+	if err != nil {
+		return err
+	}
+
+	n.Labels[mappingLabel] = fmt.Sprintf("%s-%s", n.Name, clustedID)
+
+	if n.GetAnnotations() == nil {
+		n.Annotations = map[string]string{}
+	}
+	n.GetAnnotations()[mappingAnnotationDefaulting] = fmt.Sprintf("You have not specified a name for your remote Namespace, this is your default name: [%s]. Please read our documentation for more info [%s]",
+		n.GetLabels()[mappingLabel], liqocontrollerutils.DocumentationUrl)
+
+	if err = r.Update(context.TODO(), n); err != nil {
+		klog.Errorf("%s --> Unable to update '%s' label on Namespace '%s'", err, mappingLabel, n.GetName())
+		return err
+	}
+	klog.Infof("'%s' of namespace '%s' updated with success", mappingLabel, n.GetName())
+	return nil
+}
+
 // Checks if Namespace has all offloading Labels of a specific node
 func checkOffloadingLabels(na *corev1.Namespace, n *corev1.Node) bool {
 	for key := range n.GetLabels() {
 		if strings.HasPrefix(key, offloadingPrefixLabel) {
 			if _, ok := na.GetLabels()[key]; !ok {
-				klog.Infof(" Namespace '%s' cannot be offloaded on remote cluster: %s", na.GetName(), n.Annotations[const_ctrl.VirtualNodeClusterId])
+				klog.Infof(" Namespace '%s' cannot be offloaded on remote cluster: %s", na.GetName(), n.Annotations[liqocontrollerutils.VirtualNodeClusterId])
 				return false
 			}
 		}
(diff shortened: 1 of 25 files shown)


pkg/liqo-controller-manager/namespace-controller/manage_remote_namespaces.go
```diff
@@ -42,16 +44,23 @@ func (r *NamespaceReconciler) addDesiredMapping(n *corev1.Namespace, remoteName
...
-		// if entries are already here, but mappingLabel has a different value from the previous one, we force again the old value.
-		// Common user cannot change remote namespace name while the namespace is offloaded onto remote clusters
+		// if entry is already here, but mappingLabel has a different value from the previous one, we force again the old value.
+		// User's cannot change remote namespace name while the namespace is offloaded onto remote clusters
 		if oldValue != remoteName {
 			n.GetLabels()[mappingLabel] = oldValue
+			if n.GetAnnotations() == nil {
+				n.Annotations = map[string]string{}
+			}
+			n.GetAnnotations()[mappingAnnotationRenaming] = fmt.Sprintf("You cannot change the value of label [%s] because namespace's offload has already started with name [%s]. Please read our documentation for more info [%s]",
```
giorio94, line 54:
```suggestion
			n.GetAnnotations()[mappingAnnotationRenaming] = fmt.Sprintf("You cannot change the value of label [%s] because the namespace is already offloaded with name [%s]. Please read our documentation for more info [%s]",
```

pkg/utils/cluster_info.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	clusterIDConfMap = "cluster-id"
+	// Todo: Is there a place where i can get this?
+	configMapNamespace = "liqo"
```
giorio94, line 14: This should be resolved before merging

pkg/utils/cluster_info.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func GetClusterID(c client.Client) (string, error) {
+
+	configMaps := &corev1.ConfigMapList{}
+	// possibility to specify "Liqo" namespace if it is possible to get it somewhere
+	if err := c.List(context.TODO(), configMaps, client.InNamespace(configMapNamespace)); err != nil {
```
giorio94, line 21: Why are you listing all configmaps and not getting directly the correct one (by name or by label)? This seems very fragile to me

pkg/utils/cluster_info.go
```diff
@@ -0,0 +1,34 @@
...
+		klog.Error(err, "unable to get ConfigMapList")
+		return "", err
+	}
+	if len(configMaps.Items) != 1 {
+		err := fmt.Errorf("there are %d ConfigMaps at the moment, unstable condition", len(configMaps.Items))
```
giorio94, line 26: What does "unstable" mean? ATM there are two configmaps in the liqo namespace in our deployment. This piece of code would never work

pkg/utils/const.go
```diff
@@ -1,9 +1,10 @@
-package liqo_controller_manager
+package utils
 
 const (
 	VirtualNodeClusterId            = "cluster-id" //"remote.liqo.io/clusterId"
 	MapNamespaceName                = "default"
 	TypeLabel                       = "liqo.io/type"
 	TypeNode                        = "virtual-node"
 	NamespaceMapControllerFinalizer = "namespacemap-controller.liqo.io/finalizer"
+	DocumentationUrl                = "https://github.com/liqotech/liqo/tree/master/docs"
```
giorio94, line 9:
```suggestion
	DocumentationUrl                = "https://doc.liqo.io/"
```

pkg/liqo-controller-manager/namespace-controller/manage_namespace_labels.go
```diff
@@ -10,12 +12,35 @@ import (
...
+		return err
+	}
+
+	n.Labels[mappingLabel] = fmt.Sprintf("%s-%s", n.Name, clustedID)
+
+	if n.GetAnnotations() == nil {
+		n.Annotations = map[string]string{}
+	}
+	n.GetAnnotations()[mappingAnnotationDefaulting] = fmt.Sprintf("You have not specified a name for your remote Namespace, this is your default name: [%s]. Please read our documentation for more info [%s]",
+		n.GetLabels()[mappingLabel], liqocontrollerutils.DocumentationUrl)
```
giorio94, line 28: I feel referring to the home-page of the documentation is mostly useless, as the user would not know where to look for the information about this specific issue. Additionally, ATM there is no documentation about this feature, hence it is even more misleading.

giorio94 (commented):
For the future, please split the refactoring in a separate commit, to simplify the review and highlight your changes

pkg/utils/cluster_info.go
Andreagit97: I have understood that listing resources was a best practise in order to avoid transient condition, just in case someone replicates the resources or a new resource is spawn due to transient errors. During my development I see only one ConfigMap in the liqo namespace, my fault, thank you.

pkg/utils/cluster_info.go
Andreagit97: Yes, I have already talked with Alex about this, ATM it's ok to have a costant for it.

pkg/liqo-controller-manager/namespace-controller/manage_remote_namespaces.go
Andreagit97: Thank you

pkg/utils/const.go
Andreagit97: thank you

pkg/utils/cluster_info.go
Andreagit97: Now I have understood the point, thank you

pkg/liqo-controller-manager/namespace-controller/manage_namespace_labels.go
Andreagit97: Yes you are right, thank you

PR #590, opened 2021-05-01
Linting Improvement

palexster:
# Description

This PR reinforces the linting checks on Liqo code, by adding more controls and linters.


.golangci.yml
```diff
@@ -0,0 +1,118 @@
+run:
+  timeout: 10m
+  skip-files:
+    - "zz_generated.*.go"
+    - ".*_test.go"
```
giorio94, line 5: Woudn't skip test files

.golangci.yml
```diff
@@ -0,0 +1,118 @@
...
+  exclude-rules:
+    - linters:
+        - golint
+      # Excluding, as many packages are currently affected by this
+      text: "don't use an underscore in package name"
+
+    - linters:
+        - stylecheck
+      # Excluding, as many packages are currently affected by this
+      text: "ST1003: should not use underscores in package names"
```
giorio94, line 115: Are those exclude-rules required?

.golangci.yml
```diff
@@ -0,0 +1,118 @@
...
+    - goimports
+    - golint
+    - gomodguard
+  # - gomnd
+    - goprintffuncname
+    - gosec
+    - gosimple
+    - govet
+    - ineffassign
+  # - lll
```
giorio94, line 77: I would personally enable lll, to prevent excessively long lines (probably 120 is a good compromise)

giorio94 (approved):

.golangci.yml
palexster: I would skip tests in this first phase.

giorio94: Is it there any specific reason? Do you expect those to be characterized by more linting errors or is it just a matter of reducing the amount of files to be fixed?

PR #591, opened 2021-05-01
Changing Webhook behavior

Andreagit97:
# Description

When a new `Pod` is deployed in a `Namespace` labeled with `liqo.io/scheduling="true"`, the Webhook must set the right `Pod Affinity` according to what the admin has specified in the `NamespaceOffloading.Spec.ClusterSelector` field. If the user deploys a `Pod` with some affinities, these are merged with the ones selected by the administrator in the `ClusterSelector`; if there is a conflict between the two sets of `Affinity` the `Pod` remains pending until someone deletes it.

### DIFF
--- a/cmd/liqo-webhook/main.go
+++ b/cmd/liqo-webhook/main.go
@@ -1,24 +1,47 @@
 package main
 
 import (
-	"log"
+	"context"
+	"os/signal"
+	"syscall"
+	"time"
 
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
 
 	"github.com/liqotech/liqo/pkg/mutate"
 )
 
+const gracefulPeriod = 5 * time.Second
+
 func main() {
 	config := &mutate.MutationConfig{}
-
 	setOptions(config)
 
-	log.Println("Starting server ...")
+	klog.Info("Starting server ...")
 
-	s, err := mutate.NewMutationServer(config)
+	ctx, cancel := context.WithCancel(context.Background())
+	ctxSignal, stop := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+
+	s, err := mutate.NewMutationServer(ctx, config)
 	if err != nil {
 		klog.Fatal(err)
 	}
 
+	go func() {
+		defer cancel()
+
+		<-ctxSignal.Done()
+		// Restore default signal handler.
+		stop()
+
+		ctxShutdown, cancelShutdown := context.WithTimeout(ctx, gracefulPeriod)
+		defer cancelShutdown()
+
+		klog.Info("Received signal, shutting down")
+		s.Shutdown(ctxShutdown)
+	}()
+
 	s.Serve()
+	<-ctx.Done()
+	klog.Info("Liqo webhook cleanly shutdown")
 }
--- a/pkg/consts/namespace_mapping.go
+++ b/pkg/consts/namespace_mapping.go
@@ -8,6 +8,7 @@ const (
 	// TypeLabel is the key of a Liqo label that identifies different types of nodes.
 	TypeLabel = "liqo.io/type"
 	// TypeNode is the value of a Liqo label that identifies Liqo virtual nodes.
+	// todo: change to VirtualNodeType
 	TypeNode = "virtual-node"
 	// NamespaceMapControllerFinalizer is the finalizer inserted on NamespaceMap by NamespaceMap Controller.
 	// todo: has to be removed after VirtualNode Controller refactor
--- a/pkg/consts/webhook.go
+++ b/pkg/consts/webhook.go
@@ -1,0 +1,7 @@
+package consts
+
+const (
+	// VirtualNodeTolerationKey all Pods that have to be scheduled on virtual nodes must have this toleration
+	// to Liqo taint.
+	VirtualNodeTolerationKey = "virtual-node.liqo.io/not-allowed"
+)
--- a/pkg/mutate/manage_pod_mutation.go
+++ b/pkg/mutate/manage_pod_mutation.go
@@ -1,0 +1,158 @@
+package mutate
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+
+	offv1alpha1 "github.com/liqotech/liqo/apis/offloading/v1alpha1"
+	liqoconst "github.com/liqotech/liqo/pkg/consts"
+)
+
+// getVirtualNodeToleration returns a new Toleration for the Liqo's virtual-nodes.
+func getVirtualNodeToleration() corev1.Toleration {
+	return corev1.Toleration{
+		Key:      liqoconst.VirtualNodeTolerationKey,
+		Operator: corev1.TolerationOpExists,
+		Effect:   corev1.TaintEffectNoExecute,
+	}
+}
+
+// createTolerationFromNamespaceOffloading creates a new virtualNodeToleration in case of LocalAndRemotePodOffloadingStrategyType
+// or RemotePodOffloadingStrategyType. In case of PodOffloadingStrategyType not recognized, returns an error.
+func createTolerationFromNamespaceOffloading(strategy offv1alpha1.PodOffloadingStrategyType) (corev1.Toleration, error) {
+	var toleration corev1.Toleration
+	switch {
+	case strategy == offv1alpha1.LocalAndRemotePodOffloadingStrategyType, strategy == offv1alpha1.RemotePodOffloadingStrategyType:
+		// The virtual-node toleration must be added.
+		toleration = getVirtualNodeToleration()
+	default:
+		err := fmt.Errorf("PodOffloadingStrategyType '%s' not recognized", strategy)
+		klog.Error(err)
+		return corev1.Toleration{}, err
+	}
+	return toleration, nil
+}
+
+// createNodeSelectorFromNamespaceOffloading creates the right NodeSelector according to the PodOffloadingStrategy chosen.
+func createNodeSelectorFromNamespaceOffloading(nsoff *offv1alpha1.NamespaceOffloading) (corev1.NodeSelector, error) {
+	nodeSelector := nsoff.Spec.ClusterSelector
+	switch {
+	case nsoff.Spec.PodOffloadingStrategy == offv1alpha1.RemotePodOffloadingStrategyType:
+		// To ensure that the pod is not scheduled on local nodes is necessary to add to every NodeSelectorTerm a
+		// new NodeSelectorRequirement. This NodeSelectorRequirement requires explicitly the label
+		// "liqo.io/type=virtual-node" to exclude local nodes from the scheduler choice.
+		for i := range nodeSelector.NodeSelectorTerms {
+			nodeSelector.NodeSelectorTerms[i].MatchExpressions = append(nodeSelector.NodeSelectorTerms[i].MatchExpressions,
+				corev1.NodeSelectorRequirement{
+					Key:      liqoconst.TypeLabel,
+					Operator: corev1.NodeSelectorOpIn,
+					Values:   []string{liqoconst.TypeNode},
+				})
+		}
+	case nsoff.Spec.PodOffloadingStrategy == offv1alpha1.LocalAndRemotePodOffloadingStrategyType:
+		// A new NodeSelectorTerm that allows scheduling the pod also on local nodes is added.
+		newNodeSelectorTerm := corev1.NodeSelectorTerm{
+			MatchExpressions: []corev1.NodeSelectorRequirement{{
+				Key:      liqoconst.TypeLabel,
+				Operator: corev1.NodeSelectorOpNotIn,
+				Values:   []string{liqoconst.TypeNode},
+			}},
+		}
+		nodeSelector.NodeSelectorTerms = append(nodeSelector.NodeSelectorTerms, newNodeSelectorTerm)
+	default:
+		err := fmt.Errorf("PodOffloadingStrategyType '%s' not recognized", nsoff.Spec.PodOffloadingStrategy)
+		klog.Error(err)
+		return corev1.NodeSelector{}, err
+	}
+	return nodeSelector, nil
+}
+
+// getMergedNodeSelector gets the old PodNodeSelector and merges it with the ImposedNodeSelector.
+// Every MatchExpression of the PodNodeSelector must be merged with all the MatchExpressions of the ImposedNodeSelector:
+// n PodNodeSelector MatchExpressions.
+// m ImposedNodeSelector MatchExpressions.
+// m * n MergedNoseSelector MatchExpressions.
+func getMergedNodeSelector(podNodeSelector *corev1.NodeSelector,
+	imposedNodeSelector *corev1.NodeSelector) corev1.NodeSelector {
+	mergedNodeSelector := corev1.NodeSelector{NodeSelectorTerms: []corev1.NodeSelectorTerm{}}
+	for i := range podNodeSelector.NodeSelectorTerms {
+		for j := range imposedNodeSelector.NodeSelectorTerms {
+			newMatchExpression := append(imposedNodeSelector.NodeSelectorTerms[j].MatchExpressions,
+				podNodeSelector.NodeSelectorTerms[i].MatchExpressions...)
+			mergedNodeSelector.NodeSelectorTerms = append(mergedNodeSelector.NodeSelectorTerms, corev1.NodeSelectorTerm{
+				MatchExpressions: newMatchExpression,
+			})
+		}
+	}
+	return mergedNodeSelector
+}
+
+// fillPodWithTheNewNodeSelector gets the previously computed NodeSelector imposed by the PodOffloadingStrategy and
+// merges it with the Pod NodeSelector if it is already present. It simply adds it to the Pod if previously unset.
+func fillPodWithTheNewNodeSelector(imposedNodeSelector *corev1.NodeSelector, pod *corev1.Pod) {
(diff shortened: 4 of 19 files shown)


test/unit/webhook/manage_affinity_test.go
```diff
@@ -0,0 +1,163 @@
+package webhookLogic
```
palexster, line 1: Tests should be formatted with Ginkgo/Gomega libraries.

pkg/mutate/manage_nodeaffinity.go
```diff
@@ -0,0 +1,115 @@
...
+// ManageAffinity sets right Pod Affinity according to the PodOffloadingStrategy specified in NamespaceOffloading resource.
+func ManageAffinity(namespaceOffloading *offv1alpha1.NamespaceOffloading, pod *corev1.Pod) {
+	// This newNodeSelector will substitute the old nodeSelector if the strategy is not "LocalPodOffloadingStrategy".
+	newNodeSelector := &corev1.NodeSelector{NodeSelectorTerms: []corev1.NodeSelectorTerm{}}
+
+	// 1 - Fill the newNodeSelector according to the ClusterSelector field of NamespaceOffloading.
+	switch {
+	case namespaceOffloading.Spec.PodOffloadingStrategy == offv1alpha1.LocalAndRemotePodOffloadingStrategyType:
+		// 1 - Liqo toleration
+		klog.Info("LocalAndRemotePodOffloadingStrategy")
```
palexster, line 25: Some suggestions:
1. We should understand which resource is concerned by this log
2. Maybe it should be printed only when the verbosity is increased.

pkg/mutate/manage_nodeaffinity.go
```diff
@@ -0,0 +1,115 @@
...
+
+		pod.Spec.Tolerations = append(pod.Spec.Tolerations, corev1.Toleration{
+			Key:      liqoconst.VirtualNodeTolerationKey,
+			Operator: corev1.TolerationOpExists,
+			Effect:   corev1.TaintEffectNoExecute,
+		})
+
+		klog.Info("toleration inserted")
```
palexster, line 33: Ditto.

pkg/mutate/manage_nodeaffinity.go
```diff
@@ -0,0 +1,115 @@
...
+
+// todo: if the clusterSelcetor is empty at NamespaceOffloading creation time, in the NamespaceOffloading Controller
+//       we have to set with a default matchexpression for virtual nodes only, so this is never empty
+
+// This function is global in order to access it with test in test/unit/webhookLogic
+
+// ManageAffinity sets right Pod Affinity according to the PodOffloadingStrategy specified in NamespaceOffloading resource.
+func ManageAffinity(namespaceOffloading *offv1alpha1.NamespaceOffloading, pod *corev1.Pod) {
+	// This newNodeSelector will substitute the old nodeSelector if the strategy is not "LocalPodOffloadingStrategy".
+	newNodeSelector := &corev1.NodeSelector{NodeSelectorTerms: []corev1.NodeSelectorTerm{}}
```
palexster, line 19: I think this may be acceptable in the first place but then I would tend to merge the existing selector with the Liqo one. @giorio94 what do you think about it?

Andreagit97: yes it is already merged, but I'll make it clearer in the next commit.

pkg/liqonet/natmappinginflater/natMappingInflater_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/fake"
 
-	netv1alpha1 "github.com/liqotech/liqo/apis/net/v1alpha1"
+	liqonetapi "github.com/liqotech/liqo/apis/net/v1alpha1"
```
aleoli, line 16: Why did you change the import name?

pkg/mutate/manage_pod_mutation.go
```diff
@@ -0,0 +1,132 @@
...
+
+// createTolerationFromNamespaceOffloading creates a new virtualNodeToleration in case of LocalAndRemotePodOffloadingStrategyType
+// or RemotePodOffloadingStrategyType. In case of LocalPodOffloadingStrategyType returns an empty Toleration.
+func createTolerationFromNamespaceOffloading(strategy offv1alpha1.PodOffloadingStrategyType) corev1.Toleration {
+	var toleration corev1.Toleration
+	switch {
+	case strategy == offv1alpha1.LocalAndRemotePodOffloadingStrategyType, strategy == offv1alpha1.RemotePodOffloadingStrategyType:
+		// The virtual-node toleration must be added.
+		toleration = getVirtualNodeToleration()
+	default:
```
aleoli, line 28: In my opinion, we should add a case for `LocalPodOffloadingStrategyType` and print and error in the default case.

This should prevent future errors, for example, if we will add other strategies and we will forget to handle them in this switch.

pkg/mutate/manage_pod_mutation.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	case nsoff.Spec.PodOffloadingStrategy == offv1alpha1.LocalAndRemotePodOffloadingStrategyType:
+		nodeSelector = nsoff.Spec.ClusterSelector
+		// A new NodeSelectorTerm terms that allows scheduling the pod also on local nodes is added.
+		newNodeSelectorTerm := corev1.NodeSelectorTerm{
+			MatchExpressions: append([]corev1.NodeSelectorRequirement{}, corev1.NodeSelectorRequirement{
+				Key:      liqoconst.TypeLabel,
+				Operator: corev1.NodeSelectorOpNotIn,
+				Values:   []string{liqoconst.TypeNode},
+			}),
```
aleoli, line 60:
```suggestion
			MatchExpressions: []corev1.NodeSelectorRequirement{
				{
					Key:      liqoconst.TypeLabel,
					Operator: corev1.NodeSelectorOpNotIn,
					Values:   []string{liqoconst.TypeNode},
				}
			}),
```

pkg/mutate/manage_pod_mutation.go
```diff
@@ -0,0 +1,132 @@
...
+// createNodeSelectorFromNamespaceOffloading creates the right NodeSelector according to the PodOffloadingStrategy chosen.
+func createNodeSelectorFromNamespaceOffloading(nsoff *offv1alpha1.NamespaceOffloading) corev1.NodeSelector {
+	var nodeSelector corev1.NodeSelector
+	switch {
+	case nsoff.Spec.PodOffloadingStrategy == offv1alpha1.RemotePodOffloadingStrategyType:
+		// To ensure that the pod is not scheduled on local nodes is necessary to add a new NodeSelectorRequirement which
+		// requires the label "liqo.io/type=virtual-node", to every NodeSelectorTerm.
+		nodeSelector = nsoff.Spec.ClusterSelector
+		for i := range nodeSelector.NodeSelectorTerms {
+			nodeSelector.NodeSelectorTerms[i].MatchExpressions = append(nodeSelector.NodeSelectorTerms[i].MatchExpressions,
```
aleoli, line 45: Can you add comments on the reason why you are adding this expression to each term?

pkg/mutate/manage_pod_mutation.go
```diff
@@ -0,0 +1,132 @@
...
+		nodeSelector.NodeSelectorTerms = append(nodeSelector.NodeSelectorTerms, newNodeSelectorTerm)
+	case nsoff.Spec.PodOffloadingStrategy == offv1alpha1.LocalPodOffloadingStrategyType:
+		// In local strategy there is no necessity to impose a particular NodeSelector.
+		nodeSelector = corev1.NodeSelector{NodeSelectorTerms: []corev1.NodeSelectorTerm{}}
+	}
+	return nodeSelector
+}
+
+// fillPodWithImposedNodeSelector gets the previously computed NodeSelector imposed by the PodOffloadingStrategy and
+// merges it with the Pod NodeSelector if it is already present. It simply adds it to the Pod if there is merge to do.
```
giorio94, line 71:
```suggestion
// merges it with the Pod NodeSelector if it is already present. It simply adds it to the Pod if previously unset.
```

pkg/mutate/manage_pod_mutation.go
```diff
@@ -0,0 +1,132 @@
...
+func fillPodWithImposedNodeSelector(imposedNodeSelector *corev1.NodeSelector, pod *corev1.Pod) {
+	// To preserve the Pod Affinity content, it is necessary to add the imposedNodeSelector according to what
+	// is already present in the Pod Affinity.
+	switch {
+	case pod.Spec.Affinity == nil:
+		pod.Spec.Affinity = &corev1.Affinity{
+			NodeAffinity: &corev1.NodeAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{},
```
giorio94, line 79: You could directly assign the `imposedNameSelector` here, instead of creating a placeholder. The same below

pkg/mutate/manage_pod_mutation.go
```diff
@@ -0,0 +1,132 @@
...
+		oldNodeSelector := *pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+		tmpSelector := corev1.NodeSelector{NodeSelectorTerms: []corev1.NodeSelectorTerm{}}
+		for i := range oldNodeSelector.NodeSelectorTerms {
+			for j := range imposedNodeSelector.NodeSelectorTerms {
+				newMatchExpression := append(imposedNodeSelector.NodeSelectorTerms[j].MatchExpressions, oldNodeSelector.NodeSelectorTerms[i].MatchExpressions...)
+				tmpSelector.NodeSelectorTerms = append(tmpSelector.NodeSelectorTerms, corev1.NodeSelectorTerm{
+					MatchExpressions: newMatchExpression,
+				})
+			}
+		}
```
giorio94, line 98: IMHO, this logic deserves to be extracted to a separate function, for the sake of clarity and to simplify testing. More comments should also be added.

pkg/mutate/webhook_test.go
```diff
@@ -0,0 +1,617 @@
...
+							Values:   []string{"C,D"},
+						}, {
+							Key:      "NotProvider",
+							Operator: corev1.NodeSelectorOpExists,
+						},
+						},
+					},
+				},
+				},
+			},
```
giorio94, line 80: It seems to me this is copyed and pasted three times, while changing the `PodOffloadingStrategy` only. You should use a function to generate the three items, avoiding code duplication and highlighting the differences.

pkg/mutate/webhook_test.go
```diff
@@ -0,0 +1,617 @@
...
+						Key:      "NotProvider",
+						Operator: corev1.NodeSelectorOpExists,
+					}, {
+						Key:      liqoconst.TypeLabel,
+						Operator: corev1.NodeSelectorOpIn,
+						Values:   []string{liqoconst.TypeNode},
+					},
+					},
+				},
+			}},
```
giorio94, line 184: This could also be moved to a separate function (leveraged by the one above). Similar for the other occurrences of duplicated code below.

pkg/mutate/webhook_test.go
```diff
@@ -0,0 +1,617 @@
...
+
+		It("Test that the PodAffinity in the case 2 is preserved", func() {
+			newPod := pods[2].DeepCopy()
+			newNodeSelector := imposedNodeSelector.DeepCopy()
+			fillPodWithImposedNodeSelector(newNodeSelector, newPod)
+			Expect(*newPod.Spec.Affinity.PodAffinity).To(Equal(podAffinity))
+		})
+	})
+
+	Context("4 - Call the mutatePod function and observe is pod is correctly mutated", func() {
```
giorio94, line 449:
```suggestion
	Context("4 - Call the mutatePod function and observe the pod is correctly mutated", func() {
```

pkg/liqonet/natmappinginflater/natMappingInflater_test.go
Andreagit97: GCI tool reported to me there were 2 identical imports called with a different name. So I have removed one of these. I don't know why in the old version it only reports one now. Maybe I will try to change again the name. Thank you

pkg/mutate/manage_pod_mutation.go
Andreagit97: Great idea, thank you

pkg/mutate/manage_pod_mutation.go
Andreagit97: I have modified a little bit the previous comment, I hope it is clearer now.

pkg/consts/namespace_mapping.go
```diff
@@ -8,6 +8,7 @@ const (
 	// TypeLabel is the key of a Liqo label that identifies different types of nodes.
 	TypeLabel = "liqo.io/type"
 	// TypeNode is the value of a Liqo label that identifies Liqo virtual nodes.
+	// todo: change to VirtualNodeType
```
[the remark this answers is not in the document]
Andreagit97: it will be addressed in the PR that will activate the new Namespace Mapping logic

cmd/liqo-webhook/main.go
```diff
@@ -13,12 +13,16 @@ func main() {
 
 	setOptions(config)
 
-	log.Println("Starting server ...")
+	klog.Info("Starting server ...")
 
-	s, err := mutate.NewMutationServer(config)
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
Andreagit97: thank you I have adopted this approach, it should work

pkg/consts/webhook.go
```diff
@@ -0,0 +1,7 @@
+package consts
+
+const (
+	// VirtualNodeTolerationKey all Pods that have to be scheduled on virtual nodes must have this toleration
+	// to Liqo taint.
+	VirtualNodeTolerationKey = "virtual-node.liqo.io/not-allowed"
```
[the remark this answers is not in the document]
Andreagit97: I only found one recurrence of this constant throughout the whole project, the one that you have indicated above.

pkg/mutate/server.go
```diff
@@ -18,13 +27,56 @@ type MutationServer struct {
 	mux    *http.ServeMux
 	server *http.Server
 
-	config *MutationConfig
+	webhookClient client.Client
+	config        *MutationConfig
 }
 
-func NewMutationServer(c *MutationConfig) (*MutationServer, error) {
+func getWebhookClient(ctx context.Context) client.Client {
```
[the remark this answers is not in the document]
Andreagit97: you are right

pkg/mutate/manage_pod_mutation.go
```diff
@@ -0,0 +1,132 @@
...
+	klog.V(5).Infof("Generated Toleration: %s", toleration)
+
+	// Create the right NodeSelector according to the PodOffloadingStrategy case.
+	imposedNodeSelector := createNodeSelectorFromNamespaceOffloading(namespaceOffloading)
+	klog.V(5).Infof("ImposedNodeSelector: %s", imposedNodeSelector)
+
+	// If there is no toleration, return (LocalPodOffloadingStrategy case).
+	if toleration.MatchToleration(&corev1.Toleration{}) {
+		return
+	}
```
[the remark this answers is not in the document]
Andreagit97: yes it is a good idea, thank you

pkg/mutate/manage_pod_mutation.go
Andreagit97: you are right, I have created a new function

pkg/mutate/server.go
```diff
@@ -18,13 +27,57 @@ type MutationServer struct {
 	mux    *http.ServeMux
 	server *http.Server
 
-	config *MutationConfig
+	webhookClient client.Client
+	config        *MutationConfig
+}
+
+func getWebhookClient(ctx context.Context) (client.Client, error) {
```
palexster, line 34: Can be this method factorized in the utils package externalizing the scheme? I think it can be reused outside the webhook.

palexster (approved):

giorio94 (commented):
/lgtm

Just a couple of very minor nits inline.
I have a doubt concerning the server shutdown part, but it is better if we discuss in person about that.

pkg/mutate/manage_pod_mutation.go
```diff
@@ -0,0 +1,157 @@
...
+				imposedNodeSelector)
+	}
+}
+
+// mutatePod checks the NamespaceOffloading CR associated with the Pod's Namespace.
+// The pod is modified in different ways according to the PodOffloadingStrategyType
+// chosen in the CR. Two possible modifications:
+// - The VirtualNodeToleration is added to the Pod Toleration if necessary.
+// - The old Pod NodeSelector is substituted with a new one according to the PodOffloadingStrategyType.
+func mutatePod(namespaceOffloading *offv1alpha1.NamespaceOffloading, pod *corev1.Pod) {
```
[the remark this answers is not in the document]
Andreagit97: good idea, thank you

pkg/mutate/server.go
Andreagit97: Sure, I have created a new method in liqo utils.

pkg/utils/clientWithCache/get_client_with_cache.go
```diff
@@ -0,0 +1,52 @@
+// Package clientwithcache contains utility methods to create a new controller runtime client with cache.
+package clientwithcache
```
giorio94, line 2:
```suggestion
package cachedclient
```

Just a proposal though. The same for the filename.

giorio94 (approved):
/lgtm

Just squash the commits before merging

PR #594, opened 2021-05-05
Refactor of Liqo controllers constants

Andreagit97:
# Description

This PR creates a new folder for Liqo controllers constants (from liqo/pkg/liqo-controller-manager to liqo/pkg/consts). Constants will be divided in files, which will specify their context.

### DIFF
--- a/pkg/consts/namespaceMapping.go
+++ b/pkg/consts/namespaceMapping.go
@@ -1,9 +1,10 @@
-package liqo_controller_manager
+package consts
 
 const (
-	VirtualNodeClusterId            = "cluster-id" //"remote.liqo.io/clusterId"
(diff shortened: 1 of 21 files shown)


pkg/consts/namespaceMapping.go
```diff
@@ -1,9 +1,10 @@
-package liqo_controller_manager
+package consts
 
 const (
 	VirtualNodeClusterId            = "cluster-id" //"remote.liqo.io/clusterId"
```
palexster, line 4: You need a space after the //

palexster (approved):

pkg/consts/namespaceMapping.go
Andreagit97: Thank you.

aleoli (approved):

Issue #548, opened 2021-03-30
Broadcaster refactoring


giuse2596:
This issue describes all steps to follow to get a complete refactor of Peering phase between two clusters. 
PeeringRequest CR will be replaced by ResourceRequest  CR, Advertisement CR will be replaced by ResourceOffer CR, a new controller will be created to Reconcile ResourceRequests and create new ResourceOffer locally. A new version of Broadcaster has the function of starting an Informer to watch nodes and pods and then export all cluster resources available which can be offered in ResourceOffer. Finally an update logic of ResourceOffer will be created. 

TODO list:

- [x] APIs for  Resource Request and Resource Offer #552 
- [x] Resource Offer controller with static resources #560 
- [x] Logic to watch nodes #595 
- [x] Resource Offer updates
- [x] Pods Updates #654 
- [x] Switch to the new version

palexster:
Closing as completely addressed.

PR #595, opened 2021-05-05
Broadcaster Logic

giuse2596:
# Description

This PR introduce a new component: the Broadcaster. It starts an informer to watch nodes in the cluster and offer some methods to read all resource allocatable scaled by a parameter in clusterConfig, used to generate a complete resouceOffer. 

Refers to third point of Issue #548

### DIFF
--- a/cmd/advertisement-operator/main.go
+++ b/cmd/advertisement-operator/main.go
@@ -13,6 +13,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
 	"sync"
@@ -172,10 +173,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	newBroadcaster := &resourceRequestOperator.Broadcaster{}
+
+	if err := newBroadcaster.SetupBroadcaster(clientset, time.Duration(resyncPeriod)); err != nil {
+		klog.Error(err)
+		os.Exit(1)
+	}
+
 	resourceRequestReconciler := &resourceRequestOperator.ResourceRequestReconciler{
-		Client:    mgr.GetClient(),
-		Scheme:    mgr.GetScheme(),
-		ClusterID: clusterId,
+		Client:      mgr.GetClient(),
+		Scheme:      mgr.GetScheme(),
+		ClusterID:   clusterId,
+		Broadcaster: newBroadcaster,
 	}
 
 	if err = resourceRequestReconciler.SetupWithManager(mgr); err != nil {
@@ -198,16 +207,18 @@ func main() {
 		os.Exit(1)
 	}
 
-	c := make(chan struct{})
 	var wg = &sync.WaitGroup{}
 	client, err := advertisementReconciler.InitCRDClient(localKubeconfig)
 	if err != nil {
 		os.Exit(1)
 	}
-	wg.Add(3)
-	go advertisementReconciler.CleanOldAdvertisements(c, wg)
+	wg.Add(5)
+	ctx, cancel := context.WithCancel(context.Background())
+	go advertisementReconciler.CleanOldAdvertisements(ctx.Done(), wg)
 	// TODO: this configuration watcher will be refactored before the release 0.3
 	go advertisementReconciler.WatchConfiguration(localKubeconfig, client, wg)
+	go newBroadcaster.WatchConfiguration(localKubeconfig, client, wg)
+	go newBroadcaster.StartBroadcaster(ctx, wg)
 	go resourceOfferReconciler.WatchConfiguration(localKubeconfig, client, wg)
 
 	klog.Info("starting manager as advertisementoperator")
@@ -215,7 +226,8 @@ func main() {
 		klog.Error(err)
 		os.Exit(1)
 	}
-	close(c)
+
 	close(client.Stop)
+	cancel()
 	wg.Wait()
 }
--- a/internal/resource-request-operator/broadcaster.go
+++ b/internal/resource-request-operator/broadcaster.go
@@ -1,0 +1,194 @@
+package resourcerequestoperator
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	crdclient "github.com/liqotech/liqo/pkg/crdClient"
+
+	configv1alpha1 "github.com/liqotech/liqo/apis/config/v1alpha1"
+	"github.com/liqotech/liqo/pkg/utils"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+// Broadcaster is an object which is used to get resources of the cluster.
+type Broadcaster struct {
+	allocatable   corev1.ResourceList
+	clusterConfig configv1alpha1.ClusterConfig
+	offerMutex    sync.RWMutex
+	configMutex   sync.RWMutex
+	informer      cache.SharedInformer
+}
+
+// SetupBroadcaster create the informer e run it to signal node changes updating Offers.
+func (b *Broadcaster) SetupBroadcaster(clientset *kubernetes.Clientset, resyncPeriod time.Duration) error {
+	b.allocatable = corev1.ResourceList{}
+	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
+	b.informer = factory.Core().V1().Nodes().Informer()
+	b.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    b.onAdd,
+		UpdateFunc: b.onUpdate,
+		DeleteFunc: b.onDelete,
+	})
+
+	return nil
+}
+
+// StartBroadcaster starts a new sharedInformer to watch nodes resources.
+func (b *Broadcaster) StartBroadcaster(ctx context.Context, group *sync.WaitGroup) {
+	defer group.Done()
+	b.informer.Run(ctx.Done())
+}
+
+// WatchConfiguration starts a new watcher to get clusterConfig.
+func (b *Broadcaster) WatchConfiguration(localKubeconfig string, crdClient *crdclient.CRDClient, wg *sync.WaitGroup) {
+	defer wg.Done()
+	utils.WatchConfiguration(b.setConfig, crdClient, localKubeconfig)
+}
+
+func (b *Broadcaster) setConfig(configuration *configv1alpha1.ClusterConfig) {
+	b.configMutex.Lock()
+	defer b.configMutex.Unlock()
+	b.clusterConfig = *configuration
+}
+
+func (b *Broadcaster) getConfig() *configv1alpha1.ClusterConfig {
+	b.configMutex.RLock()
+	defer b.configMutex.RUnlock()
+	configCopy := b.clusterConfig.DeepCopy()
+	return configCopy
+}
+
+// react to a Node Creation/First informer run.
+func (b *Broadcaster) onAdd(obj interface{}) {
+	node := obj.(*corev1.Node)
+	if node.Status.Phase == corev1.NodeRunning {
+		toAdd := &node.Status.Allocatable
+		currentResources := b.allocatable.DeepCopy()
+		addResources(currentResources, *toAdd)
+
+		if err := b.writeResources(currentResources); err != nil {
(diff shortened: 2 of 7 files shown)


internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,158 @@
...
+		klog.Errorf("OnAdd error: unable to write allocatable of Node %s: %s", node.Name, err)
+	}
+}
+
+// write cluster resources in thread safe mode.
+func (b *Broadcaster) writeResources(newResources corev1.ResourceList) error {
+	b.offerMutex.Lock()
+	defer b.offerMutex.Unlock()
+	if newResources != nil {
+		b.allocatable = newResources
```
aleoli, line 151: Can we add a DeepCopy function here?

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,183 @@
...
+	oldNode := oldObj.(*corev1.Node)
+	newNode := newObj.(*corev1.Node)
+	oldNodeResources := oldNode.Status.Allocatable
+	newNodeResources := newNode.Status.Allocatable
+	currentResources := b.allocatable.DeepCopy()
+	if newNode.Status.Phase == corev1.NodeRunning {
+		// node was already Running, update with possible new resources.
+		if oldNode.Status.Phase == corev1.NodeRunning {
+			for resourceName, quantity := range newNodeResources {
+				oldQuantity := oldNodeResources[resourceName]
```
aleoli, line 92: This function seems to be very complex, and with similar/repeated codes, can we split it into different functions?

Then I would suggest something like this to simplify the logic:
* if old is running subtract the resources
* if new is running add the resources

we don't care about the combination of the two, if the result is different we will update the stored value

internal/resource-request-operator/resourceRequest_controller_test.go
```diff
@@ -27,8 +30,39 @@ var _ = Describe("ResourceRequest controller", func() {
...
 			ctx := context.Background()
+			By("Creating mock nodes")
+			resources := corev1.ResourceList{}
+			resources[corev1.ResourceCPU] = *resource.NewQuantity(2, resource.DecimalSI)
+			resources[corev1.ResourceMemory] = *resource.NewQuantity(1, resource.BinarySI)
+			resources[corev1.ResourceLimitsCPU] = *resource.NewQuantity(3, resource.DecimalSI)
+			node1 := &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-node1",
+					Namespace: "default",
```
aleoli, line 42: The nodes are cluster scoped resources, they do not require setting the namespace

internal/resource-request-operator/broadcaster.go
giuse2596: And what about the case when old is running and new is running too but allocatable resources change? I think we have to update subtracting or adding the difference between old and new but probably I am wrong.

aleoli: ok, you are right

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,184 @@
...
+		return
+	}
+
+	quantity.Set(quantity.Value() * percentage / 100)
+}
+
+// addResources is an utility function to add resources.
+func addResources(currentResources, toAdd *corev1.ResourceList) {
+	for resourceName, quantity := range *toAdd {
+		value := (*currentResources)[resourceName]
```
aleoli, line 154: I would add a check to ensure that the resource exists in the map

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		(*currentResources)[resourceName] = value
+	}
+}
+
+// subResources is an utility function to subtract resources.
+func subResources(currentResources, toSub *corev1.ResourceList) {
+	for resourceName, quantity := range *toSub {
+		value := (*currentResources)[resourceName]
+		if !quantity.IsZero() {
```
aleoli, line 166: We check many times in the code that the quality has to be different from zero, is it a problem to add/subtract zero quantities?

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,184 @@
...
+			value.Sub(quantity)
+		}
+		(*currentResources)[resourceName] = value
+	}
+}
+
+// updateResources is an utility function to update resources.
+func updateResources(currentResources, oldResources, newResources *corev1.ResourceList) {
+	for resourceName, quantity := range *newResources {
+		oldQuantity := (*oldResources)[resourceName]
```
aleoli, line 176: some as before, existence check

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,191 @@
...
+	})
+
+	config.ContentConfig.GroupVersion = &configv1alpha1.GroupVersion
+	config.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
+	config.UserAgent = rest.DefaultKubernetesUserAgent()
+	CRDclient, err := crdclient.NewFromConfig(config)
+	if err != nil {
+		return err
+	}
+	go utils.WatchConfiguration(b.setConfig, CRDclient, "")
```
palexster, line 49: I would put this watch configuration outside the SetupBroadcaster, where other controllers do the same operation.

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,191 @@
...
+	go informer.Run(stopper)
```
palexster, line 50: I would start the broadcaster in a separate method that is not dedicated to the setup.

cmd/advertisement-operator/main.go
```diff
@@ -163,10 +163,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	newBroadcaster := &resourceRequestOperator.Broadcaster{}
+	componentStopper := make(chan struct{})
```
giorio94, line 167: You should be probably using contexts (WithCancel) instead of just channels if you need to stop a component

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,194 @@
...
+	clusterConfig configv1alpha1.ClusterConfig
+	offerMutex    sync.RWMutex
+	configMutex   sync.RWMutex
+	informer      cache.SharedInformer
+}
+
+// SetupBroadcaster create the informer e run it to signal node changes updating Offers.
+func (b *Broadcaster) SetupBroadcaster(clientset *kubernetes.Clientset) error {
+	b.allocatable = corev1.ResourceList{}
+	factory := informers.NewSharedInformerFactory(clientset, 0)
```
giorio94, line 32: The resync period should not be hardcoded to 0

cmd/advertisement-operator/main.go
```diff
@@ -175,22 +184,23 @@ func main() {
 
 	// +kubebuilder:scaffold:builder
 
-	c := make(chan struct{})
 	var wg = &sync.WaitGroup{}
 	client, err := advertisementReconciler.InitCRDClient(localKubeconfig)
 	if err != nil {
 		os.Exit(1)
 	}
 	wg.Add(2)
```
palexster, line 192: Should be increased to take in account newly added routines

cmd/advertisement-operator/main.go
```diff
@@ -175,22 +184,23 @@ func main() {
...
-	go advertisementReconciler.CleanOldAdvertisements(c, wg)
+	go advertisementReconciler.CleanOldAdvertisements(componentStopper, wg)
 	go advertisementReconciler.WatchConfiguration(localKubeconfig, client, wg)
+	go newBroadcaster.WatchConfiguration(localKubeconfig, client, wg)
+	newBroadcaster.StartBroadcaster(componentStopper)
```
palexster, line 196: Should be part of the WaitGroup in order to have a clean collection when the process is terminated.

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,194 @@
...
+
+	quantity.Set(quantity.Value() * percentage / 100)
+}
+
+// addResources is an utility function to add resources.
+func addResources(currentResources, toAdd *corev1.ResourceList) {
+	for resourceName, quantity := range *toAdd {
+		if value, exists := (*currentResources)[resourceName]; exists {
+			value.Add(quantity)
+			(*currentResources)[resourceName] = value
```
giorio94, line 164: AFAIU, explicit dereferencing should not be required with arrays

internal/resource-request-operator/utils.go
```diff
@@ -17,7 +16,7 @@ import (
 
 // this function generate an empty offer.
```
giorio94, line 17: The comment should be updated

internal/resource-request-operator/suite_test.go
```diff
@@ -4,6 +4,10 @@ import (
 	"path/filepath"
```
giorio94, line 4: Imports should be regrouped in three block only

internal/resource-request-operator/resourceRequest_controller_test.go
```diff
@@ -4,6 +4,9 @@ import (
 	"context"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
```
giorio94, line 8: Should be grouped with the other imports

internal/resource-request-operator/broadcaster.go
giuse2596: I have already tried but it doesn't work. Compiler says "type *corev1.ResourceList does not support indexing"

giorio94: As maps are always passed by reference, you should be able to drop the pointer with no side effects, increasing readability

internal/resource-request-operator/utils.go
```diff
@@ -15,9 +14,9 @@ import (
 	"github.com/liqotech/liqo/pkg/discovery"
 )
 
-// this function generate an empty offer.
+// this function generate a new local ResourceOffer.
```
giorio94, line 17:
```suggestion
// generateResourceOffer generates a new local ResourceOffer.
```

cmd/advertisement-operator/main.go
giorio94: This should no longer be needed if you use contexts

cmd/advertisement-operator/main.go
```diff
@@ -204,10 +214,14 @@ func main() {
 	if err != nil {
 		os.Exit(1)
 	}
-	wg.Add(3)
-	go advertisementReconciler.CleanOldAdvertisements(c, wg)
+	wg.Add(5)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
giorio94, line 219: As you are deferring, you are causing a deadlock due to circular dependency (with `wg`). You should explicitly call `cancel()` before `wg.Wait`

internal/resource-request-operator/broadcaster.go
```diff
@@ -0,0 +1,197 @@
...
+	clusterConfig configv1alpha1.ClusterConfig
+	offerMutex    sync.RWMutex
+	configMutex   sync.RWMutex
+	informer      cache.SharedInformer
+}
+
+// SetupBroadcaster create the informer e run it to signal node changes updating Offers.
+func (b *Broadcaster) SetupBroadcaster(clientset *kubernetes.Clientset, resyncPeriod time.Duration) error {
+	b.allocatable = corev1.ResourceList{}
+	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
```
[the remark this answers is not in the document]
palexster: Ideally Yes. But I feel that this would be part of a major refactor when we move to the Liqo controller-manager.

cmd/advertisement-operator/main.go
```diff
@@ -217,5 +231,6 @@ func main() {
 	}
 	close(c)
```
palexster, line 232: Is c still used somewhere?

internal/resource-request-operator/broadcaster.go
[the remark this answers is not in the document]
giorio94: Definitely. I was just wondering whether to start moving this creation to the caller, clearly used only for this component. Just to be controller-manager proof :)

cmd/advertisement-operator/main.go
giuse2596: No, merge it is just a merge oversight. I will drop it.

giorio94 (approved):

palexster (approved):

PR #607, opened 2021-05-10
Fix Discovery Linters

aleoli:
# Description

This PR fixes some checks for the new listing rules

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -97,6 +97,7 @@ linters:
   # - wsl
 
 issues:
+  max-issues-per-linter: 0
   # Disable the default exclude patterns (as they disable the mandatory comments)
   exclude-use-default: false
   exclude:
@@ -107,3 +108,8 @@ issues:
     - linters:
         - govet
       text: 'declaration of "(err|ctx)" shadows declaration at'
+    - linters:
+        - gosec
+      # Disable the check to test that HTTP clients are not using an insecure TLS connection.
+      # We need it to contact the remote authentication services exposing a self-signed certificate
+      text: TLS InsecureSkipVerify set true.
--- a/internal/discovery/foreign-cluster-operator/auth.go
+++ b/internal/discovery/foreign-cluster-operator/auth.go
@@ -21,6 +21,7 @@ import (
 
 	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
 	"github.com/liqotech/liqo/pkg/auth"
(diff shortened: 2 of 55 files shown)


palexster (approved):

pkg/consts/crdClient.go
```diff
@@ -0,0 +1,3 @@
+package consts
+
+const Apis = "/apis"
```
giorio94, line 3:
```suggestion
const ApisPath = "/apis"
```

internal/discovery/foreign-cluster-operator/auth.go
```diff
@@ -11,19 +11,19 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	client_scheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
-
-	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
-	"github.com/liqotech/liqo/pkg/auth"
-	"github.com/liqotech/liqo/pkg/crdClient"
-	"github.com/liqotech/liqo/pkg/discovery"
-	"github.com/liqotech/liqo/pkg/kubeconfig"
```
giorio94, line 26: This should be reverted

.golangci.yml
```diff
@@ -107,3 +108,6 @@ issues:
     - linters:
         - govet
       text: 'declaration of "(err|ctx)" shadows declaration at'
+    - linters:
+        - gosec
+      text: TLS InsecureSkipVerify set true.
```
giorio94, line 113: It would be better if there were a comment explaining why this is needed

pkg/discovery/const.go
```diff
@@ -1,44 +1,67 @@
...
 const (
+	// TenantControlNamespaceLabel used to mark the tenant control namespaces
 	TenantControlNamespaceLabel = "discovery.liqo.io/tenant-control-namespace"
-	ClusterRoleLabel            = "discovery.liqo.io/cluster-role"
+	// ClusterRoleLabel used all the Liqo cluster roles
+	ClusterRoleLabel = "discovery.liqo.io/cluster-role"
 
-	ClusterIdLabel      = "discovery.liqo.io/cluster-id"
-	AuthTokenLabel      = "discovery.liqo.io/auth-token"
+	// ClusterIDLabel
```
giorio94, line 9: The comment is empty

PR #621, opened 2021-05-12
Add a script for checking imports only on .go files

Andreagit97:
# Description

This Script allows to ignore imports of auto-generated files ("zz_generated..")

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ rbacs: controller-gen
 # Run go fmt against code
 fmt:
 	go fmt ./...
-	goimports -local github.com/liqotech/liqo -w .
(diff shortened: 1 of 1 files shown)


scripts/utils/goimports_ignore.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+find "$(pwd)" -type f -name '*.go' -a ! -name '*zz_generated*' -exec goimports -local github.com/liqotech/liqo -w {} \;
```
giorio94, line 3: Would move directly to the makefile, as a one-liner

giorio94 (approved):

palexster (approved):

scripts/utils/goimports_ignore.sh
Andreagit97: yes you are right, thank you

PR #634, opened 2021-05-18
Read Remote Tenant Namespace Name

aleoli:
# Description

This pr fixes a bug that causes the impossibility to fill the remote tenant namespace field for foreign clusters with an already available identity.

With this fix, the foreign cluster operator reads it for the certificate secret.

# How Has This Been Tested?

- [x] add unit test
- [x] locally on KinD


internal/discovery/foreign-cluster-operator/auth.go
```diff
@@ -107,6 +107,28 @@ func (r *ForeignClusterReconciler) getRemoteClient(
 	return nil, nil
 }
 
+func (r *ForeignClusterReconciler) getRemoteTenantNamespace(fc *discoveryv1alpha1.ForeignCluster) (bool, error) {
```
palexster, line 110: Minor comment: I was a bit confused by the name of the method. Actually, this method is not getting and returning the remoteTenantNamespace here. On the contrary, it is setting it in the foreignCluster. Maybe rename it with something like tryFetchRemoteTenantNamespace?

internal/discovery/foreign-cluster-operator/auth.go
```diff
@@ -107,6 +107,28 @@ func (r *ForeignClusterReconciler) getRemoteClient(
...
+	}
+
+	remoteNamespace, err := r.identityManager.GetRemoteTenantNamespace(
+		fc.Spec.ClusterIdentity.ClusterID, fc.Status.TenantControlNamespace.Local)
+	if err != nil {
+		klog.Error(err)
+		return false, err
+	}
+
+	if fc.Status.TenantControlNamespace == nil {
```
palexster, line 122: Is this condition plausible? Should it not be defaulted at the API level?

palexster (approved):
Just a couple of comments

internal/discovery/foreign-cluster-operator/auth.go
giorio94: Honestly I am not sure whether it is possible to avoid this as the field is in the status. Should be tested

aleoli: ok, thanks @giorio94, I will test it

pkg/identityManager/client.go
```diff
@@ -73,3 +73,31 @@ func (certManager *certificateIdentityManager) GetConfig(remoteClusterID, namesp
 		},
 	}, nil
 }
+
+// GetRemoteTenantNamespace returns the tenant control namespace that
+// the remote cluster assigned to this peering.
+func (certManager *certificateIdentityManager) GetRemoteTenantNamespace(remoteClusterID, namespace string) (string, error) {
```
giorio94, line 79:
```suggestion
func (certManager *certificateIdentityManager) GetRemoteTenantNamespaceName(remoteClusterID, localTenantNamespaceName string) (string, error) {
```

Just a proposal to remark better what the function is doing (IIUC)

internal/discovery/foreign-cluster-operator/auth.go
giorio94: Agree that the current name appears to be misleading (AFAIU you are checking if this piece of information is updated, rather than getting something). Would also add a comment to the function (although not strictly required by the linter)

giorio94 (approved):
/lgtm

PR #635, opened 2021-05-19
Support for custom routing table and policy routing rules

alacuku:
# Description
When routes are configured on each node to send the traffic for a given remote cluster toward the gateway we use a custom routing table. And for each route we also insert a policy routing rules to lookup the our custom routing table when traffic has to reach a remote network. Doing such way we isolate the configuration needed by liqo, and the new routing table is the only truth for the configuration. Previously we used to keep the route configuration in memory but it is cumbersome to keep the cache data and the real config in sync. 

Fixes #(issue)

# How Has This Been Tested?

Unit tests have been written to test the functions.

### DIFF
--- a/pkg/liqonet/routing/common.go
+++ b/pkg/liqonet/routing/common.go
@@ -1,0 +1,308 @@
+package routing
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"reflect"
+
+	"github.com/liqotech/liqo/apis/net/v1alpha1"
+	"github.com/liqotech/liqo/pkg/liqonet"
+
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+	"k8s.io/klog/v2"
+)
+
+var (
+	routingTableID = 18952
+)
+
+// WrongParameter it is returned when parameters passed to a function are not correct.
+type WrongParameter struct {
+	Type string
+	Text string
+}
+
+func (wp *WrongParameter) Error() string {
+	return "wrong " + wp.Type + " parameters: " + wp.Text
+}
+
+// ParseIPError it is returned when net.ParseIP() fails to parse and ip address.
+type ParseIPError struct {
+	Type         string
+	IPToBeParsed string
+}
+
+func (pie *ParseIPError) Error() string {
+	return "please check that the IP address is in che correct format" + pie.Type + ": " + pie.IPToBeParsed
+}
+
+func addRoute(dstNet, gwIP string, iFaceIndex, tableID int) (bool, error) {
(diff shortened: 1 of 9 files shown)


pkg/liqonet/routing/common.go
```diff
@@ -0,0 +1,314 @@
...
+func addRoute(dstNet, gwIP string, iFaceIndex, tableID int) (bool, error) {
+	var route *netlink.Route
+	var gatewayIP net.IP
+	// Convert destination in *net.IPNet.
+	_, destinationNet, err := net.ParseCIDR(dstNet)
+	if err != nil {
+		return false, err
+	}
+	// If gwIP is not set then skip this section.
+	if gwIP != "" {
```
palexster, line 50: Would wrap this logic in a utility function for factorization

pkg/liqonet/routing/common.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	return nil
+}
+
+func addPolicyRoutingRule(fromSubnet, toSubnet string, tableID int) (bool, error) {
+	var destinationNet, sourceNet *net.IPNet
+	var err error
+
+	// Check that at least one between source and destination networks are defined.
+	if fromSubnet == "" && toSubnet == "" {
```
palexster, line 152: Would factorize this logic in a dedicated method

palexster (approved):

pkg/liqonet/routing/common_test.go
```diff
@@ -0,0 +1,489 @@
...
+				Expect(index).To(BeZero())
+				Expect(err).Should(Equal(&ParseIPError{Type: "IP address", IPToBeParsed: gwIPWrong}))
+			})
+		})
+
+		Context("when there is a route for the given IP address", func() {
+			It("should return nil and a correct link index", func() {
+				index, err := getIFaceIndexForIP(gwIPCorrect)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(index).Should(BeNumerically("==", dummylink1.Attrs().Index))
```
aleoli, line 444: Just a curiosity, why not a simple `Be(Equal(...))` instead of the `BeNumerically`?

aleoli (approved):

pkg/liqonet/routing/common_test.go
alacuku: According to the [Gomega documentation](https://onsi.github.io/gomega/#benumericallycomparator-string-compareto-interface) it's the recommended way to make assertions with numerical types.

aleoli: 👍 thanks

PR #638, opened 2021-05-20
NAT rules for ExternalCIDR (3/4)

davidefalcone1:
# Description

This PR represent the 1st of a couple of steps which are necessary in order to integrate the management of NAT routes related to ExternalCIDR IP address. It focuses on the IPAM module management of a new resource, called NatMapping. The following step will make the Gateway able to reconcile the CR by adding the NAT rules.

As mentioned previously, a new per-cluster resource NatMapping has been introduced. Its goal is to keep track of the mappings made by the IPAM between an IP belonging to ExternalCIDR and the real IP address. For the time being, it includes only the Spec section. The Spec includes the clusterID the resource refers to and a set of mappings.

A new module, NatMappingInflater, has been introduced to manage the insertion and deletion of mappings through the CR. Therefore, the IPAM has a new field of type NatMappingInflater.
The Inflater, in addition to update NatMapping resources for each cluster, updates an in-memory structure that holds NAT mappings for each cluster. This means reading a NAT mapping will be faster, since the in-memory structure will be used. Write operations are performed both on the resource and on the in-memory structure.
In order to avoid the deletion of the resource by the user, the Inflater manages also an Informer on NatMapping resource that re-creates a resource when the user wrongly deletes it.

File pkg/liqonet/errors.go has been moved to /pkg/liqonet/errors/ directory and a package called errors has been created. The reason is that otherwise an import cycle would happen if we try to assess error.go(pkg liqonet) within the NatMappingInflater(imported by liqonet).


pkg/liqonet/ipam.go
```diff
@@ -562,6 +570,78 @@ func (liqoIPAM *IPAM) FreeSubnetsPerCluster(clusterID string) error {
...
+	clusterSubnets, err := liqoIPAM.ipamStorage.getClusterSubnets()
+	if err != nil {
+		return fmt.Errorf("cannot get cluster subnets: %w", err)
+	}
+	subnets, exists := clusterSubnets[clusterID]
+	if !exists {
+		return fmt.Errorf("subnets of cluster %s are not set", clusterID)
+	}
+	var externalCIDR string
+	if subnets.LocalNATExternalCIDR == liqoconst.DefaultCIDRValue {
```
palexster, line 585: Can you comment more on this logic?

pkg/liqonet/natMappingInflater.go
```diff
@@ -0,0 +1,208 @@
...
+	natMappings.Spec.Mappings[oldIP] = newIP
+	// Update resource
+	if err := inflater.updateNatMappings(natMappings); err != nil {
+		return fmt.Errorf("cannot update natMapping resource for cluster %s: %w", clusterID, err)
+	}
+	return nil
+}
+
+// RemoveMapping deletes a mapping from the resource related to a remote cluster.
+func (inflater *NatMappingInflater) RemoveMapping(oldIP, clusterID string) error {
```
palexster, line 138: I was wondering: Would be useful to model the return values (bool, error) as @alacuku did in #652 to know if an effective change has been triggered and not just the absence of errors? @alacuku @davidefalcone1 what do you think?

palexster (approved):

PR #643, opened 2021-05-21
NamespaceMap Controller refactor

Andreagit97:
# Description

NamespaceMap controller has to create and delete remote Namespaces when entries are inserted or removed from NamespaceMaps. In particular, this PR adds this main features: 
- When there is a problem in remote namespace creation onto a remote cluster, the associated NamespaceMap is updated with a Refused phase for that namespace (Status->CurrentMapping->NamespaceEntry->RemoteStatus->Phase=Refused). The NamespaceMap Controller will try to create that Namespace until the problem will be solved with exponential backoff.

Ref. #572 **STEP 9**


pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
```diff
@@ -58,49 +59,58 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
 }
 
 // This function checks if there are Namespaces that have to be created or deleted, in accordance with DesiredMapping
-// and CurrentMapping fields.
-func (r *NamespaceMapReconciler) manageRemoteNamespaces(nm *mapsv1alpha1.NamespaceMap) error {
+// and CurrentMapping fields. It updates also NamespaceOffloading status in according to the remote Namespace Status.
+func (r *NamespaceMapReconciler) manageRemoteNamespaces(nm *mapsv1alpha1.NamespaceMap, requeue *bool) error {
```
aleoli, line 63: In my opinion, it's better to use a second return argument instead of a reference parameter, since it's more clear and readable. What do you think about that?

pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
```diff
@@ -58,49 +59,58 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
-		if err := r.deleteRemoteNamespace(nm.Labels[liqoconst.RemoteClusterID],
-			remoteStatus.RemoteNamespace); err != nil {
-			return err
+		// If exist, delete remote Namespaces
+		if remoteStatus.Phase == mapsv1alpha1.MappingAccepted {
+			if err := r.deleteRemoteNamespace(nm.Labels[liqoconst.RemoteClusterID],
+				remoteStatus.RemoteNamespace); err != nil {
+				return err
+			}
+			klog.Infof("Remote Namespace '%s' successfully deleted.", remoteStatus.RemoteNamespace)
```
aleoli, line 104: can we add more detailed logging here? For example, adding the local namespace name and the remote cluster id?

pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
Andreagit97: yes it's a great advice, thank you.

pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
Andreagit97: yes sure

Andreagit97: sorry the right log is at line 57, this is a duplicate i have removed it, thank you

pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller.go
```diff
@@ -45,42 +46,46 @@ const (
...
 
+	if requeue {
+		return ctrl.Result{}, fmt.Errorf("unable to create some remote namespace")
+	}
 	return ctrl.Result{}, nil
 }
 
-// The Controller is triggered only when the number of entries in DesiredMapping changes, so only when
-// a namespace's request is added or removed.
+// Events not filtered:
```
palexster, line 87: The refactor of this comment is not really easy to understand IMHO.

pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
```diff
@@ -58,49 +59,58 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
 	}
 
 	// if DesiredMapping field has more entries than CurrentMapping, is necessary to create new remote namespaces
 	for localName, remoteName := range nm.Spec.DesiredMapping {
-		if _, ok := nm.Status.CurrentMapping[localName]; ok {
+		if remoteStatus, ok := nm.Status.CurrentMapping[localName]; ok && remoteStatus.Phase == mapsv1alpha1.MappingAccepted {
 			continue
 		}
+		namespacePhase := mapsv1alpha1.MappingAccepted
+		message := "Your remote namespace is correctly created on this cluster"
```
palexster, line 75: I would use the 3rd person with a message like "Created remote namespace xyz". @giorio94 what do you think?

pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
```diff
@@ -58,49 +59,58 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
+func (r *NamespaceMapReconciler) manageRemoteNamespaces(nm *mapsv1alpha1.NamespaceMap) (bool, error) {
+	requeue := false
 	if nm.Status.CurrentMapping == nil {
 		nm.Status.CurrentMapping = map[string]mapsv1alpha1.RemoteNamespaceStatus{}
 	}
 
 	// if DesiredMapping field has more entries than CurrentMapping, is necessary to create new remote namespaces
 	for localName, remoteName := range nm.Spec.DesiredMapping {
-		if _, ok := nm.Status.CurrentMapping[localName]; ok {
+		if remoteStatus, ok := nm.Status.CurrentMapping[localName]; ok && remoteStatus.Phase == mapsv1alpha1.MappingAccepted {
```
palexster, line 71: May we at least cross-check that the namespace is still available and is labeled as expected?

pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
giorio94: Definitely agree

pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller.go
Andreagit97: I will rewrite it, thank you

pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
Andreagit97: I will change it, thank you

pkg/liqo-controller-manager/namespaceMap-controller/manage_remoteNamespaces.go
Andreagit97: yes sure

pkg/liqo-controller-manager/namespaceMap-controller/update_namespaceoffloading_status.go
```diff
@@ -0,0 +1,150 @@
...
+func (r *NamespaceMapReconciler) checkNamespaceOffloadingStatus(namespaceOffloading *offv1alpha1.NamespaceOffloading) error {
+	virtualNodes := &corev1.NodeList{}
+	if err := r.List(context.TODO(), virtualNodes,
+		client.MatchingLabels{liqoconst.TypeLabel: liqoconst.TypeNode}); err != nil {
+		klog.Error(err, " --> Unable to List all virtual nodes")
+		return err
+	}
+
+	if len(virtualNodes.Items) == 0 {
+		err := fmt.Errorf(" No VirtualNode at the moment")
```
palexster, line 29: AFAIK This should not be an error condition. It is legitimate not to have an active peering in a given moment

pkg/liqo-controller-manager/namespaceMap-controller/update_namespaceoffloading_status.go
```diff
@@ -0,0 +1,150 @@
...
+			if condition.Status == corev1.ConditionTrue {
+				ready++
+			} else {
+				notReady++
+			}
+		}
+
+		switch {
+		case ready == 0:
+			namespaceOffloading.Status.OffloadingPhase = offv1alpha1.AllFailedOffloadingPhaseType
```
palexster, line 53: Do we need to differentiate among SomeFailed and AllFailed? What do you think @giorio94?

giorio94: I would personally differentiate, since SomeFailed means you can still perform offloading, but some name conflict arose, while AllFailed is sorta catastrophic. Names could be clearer, but I couldn't come up with better ideas at the time :(

pkg/liqo-controller-manager/namespaceMap-controller/update_namespaceoffloading_status.go
```diff
@@ -0,0 +1,150 @@
...
+// The remote Namespace is requested to be deleted so the Ready Condition becomes "false". At the moment a remote
+// namespace is requested to be deleted only when the corresponding NamespaceOffloading resource is deleted, so
+// the global offloading phase is set to "Terminating".
+func (r *NamespaceMapReconciler) setNotReadyNamespaceCondition(localName, clusterID string) error {
+	namespaceOffloading := &offv1alpha1.NamespaceOffloading{}
+	if err := r.Get(context.TODO(), types.NamespacedName{
+		Namespace: localName,
+		Name:      liqoconst.DefaultNamespaceOffloadingName,
+	}, namespaceOffloading); err != nil {
+		klog.Errorf("%s -> unable to get namespaceOffloading for the namespace '%s'", err, localName)
```
palexster, line 112: Not Found is a possible error here? Should be caught differently?

pkg/liqo-controller-manager/namespaceMap-controller/update_namespaceoffloading_status.go
palexster: Makes sense 👍 thanks!

pkg/liqo-controller-manager/namespaceMap-controller/update_namespaceoffloading_status.go
Andreagit97: If the NamespaceMap involved in the reconcile process exists, at least the associated virtualNode must exist, I have added a check of NotFound condition at the beginning of the reconcile function.

pkg/liqo-controller-manager/namespaceMap-controller/update_namespaceoffloading_status.go
palexster: If the namespace is deleted, you will get a not found, or am I missing something?

Andreagit97: yes you are right, thank you.

palexster (approved):

pkg/liqo-controller-manager/namespaceMap-controller/update_namespaceoffloading_status.go
```diff
@@ -0,0 +1,167 @@
...
+// with setNamespaceOffloadingStatus function.
+func (r *NamespaceMapReconciler) setReadyNamespaceCondition(localName, clusterID string, status error) error {
+	namespaceOffloading := &offv1alpha1.NamespaceOffloading{}
+	if err := r.Get(context.TODO(), types.NamespacedName{
+		Namespace: localName,
+		Name:      liqoconst.DefaultNamespaceOffloadingName,
+	}, namespaceOffloading); err != nil {
+		klog.Errorf("%s -> unable to get namespaceOffloading for the namespace '%s'", err, localName)
+		return err
+	}
```
giorio94, line 90: Would prevent code duplication with the `setNotReadyNamespaceCondition` and `removeRemoteNamespaceConditions` functions, to prevent issues during future modifications.

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
```diff
@@ -18,69 +18,62 @@ package namespacemapctrl
...
-	if len(namespaceMap.Status.CurrentMapping) == 0 {
-		ctrlutils.RemoveFinalizer(namespaceMap, liqoconst.NamespaceMapControllerFinalizer)
-		if err := r.Update(context.TODO(), namespaceMap); err != nil {
-			klog.Errorf("%s --> Unable to remove '%s' from NamespaceMap '%s'", err,
-				liqoconst.NamespaceMapControllerFinalizer, namespaceMap.GetName())
-			return ctrl.Result{}, err
-		}
-		klog.Infof("Finalizer correctly removed from NamespaceMap '%s'", namespaceMap.GetName())
+	// Only when all remote namespaces associated with this NamespaceMap are deleted, remove finalizer
+	if err := r.RemoveNamespaceMapControllerFinalizer(ctx, namespaceMap); err != nil {
```
giorio94, line 68: IIUC, it is impossible to delete a NamespaceMap entry, unless all desired mapping are previously removed manually. Is this behavior desired? I doubt so

pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller_test.go
```diff
@@ -0,0 +1,418 @@
...
+			Eventually(func() bool {
+				if err := homeClient.List(context.TODO(), nms, client.MatchingLabels{liqoconst.RemoteClusterID: remoteClusterId2}); err != nil {
+					return false
+				}
+				Expect(len(nms.Items) == 1).To(BeTrue())
+				nms.Items[0].Status.CurrentMapping = nil
+				if err := homeClient.Update(context.TODO(), nms.Items[0].DeepCopy()); err != nil {
+					return false
+				}
+				return true
```
giorio94, line 90: Same here (many other repetitions below)

pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller_test.go
```diff
@@ -0,0 +1,418 @@
...
+					Namespace: namespace1Name,
+					Name:      liqoconst.DefaultNamespaceOffloadingName,
+				}, namespaceOffloading1); err != nil {
+					return false
+				}
+				if namespaceOffloading1.Status.RemoteNamespacesConditions[remoteClusterId2][0].Type != offv1alpha1.NamespaceReady ||
+					namespaceOffloading1.Status.RemoteNamespacesConditions[remoteClusterId2][0].Status != corev1.ConditionTrue {
+					return false
+				}
+				return true
```
giorio94, line 143:
```suggestion
return namespaceOffloading ...
```

pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller_test.go
```diff
@@ -0,0 +1,418 @@
...
+				if err := homeClient.Get(context.TODO(), types.NamespacedName{
+					Namespace: namespace1Name,
+					Name:      liqoconst.DefaultNamespaceOffloadingName,
+				}, namespaceOffloading1); err != nil {
+					return false
+				}
+				if namespaceOffloading1.Status.OffloadingPhase != offv1alpha1.ReadyOffloadingPhaseType {
+					return false
+				}
+				return true
```
giorio94, line 157:
```suggestion
				return namespaceOffloading1.Status.OffloadingPhase == offv1alpha1.ReadyOffloadingPhaseType
```

giorio94 (commented):
Some comments inline.

Please, pay attention to (function and variable) names, and try to make the overall logic clear. Currently, I feel it is quite hard to follow your workflow.

pkg/liqo-controller-manager/namespaceMap-controller/manage_namespacemap_controller_finalizer.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	return nil
+}
+
+// RemoveNamespaceMapControllerFinalizer remove NamespaceMap controller finalizer.
+func (r *NamespaceMapReconciler) RemoveNamespaceMapControllerFinalizer(ctx context.Context,
+	nm *mapsv1alpha1.NamespaceMap) error {
+	if len(nm.Status.CurrentMapping) == 0 {
+		ctrlutils.RemoveFinalizer(nm, liqoconst.NamespaceMapControllerFinalizer)
+		if err := r.Update(ctx, nm); err != nil {
```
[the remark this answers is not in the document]
Andreagit97: AFAIK this implementation of Patch doesn't allow to remove object fields, only to add and replace them with MergePatchType, probably I have missed something...

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
```diff
@@ -12,29 +14,28 @@ import (
...
 
-	namespace := &corev1.Namespace{
+	remoteNamespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: remoteName,
 		},
 	}
 
-	if err := r.RemoteClients[clusterID].Create(context.TODO(), namespace); err != nil {
-		klog.Errorf("unable to create namespace '%s' onto remote cluster: '%s'", namespace.Name, clusterID)
```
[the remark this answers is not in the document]
Andreagit97: Now the error message is managed outside this function, where the situation is more clear... if the namespace already exists because I have previously created it or it was already present on that cluster

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: It is impossible to delete a NamespaceMap, unless all remote Namespaces of that cluster are previously deleted, the finalizer doesn't care about DesiredMapping. So the finalizer is removed only when there are no more remote Namespace and the NamespaceMap is empty

pkg/liqo-controller-manager/namespaceMap-controller/update_namespaceoffloading_status.go
Andreagit97: you are right, thank you

pkg/liqo-controller-manager/namespaceMap-controller/namespaceMap_controller_test.go
```diff
@@ -0,0 +1,418 @@
...
+				Status:  corev1.ConditionFalse,
+				Reason:  "ClusterNotSelected",
+				Message: "You have not selected this cluster through ClusterSelector fields",
+			}, remoteClusterId1)
+
+			Eventually(func() bool {
+				if err := homeClient.Update(context.TODO(), namespaceOffloading1); err != nil {
+					return false
+				}
+				return true
```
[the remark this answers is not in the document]
Andreagit97: thank you

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
```diff
@@ -44,63 +45,89 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
+			klog.Infof("Namespace '%s' correctly deleted onto remote cluster: '%s'", remoteName, clusterID)
+			return nil
+		}
 		klog.Errorf("unable to get remote namespace '%s'", remoteName)
 		return err
 	}
 
-	if err := r.RemoteClients[clusterID].Delete(context.TODO(), remoteNamespace); err != nil {
-		klog.Errorf("unable to delete namespace '%s' onto remote cluster: '%s'", remoteName, clusterID)
-		return err
```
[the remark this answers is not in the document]
giorio94: Got it. I feel you could also just issue the deletion call and assume Kubernetes will eventually delete it, but also this strategy seems fine to me

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
```diff
@@ -44,63 +45,89 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
-		klog.Infof("Delete remote Namespace '%s'.", remoteStatus.RemoteNamespace)
+		delete(nm.Status.CurrentMapping, localName)
 	}
-	return nil
+
+	// Several conflicts are possible here
+	if err := r.Update(context.TODO(), nm); err != nil {
+		klog.Errorf("%s -> unable to update NamespaceMap '%s' Status", err, nm.Name)
+		return true
+	}
```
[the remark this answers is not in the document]
giorio94: If I am not wrong, Patch does not check for the resource version (and indeed it is not sent as part of the payload)

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
giorio94: I personally feed this pretty strange, as it would be impossible to manually delete the resource. I would expect all mappings to be deleted upon termination, and then the finalizer to be removed. @palexster WDYT?

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
```diff
@@ -44,63 +45,89 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
-	klog.Infof("Namespace '%s' correctly deleted onto remote cluster: '%s'", remoteName, clusterID)
-	return nil
+	return fmt.Errorf("remote Namespace '%s' terminating", remoteName)
 }
 
 // This function checks if there are Namespaces that have to be created or deleted, in accordance with DesiredMapping
-// and CurrentMapping fields.
-func (r *NamespaceMapReconciler) manageRemoteNamespaces(nm *mapsv1alpha1.NamespaceMap) error {
+// and CurrentMapping fields. It updates also NamespaceOffloading status in according to the remote Namespace Status.
+func (r *NamespaceMapReconciler) ensureRemoteNamespacesExistence(nm *mapsv1alpha1.NamespaceMap) bool {
```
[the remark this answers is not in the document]
giorio94: Got it

pkg/liqo-controller-manager/namespaceMap-controller/manage_namespacemap_controller_finalizer.go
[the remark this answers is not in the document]
giorio94: Definitely not an expert, but patch does allow to remove fields. If possible I would go for that just for consistency

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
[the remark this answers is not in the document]
giorio94: Gotcha

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
```diff
@@ -44,63 +45,89 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
-	// if DesiredMapping field has more entries than CurrentMapping, is necessary to create new remote namespaces
+	// For every entry of DesiredMapping crete remote Namespace if it has not already being created
 	for localName, remoteName := range nm.Spec.DesiredMapping {
-		if _, ok := nm.Status.CurrentMapping[localName]; ok {
-			continue
-		}
-		if err := r.createRemoteNamespace(nm.Labels[liqoconst.RemoteClusterID], remoteName); err != nil {
-			return err
+		namespacePhase := mapsv1alpha1.MappingAccepted
+		var status error
```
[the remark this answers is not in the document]
giorio94: I personally feel err would be clearer (or creationError to be more verbose). Also the check `status == nil` is a bit strange.
Still, this is mostly a matter of personal preference

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
```diff
@@ -44,63 +45,89 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
+		if status = r.createRemoteNamespace(nm.Labels[liqoconst.RemoteClusterID], remoteName); status != nil {
+			if apierrors.IsAlreadyExists(status) && nm.Status.CurrentMapping[localName].Phase == mapsv1alpha1.MappingAccepted {
+				continue
+			}
+			namespacePhase = mapsv1alpha1.MappingCreationLoopBackOff
```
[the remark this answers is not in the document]
giorio94: Sorry, I missed that

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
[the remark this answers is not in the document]
Andreagit97: @palexster what do you think about it ?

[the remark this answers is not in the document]
palexster: Personally, as a user, I would prefer to be aware if the deleted namespaces are still there or not, and be informed of the fact that they hang indefinetly.

pkg/liqo-controller-manager/namespaceMap-controller/manage_namespacemap_controller_finalizer.go
```diff
@@ -0,0 +1,42 @@
...
+		if err := r.Patch(ctx, nm, client.Merge); err != nil {
+			klog.Errorf("%s --> Unable to add finalizer to the NamespaceMap '%s'", err, nm.GetName())
+			return err
+		}
+		klog.Infof("Finalizer correctly added on NamespaceMap '%s'", nm.GetName())
+	}
+	return nil
+}
+
+// RemoveNamespaceMapControllerFinalizer remove NamespaceMap controller finalizer.
```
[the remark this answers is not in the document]
Andreagit97: thank you

pkg/liqo-controller-manager/namespaceMap-controller/create_remote_clients.go
```diff
@@ -15,6 +15,10 @@ import (
 	discoveryV1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
 )
 
+const (
+	clusterIDForeign = "discovery.liqo.io/cluster-id"
```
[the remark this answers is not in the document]
Andreagit97: I have placed it in `liqo/pkg/consts/discovery.go`.

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
```diff
@@ -12,29 +14,28 @@ import (
 	liqoconst "github.com/liqotech/liqo/pkg/consts"
 )
 
-// This function creates a new Namespace onto the remote cluster, the right client to use is chosen
-// by NamespaceMap's cluster-id.
+// This function creates a remote Namespace onto the remote cluster, if it doesn't exist yet.
```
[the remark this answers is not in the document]
Andreagit97: thank you

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
```diff
@@ -44,63 +45,89 @@ func (r *NamespaceMapReconciler) deleteRemoteNamespace(clusterID, remoteName str
...
-func (r *NamespaceMapReconciler) manageRemoteNamespaces(nm *mapsv1alpha1.NamespaceMap) error {
+// and CurrentMapping fields. It updates also NamespaceOffloading status in according to the remote Namespace Status.
+func (r *NamespaceMapReconciler) ensureRemoteNamespacesExistence(nm *mapsv1alpha1.NamespaceMap) bool {
+	requeue := false
 	if nm.Status.CurrentMapping == nil {
 		nm.Status.CurrentMapping = map[string]mapsv1alpha1.RemoteNamespaceStatus{}
 	}
 
-	// if DesiredMapping field has more entries than CurrentMapping, is necessary to create new remote namespaces
+	// For every entry of DesiredMapping crete remote Namespace if it has not already being created
```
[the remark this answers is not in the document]
Andreagit97: thank you

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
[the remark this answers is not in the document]
Andreagit97: thank you for the advice.

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
```diff
@@ -18,69 +18,62 @@ package namespacemapctrl
...
 	}
 
-	if !ctrlutils.ContainsFinalizer(namespaceMap, liqoconst.NamespaceMapControllerFinalizer) {
-		ctrlutils.AddFinalizer(namespaceMap, liqoconst.NamespaceMapControllerFinalizer)
-		if err := r.Patch(context.TODO(), namespaceMap, client.Merge); err != nil {
-			klog.Errorf("%s --> Unable to add finalizer to the NamespaceMap '%s'", err, namespaceMap.GetName())
-			return ctrl.Result{}, err
-		}
-		klog.Infof("Finalizer correctly added on NamespaceMap '%s'", namespaceMap.GetName())
+	// If someone deletes namespaceMap is necessary to remove all remote namespace
```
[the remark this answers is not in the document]
Andreagit97: thank you

pkg/liqo-controller-manager/namespaceMap-controller/manage_remote_namespaces.go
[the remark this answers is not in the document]
Andreagit97: I have changed it with Patch MergeFrom strategy, thank you.

pkg/liqo-controller-manager/namespaceMap-controller/manage_namespacemap_controller_finalizer.go
```diff
@@ -0,0 +1,52 @@
+package namespacemapctrl
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"k8s.io/klog"
+	ctrlutils "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
```
giorio94, line 9: Should be merged in a single block

pkg/liqo-controller-manager/namespaceMap-controller/manage_namespacemap_controller_finalizer.go
```diff
@@ -0,0 +1,52 @@
...
+
+// TryToRemoveNamespaceMapControllerFinalizer remove the NamespaceMapController finalizer if it is possible.
+// If the finalizer is correctly removed returns "true".
+func (r *NamespaceMapReconciler) TryToRemoveNamespaceMapControllerFinalizer(ctx context.Context,
+	nm *mapsv1alpha1.NamespaceMap) (bool, error) {
+	removed := false
+	if len(nm.Status.CurrentMapping) == 0 {
+		patch := nm.DeepCopy()
+		ctrlutils.RemoveFinalizer(nm, namespaceMapControllerFinalizer)
+		// MergeFrom force the resource patch, without conflicts
```
giorio94, line 42:
```suggestion
		// MergeFrom forces the resource patch, without conflicts
```

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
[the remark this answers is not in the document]
giorio94: > 

This comment still applies

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
```diff
@@ -29,61 +30,69 @@ import (
...
+// 2 -- the NamespaceMap is deleted and has the NamespaceMapControllerFinalizer.
 func manageDesiredMappings() predicate.Predicate {
 	return predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
+			// If the NamespaceMap is deleted and has my finalizer.
+			if !e.ObjectNew.GetDeletionTimestamp().IsZero() &&
+				ctrlutils.ContainsFinalizer(e.ObjectNew, namespaceMapControllerFinalizer) {
+				return true
+			}
+			// If an entry is added or removed in DesiredMapping
```
giorio94, line 95: Commenting here, because I cannot below. As the CreateFunc always returns false, the resource will never be reconciled in case of creation, and if desired mappings are present at that time they will be ignored.

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
```diff
@@ -29,61 +30,69 @@ import (
...
 			return len(e.ObjectOld.(*mapsv1alpha1.NamespaceMap).Spec.DesiredMapping) !=
 				len(e.ObjectNew.(*mapsv1alpha1.NamespaceMap).Spec.DesiredMapping)
```
giorio94, line 97: Shouldn't you also perform a reconcile if a mapping is modified (and the overall array length remains the same)?

pkg/liqo-controller-manager/namespaceMap-controller/manage_namespacemap_controller_finalizer.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	return nil
+}
+
+// TryToRemoveNamespaceMapControllerFinalizer remove the NamespaceMapController finalizer if it is possible.
+// If the finalizer is correctly removed returns "true".
+func (r *NamespaceMapReconciler) TryToRemoveNamespaceMapControllerFinalizer(ctx context.Context,
+	nm *mapsv1alpha1.NamespaceMap) (bool, error) {
+	removed := false
+	if len(nm.Status.CurrentMapping) == 0 {
```
giorio94, line 39: I would personally move this check to the caller, as I feel the function name is still difficult to understand. The workflow, in that case, would be clearer.

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
```diff
@@ -0,0 +1,203 @@
...
+// This function tries to create all the remote namespaces requested in DesiredMapping (NamespaceMap->Spec->DesiredMapping).
+func (r *NamespaceMapReconciler) ensureRemoteNamespacesExistence(nm *mapsv1alpha1.NamespaceMap) int {
+	errorCondition := 0
+	for localName, remoteName := range nm.Spec.DesiredMapping {
+		if err := r.createRemoteNamespace(nm.Labels[liqoconst.RemoteClusterID], remoteName); err != nil {
+			nm.Status.CurrentMapping[localName] = mapsv1alpha1.RemoteNamespaceStatus{
+				RemoteNamespace: remoteName,
+				Phase:           mapsv1alpha1.MappingCreationLoopBackOff,
+			}
+			errorCondition = 1
```
giorio94, line 85: If you want to return a counter this should be incremented instead of set to one, otherwise use a boolean variable.

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
```diff
@@ -0,0 +1,203 @@
...
+			RemoteNamespace: remoteName,
+			Phase:           mapsv1alpha1.MappingAccepted,
+		}
+	}
+	return errorCondition
+}
+
+// This function deletes a remote Namespace from the remote cluster, the right client to use is chosen
+// by NamespaceMap's cluster-id. This function return nil (success) only when the remote Namespace is really deleted,
+// so the get Api return a "NotFound".
```
giorio94, line 100:
```suggestion
// so the get Api returns a "NotFound".
```

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
```diff
@@ -0,0 +1,203 @@
...
+	for localName, remoteStatus := range nm.Status.CurrentMapping {
+		if _, ok := nm.Spec.DesiredMapping[localName]; ok && nm.DeletionTimestamp.IsZero() {
+			continue
+		}
+		if err := r.deleteRemoteNamespace(nm.Labels[liqoconst.RemoteClusterID], remoteStatus.RemoteNamespace); err != nil {
+			nm.Status.CurrentMapping[localName] = mapsv1alpha1.RemoteNamespaceStatus{
+				RemoteNamespace: remoteStatus.RemoteNamespace,
+				Phase:           mapsv1alpha1.MappingTerminating,
+			}
+			errorCondition = 1
```
giorio94, line 149: If you want to return a counter this should be incremented instead of set to one, otherwise use a boolean variable.

giorio94 (commented):
Some final comments inline

pkg/liqo-controller-manager/namespaceMap-controller/manage_namespacemap_controller_finalizer.go
Andreagit97: thank you

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
Andreagit97: I have used a counter because in the calling function "ensureRemoteNamespaces"  I have summed this int value in order to understand if in one of the two called functions there was an error or not. So here I can use a boolean you are right, but in the calling function I need this int value, so I can increment the counter here even if I don't care the number of error that I have, just one is enough...

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
Andreagit97: same as before

pkg/liqo-controller-manager/namespaceMap-controller/manage_namespacemap_controller_finalizer.go
Andreagit97: you are right, thank you

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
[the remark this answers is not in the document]
Andreagit97: sorry my fault

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: Yes, it is a quite strange scenario, the controller creates the NamespaceMap empty and we would deny the user to create his NamespaceMap manually in the next future, so I'm not sure in this case...

Andreagit97: What do you think about that @palexster ?

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: yes, you are right maybe an entry is removed and another is added immediately when the event is evaluated the length is the same

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
Andreagit97: I have incremented the counters at the end

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
palexster: I had completely missed that. Why do you filter out creation/deletion events? It should not be reconciled at creation?

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
palexster: > I have used a counter because in the calling function "ensureRemoteNamespaces" I have summed this int value in order to understand if in one of the two called functions there was an error or not. So here I can use a boolean you are right, but in the calling function I need this int value, so I can increment the counter here even if I don't care the number of errors that I have, just one is enough...

I also think that int are a bit strange in this situation. You can obtain the same with the OR operand.

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: I have removed all events.

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
Andreagit97: I have used boolean with OR in the calling function

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
```diff
@@ -0,0 +1,95 @@
...
+type NamespaceMapReconciler struct {
+	client.Client
+	Scheme         *runtime.Scheme
+	Mapper         meta.RESTMapper
+	RemoteClients  map[string]client.Client
+	LocalClusterID string
+}
+
+// cluster-role
+// +kubebuilder:rbac:groups=discovery.liqo.io,resources=foreignclusters,verbs=get;list
```
giorio94, line 42: Why is it required to have permissions on foreigncluster resources?

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
```diff
@@ -0,0 +1,95 @@
...
+// not necessary but it must have role to create/delete remote Namespaces
+// +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;create;delete
```
giorio94, line 44: Not clear to me the meaning of the comment. Additionally, it seems to me the permissions on namespaces are required in the remote clusters, not the local one (hence, they do not need to be added to this role)

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
```diff
@@ -0,0 +1,95 @@
...
+// role
+// +kubebuilder:rbac:groups=virtualKubelet.liqo.io,namespace="do-not-care",resources=namespacemaps,verbs=get;watch;update;patch
```
giorio94, line 46: Shouldn't this be cluster wide?

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
```diff
@@ -0,0 +1,95 @@
...
+// +kubebuilder:rbac:groups=core,namespace="do-not-care",resources=configmaps,verbs=get
```
giorio94, line 47: Why do you need to access configmaps?

giorio94 (commented):
The code looks ok to me. Just some final doubts about the permissions you added in the last iteration. Additionally, if you specify permissions, you should add the generation to the Makefile and correctly import them in the helm chart

Andreagit97:
> The code looks ok to me. Just some final doubts about the permissions you added in the last iteration. Additionally, if you specify permissions, you should add the generation to the Makefile and correctly import them in the helm chart

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: because I have to create the kubeconfig for the remote client, in the foreignCluster I have the reference to the right secret to use

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: yes is not clear the comment means exactly what you have said, that marker is only a reminder, that we have to add permission in the remote clusters. I will remove that.

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: NamespaceMaps should be all created in the technical Namespace, so I have thought of a simple role and not a cluster role, NamespaceMaps are Namespaced resources.

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: in order to get the local clusterID

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
giorio94: Got it. Hence I suppose you should also have the permission to read that secret

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: Yes it is exactly like you have said

pkg/liqo-controller-manager/namespaceMap-controller/namespacemap_controller.go
Andreagit97: yes you are right thank you

giorio94 (approved):
/lgtm

Rebase and squash, and this should be good to go

PR #648, opened 2021-05-23
Fix: ExternalCIDR initial setting could affect PodCIDR/ServiceCIDR setting

davidefalcone1:
# Description

Currently, the network manager first tries to set the ExternalCIDR and then the PodCIDR and ServiceCIDR. This could generate a failure if the the pod subnet IP is set to 10.0.0.0, since the network is first assigned to ExternalCIDR.


palexster:
Is it possible to cover this case with a dedicated test?

davidefalcone1:
> Is it possible to cover this case with a dedicated test?

Sure!

palexster (approved):

alacuku (approved):

PR #654, opened 2021-05-26
Broadcaster pod Informer

giuse2596:
This pr update broadcaster logic to be able to watch pods offloaded by related peered cluster useful to offer right resources in ResourceOffer.

Refers to fourth point of Issue #548


internal/resource-request-operator/broadcaster.go
```diff
@@ -101,44 +131,104 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
 		(newNode.Status.Phase == corev1.NodeTerminated || newNode.Status.Phase == corev1.NodePending) {
 		subResources(currentResources, oldNodeResources)
 	}
-	if err := b.writeResources(currentResources); err != nil {
+	if err := b.writeResources("", currentResources); err != nil {
```
palexster, line 134: Can we map this value with a constant explaining the meaning? e.g., something like "ResourcesForAllClusters"

giuse2596: Do you mean to substitute the clusterID with this constant?

internal/resource-request-operator/suite_test.go
```diff
@@ -103,13 +113,100 @@ var _ = BeforeSuite(func(done Done) {
...
+					Name: "test-container2",
+					Resources: corev1.ResourceRequirements{
+						Limits:   resources,
+						Requests: resources,
+					},
+				},
+			},
+		},
+	}
+	_, err = clientset.CoreV1().Nodes().Create(ctx, node1, metav1.CreateOptions{})
```
aleoli, line 195: The test-related resources should not be global variables and they should be defined inside the test-case

In the beforeEach you can create the cluster, in the afterEach delete it, and in each It case create the specific resources (you can add a function to do that if they are the same resources every time)

See the ResourceOffer operator as an example
https://github.com/liqotech/liqo/blob/master/pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller_test.go

internal/resource-request-operator/broadcaster.go
palexster: In case this field has to use a empty string, I would prefer to have a constant that tells the semantic meaning of "". (IIUC in this case, we have not to specify a specific clusterID and therefore with pass an empty string)

internal/resource-request-operator/broadcaster.go
```diff
@@ -6,46 +6,78 @@ import (
...
+		AddFunc:    b.onNodeAdd,
+		UpdateFunc: b.onNodeUpdate,
+		DeleteFunc: b.onNodeDelete,
+	})
+
+	factoryWithOptions := informers.NewSharedInformerFactoryWithOptions(clientset, resyncPeriod, informers.WithTweakListOptions(setLabelSelectors))
+	b.podInformer = factoryWithOptions.Core().V1().Pods().Informer()
+	b.podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    b.onPodAdd,
+		UpdateFunc: b.onPodUpdate,
```
giorio94, line 59: Since Pod resources are immutable, you do not need to watch for updates

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,44 +133,104 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
+		if err := b.writeResources(resourceForAllClusters, currentResources); err != nil {
 			klog.Errorf("OnAdd error: unable to write allocatable of Node %s: %s", node.Name, err)
 		}
 	}
 }
 
+func (b *Broadcaster) onPodAdd(obj interface{}) {
+	pod := obj.(*corev1.Pod)
+	fmt.Printf("Entrato per pod %s\n", pod.Name)
+	if clusterID := pod.Labels[forge.LiqoOriginClusterID]; clusterID != "" {
```
giorio94, line 157: Would add a warning message in case the ClusterID is not present

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,44 +133,104 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
+
+		if err = b.writeResources(clusterID, currentResources); err != nil {
+			klog.Errorf("OnUpdatePod error: unable to write resources of Pod %s: %s", newPod.Name, err)
+		}
+	}
+}
+
+func (b *Broadcaster) onPodDelete(obj interface{}) {
+	pod := obj.(*corev1.Pod)
+	if clusterID := pod.Labels[forge.LiqoOriginClusterID]; clusterID != "" {
```
giorio94, line 195: Would add a warning message in case the ClusterID is not present

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,44 +133,104 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
+		toAdd, err := extractPodResources(pod)
+		currentResources := b.readPodResources(clusterID)
+		subResources(currentResources, toAdd)
+		if err != nil {
+			klog.Errorf("OnDeletePod error: unable to extract resources from Pod %s: %s", pod.Name, err)
+		}
```
giorio94, line 201: Should be moved above, close to the error generation

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,44 +133,104 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
+	pod := obj.(*corev1.Pod)
+	fmt.Printf("Entrato per pod %s\n", pod.Name)
+	if clusterID := pod.Labels[forge.LiqoOriginClusterID]; clusterID != "" {
+		fmt.Printf("Entrato per pod %s cluster id control\n", pod.Name)
+		toAdd, err := extractPodResources(pod)
+		currentResources := b.readPodResources(clusterID)
+		addResources(currentResources, toAdd)
+		if err != nil {
+			klog.Errorf("OnAddPod error: unable to extract resources from Pod %s: %s", pod.Name, err)
+		}
```
giorio94, line 164: Should be moved close to the error generation

giorio94 (commented):
Some minor comments about the main logic

internal/resource-request-operator/broadcaster.go
```diff
@@ -192,3 +300,23 @@ func updateResources(currentResources, oldResources, newResources corev1.Resourc
 		}
 	}
 }
+
+func setLabelSelectors(options *metav1.ListOptions) {
```
giorio94, line 304: As this function is also used in other parts of the code, I would generalize that and avoid creating a duplicate

internal/resource-request-operator/broadcaster.go
```diff
@@ -147,6 +239,22 @@ func (b *Broadcaster) ReadResources() (corev1.ResourceList, error) {
...
+	defer b.podMutex.RUnlock()
+	if toRead, exists := b.resourcePodMap[clusterID]; exists {
+		return toRead.DeepCopy()
+	}
+	return corev1.ResourceList{}
+}
+
 func (b *Broadcaster) scaleResources(quantity *resource.Quantity) {
 	percentage := int64(b.getConfig().Spec.AdvertisementConfig.OutgoingConfig.ResourceSharingPercentage)
 	if percentage == 0 {
```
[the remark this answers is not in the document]
giuse2596: It was my assumption but I agree it should be dropped. To offer empty resources is better and also in error case is more clear. Thank you.

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,58 +124,100 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
 		subResources(currentResources, *toDelete)
-		if err := b.writeResources(currentResources); err != nil {
-			klog.Errorf("OnAdd error: unable to write allocatable of Node %s: %s", node.Name, err)
+		b.writeNodesResources(currentResources)
+	}
+}
+
+func (b *Broadcaster) onPodAdd(obj interface{}) {
+	pod := obj.(*corev1.Pod)
+	klog.V(2).Infof("OnPodAdd: Add for pod %s\n", pod.Name)
```
giorio94, line 143: Would personally raise to V(4) or V(5), as debug/trace level. The same for the one below

internal/resource-request-operator/broadcaster.go
```diff
@@ -6,46 +6,72 @@ import (
...
+		DeleteFunc: b.onPodDelete,
 	})
 
 	return nil
 }
 
-// StartBroadcaster starts a new sharedInformer to watch nodes resources.
+// StartBroadcaster starts two shared Informers, one for nodes and one for pods launching two separated goroutines.
 func (b *Broadcaster) StartBroadcaster(ctx context.Context, group *sync.WaitGroup) {
+	group.Add(2)
```
giorio94, line 62: I suppose it is more idiomatic to increment this in the actual functions, just before deferring the Done

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,58 +124,100 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
-	if b.allocatable == nil {
-		return nil, fmt.Errorf("error getting cluster resources")
+// write pods resources in thread safe mode.
+func (b *Broadcaster) writePodsResources(clusterID string, newResources corev1.ResourceList) error {
+	b.podMutex.Lock()
+	defer b.podMutex.Unlock()
+	if clusterID != "" {
+		b.resourcePodMap[clusterID] = newResources.DeepCopy()
+		return nil
 	}
```
giorio94, line 187: I would personally avoid making this check, as you are already ensuring this condition cannot happen if you land here. @palexster WDYT?

internal/resource-request-operator/suite_test.go
```diff
@@ -1,115 +0,0 @@
-package resourcerequestoperator
```
giorio94, line 1: Is it desired this file is deleted? It seems to me it is pretty standard when dealing with tests

giuse2596: I have done it to try to fix the problems I am having with tests and now I create the environment with BeforeEach() and destroy it in AfterEach() but at moment seems to be useless even if until now tests always pass here.

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,58 +120,94 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
-// write cluster resources in thread safe mode.
-func (b *Broadcaster) writeResources(newResources corev1.ResourceList) error {
-	b.offerMutex.Lock()
-	defer b.offerMutex.Unlock()
-	if newResources != nil {
-		b.allocatable = newResources.DeepCopy()
-		return nil
+func (b *Broadcaster) onPodAdd(obj interface{}) {
+	pod := obj.(*corev1.Pod)
+	klog.V(4).Infof("OnPodAdd: Add for pod %s\n", pod.Name)
```
giorio94, line 139: Objects are typically shown as namespace/name

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,58 +120,94 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
+	podResources := extractPodResources(pod)
+	currentResources := b.readClusterResources()
+	if clusterID := pod.Labels[forge.LiqoOriginClusterID]; clusterID != "" {
+		klog.V(4).Infof("OnPodAdd: Pod %s passed ClusterID check\n", pod.Name)
```
giorio94, line 143: Same as before, and the log should also include the ClusterID. Otherwise the phrase `passed ClusterID check` does not prove to be of any use in case of debugging.

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,58 +120,94 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
+		currentPodsResources := b.readPodResources(clusterID)
+		addResources(currentPodsResources, podResources)
+		b.writePodsResources(clusterID, currentPodsResources)
 	}
+	subResources(currentResources, podResources)
+	b.writeClusterResources(currentResources)
```
giorio94, line 149: Would move this above the if check. As this logic is not intuitive at first sight, I feel it would be better to add a couple of comments about why you are subtracting the resources from an accumulator and, if the cluster ID matches, adding them to another one.

internal/resource-request-operator/broadcaster.go
```diff
@@ -101,58 +120,94 @@ func (b *Broadcaster) onUpdate(oldObj, newObj interface{}) {
...
+	podResources := extractPodResources(pod)
+	currentResources := b.readClusterResources()
+	if clusterID := pod.Labels[forge.LiqoOriginClusterID]; clusterID != "" {
+		klog.V(4).Infof("OnPodDelete: Pod %s passed ClusterID check\n", pod.Name)
+		currentPodsResources := b.readPodResources(clusterID)
+		subResources(currentPodsResources, podResources)
+		b.writePodsResources(clusterID, currentPodsResources)
+	}
+	addResources(currentResources, podResources)
+	b.writeClusterResources(currentResources)
```
giorio94, line 164: Same here

giorio94 (approved):
/lgtm Just a couple of comments inline

palexster (approved):

PR #655, opened 2021-05-27
Adding support for custom network namespace in the gateway operator

alacuku:
# Description

Supporting the creation and configuration of a new network namespace for the gateway operator. The idea is to move the gateway pod in host network. In order to not put the iptables configuration in the node's namespace a new one is created. In the custom network namespace we put all the liqo configuration such as vpn tunnel and iptables rulespecs.

Fixes #(issue)

# How Has This Been Tested?

Unit tests have been written for all the new functions.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -222,7 +222,9 @@ github.com/containerd/typeurl v0.0.0-20190911142611-5eb25027c9fd/go.mod h1:GeKYz
 github.com/containerd/typeurl v1.0.1/go.mod h1:TB1hUtrpaiO88KEK56ijojHS1+NeF0izUACaJW2mdXg=
 github.com/containerd/zfs v0.0.0-20200918131355-0a33824f23a2/go.mod h1:8IgZOBdv8fAgXddBT4dBXJPtxyRsejFIpXoklgxgEjw=
 github.com/containernetworking/cni v0.7.1/go.mod h1:LGwApLUm2FpoOfxTDEeq8T9ipbpZ61X79hmU3w8FmsY=
+github.com/containernetworking/cni v0.8.0 h1:BT9lpgGoH4jw3lFC7Odz2prU5ruiYKcgAjMCbgybcKI=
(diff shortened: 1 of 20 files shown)


go.sum
```diff
@@ -222,7 +222,9 @@ github.com/containerd/typeurl v0.0.0-20190911142611-5eb25027c9fd/go.mod h1:GeKYz
 github.com/containerd/typeurl v1.0.1/go.mod h1:TB1hUtrpaiO88KEK56ijojHS1+NeF0izUACaJW2mdXg=
```
palexster, line 222: Can you perform a go mod tidy?

pkg/liqonet/netns/netns.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	var createVethPair = func(hostNS ns.NetNS) error {
+		_, _, err := ip.SetupVethWithName(originVethName, dstVethName, linkMTU, dstNetns)
+		if err != nil {
+			klog.Errorf("an error occurred while creating veth pair between host and gateway namespace: %v", err)
+			return err
+		}
+		return nil
+	}
+	retryiable := func(err error) bool {
```
palexster, line 71: Why do you need a retriable here?

palexster (approved):

pkg/liqonet/netns/netns.go
alacuku: The veth pair has one one device on the host network and the other one on the custom namespace. Some times if we delete the custom namespace it wold take some time for the kernel to remove the veth device on the host network, and a conflict could arise while reinstalling the veth device in the host network. I also added a comment.

pkg/liqonet/routing/common.go
```diff
@@ -2,8 +2,12 @@ package routing
 
 import (
 	"errors"
+	"io/ioutil"
 	"net"
 	"reflect"
+	"strings"
+
+	"k8s.io/client-go/util/retry"
```
aleoli, line 10: this import can be joined with the last import block

aleoli (approved):

Issue #474, opened 2021-02-08
Virtual kubelet node controller


mlavacca:
The node management of the Virtual kubelet has to be refactored and re-arranged. To accomplish this goal, three parts have to be changed:
* NodeController (code reformatting and change in logging library)
* NodeProvider (setup of nodeNotifyStatus triggered by advertisement and tunnelEndpoint)
* Provider (deletion of nodeNotifyStatus triggered by advertisement and tunnelEndpoint parts since they are in the wrong place)

The issue can be split up into 6 different sub-issues that can be faced  individually:
- [x] group and move all node provider files under `pkg`, replace `log` with `klog`
- [x] move `NodeErrorHandler`, and `NodeFromProvider` functions from `/cmd` to `/pkg/...`
- [x] replace `NaiveNodeProvider` with `LiqoNodeProvider` and setup all the data structures for allowing the node controller to work with `advertisement` and `tunnelEndpoint`.
- [x]  move `advertisement` reconcilier functionality from `LiqoProvider` to `LiqoNodeProvider`
- [x]  move `tunnelEndpoint` reconcilier functionality from `LiqoProvider` to `LiqoNodeProvider`
- [x] remove entirely the virtual node management from `LiqoProvider`

PR #657, opened 2021-05-28
Update the VirtualNodeUpdater to use new Resources

aleoli:
# Description

This pr moves the logic of node status update in a new `LiqoNodeProvider` and adds the logic to the virtual node updater to handle the new Liqo resources (i.e. ResourceOffers).

Ref. #573 

# How Has This Been Tested?

- [x] add unit test
- [x] locally on KinD


palexster:
Ref. #474

cmd/virtual-kubelet/root/root.go
```diff
@@ -44,6 +44,8 @@ import (
 	"github.com/liqotech/liqo/pkg/virtualKubelet/manager"
 	"github.com/liqotech/liqo/pkg/virtualKubelet/node/module"
 	nodeProvider "github.com/liqotech/liqo/pkg/virtualKubelet/node/provider"
+	nodeprovider "github.com/liqotech/liqo/pkg/virtualKubelet/nodeProvider"
```
giorio94, line 47: I suppose one of the two should be dropped

pkg/virtualKubelet/nodeProvider/resourceWatcher.go
```diff
@@ -0,0 +1,95 @@
...
+	namespace := ""
+	if p.useNewAuth {
+		resource = "resourceoffers"
+		namespace = p.kubeletNamespace
+	} else {
+		resource = "advertisements"
+		namespace = v1.NamespaceAll
+	}
+
+	sharingInformerFactory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(p.dynClient, 0, namespace, func(opt *metav1.ListOptions) {
```
giorio94, line 33: Resync period should not be hard-coded to 0

pkg/virtualKubelet/nodeProvider/resourceWatcher.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+	sharingInformer := sharingInformerFactory.ForResource(sharingv1alpha1.GroupVersion.WithResource(resource)).Informer()
+	sharingInformer.AddEventHandler(getEventHandler(func(event watch.Event) error {
+		if p.useNewAuth {
+			return p.reconcileNodeFromResourceOffer(event)
+		}
+		return p.reconcileNodeFromAdv(event)
+	}))
+
+	tepInformerFactory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(p.dynClient, 0, namespace, func(opt *metav1.ListOptions) {
```
giorio94, line 44: Same here

pkg/virtualKubelet/nodeProvider/reconciler.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	for {
+		if err := p.updateFromResourceOffer(&resourceOffer); err == nil {
+			klog.Info("node correctly updated from resourceOffer")
+			break
+		} else {
+			klog.Errorf("node update from resourceOffer %v failed for reason %v; retry...", resourceOffer.Name, err)
+		}
+	}
```
giorio94, line 66: This risks to be a busy loop in case the error is never solved.
1) I would add an exponential back-off sleep mechanism;
2) I would check whether it is ok if this function never returns trying to solve the error, hence blocking this thread.
The same applies to the same logic below

giorio94 (approved):
As far as I understand, this look OK to me.

pkg/virtualKubelet/nodeProvider/reconciler.go
aleoli: I moved the retry logic with an exponential backoff in the informer handler funcs, can you please see if does it make sense in your opinion?

cmd/virtual-kubelet/root/root.go
aleoli: no, the first one is vk provided package, with utils functions. The second one is the package containing the liqo node provider (I put it in the same folder where the liqo pod provider package was)

giorio94: Ok, then I would rename the latter to liqoNodeProvider, otherwise it is very hard to distinguish the two

pkg/virtualKubelet/nodeProvider/reconciler.go
giorio94: Looks fine to me

palexster (approved):

PR #658, opened 2021-05-30
IPTables module refactoring (1/2)

davidefalcone1:
# Description

This PR modifies the IPTables driver, consumed by the tunnel-operator in order to carry out NAT and Filtering operations.
List of changes:
- A new package has been created for IPTables. 
- All functions have been tested.
- Constants have been moved in constants pkg.
- A new function to create chains per cluster (es. LIQO-PRRT-CLS-cluster1) has been created (before was included in EnsureChainRulespecs)
- A new function to remove iptables configuration after an un-peer procedure has been added.
- A deletion of all rules related to Liqo has been added through a dedicated func.

The old IPTables module is still used and will be removed in the next PR.


pkg/liqonet/iptables/iptables.go
```diff
@@ -0,0 +1,668 @@
...
+	chains := getChainsPerCluster(clusterID)
+	for _, chain := range chains {
+		if err := h.createIptablesChainIfNotExists(getTableFromChain(chain), chain); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func getTableFromChain(chain string) string {
```
palexster, line 331: I would add a comment on this method since it is frequently called but it is not straightforward to get what should return.

pkg/liqonet/iptables/iptables.go
```diff
@@ -0,0 +1,668 @@
...
+
+// EnsureChainRulesPerCluster reads TunnelEndpoint resource and
+// makes sure that chain rules for the given cluster exist.
+func (h IPTHandler) EnsureChainRulesPerCluster(tep *netv1alpha1.TunnelEndpoint) error {
+	chainRules, err := getChainRulesPerCluster(tep)
+	if err != nil {
+		return err
+	}
+	for chain, newRules := range chainRules {
+		// Invocking updateRulesPerChain would be an error
```
palexster, line 261:
```suggestion
		// Invoking updateRulesPerChain would be an error
```

pkg/liqonet/iptables/iptables.go
```diff
@@ -0,0 +1,668 @@
...
+		// since it would replace all the rules in a Liqo chain
+		// with rules related to this remote cluster.
+		// Therefore, first get existing rules related to remote cluster
```
palexster, line 264:
```suggestion
		// Therefore, we need to first get existing rules related to remote cluster
```

pkg/liqonet/iptables/iptables.go
```diff
@@ -0,0 +1,668 @@
...
+	for chain, rule := range liqoRules {
+		if err := h.deleteRulesInChain(chain, []string{rule}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Function that deletes Liqo chains.
+func (h IPTHandler) deleteLiqoChains() error {
```
palexster, line 190: IIUC, The code inside this method is repeated and can be factorized.

alacuku (approved):
/lgtm

build/liqo-test/Dockerfile
```diff
@@ -12,6 +12,10 @@ RUN curl --fail -sSLo envtest-bins.tar.gz "https://storage.googleapis.com/kubebu
 RUN mkdir /usr/local/kubebuilder/
 RUN tar -C /usr/local/kubebuilder/ --strip-components=1 -zvxf envtest-bins.tar.gz
 
+# Install iptables
```
palexster, line 15: Can you do that in the same line?

palexster (approved):

PR #661, opened 2021-06-01
Added new IPAM gRPC

davidefalcone1:
# Description

This PR upgrades the IPAM module with a new gRPC call. It is used to retrieve the home Pod IP starting from the IP of a Pod living in the remote cluster.


cmd/virtual-kubelet/provider/store.go
```diff
@@ -73,6 +73,7 @@ type InitConfig struct {
 	RemoteKubeConfig     string
 	HomeClusterID        string
 	RemoteClusterID      string
+	GRPCServer           string
```
palexster, line 76: I would not point out the technology we are using, but instead, why do we need this server. In this case, something like LiqoIpamEndpoint would be a lot easier to understand.

cmd/virtual-kubelet/register.go
```diff
@@ -16,6 +16,7 @@ func registerKubernetes(s *provider.Store) error {
 			cfg.HomeKubeConfig,
 			cfg.RemoteKubeConfig,
 			cfg.InformerResyncPeriod,
+			cfg.GRPCServer,
```
palexster, line 19: Ditto.

pkg/virtualKubelet/forge/apiForger.go
```diff
@@ -1,14 +1,19 @@
...
+	"fmt"
 	"reflect"
 
 	"github.com/pkg/errors"
+	"google.golang.org/grpc"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog"
```
palexster, line 13: should be klogv2

palexster:
Hi @davidefalcone1, thanks a lot for this PR! I think it is really nice to have finally correctly split the network and the virtual kubelet logic. 

I left some comments inline. Please, pay attention to only use klogv2 as the logging library (we do not have unfortunately a lint check for that)

aleoli (approved):

palexster (approved):

PR #665, opened 2021-06-04
ForeignCluster Operator Refactoring - 1

aleoli:
# Description

This pr starts the refactoring of the ForeignCluster Operator to improve readability and stability.

In particular:
* some functions have been moved away and will be moved to a ForeignCluster webhook in a future pr
* some fields in the ForeignCluster resource are now deprecated, since no more need with the new authentication mechanism
* the reconcile function has been refactored to use the controller runtime client and to be more readable

# How Has This Been Tested?

- [x] locally on KinD
- [x] add unit test
- [ ] other unit test will be added in a future dedicated pr

### DIFF
--- a/apis/discovery/v1alpha1/foreigncluster_methods.go
+++ b/apis/discovery/v1alpha1/foreigncluster_methods.go
@@ -88,5 +88,5 @@ func (fc *ForeignCluster) IsExpired() bool {
 		return true
 	}
 	now := time.Now().Unix()
-	return int64(lu+int(fc.Status.TTL)) < now
+	return int64(lu+fc.Spec.TTL) < now
 }
--- a/apis/discovery/v1alpha1/foreigncluster_types.go
+++ b/apis/discovery/v1alpha1/foreigncluster_types.go
@@ -29,6 +29,21 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// PeeringPhaseType indicates the phase of a peering with a remote cluster.
+type PeeringPhaseType string
+
+const (
+	// PeeringPhaseNone indicates that there is no peering.
+	PeeringPhaseNone PeeringPhaseType = "None"
+	// PeeringPhasePending indicates that the peering is pending,
+	// and we are waiting fore the remote cluster feedback.
+	PeeringPhasePending PeeringPhaseType = "Pending"
+	// PeeringPhaseEstablished indicates that the peering has been established.
+	PeeringPhaseEstablished PeeringPhaseType = "Established"
+	// PeeringPhaseDisconnecting indicates that the peering is being deleted.
+	PeeringPhaseDisconnecting PeeringPhaseType = "Disconnecting"
+)
+
 // ForeignClusterSpec defines the desired state of ForeignCluster.
 type ForeignClusterSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
@@ -36,7 +51,7 @@ type ForeignClusterSpec struct {
 
 	// Foreign Cluster Identity.
 	ClusterIdentity ClusterIdentity `json:"clusterIdentity,omitempty"`
-	// Namespace where Liqo is deployed.
+	// Namespace where Liqo is deployed. (Deprecated)
 	Namespace string `json:"namespace,omitempty"`
 	// Enable join process to foreign cluster.
 	// +kubebuilder:default=false
@@ -51,6 +66,10 @@ type ForeignClusterSpec struct {
 	// +kubebuilder:default="Unknown"
 	// Indicates if this remote cluster is trusted or not.
 	TrustMode discovery.TrustMode `json:"trustMode,omitempty"`
+	// If discoveryType is LAN or WAN and this indicates the number of seconds after that
+	// this ForeignCluster will be removed if no updates have been received.
+	// +kubebuilder:validation:Minimum=0
+	TTL int `json:"ttl,omitempty"`
 }
 
 // ClusterIdentity contains the information about a remote cluster (ID and Name).
@@ -72,8 +91,6 @@ type ForeignClusterStatus struct {
 
 	Outgoing Outgoing `json:"outgoing,omitempty"`
 	Incoming Incoming `json:"incoming,omitempty"`
-	// If discoveryType is LAN and this counter reach 0 value, this FC will be removed.
-	TTL uint32 `json:"ttl,omitempty"`
 	// It stores most important network statuses.
 	Network Network `json:"network,omitempty"`
 	// Authentication status.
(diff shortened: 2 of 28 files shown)


apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -29,14 +29,29 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// JoinPhase indicates the phase of a peering with a remote cluster.
+type JoinPhase string
```
palexster, line 33: I would use PeeringPhase and drop all the "Join"-references (replacing them by Peer). What do you think @giorio94 @frisso?

internal/discovery/foreign-cluster-operator/webhook.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
+	"github.com/liqotech/liqo/pkg/discovery"
+)
+
+// webhook contains those functions the will be moved to a mutating webhook.
+func (r *ForeignClusterReconciler) webhook(ctx context.Context,
```
palexster, line 14: Maybe we can call this function with something like ValidateFC?

internal/discovery/foreign-cluster-operator/auth.go
```diff
@@ -69,18 +56,10 @@ func (r *ForeignClusterReconciler) getRemoteClient(
 	}
 
 	// not existing role
-	isPending := fc.Status.AuthStatus == discovery.AuthStatusPending
+	isPending := fc.Status.AuthStatus == discovery.AuthStatusPending || fc.Status.AuthStatus == ""
```
palexster, line 59: Can we bypass this new condition by defaulting the status resource?

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -107,326 +113,110 @@ type ForeignClusterReconciler struct {
...
-			klog.Error(err)
-			return ctrl.Result{
-				Requeue:      true,
-				RequeueAfter: r.RequeueAfter,
-			}, err
-		} else if update {
-			requireUpdate = true
-		}
+	// create the local TenantControlNamespace if it has not been created
+	if result, err := r.createLocalTenantControlNamespace(ctx, &foreignCluster); err != nil {
```
palexster, line 137: what about ensureLocalTenantControlNamespace?

apis/discovery/v1alpha1/foreigncluster_methods.go
```diff
@@ -88,5 +88,5 @@ func (fc *ForeignCluster) IsExpired() bool {
 		return true
 	}
 	now := time.Now().Unix()
-	return int64(lu+int(fc.Status.TTL)) < now
+	return int64(lu+fc.Spec.TTL) < now
```
giorio94, line 91: Unrelated from this specific line, but unclear to me why this file is in the APIs folder.

apis/discovery/v1alpha1/foreigncluster_types.go
giorio94: Definitely agree

apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -29,14 +29,29 @@ import (
...
+const (
+	// JoinPhaseNone indicates that there is no peering.
+	JoinPhaseNone JoinPhase = "None"
+	// JoinPhasePending indicates that the peering is pending,
+	// and we are waiting fore the remote cluster feedback.
+	JoinPhasePending JoinPhase = "Pending"
+	// JoinPhaseEstablished indicates that the peering has been established.
+	JoinPhaseEstablished JoinPhase = "Established"
+	// JoinPhaseDeleting indicates that the peering is being deleted.
+	JoinPhaseDeleting JoinPhase = "Deleting"
```
giorio94, line 44: I would propose Disconnecting instead of Deleting, as you are not deleting an object, but disconnecting from a peer.

apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -51,6 +66,10 @@ type ForeignClusterSpec struct {
 	// +kubebuilder:default="Unknown"
 	// Indicates if this remote cluster is trusted or not.
 	TrustMode discovery.TrustMode `json:"trustMode,omitempty"`
+	// If discoveryType is LAN or WAN and this indicates the number of seconds after that
+	// this ForeignCluster will be removed if no updated has been received.
+	// +kubebuilder:validation:Minimum=0
+	TTL int `json:"ttl,omitempty"`
```
giorio94, line 72: Not clear to me whether this field is necessary. Is it possible to be manage this at the parent level (the resource which specifies the DNS domain), to avoid continuously triggering the reconciliation of this resource? As this is an user-facing API I would avoid adding fields that may be misleading (I am also not completely sure whether the DIscoveryType field is required, as well as the ClusterIdentity one, the latter looking to me more part of the status).

apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -102,50 +119,56 @@ type ResourceLink struct {
...
 }
 
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:scope=Cluster
+// +kubebuilder:subresource:status
 
 // ForeignCluster is the Schema for the foreignclusters API.
-// +kubebuilder:printcolumn:name="Outgoing joined",type=string,JSONPath=`.status.outgoing.joined`
-// +kubebuilder:printcolumn:name="Incoming joined",type=string,JSONPath=`.status.incoming.joined`
+// +kubebuilder:printcolumn:name="Outgoing joined",type=string,JSONPath=`.status.outgoing.joinPhase`
```
giorio94, line 168:
```suggestion
// +kubebuilder:printcolumn:name="Outgoing peering phase",type=string,JSONPath=`.status.outgoing.joinPhase`
```

Seems more clear to me. Same for the one below.

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -107,326 +113,110 @@ type ForeignClusterReconciler struct {
...
-		klog.Infof("%s deleted, discovery type %s has no active connections", fc.Name, fc.Spec.DiscoveryType)
-		klog.V(4).Infof("ForeignCluster %s successfully reconciled", fc.Name)
+		klog.V(4).Infof("ForeignCluster %s successfully reconciled", foreignCluster.Name)
 		return ctrl.Result{}, nil
 	}
 
-	// TODO: it's probably better to refactor it when we will remove the old authentication, we will need no more a remote client
-	foreignDiscoveryClient, err := r.getRemoteClient(fc, &discoveryv1alpha1.GroupVersion)
-	if err != nil {
+	// create and fetch the identity form the remote cluster
```
giorio94, line 173:
```suggestion
	// create and fetch the identity from the remote cluster
```

giorio94 (commented):
Just a couple of initial comments. Still have to look through the rest of the code

giorio94 (commented):
The comments we were talking about this morning

palexster:
Ref. #670

internal/discovery/foreign-cluster-operator/auth.go
aleoli: I tried to set it as optional, both the field and the whole status struct in this way
```go
type ForeignClusterStatus struct {
	...
	// Authentication status.
	// +kubebuilder:validation:Enum="Pending";"Accepted";"Refused";"EmptyRefused"
	// +kubebuilder:default="Pending"
	// +kubebuilder:validation:Optional
	AuthStatus discovery.AuthStatus `json:"authStatus"`
}
```
```go
type ForeignCluster struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec ForeignClusterSpec `json:"spec,omitempty"`
	// +kubebuilder:validation:Optional
	Status ForeignClusterStatus `json:"status"`
}
```

but it is still not working, since when a new ForeignCluster has been created it is in this form
```yaml
apiVersion: discovery.liqo.io/v1alpha1
kind: ForeignCluster
metadata:
  annotations:
    LastUpdate: "1623332772"
  creationTimestamp: "2021-06-10T13:45:23Z"
  generation: 1
  labels:
    discovery.liqo.io/cluster-id: a50a67a1-9986-47d1-9127-20e1dd5f5838
    discovery.liqo.io/discovery-type: LAN
  name: a50a67a1-9986-47d1-9127-20e1dd5f5838
  resourceVersion: "10133"
  selfLink: /apis/discovery.liqo.io/v1alpha1/foreignclusters/a50a67a1-9986-47d1-9127-20e1dd5f5838
  uid: 99c7ba81-1d3b-4b93-8011-6e9134efb941
spec:
  authUrl: https://172.18.0.3:31383
  clusterIdentity:
    clusterID: a50a67a1-9986-47d1-9127-20e1dd5f5838
  discoveryType: LAN
  join: true
  namespace: liqo
  trustMode: Untrusted
  ttl: 90
```

where the status is not present at all. With this resource, when the client retrieves the status from the API Server, it sees as nil and fills the authStatus as an empty string...

I propose to not change it at the moment since the field will completely be deleted in the refactoring proposed in #670 
@palexster @giorio94

apis/discovery/v1alpha1/foreigncluster_types.go
aleoli: Ref #670

internal/discovery/foreign-cluster-operator/auth.go
giorio94: That's fine from my point of view. I'm honestly not sure it is possible to default status fields

palexster (approved):

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -557,40 +354,74 @@ func (r *ForeignClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		klog.Error(err)
+		return err
+	}
+
+	foreignCluster.Status.Outgoing.PeeringPhase = getJoinPhase(&outgoingResourceRequestList)
+	foreignCluster.Status.Incoming.PeeringPhase = getJoinPhase(&incomingResourceRequestList)
+	return nil
+}
+
+func getJoinPhase(resourceRequestList *discoveryv1alpha1.ResourceRequestList) discoveryv1alpha1.PeeringPhaseType {
```
giorio94, line 397:
```suggestion
func getPeeringPhase(resourceRequestList *discoveryv1alpha1.ResourceRequestList) discoveryv1alpha1.PeeringPhaseType {
```

I suppose

internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -557,40 +354,74 @@ func (r *ForeignClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}); err != nil {
+		klog.Error(err)
+		return err
+	}
+
+	if len(outgoingResourceRequestList.Items) > 1 {
+		err := fmt.Errorf("[%v] more than one outgoing resource request found in namespace %v", remoteClusterID, localNamespace)
+		klog.Error(err)
+		return err
+	}
```
giorio94, line 385: Probably you could move this check to the `getPeeringPhase` function, using a switch statement

giorio94 (approved):

PR #677, opened 2021-06-13
IPTables refactoring (2/2)

davidefalcone1:
# Description

This PR concludes the IPTables module refactoring, by introducing exported functions to ensure prerouting and postrouting
rules. Moreover, the old IPTables module is substituted by the new one.


internal/liqonet/tunnel-operator/tunnel-operator.go
```diff
@@ -39,14 +39,15 @@ import (
 	netv1alpha1 "github.com/liqotech/liqo/apis/net/v1alpha1"
 	liqoconst "github.com/liqotech/liqo/pkg/consts"
 	"github.com/liqotech/liqo/pkg/liqonet"
+	"github.com/liqotech/liqo/pkg/liqonet/iptables"
 	liqonetns "github.com/liqotech/liqo/pkg/liqonet/netns"
 	"github.com/liqotech/liqo/pkg/liqonet/overlay"
 	liqorouting "github.com/liqotech/liqo/pkg/liqonet/routing"
 	"github.com/liqotech/liqo/pkg/liqonet/tunnel"
+	"github.com/liqotech/liqo/pkg/liqonet/utils"
```
palexster, line 48: You should to sort the imports

pkg/liqonet/iptables/iptables.go
```diff
@@ -429,6 +429,68 @@ func (h IPTHandler) removeChainsPerCluster(clusterID string) error {
 	return nil
 }
 
+// EnsurePostroutingRules makes sure that the postrouting rules for a given cluster are in place and updated.
+func (h IPTHandler) EnsurePostroutingRules(isGateway bool, tep *netv1alpha1.TunnelEndpoint) error {
```
alacuku, line 433: `isGateway` flag is not useful anymore. It was used when the iptables rules were configured also on the nodes by the `liqo-route` operator. We do not touch anymore the iptables rules of the nodes, so we can drop the flag.

pkg/liqonet/iptables/iptables.go
```diff
@@ -570,6 +641,68 @@ func (h IPTHandler) insertRulesIfNotPresent(table, chain string, rules []string)
 	return nil
 }
 
+func getPostroutingRules(isGateway bool, tep *netv1alpha1.TunnelEndpoint) ([]string, error) {
+	clusterID := tep.Spec.ClusterID
```
palexster, line 645: Can this parameter check be factorized with getPreRoutingRules with a common method?

pkg/liqonet/iptables/iptables.go
```diff
@@ -570,6 +641,68 @@ func (h IPTHandler) insertRulesIfNotPresent(table, chain string, rules []string)
...
+	if clusterID == "" {
+		return nil, &errors.WrongParameter{
+			Parameter: consts.ClusterIDLabelName,
+			Reason:    errors.StringNotEmpty,
+		}
+	}
+	localPodCIDR := tep.Status.LocalPodCIDR
+	localRemappedPodCIDR, remotePodCIDR := utils.GetPodCIDRS(tep)
+	if err := utils.IsValidCIDR(localPodCIDR); err != nil {
```
alacuku, line 654: You have introduced the validations of the parameters before trying to insert the rules, that's a nice! What do you think if we introduce a function to check the correctness of the `tep`? Such way we do it once.

pkg/liqonet/utils/utils.go
```diff
@@ -245,3 +245,14 @@ func GetFirstIP(network string) (string, error) {
 	}
 	return firstIP.String(), nil
 }
+
+// MustGetFirstIP returns the first IP address of a network.
+// It panics in case of errors.
+func MustGetFirstIP(network string) string {
```
alacuku, line 251: Is this function called only for testing purposes? Since it panics in case of error I would move it on the testing package where it is called.

alacuku (approved):
/lgtm

palexster (approved):

PR #681, opened 2021-06-15
Updated liqonet/utils interface

davidefalcone1:
Deleted functions RemoveString and ContainsString. They used to be called to check/remove finalizers. Therefore, they have been deleted and substituted with correspondent functions in k8s.controllerutil.


palexster (approved):

alacuku (approved):
/lgtm

PR #685, opened 2021-06-16
Fixed pre-delete logic of NetworkConfig resource

davidefalcone1:
In TunnelEndpointOperator, functions invocations of pre-delete logic of NetworkConfig have been re-organized so that functions are called before the finalizer is removed.


alacuku (approved):
/lgtm

palexster (approved):

PR #686, opened 2021-06-17
Symmetric Routing for vxlan overlay

alacuku:
# Description
The operator for a given pod, based on the node where the pod is running, configures a route to send the traffic through the overlay network. The route is used only for the traffic generated on remote clusters with destination address a pod running on the local cluster. Those route are configured only on the node where the `gateway` pod is running. The traffic for local pods should be handled by the local CNI but if the reverse path filtering is in strict mode (default) the traffic coming from a network interface where the source address is not routable by the same interface is dropped. To solve the issue it requires to set the reverse path filtering to loose mode, but some times it is not possible. To overcome the problem all the traffic sent to or coming from a peering cluster has to be routed using the overlay network.

# How Has This Been Tested?

Unit test for each function.

### DIFF
--- a/internal/liqonet/route-operator/overlayOperator_test.go
+++ b/internal/liqonet/route-operator/overlayOperator_test.go
@@ -12,11 +12,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
+
+	"k8s.io/apimachinery/pkg/types"
 
 	liqoerrors "github.com/liqotech/liqo/pkg/liqonet/errors"
 	"github.com/liqotech/liqo/pkg/liqonet/overlay"
@@ -39,13 +39,18 @@ var (
 	/*** EnvTest Section ***/
 	overlayScheme  = runtime.NewScheme()
 	overlayEnvTest *envtest.Environment
-	overlayClient  client.Client
 )
 
 var _ = Describe("OverlayOperator", func() {
 	// Before each test we create an empty pod.
 	// The right fields will be filled according to each test case.
 	JustBeforeEach(func() {
+		overlayReq = ctrl.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: overlayNamespace,
+				Name:      overlayPodName,
+			},
+		}
 		// Create the test pod with the labels already set.
 		overlayTestPod = &corev1.Pod{
 			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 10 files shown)


internal/liqonet/route-operator/overlayOperator_test.go
```diff
@@ -6,16 +6,16 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"k8s.io/apimachinery/pkg/types"
+
```
palexster, line 10: Should be merged with the next group

internal/liqonet/route-operator/symmetricRoutingOperator.go
```diff
@@ -0,0 +1,165 @@
...
+
+// Reconcile for a given pod, based on the node where the pod is running, configures a route to send the traffic
+// through the overlay network. The route is used only for the traffic generated on remote clusters with
+// destination address a pod running on the local cluster. The traffic for local pods should be handled by
+// the local CNI but if the reverse path filtering is in strict mode (default) the traffic coming from a
+// network interface where the source address is not routable by the same interface is dropped. To solve the
+// issue it requires to set the reverse path filtering to loose mode, but some times it is not possible. To
+// overcome the problem all the traffic sent to or coming from a peering cluster has to be routed using the
+// overlay network.
+func (src *SymmetricRoutingController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
palexster, line 46: Question: Is resync configured? How often?

internal/liqonet/route-operator/symmetricRoutingOperator.go
```diff
@@ -0,0 +1,165 @@
...
+		nodesLock:      nodesLock,
+		vxlanNodes:     vxlanNodes,
+		routes:         map[string]string{},
+	}, nil
+}
+
+// addPeer adds route for the given pod. It return true when the route dose not exists
+// and is added, false if the route does already exist. An error is return if something
+// goes wrong.
+func (src *SymmetricRoutingController) addRoute(req ctrl.Request, p *corev1.Pod) (bool, error) {
```
palexster, line 105: Not sure if the lock is necessary. AFAIK Reconcile is by default executed without any parallelization.

internal/liqonet/route-operator/symmetricRoutingOperator.go
```diff
@@ -0,0 +1,165 @@
...
+		return deleted, err
+	}
+	delete(src.routes, req.String())
+	return deleted, nil
+}
+
+// podFilter used to filter out all the pods that are running on the same node
+// as the operator or pods that are not ready. We do not need to configure routes
+// for pods running on same node.
+func (src *SymmetricRoutingController) podFilter(obj client.Object) bool {
```
palexster, line 141: Does the operator correspond to the liqo-gateway?

internal/liqonet/route-operator/symmetricRoutingOperator.go
alacuku: Resync is not configured. The operator uses a custom routing table. The idea is to avoid the possibility that external processes mess with the configuration. The operator handles all the existing pods on the cluster, I would prefer to not have the rysync option.

palexster (approved):
/lgtm
Some minor comments and questions added.

internal/liqonet/route-operator/symmetricRoutingOperator.go
alacuku: The lock is need for the `src.vxlanNodes` map. The map is populated (write access) by another go routine. The operator here accesses it in `read` mode.

internal/liqonet/route-operator/symmetricRoutingOperator.go
alacuku: The current operator runs inside the `liqo-route` pod.

internal/liqonet/route-operator/symmetricRoutingOperator.go
palexster: Yes, I was thinking of the scenario when a route is added for the creation of one pod. Then, if when the pod is deleted, the operator is not available: how can we remove the "orphaned" route? Does the operator flush the entire table at each restart?

internal/liqonet/route-operator/symmetricRoutingOperator.go
```diff
@@ -0,0 +1,165 @@
+package routeoperator
+
+import (
```
palexster, line 3: gomimports should be fixed here also

internal/liqonet/route-operator/symmetricRoutingOperator.go
alacuku: Yes, that is right, the custom routing table is flushed at exit time by the operator.

aleoli (approved):

PR #687, opened 2021-06-17
Vxlan operator nodes map

alacuku:
# Description

The vxlan operator populates a map for each node of the k8s cluster saving its IP address. The map is consumed by the `symmetric-routing operator` introduced in #686.

# How Has This Been Tested?
 Unit tests have been modified to take in account the modified code.

### DIFF
--- a/internal/liqonet/route-operator/overlayOperator.go
+++ b/internal/liqonet/route-operator/overlayOperator.go
@@ -3,6 +3,7 @@ package routeoperator
 import (
 	"context"
 	"net"
+	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	k8sApiErrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


palexster (approved):

internal/liqonet/route-operator/overlayOperator.go
```diff
@@ -104,8 +110,8 @@ func (ovc *OverlayController) Reconcile(ctx context.Context, req ctrl.Request) (
 }
 
 // NewOverlayController returns a new controller ready to be setup and started with the controller manager.
-func NewOverlayController(podName, podIP string, podSelector *metav1.LabelSelector,
-	vxlanDevice overlay.VxlanDevice, cl client.Client) (*OverlayController, error) {
+func NewOverlayController(podName, podIP string, podSelector *metav1.LabelSelector, vxlanDevice overlay.VxlanDevice,
+	nodesLock *sync.RWMutex, vxlanNodes map[string]string, cl client.Client) (*OverlayController, error) {
```
aleoli, line 114: Can this mutex be created inside this function instead of in the outside and then passed by reference? I think that there is no reason to share the mutex with other components, and to pass it by reference sounds strange to me.
Or am I missing something?

aleoli (approved):

internal/liqonet/route-operator/overlayOperator.go
alacuku: The mutex is used for the `vxlanNodes` map. It is shared by two different go routines. To each go routine we passe the mutex and the map. A better solution would be to create a new `data structure` that holds both the map and the mutex. For the moment i would prefer to keep it simple.

aleoli: ok, thanks

PR #690, opened 2021-06-17
Some improvements to the resourceRequest controller.

giuse2596:
# Description

- [x] Add Owns() and filter options to For() in SetuoWithManager function.
- [x] Create a new ForeignCluster if it does not exists.
- [x] Fix Node Informers Running check substituting Node.Status.Phase (deprecated and never used) with Node.Status.Conditions lookup.
- [x] Fix klog import in Broadcaster.

### DIFF
--- a/internal/resource-request-operator/broadcaster.go
+++ b/internal/resource-request-operator/broadcaster.go
@@ -10,7 +10,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
 	resourcehelper "k8s.io/kubectl/pkg/util/resource"
 
 	configv1alpha1 "github.com/liqotech/liqo/apis/config/v1alpha1"
@@ -92,7 +92,7 @@ func (b *Broadcaster) getConfig() *configv1alpha1.ClusterConfig {
 // react to a Node Creation/First informer run.
 func (b *Broadcaster) onNodeAdd(obj interface{}) {
 	node := obj.(*corev1.Node)
-	if node.Status.Phase == corev1.NodeRunning {
+	if utils.IsNodeReady(node) {
 		klog.V(4).Infof("Adding Node %s\n", node.Name)
 		toAdd := &node.Status.Allocatable
 		currentResources := b.readClusterResources()
@@ -109,16 +109,16 @@ func (b *Broadcaster) onNodeUpdate(oldObj, newObj interface{}) {
 	newNodeResources := newNode.Status.Allocatable
 	currentResources := b.readClusterResources()
 	klog.V(4).Infof("Updating Node %s in %v\n", oldNode.Name, newNode)
-	if newNode.Status.Phase == corev1.NodeRunning {
-		// node was already Running, update with possible new resources.
-		if oldNode.Status.Phase == corev1.NodeRunning {
+	if utils.IsNodeReady(newNode) {
+		// node was already Ready, update with possible new resources.
+		if utils.IsNodeReady(oldNode) {
 			updateResources(currentResources, oldNodeResources, newNodeResources)
 			// node is starting, add all its resources.
 		} else {
 			addResources(currentResources, newNodeResources)
 		}
 		// node is terminating or stopping, delete all its resources.
-	} else if oldNode.Status.Phase == corev1.NodeRunning && newNode.Status.Phase != corev1.NodeRunning {
+	} else if utils.IsNodeReady(oldNode) && !utils.IsNodeReady(newNode) {
 		subResources(currentResources, oldNodeResources)
 	}
 	b.writeClusterResources(currentResources)
@@ -129,7 +129,7 @@ func (b *Broadcaster) onNodeDelete(obj interface{}) {
 	node := obj.(*corev1.Node)
 	toDelete := &node.Status.Allocatable
 	currentResources := b.readClusterResources()
-	if node.Status.Phase == corev1.NodeRunning {
+	if utils.IsNodeReady(node) {
 		klog.V(4).Infof("Deleting Node %s\n", node.Name)
 		subResources(currentResources, *toDelete)
 		b.writeClusterResources(currentResources)
--- a/internal/resource-request-operator/resourceRequest_controller.go
+++ b/internal/resource-request-operator/resourceRequest_controller.go
@@ -7,10 +7,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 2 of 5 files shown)


internal/resource-request-operator/resourceRequest_controller.go
```diff
@@ -41,6 +44,25 @@ func (r *ResourceRequestReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	err = r.List(ctx, foreignClusterList, client.MatchingLabels{
+		discovery.ClusterIDLabel: resourceRequest.Spec.ClusterIdentity.ClusterID,
+	})
+
+	if err != nil {
+		klog.Errorf("%s -> unable to List foreignCluster with clusterID %s: %s", r.ClusterID, resourceRequest.Spec.ClusterIdentity.ClusterID, err)
+		return ctrl.Result{}, nil
+	}
+
+	if len(foreignClusterList.Items) == 0 {
```
palexster, line 57: Does deserve a bit more commentary? Is this related to the Incoming peer?

internal/resource-request-operator/utils.go
```diff
@@ -51,3 +52,29 @@ func (r *ResourceRequestReconciler) generateResourceOffer(request *discoveryv1al
 	klog.Infof("%s -> %s Offer: %s", r.ClusterID, op, offer.ObjectMeta.Name)
 	return nil
 }
+
+func (r *ResourceRequestReconciler) createForeignCluster(resourceRequest *discoveryv1alpha1.ResourceRequest) error {
```
palexster, line 56: Ditto.

internal/resource-request-operator/broadcaster.go
```diff
@@ -268,3 +268,12 @@ func extractPodResources(pod *corev1.Pod) corev1.ResourceList {
 	resourcesToExtract, _ := resourcehelper.PodRequestsAndLimits(pod)
 	return resourcesToExtract
 }
+
+func isNodeRunning(node *corev1.Node) bool {
```
giorio94, line 272:
```suggestion
func isNodeReady(node *corev1.Node) bool {
```

Since this function may be of use also in other situations (e.g. VK node controller), I would propose to move this outside of this specific package

internal/resource-request-operator/utils.go
```diff
@@ -51,3 +52,29 @@ func (r *ResourceRequestReconciler) generateResourceOffer(request *discoveryv1al
...
+		Spec: discoveryv1alpha1.ForeignClusterSpec{
+			ClusterIdentity: resourceRequest.Spec.ClusterIdentity,
+			Namespace:       resourceRequest.Namespace,
+			Join:            false,
+			DiscoveryType:   discovery.IncomingPeeringDiscovery,
+			AuthURL:         resourceRequest.Spec.AuthURL,
+		},
+	}
+
+	err := r.Create(context.Background(), foreignCluster)
```
giorio94, line 74: It seems to me at this point the resourcerequest should become the owner of this foreigncluster to ensure eventual deletion. @aleoli Is that correct?

internal/resource-request-operator/utils.go
```diff
@@ -51,3 +52,29 @@ func (r *ResourceRequestReconciler) generateResourceOffer(request *discoveryv1al
...
+			},
+		},
+		Spec: discoveryv1alpha1.ForeignClusterSpec{
+			ClusterIdentity: resourceRequest.Spec.ClusterIdentity,
+			Namespace:       resourceRequest.Namespace,
+			Join:            false,
+			DiscoveryType:   discovery.IncomingPeeringDiscovery,
+			AuthURL:         resourceRequest.Spec.AuthURL,
+		},
+	}
```
giorio94, line 72: @aleoli Is it sufficient to create this foreigncluster or its definition should also be enforced in case of modifications?

giorio94 (commented):
A couple of comments inline.
I definitely agree with @palexster that more comments are needed to clarify the workflow

internal/resource-request-operator/utils.go
aleoli: No, it can't, it is a namespaced resource while the ForeignCluster is cluster-wide. The deletion logic is inside the ForeignCluster Operator.
It has to be set the opposite, the ResourceRequest ahs to be owned by the ForeignClsuter to ensure FC Status updates on incoming peering chages.

internal/resource-request-operator/utils.go
aleoli: It should be enforced, but only if the discovery type is still IncomingPeering, since it can be discovered in other ways (i.e. mDNS, DNS) that can overwrite this method and change its content.
And then we have to take care and not enforce the join flag, to preserve user chooses.

giorio94: Got it. As agreed, for the moment it is probably better not to enforce it to avoid problems, since all fields should be anyway immutable (at least conceptually)

internal/resource-request-operator/utils.go
giorio94: Got it. Indeed, as discussed, it should be configured the reverse owner-reference (the resource request is owned by the foreigncluster), to ensure the status is correctly reconciled. Once those two operations are performed, it is probably better to return and create the resourceoffer in the next reconcile, to prevent self-created conflicts

internal/resource-request-operator/resourceRequest_controller.go
```diff
@@ -41,7 +44,29 @@ func (r *ResourceRequestReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		// if does not exist any foreignCluster which satisfies above conditions create a new one.
```
giorio94, line 60: IMHO, this comment has limited value, since it mostly repeats what is obvious from the code. In comments you should mostly elaborate on high level logic, why it is needed and why certain conditions can occur (e.g. why the foreigncluster may/may not exist), without repeating low level details.

internal/resource-request-operator/resourceRequest_controller.go
```diff
@@ -41,7 +44,29 @@ func (r *ResourceRequestReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, nil
 	}
 
-	offerErr := r.generateResourceOffer(&resourceRequest)
+	// check if a foreignCluster with an incoming peering
```
giorio94, line 47: The comment mentions the incoming peering, but it seems to me this is not checked anywhere (correctly AFAIU). In that case the comment should be rephrased.

pkg/utils/nodeUtils.go
```diff
@@ -0,0 +1,13 @@
+package utils
+
+import corev1 "k8s.io/api/core/v1"
+
+// IsNodeReady return true if the passed node has the NodeReady condition = True, false otherwise.
```
giorio94, line 5:
```suggestion
// IsNodeReady returns true if the passed node has the NodeReady condition equal to True, false otherwise.
```

palexster (approved):

giorio94 (approved):

PR #692, opened 2021-06-18
NAT rules for ExternalCIDR (4/4)

davidefalcone1:
# Description
This PR adds a new controller that reconciles the resource NatMapping. It ensures the appropriate set of NAT rules for supporting ExternalCIDR traffic is in place and updated.

### DIFF
--- a/internal/liqonet/tunnel-operator/natmapping_operator_test.go
+++ b/internal/liqonet/tunnel-operator/natmapping_operator_test.go
@@ -1,0 +1,251 @@
+package tunneloperator
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/kubernetes/pkg/util/slice"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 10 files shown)


internal/liqonet/tunnel-operator/natmapping_operator_test.go
```diff
@@ -0,0 +1,269 @@
...
+	})
+	AfterEach(func() {
+		err := k8sClient.DeleteAllOf(context.Background(), &v1alpha1.NatMapping{}, &client.DeleteAllOfOptions{
+			ListOptions: client.ListOptions{
+				Namespace: namespace,
+			},
+		})
+		Expect(err).To(BeNil())
+	})
+	Context("If a natmapping resource is deleted", func() {
```
alacuku, line 65: The operator does not add finalizers to the resources, hence when a resource is deleted the operator just can not fetch it and will receive an `notFoundError`. This test case is the same as the one in the context `Reconcile on a resource that does not exist`

internal/liqonet/tunnel-operator/tunnel_operator_suite_test.go
```diff
@@ -1,17 +1,176 @@
...
+	iptNetns, err = netns.CreateNetns(iptNetnsName)
+	Expect(err).To(BeNil())
+
+	// Config IPTables for remote clusters
+	err = initIPTables()
+	Expect(err).To(BeNil())
+
+	err = netv1alpha1.AddToScheme(scheme.Scheme)
+	Expect(err).To(BeNil())
+	//+kubebuilder:scaffold:scheme
```
alacuku, line 98: we should remove it.

palexster (approved):

alacuku (approved):

PR #708, opened 2021-06-24
Enabled traffic toward ExternalCIDR

davidefalcone1:
# Description
This PR enables peered clusters to exchange traffic toward ExternalCIDR, as they already do with PodCIDR. To solve this problem, routing drives supported by Liqo have been modified in such a way that now routing tables hold not only PodCIDR destinations but also ExternalCIDR destinations. SNAT iptables rule for traffic toward ExternalCIDR was added as well.

### DIFF
--- a/internal/liqonet/tunnel-operator/tunnel-operator.go
+++ b/internal/liqonet/tunnel-operator/tunnel-operator.go
@@ -50,6 +50,10 @@ var (
 	result = ctrl.Result{}
 )
 
+// Constant used for add/deletion of policy routing rules
+// to specify any network.
+const anyNetwork = ""
+
 // TunnelController type of the tunnel controller.
 type TunnelController struct {
 	client.Client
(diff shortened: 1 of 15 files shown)


internal/liqonet/tunnel-operator/tunnel-operator.go
```diff
@@ -208,6 +216,14 @@ func (tc *TunnelController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		if deleted {
 			klog.Infof("%s -> policy routing rule for subnet {%s} correctly removed", clusterID, remotePodCIDR)
 		}
+		deleted, err = liqorouting.DelPolicyRoutingRule(remoteExternalCIDR, "", liqoconst.RoutingTableID)
```
palexster, line 219: Can be interesting to have a constant to map "" with a semantic meaning?

palexster (approved):

Issue #598, opened 2021-05-07
Pods hang in terminating status


giorio94:
**Describe the bug**

When a pod is deleted in the local cluster (manually, or through a scale operations), it triggers the transition into terminating status and the removal of the remote replicaset. Although the remote replicaset is correctly deleted, and the finalizer of the local pod is removed, it nonetheless continues hanging in the terminating status, until the termination grace period expires.

**To Reproduce**
1. Establish a liqo peering with a remote cluster
2. Create a pod, forcing the scheduling in the remote cluster (optionally, increase the termination grace period to better observe the behavior) 
3. Delete the pod
4. Observe it remains in terminating status until the termination grace period expires.

**Expected behavior**
The pod should be correctly deleted when the remote replicaset removal succeeded

filippoprojetto:
I am investigating this issue. 

I have a question about the desired behaviour when handling pods deletion. From the implementation it looks like that receiving a `nil` pod object should make the kubelet raise an error but the tests assert the opposite. The test does not fail because the `CreatePod` function is called instead of `DeletePod` - fixing that the test fails as expected. 
Impl: https://github.com/liqotech/liqo/blob/5b7d99ad29ef7de2eda628e99c5b8b487e291780/pkg/virtualKubelet/provider/pods.go#L102 
Test: https://github.com/liqotech/liqo/blob/5b7d99ad29ef7de2eda628e99c5b8b487e291780/pkg/virtualKubelet/provider/pods_test.go#L135 

I see that both the Create and Update functions do not return an error when a nil pod is received. Is there a reason for the DeletePod function to deviate from this behaviour?

filippoprojetto:
It looks like these 3 functions where all returning an error when getting a `nil` pod but it was changed with [this](https://github.com/liqotech/liqo/pull/500) PR. Was it intentional to let the `DeletePod` function behind?

giorio94:
Trying to answer, although I wasn't involved in the virtual kubelet at that time and so I may be wrong.

[...]

PR #710, opened 2021-06-24
test: fix deletepod testcase

filippoprojetto:
Trigger the function that was intended to be tested, ie. DeletePod.
We are not sure what is the behaviour we want when a nil pod reaches
this function, if that's possible at all. Further analysis and work is
needed.

Also, note that Create and Update functions show a different behaviour
introduced as quick fix to race condition:
https://github.com/liqotech/liqo/issues/598#issuecomment-867572624. 
To re-evaluate.

### DIFF
--- a/pkg/virtualKubelet/provider/pods_test.go
+++ b/pkg/virtualKubelet/provider/pods_test.go
@@ -133,8 +133,8 @@ var _ = Describe("Pods", func() {
 		})
 
 		It("delete pod", func() {
-			err := provider.CreatePod(context.TODO(), nil)
-			Expect(err).NotTo(HaveOccurred())
+			err := provider.DeletePod(context.TODO(), nil)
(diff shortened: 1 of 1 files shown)


giorio94 (approved):

palexster (approved):

pkg/virtualKubelet/provider/pods_test.go
```diff
@@ -133,8 +133,8 @@ var _ = Describe("Pods", func() {
 		})
 
 		It("delete pod", func() {
-			err := provider.CreatePod(context.TODO(), nil)
-			Expect(err).NotTo(HaveOccurred())
+			err := provider.DeletePod(context.TODO(), nil)
+			Expect(err).To(HaveOccurred())
```
palexster, line 137: Why should we need to change this Expect?

PR #711, opened 2021-06-24
Fix race condition causing netns tests to fail

alacuku:
# Description

Some times `netns` tests fails due to a race conditions causing the clean up function to fail.

Fixes #(issue)
Fix race condition causing unit tests of the `netns` package to fail.


filippoprojetto (approved):
🚀

palexster (approved):

PR #718, opened 2021-06-30
Virtual Node Cleanup

aleoli:
# Description

This pr adds the logic to clean up the node when the peering is tearing down.

This is composed of 4 steps:
- node cordoning
- eviction of pods using the Eviction API
- waiting for pod deletion
- delete the node

# How Has This Been Tested?

- [x] locally on KinD
- [x] add unit test


pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller_test.go
```diff
@@ -228,25 +229,14 @@ var _ = Describe("ResourceOffer Controller", func() {
 				return reflect.DeepEqual(deploymentList.Items[0], *vkDeploy)
 			}, timeout, interval).Should(BeTrue())
 
-			// check tht the deployment has the controller reference
-			Eventually(func() []metav1.OwnerReference {
+			// check tht the deployment has the controller reference annotation
```
giorio94, line 232:
```suggestion
			// check that the deployment has the controller reference annotation
```

pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller_methods.go
```diff
@@ -117,7 +118,7 @@ func (r *ResourceOfferReconciler) createVirtualKubeletDeployment(
 	// create the base resources
 	vkServiceAccount := forge.VirtualKubeletServiceAccount(name, namespace)
 	op, err := controllerutil.CreateOrUpdate(ctx, r.Client, vkServiceAccount, func() error {
-		return controllerutil.SetOwnerReference(resourceOffer, vkServiceAccount, r.Scheme)
+		return nil
```
giorio94, line 121: Will this be deleted when removing the namespace?

pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller.go
```diff
@@ -139,14 +149,74 @@ func (r *ResourceOfferReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			if req, err := getRequestFromObject(ge.Object); err == nil {
+				rli.Add(req)
+			}
+		},
+	}
+}
+
+// getRequestFromObject reads an annotation in the object and returns the request
+// to be enqueued for the reconciliation.
+func getRequestFromObject(obj client.Object) (reconcile.Request, error) {
```
giorio94, line 209:
```suggestion
func getResourceRequestFromObject(obj client.Object) (reconcile.Request, error) {
```

I would personally be more explicit, otherwise the name might be misleading.

pkg/virtualKubelet/liqoNodeProvider/cordon.go
```diff
@@ -0,0 +1,49 @@
...
+	"encoding/json"
+
+	"gomodules.xyz/jsonpatch/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+)
+
+// cordonNode cordons the controlled node setting it in the unschedulable state.
+func (p *LiqoNodeProvider) cordonNode(ctx context.Context) error {
```
giorio94, line 14: You are already using a very similar logic to update the labels of the node. I would suggest creating a generic function which given the original and patched nodes performs the actual patch, to avoid duplicating the marshal, createpatch, patch logic.

pkg/virtualKubelet/liqoNodeProvider/drain.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+
+	klog.Infof("node %v successfully drained", p.node.GetName())
+	return nil
+}
+
+// getPodsForDeletion lists the pods that are running on our virtual node.
+func (p *LiqoNodeProvider) getPodsForDeletion(ctx context.Context) (*v1.PodList, error) {
+	podList, err := p.client.CoreV1().Pods(metav1.NamespaceAll).List(context.TODO(), metav1.ListOptions{
```
giorio94, line 39:
```suggestion
	podList, err := p.client.CoreV1().Pods(metav1.NamespaceAll).List(ctx, metav1.ListOptions{
```

pkg/virtualKubelet/liqoNodeProvider/drain.go
```diff
@@ -0,0 +1,110 @@
...
+	if err := p.waitForDelete(ctx, pod); err != nil {
+		klog.Error(err)
+		errors <- err
+		return
+	}
+}
+
+// waitForDelete waits for the pod deletion.
+func (p *LiqoNodeProvider) waitForDelete(ctx context.Context, pod *v1.Pod) error {
+	return wait.PollImmediateInfinite(10*time.Second, func() (bool, error) {
```
giorio94, line 99: Would extract the duration as a constant.

pkg/virtualKubelet/liqoNodeProvider/reconciler.go
```diff
@@ -276,12 +329,36 @@ func (p *LiqoNodeProvider) updateNode() error {
...
 
 func (p *LiqoNodeProvider) handleResourceOfferDelete(resourceOffer *sharingv1alpha1.ResourceOffer) error {
+	ctx := context.TODO()
+
+	if err := p.cordonNode(ctx); err != nil {
+		klog.Errorf("error cordoning node: %v", err)
+		return err
+	}
+
+	if err := p.drainNode(ctx); err != nil {
```
giorio94, line 339: Is it fine to remain blocked here for long periods of time, or would it cause the loss of potential updates (e.g. about the status?)

giorio94 (approved):
/lgtm

Just a couple of minor comments and doubts inline.

pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller_methods.go
aleoli: No, the Tenant Namespace where the peering-related resources are created (and also the VirtualKubelet) will not be deleted. The problem setting up this owner reference is that the ServiceAccount will be deleted before the VirtualKubelet termination. Removing it we will have a ServiceAccount with no related permission that will be used by the next VK in the next peering with that cluster.

In your opinion, is it better to leave this ServiceAccount or to delete it at the same time we delete the ClusterRoleBinding? (that is handled with a finalizer)

pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller.go
aleoli: Request in my mind was the `reconcile.Request` 😅

I have definitely to be clearer

pkg/virtualKubelet/liqoNodeProvider/reconciler.go
aleoli: Yes, we will lose potential updates, but they are updates on resources/conditions over a node that we are deleting and will not exists anymore with this VK instance

palexster (approved):

pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller_methods.go
giorio94: The current approach seems fine to me, a serviceaccount without permissions does no harm.

pkg/virtualKubelet/liqoNodeProvider/reconciler.go
giorio94: I agree in the general case, but I am just a bit worried whether this would apply also in case of errors/problems.

PR #719, opened 2021-07-01
Fix generation of rbacs.

Andreagit97:
# Description

- some fixes in getCachedClient function.
- before regenerating rbacs files, removing all.


Makefile
```diff
@@ -61,6 +61,7 @@ manifests: controller-gen
 
 #Generate RBAC for each controller
 rbacs: controller-gen
+	rm -f -r deployments/liqo/files/*
```
giorio94, line 64: Is it possible to drop the `-r` flag? It is always a risk to use it inside a script, as it may lead to disasters if executed in the wrong directory...
Nit. It is a good practice to use the extended version of the flags inside the scripts, to make their semantic meaning more evident.

giorio94 (approved):

Makefile
Andreagit97: yes sure, thank you for the advise

palexster (approved):

Issue #721, opened 2021-07-01
Home pod container statuses not (gradually) updated on termination


fproj:
When a "shadow" home pod is deleted, the container statuses are updated all at once (status set to Terminated), only when the delete pod event is received from the foreign cluster. Liqo's `PodIncomingReflector` blacklists the foreign pod when it's terminating so all the foreign pod status intermediate updates are ignored.

Maybe not a problem.

PR #722, opened 2021-07-01
bugfix pod termination

### LINKED ISSUE #598: Pods hang in terminating status

filippoprojetto:
# Description

Pod deletion initiated from the home pod was not completed correctly.
Indeed, the pod's container statuses in the home was never set to
terminated so the virtual kubelet was not able to delete pod.

This fix updates the container statuses in one go, when the foreign pod
is deleted, ie. we miss to update the home pod when the foreign
containers are terminated. Fixing this problem is out of scope and would
probably require modifying the pod black listing mechanism which seems
too risky at this time. Issue described here:
#721.

Resolves: #598

# How Has This Been Tested?

- Manually on my development workstation using 2 local clusters (created with kind).
- Unit tests added for PodIncomingReflector.

### DIFF
--- a/pkg/virtualKubelet/apiReflection/reflectors/incoming/pods.go
+++ b/pkg/virtualKubelet/apiReflection/reflectors/incoming/pods.go
@@ -12,7 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/util/retry"
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
 
 	"github.com/liqotech/liqo/pkg/virtualKubelet"
 	apimgmt "github.com/liqotech/liqo/pkg/virtualKubelet/apiReflection"
@@ -22,10 +22,14 @@ import (
 	"github.com/liqotech/liqo/pkg/virtualKubelet/forge"
 )
 
+// HomePodGetter function to retrieve the shadow home pod given the foreign pod.
+type HomePodGetter func(reflector ri.APIReflector, foreignPod *corev1.Pod) (*corev1.Pod, error)
+
 // PodsIncomingReflector is the incoming reflector in charge of detecting status change in foreign pods
 // and pushing the updated object to the vk internals.
 type PodsIncomingReflector struct {
 	ri.APIReflector
+	HomePodGetter HomePodGetter
 }
 
 // SetSpecializedPreProcessingHandlers allows to set the pre-routine handlers for the PodsIncomingReflector.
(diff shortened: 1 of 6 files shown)


giorio94 (approved):

pkg/virtualKubelet/apiReflection/reflectors/incoming/pods.go
```diff
@@ -210,3 +210,27 @@ func (r *PodsIncomingReflector) isAllowed(ctx context.Context, obj interface{})
 	}
 	return !ok
 }
+
+// GetHomePod retrieves the shadow home pod given the foreign pod.
+var GetHomePod = func(pir *PodsIncomingReflector, foreignPod *corev1.Pod) (interface{}, error) {
```
giorio94, line 215: I feel this a bit weird as it is defined as a variable, instead of a proper function with the reflector as a receiver.
Still, I understand you need this for mocking during tests, so it seems fine to me.

pkg/virtualKubelet/apiReflection/reflectors/incoming/pods_test.go
```diff
@@ -1,12 +1,18 @@
...
+	"flag"
+	"fmt"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/types"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
```
giorio94, line 15:
```suggestion
	"k8s.io/klog/v2"
```

Unfortunately we never managed to make this a linter constraint

giorio94 (approved):

pkg/virtualKubelet/apiReflection/reflectors/incoming/pods.go
filippoprojetto: I have tried a different implementation that allows mocking the function. It might be prettier.

giorio94: Looks great!

giorio94 (approved):

palexster (approved):

PR #723, opened 2021-07-01
Check remote namespaces privileges.

Andreagit97:
# Description

When the NamespaceMap controller creates a new remote namespace, it must check that the right privileges are set to operate in that namespace.

Ref. #572

### DIFF
--- a/pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
+++ b/pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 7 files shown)


pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
```diff
@@ -177,3 +188,20 @@ func (r *NamespaceMapReconciler) namespaceMapDeletionProcess(ctx context.Context
...
+		klog.Errorf("%s -> unable to list roleBindings in the remote namespace '%s'", err, remoteNamespaceName)
+		return err
+	}
+	if len(roleBindingList.Items) < 2 {
+		err := fmt.Errorf("virtual kubelet will not have the necessary privileges on the remote namespace '%s'",
+			remoteNamespaceName)
+		klog.Error(err)
+		return err
+	}
+	return nil
```
giorio94, line 206: I would personally try to get the rolebinding by name or label (depending on how it is configured) and not rely on the number, since it may lead to false positives.

pkg/liqo-controller-manager/namespaceMap-controller/testUtils/namespacemap_controller_test_utils.go
```diff
@@ -0,0 +1,45 @@
...
+	roleBindingName1 = "role-binding-1"
+	roleBindingName2 = "role-binding-2"
+	roleType         = "Role"
+	roleName         = "fake"
+)
+
+// The remote namespace must have at least 2 roleBinding inside.
+
+// GetRoleBinding1ForASpecificNamespace provides a first roleBinding in the namespace passed as parameter.
+func GetRoleBinding1ForASpecificNamespace(namespaceName string) rbacv1.RoleBinding {
```
giorio94, line 18: You could join the two functions taking as parameter an index which is added as suffix to the name.

giorio94 (approved):
/lgtm. Just a couple of nits.

palexster (approved):

pkg/liqo-controller-manager/namespaceMap-controller/ensure_remote_namespaces.go
Andreagit97: Yes, I can use a label to List the roleBindings, do you agree @aleoli ?

pkg/liqo-controller-manager/namespaceMap-controller/testUtils/namespacemap_controller_test_utils.go
Andreagit97: thank you

PR #724, opened 2021-07-02
Change remote clients in NamespaceMap controller and fix some bugs.

Andreagit97:
# Description

- change remote clients from controller-runtime ones to standard ones (kubernetes.Interface).
- create Tenant with a new annotation for remote namespaces.
- add some consts to pkg/consts/namespace_mapping.go.
- namespaceMap controller now restarts with a resync period.
- change the annotation that the remote namespaces must have.

### DIFF
--- a/internal/resource-request-operator/resourceRequest_controller_methods.go
+++ b/internal/resource-request-operator/resourceRequest_controller_methods.go
@@ -11,6 +11,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
+	liqoconst "github.com/liqotech/liqo/pkg/consts"
 )
 
 const tenantFinalizer = "liqo.io/tenant"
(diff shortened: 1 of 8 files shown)


internal/resource-request-operator/resourceRequest_controller_methods.go
```diff
@@ -22,14 +23,20 @@ func requireTenantDeletion(resourceRequest *discoveryv1alpha1.ResourceRequest) b
...
 			Name: fmt.Sprintf("tenant-%v", remoteClusterID),
 		},
 	}
 	_, err = controllerutil.CreateOrUpdate(ctx, r.Client, tenant, func() error {
 		tenant.Spec = capsulev1alpha1.TenantSpec{
+			NamespacesMetadata: capsulev1alpha1.AdditionalMetadata{
+				AdditionalAnnotations: map[string]string{
+					annotationKey: liqoconst.RemoteNamespaceAnnotationValue,
+				},
+			},
```
aleoli, line 39: In my opinion, it is clearer to use a fixed key, like `liqo.io/remote-namespace` and set the clusterID as a value. What's your opinion @palexster ?

aleoli (approved):

internal/resource-request-operator/resourceRequest_controller_methods.go
Andreagit97: What about remote-namespace/liqo.io = clusterID ?

aleoli: Usually, labels and annotation are in the form domain/name

palexster: Totally agree

palexster (approved):

PR #725, opened 2021-07-02
Resources Phase Explicit Signaling

aleoli:
# Description

This pr allows peering-related resources (i.e. ResourceRequest and ResourceOffers) to have an explicit mechanism to indicate the desired state (Create os Delete) handled by the home cluster (in the resource `Spec`), and the observed state (Created od Deleted) handled by the remote cluster (in the resource `Status`). In this way, we are able the handle a graceful deletion of these resources and to have a graceful peering tear down. The VirtualKublet can now have the time required for a completed clean-up of the remote cluster.

# How Has This Been Tested?

- [x] locally on KinD


internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -178,6 +179,10 @@ func (r *ForeignClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			klog.Error(err)
 			return ctrl.Result{}, err
 		}
+	default:
```
palexster, line 182: Can we constrain the possible inputs also at the resource level?

aleoli: Sure, this value is returned by [this](https://github.com/liqotech/liqo/blob/24f23f23c589d68731f3e3e48d70fdcb9e5ec254/internal/discovery/foreign-cluster-operator/peeringPolicy.go#L15) function. It can only be one of the two values define, this default case is mostly to avoid forgetting to add a case if we will add other values

pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller_methods.go
```diff
@@ -225,11 +225,26 @@ func (r *ResourceOfferReconciler) getVirtualKubeletDeployment(
...
+	kubeletDeletePhaseNodeDeleted  kubeletDeletePhase = "NodeDeleted"
+)
+
+func getDeleteVirtualKubeletPhase(resourceOffer *sharingv1alpha1.ResourceOffer) kubeletDeletePhase {
 	notAccepted := !isAccepted(resourceOffer)
 	deleting := !resourceOffer.DeletionTimestamp.IsZero()
+	desiredDelete := !resourceOffer.Spec.WithdrawalTimestamp.IsZero()
 	nodeDrained := !controllerutil.ContainsFinalizer(resourceOffer, consts.NodeFinalizer)
-	return (notAccepted || deleting) && nodeDrained
+	if notAccepted || deleting || desiredDelete {
```
palexster, line 241: Can you please comment this logic?

palexster (approved):

giorio94 (approved):
/lgtm

PR #726, opened 2021-07-04
Activation of the new webhook logic

Andreagit97:
# Description

This PR activates the new webhook logic. When a pod is deployed in a namespace enabled for the liqo offloading, the liqo toleration and the right node selector are added to the pod according to the PodOffloadingStrategy specified in the NamespaceOffloading resource.

Ref. #572

### DIFF
--- a/deployments/liqo/values.yaml
+++ b/deployments/liqo/values.yaml
@@ -167,9 +167,6 @@ webhook:
   mutatingWebhookConfiguration:
     # -- mutatingWebhookConfiguration annotations
     annotations: {}
-    # -- The label that needs to be applied to a namespace to make it eligible for pod offloading in a remote cluster
-    namespaceSelector:
-      liqo.io/enabled: "true"
 
 peeringRequest:
   pod:
--- a/pkg/mutate/server.go
+++ b/pkg/mutate/server.go
@@ -27,12 +27,14 @@ type MutationServer struct {
 
 	webhookClient client.Client
 	config        *MutationConfig
+	ctx           context.Context
 }
 
 // NewMutationServer creates a new mutation server.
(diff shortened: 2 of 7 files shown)


giorio94 (approved):

deployments/liqo/values.yaml
```diff
@@ -169,7 +169,7 @@ webhook:
     annotations: {}
     # -- The label that needs to be applied to a namespace to make it eligible for pod offloading in a remote cluster
     namespaceSelector:
-      liqo.io/enabled: "true"
+      liqo.io/scheduling: "true"
```
aleoli, line 172: I have found an occurrence of it in `pkg/consts/namespace_mapping.go` can you please check if we can update it and if it is used?

And I think that we will update it in the documentation when releasing the 0.3 version @palexster

aleoli (approved):

pkg/mutate/server.go
```diff
@@ -27,12 +27,14 @@ type MutationServer struct {
...
 	webhookClient client.Client
 	config        *MutationConfig
+	ctx           context.Context
 }
 
 // NewMutationServer creates a new mutation server.
 func NewMutationServer(ctx context.Context, c *MutationConfig) (*MutationServer, error) {
 	s := &MutationServer{}
 	s.config = c
+	s.ctx = ctx
```
palexster, line 37: Should ctx be propagated from the main file instead of being declared here?

deployments/liqo/values.yaml
```diff
@@ -169,7 +169,7 @@ webhook:
     annotations: {}
     # -- The label that needs to be applied to a namespace to make it eligible for pod offloading in a remote cluster
     namespaceSelector:
```
palexster, line 171: I would remove this value since it is no more used.

PR #727, opened 2021-07-05
New package for IPAM module

davidefalcone1:
This PR moves the logic of the IPAM module and all its side components in a new package/folder created ad-hoc in order to improve the project directory/file organization.


pkg/liqonet/doc.go
```diff
@@ -1,2 +0,0 @@
-// Package liqonet all the related functions for the liqo network are contained in this package
```
palexster, line 1: Should it be really removed?

davidefalcone1: After this PR, directory liqonet contains only directories(which are packages) and 0 files. An external file would import a specific liqonet package, instead of liqonet.

palexster (approved):

alacuku (approved):

PR #733, opened 2021-07-06
Remove Old Peering

aleoli:
# Description

This pr removes the code related to the old peering mechanism


.github/workflows/integration.yml
```diff
@@ -5,7 +5,7 @@ on:
       - 'v*'
     branches:
       - master
-  pull_request_target:
+  pull_request:
```
giorio94, line 8: Should be reverted before merging. Same below

giorio94 (approved):

.github/workflows/integration.yml
aleoli: Sure, they are in a separate commit to remove them easily 👍

palexster (approved):

PR #734, opened 2021-07-06
Added OfferUpdater component which implement all ResourceOffer update.

giuse2596:
# Description
This P.R. adds all the missing logic concerning the ResourceOffer update. 
Introduce a new component, the Updater, which use a workqueue to process all locally existing ResourceOffer and update the offered resources to all the peered clusters. Every change in cluster nodes/pods trigger this component adding a new key (the clusterID of the remote cluster) which identifies the resourceOffer that will be updated as soon as possible.
In case of a dropped peering the key will be removed from the queue.

Refers to fifth point of Issue #548

### DIFF
--- a/cmd/advertisement-operator/main.go
+++ b/cmd/advertisement-operator/main.go
@@ -80,6 +80,7 @@ func main() {
 	var liqoNamespace, kubeletImage, initKubeletImage string
 	var resyncPeriod int64
 	var offloadingStatusControllerRequeueTime int64
+	var offerUpdateThreshold uint64
 	var namespaceMapControllerRequeueTime int64
 
 	flag.StringVar(&metricsAddr, "metrics-addr", defaultMetricsaddr, "The address the metric endpoint binds to.")
@@ -87,7 +88,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection,
 		"enable-leader-election", false,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
-
+	flag.Uint64Var(&offerUpdateThreshold, "offer-update-threshold-perc", uint64(5),
+		"Set the threshold percentage of quantity of resources modified which triggers the resourceOffer update.")
 	flag.Int64Var(&resyncPeriod, "resyncPeriod", int64(10*time.Hour), "Period after that operators and informers will requeue events.")
 	flag.Int64Var(&offloadingStatusControllerRequeueTime, "offloadingStatusControllerRequeueTime", int64(10*time.Second),
 		"Period after that the offloadingStatus Controller is awaken on every NamespaceOffloading to set its status.")
@@ -108,6 +110,11 @@ func main() {
 
 	if clusterId == "" {
 		klog.Error("Cluster ID must be provided")
+		os.Exit(1)
+	}
+
+	if offerUpdateThreshold > 100 {
+		klog.Error("offerUpdateThreshold exceeds 100")
 		os.Exit(1)
 	}
 
@@ -157,8 +164,9 @@ func main() {
 	}
 
 	newBroadcaster := &resourceRequestOperator.Broadcaster{}
-
-	if err := newBroadcaster.SetupBroadcaster(clientset, time.Duration(resyncPeriod)); err != nil {
+	updater := &resourceRequestOperator.OfferUpdater{}
+	updater.Setup(clusterId, mgr.GetScheme(), newBroadcaster, mgr.GetClient())
+	if err := newBroadcaster.SetupBroadcaster(clientset, updater, time.Duration(resyncPeriod), offerUpdateThreshold); err != nil {
 		klog.Error(err)
 		os.Exit(1)
 	}
--- a/internal/resource-request-operator/broadcaster.go
+++ b/internal/resource-request-operator/broadcaster.go
@@ -14,6 +14,7 @@ import (
 	resourcehelper "k8s.io/kubectl/pkg/util/resource"
 
 	configv1alpha1 "github.com/liqotech/liqo/apis/config/v1alpha1"
+	"github.com/liqotech/liqo/internal/resource-request-operator/interfaces"
 	crdclient "github.com/liqotech/liqo/pkg/crdClient"
 	"github.com/liqotech/liqo/pkg/utils"
 	"github.com/liqotech/liqo/pkg/virtualKubelet/forge"
@@ -21,20 +22,27 @@ import (
 
 // Broadcaster is an object which is used to get resources of the cluster.
 type Broadcaster struct {
-	allocatable    corev1.ResourceList
-	resourcePodMap map[string]corev1.ResourceList
-	clusterConfig  configv1alpha1.ClusterConfig
-	nodeMutex      sync.RWMutex
-	podMutex       sync.RWMutex
-	configMutex    sync.RWMutex
-	nodeInformer   cache.SharedIndexInformer
-	podInformer    cache.SharedIndexInformer
+	allocatable               corev1.ResourceList
+	resourcePodMap            map[string]corev1.ResourceList
+	lastReadResources         map[string]corev1.ResourceList
+	clusterConfig             configv1alpha1.ClusterConfig
+	nodeMutex                 sync.RWMutex
+	podMutex                  sync.RWMutex
+	configMutex               sync.RWMutex
+	nodeInformer              cache.SharedIndexInformer
+	podInformer               cache.SharedIndexInformer
+	updater                   interfaces.UpdaterInterface
+	updateThresholdPercentage uint64
 }
 
 // SetupBroadcaster create the informer e run it to signal node changes updating Offers.
-func (b *Broadcaster) SetupBroadcaster(clientset kubernetes.Interface, resyncPeriod time.Duration) error {
+func (b *Broadcaster) SetupBroadcaster(clientset kubernetes.Interface, updater interfaces.UpdaterInterface,
+	resyncPeriod time.Duration, offerUpdateThreshold uint64) error {
 	b.allocatable = corev1.ResourceList{}
+	b.updateThresholdPercentage = offerUpdateThreshold
+	b.updater = updater
 	b.resourcePodMap = map[string]corev1.ResourceList{}
+	b.lastReadResources = map[string]corev1.ResourceList{}
 	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
 	b.nodeInformer = factory.Core().V1().Nodes().Informer()
 	b.nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
@@ -54,6 +62,7 @@ func (b *Broadcaster) SetupBroadcaster(clientset kubernetes.Interface, resyncPer
 
 // StartBroadcaster starts two shared Informers, one for nodes and one for pods launching two separated goroutines.
 func (b *Broadcaster) StartBroadcaster(ctx context.Context, group *sync.WaitGroup) {
+	go b.updater.Start(ctx, group)
 	go b.startNodeInformer(ctx, group)
 	go b.startPodInformer(ctx, group)
 }
@@ -82,7 +91,8 @@ func (b *Broadcaster) setConfig(configuration *configv1alpha1.ClusterConfig) {
 	b.clusterConfig = *configuration
 }
 
-func (b *Broadcaster) getConfig() *configv1alpha1.ClusterConfig {
+// GetConfig returns an instance of a ClusterConfig resource.
+func (b *Broadcaster) GetConfig() *configv1alpha1.ClusterConfig {
 	b.configMutex.RLock()
 	defer b.configMutex.RUnlock()
 	configCopy := b.clusterConfig.DeepCopy()
@@ -173,8 +183,15 @@ func (b *Broadcaster) onPodDelete(obj interface{}) {
 // write nodes resources in thread safe mode.
 func (b *Broadcaster) writeClusterResources(newResources corev1.ResourceList) {
 	b.nodeMutex.Lock()
+	b.podMutex.RLock()
 	defer b.nodeMutex.Unlock()
+	defer b.podMutex.RUnlock()
 	b.allocatable = newResources.DeepCopy()
+	for clusterID := range b.resourcePodMap {
+		if b.checkThreshold(clusterID) {
+			b.enqueueForCreationOrUpdate(clusterID)
+		}
+	}
 }
 
 // write pods resources in thread safe mode.
@@ -182,6 +199,9 @@ func (b *Broadcaster) writePodsResources(clusterID string, newResources corev1.R
 	b.podMutex.Lock()
 	defer b.podMutex.Unlock()
 	b.resourcePodMap[clusterID] = newResources.DeepCopy()
+	if b.checkThreshold(clusterID) {
+		b.enqueueForCreationOrUpdate(clusterID)
+	}
 }
 
 // ReadResources return in thread safe mode a scaled value of the resources.
@@ -189,12 +209,26 @@ func (b *Broadcaster) ReadResources(clusterID string) corev1.ResourceList {
 	toRead := b.readClusterResources()
 	podsResources := b.readPodResources(clusterID)
 	addResources(toRead, podsResources)
+	b.lastReadResources[clusterID] = toRead
 	for resourceName, quantity := range toRead {
 		scaled := quantity
 		b.scaleResources(resourceName, &scaled)
 		toRead[resourceName] = scaled
 	}
 	return toRead
+}
+
+func (b *Broadcaster) enqueueForCreationOrUpdate(clusterID string) {
+	b.updater.Push(clusterID)
+}
+
+// RemoveClusterID removes a clusterID from all broadcaster internal structures
+// it is useful when a particular foreign cluster has no more peering and its ResourceRequest has been deleted.
+func (b *Broadcaster) RemoveClusterID(clusterID string) {
+	b.podMutex.Lock()
+	defer b.podMutex.Unlock()
+	delete(b.resourcePodMap, clusterID)
+	delete(b.lastReadResources, clusterID)
 }
 
 func (b *Broadcaster) readClusterResources() corev1.ResourceList {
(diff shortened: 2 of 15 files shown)


internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (u *OfferUpdater) updateOffer(clusterID string) error {
+	offerList := &sharingv1alpha1.ResourceOfferList{}
+	if err := u.Client.List(context.Background(), offerList, client.MatchingLabels{
+		discovery.ClusterIDLabel: clusterID,
+	}); err != nil {
+		return err
+	}
+	if len(offerList.Items) > 1 {
+		return fmt.Errorf("too much ResourceOffer for cluster %s", clusterID)
```
giorio94, line 92:
```suggestion
		return fmt.Errorf("too many ResourceOffers for cluster %s", clusterID)
```

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,114 @@
...
+	} else if len(offerList.Items) == 0 {
+		return fmt.Errorf("no ResourceOffer for cluster %s", clusterID)
+	}
```
giorio94, line 95: Shouldn't this condition occur the first time, when no resource offers are present yet?

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,114 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	sharingv1alpha1 "github.com/liqotech/liqo/apis/sharing/v1alpha1"
+	"github.com/liqotech/liqo/internal/resource-request-operator/interfaces"
+	"github.com/liqotech/liqo/pkg/discovery"
+)
+
+const requeueTimeout = 5 * time.Second
```
giorio94, line 19: Unclear to me why you are combining a tight requeue timeout with a push mechanism. I feel either one of the two approaches should be sufficient (if you use a push mechanism, then a requeue of half the TTL plus a random delta for de-sync should be sufficient, otherwise you drop the push mechanism and keep only the requeue)

internal/resource-request-operator/resourceRequest_controller.go
```diff
@@ -47,6 +49,17 @@ func (r *ResourceRequestReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	if resourceRequest.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(&resourceRequest, cleanBroadcasterFinalizer) {
+			controllerutil.AddFinalizer(&resourceRequest, cleanBroadcasterFinalizer)
+		}
+	} else {
+		if controllerutil.ContainsFinalizer(&resourceRequest, cleanBroadcasterFinalizer) {
+			r.Broadcaster.RemoveClusterID(remoteClusterID)
+			controllerutil.RemoveFinalizer(&resourceRequest, cleanBroadcasterFinalizer)
+		}
+	}
```
giorio94, line 61: Wondering whether the finalizer could be completely dropped. Theoretically, it should be sufficient to remove the clusterID if the get returns a notfound error at the beginning, avoiding the finalizer complexity. @palexster WDYT?

palexster: 100%

cmd/advertisement-operator/main.go
```diff
@@ -80,12 +80,17 @@ func main() {
...
+	var timeToLive int64
 
 	flag.StringVar(&metricsAddr, "metrics-addr", defaultMetricsaddr, "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection,
 		"enable-leader-election", false,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+	flag.Int64Var(&offerUploadThreshold, "offer-upload-threshold-perc", int64(5),
+		"Set the threshold percentage of quantity of resources modified which triggers the resourceOffer update.")
+	flag.Int64Var(&timeToLive, "resourceOffer-time-to-live", int64(30), "Set ResourceOffer time-to-live value")
```
giorio94, line 93: @aleoli Should the TTL still be set, or could it be completely dropped from the resource?

cmd/advertisement-operator/main.go
```diff
@@ -80,12 +80,17 @@ func main() {
...
 	var offloadingStatusControllerRequeueTime int64
+	var offerUploadThreshold int64
+	var timeToLive int64
 
 	flag.StringVar(&metricsAddr, "metrics-addr", defaultMetricsaddr, "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection,
 		"enable-leader-election", false,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+	flag.Int64Var(&offerUploadThreshold, "offer-upload-threshold-perc", int64(5),
```
giorio94, line 91:
```suggestion
	flag.Int64Var(&offerUpdateThreshold, "offer-update-threshold-perc", int64(5),
```

Should be probably unsigned, and would add a sanity check below to ensure it does not exceed 100.

internal/resource-request-operator/interfaces/broadcasterInterface.go
```diff
@@ -0,0 +1,13 @@
+package interfaces
+
+import corev1 "k8s.io/api/core/v1"
+
+// ClusterResourceInterface represents a generic subset of Broadcaster exported methods to be used instead of a direct access to
+// the Broadcaster instance and get/update some cluster resources information.
+type ClusterResourceInterface interface {
+	// ReadResources returns all free cluster resources calculated for a given clusterID scaled by a percentage value.
+	// If clusterID is not valid all scaled cluster resources will be returned.
```
giorio94, line 9: Would propose to return an error, to avoid unnoticed bugs

internal/resource-request-operator/interfaces/updaterInterface.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+	"sync"
+)
+
+// UpdaterInterface represents a generic subset of Updater exported methods to be used instead of a direct access to
+// a particular Updater instance.
+type UpdaterInterface interface {
+	// Start run an instance of an updater which will be stopped when ctx.Done() is called.
```
giorio94, line 11:
```suggestion
	// Start runs an instance of an updater which will be stopped when ctx.Done() is closed.
```

internal/resource-request-operator/interfaces/updaterInterface.go
```diff
@@ -0,0 +1,17 @@
...
+	Start(ctx context.Context, group *sync.WaitGroup)
+	// Push add the clusterID to the internal queue to be processed as soon as possible.
```
giorio94, line 13:
```suggestion
	// Push adds the clusterID to the internal queue to be processed as soon as possible.
```

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,146 @@
...
+const requeueTimeout = 5 * time.Minute
+
+// OfferUpdater is a component which wraps all ResourceOffer update logic.
+type OfferUpdater struct {
+	queue workqueue.RateLimitingInterface
+	client.Client
+	broadcasterInt interfaces.ClusterResourceInterface
+}
+
+// Setup initialize all parameters of the OfferUpdater component.
```
giorio94, line 31:
```suggestion
// Setup initializes all parameters of the OfferUpdater component.
```

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,146 @@
...
+func (u *OfferUpdater) Setup(broadcaster interfaces.ClusterResourceInterface, k8Client client.Client) {
+	u.broadcasterInt = broadcaster
+	u.Client = k8Client
+	u.queue = workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Offer update queue")
+}
+
+// Start run the OfferUpdate worker.
```
giorio94, line 38:
```suggestion
// Start runs the OfferUpdate worker.
```

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (u *OfferUpdater) isValidClusterID(clusterID string) bool {
+	resourceRequestList := &discoveryv1alpha1.ResourceRequestList{}
+	err := u.Client.List(context.Background(), resourceRequestList)
+	if err != nil || len(resourceRequestList.Items) != 1 {
+		return false
+	}
+	return true
+}
```
giorio94, line 136: Two comments here:
1. You may end up stop reconciling a resource due to temporary errors (e.g. the API server is not reachable).
2. You issue duplicate requests to the API server, since the same exact check is performed in the updateOffer function. Would be simpler to return an appropriate error there and issue forget if necessary.

For the sake of completeness, both problems are mitigated if the client is cached, nonetheless would avoid them in the first place.

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,146 @@
...
+
+func getRandomDelta() int64 {
+	max := new(big.Int)
+	max.SetInt64(int64(60))
```
giorio94, line 140: Would avoid hardcoding the value here and define an appropriate constant with the associated comment.

internal/resource-request-operator/broadcaster.go
```diff
@@ -197,6 +218,20 @@ func (b *Broadcaster) ReadResources(clusterID string) corev1.ResourceList {
 	return toRead
 }
 
+func (b *Broadcaster) enqueueForCreationOrUpdate(clusterID string) {
+	b.updater.Push(clusterID)
+}
+
+// RemoveClusterID remove a clusterID from all broadcaster internal structures
```
giorio94, line 225:
```suggestion
// RemoveClusterID removes a clusterID from all broadcaster internal structures
```

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,191 @@
...
+		var ok bool
+		if clusterID, ok = obj.(string); !ok {
+			// As the item in the workqueue is actually invalid, we call
+			// Forget here else we'd go into a loop of attempting to
+			// process a work item that is invalid.
+			u.queue.Forget(obj)
+			return fmt.Errorf("error getting object %v from OfferUpater queue. It is not a string", obj)
+		}
+		if requeue, err := u.createOrUpdateOffer(clusterID); err != nil {
+			// Put the item back on the workqueue to handle any transient errors.
```
giorio94, line 86: Would update this comment to reflect the additional check

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,191 @@
...
+			if requeue {
+				u.queue.AddRateLimited(clusterID)
+			}
```
giorio94, line 89: Should call `Remove()` if not requeued.

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,191 @@
...
+	u.queue.Add(clusterID)
+}
+
+// Remove remove a specified clusterID from the update queue and it will be no more processed.
+func (u *OfferUpdater) Remove(clusterID string) {
+	u.queue.Forget(clusterID)
+	klog.Infof("Removed cluster %s from update queue", clusterID)
+}
+
+func getRandomDelta() int64 {
```
giorio94, line 183: Nit. Could already return a duration here, to simplify the expression above.

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,191 @@
...
+			return controllerutil.SetControllerReference(&request, offer, u.scheme)
+		}
+		spec := sharingv1alpha1.ResourceOfferSpec{
+			ClusterId: u.homeClusterID,
+			Images:    []corev1.ContainerImage{},
+			ResourceQuota: corev1.ResourceQuotaSpec{
+				Hard: resources,
+			},
+			Labels: u.broadcasterInt.GetConfig().Spec.DiscoveryConfig.ClusterLabels,
+		}
```
giorio94, line 148: Doing so you would overwrite the entire spec. You should only change the fields you are controlling.

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,191 @@
...
+		},
+	}
+
+	op, err := controllerutil.CreateOrUpdate(context.Background(), u.Client, offer, func() error {
+		if offer.ObjectMeta.CreationTimestamp.IsZero() {
+			offer.Labels = map[string]string{
+				discovery.ClusterIDLabel:         request.Spec.ClusterIdentity.ClusterID,
+				crdreplicator.LocalLabelSelector: "true",
+				crdreplicator.DestinationLabel:   request.Spec.ClusterIdentity.ClusterID,
+			}
```
giorio94, line 129: Labels could always be updated, to ensure they are present. Still, you should pay attention not overwrite already existing ones.

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,191 @@
...
+			spec := sharingv1alpha1.ResourceOfferSpec{
+				ClusterId: u.homeClusterID,
+				Images:    []corev1.ContainerImage{},
+				ResourceQuota: corev1.ResourceQuotaSpec{
+					Hard: resources,
+				},
+				Labels: u.broadcasterInt.GetConfig().Spec.DiscoveryConfig.ClusterLabels,
+			}
+			offer.Spec = spec
+			return controllerutil.SetControllerReference(&request, offer, u.scheme)
```
giorio94, line 139: Also this could be always applied.

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+			spec := sharingv1alpha1.ResourceOfferSpec{
+				ClusterId: u.homeClusterID,
+				Images:    []corev1.ContainerImage{},
+				ResourceQuota: corev1.ResourceQuotaSpec{
+					Hard: resources,
+				},
+				Labels: u.broadcasterInt.GetConfig().Spec.DiscoveryConfig.ClusterLabels,
+			}
+			offer.Spec = spec
```
giorio94, line 138: This is common with the part below, so you do not need to fence it with the `IsZero()` check.

giorio94 (commented):
Looks in a good shape to me. Just some final comments about the update part.
Please, during the review phase keep separated commits, to avoid having to look through the entire code to spot the differences, then squash only in the end.

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,175 @@
...
+			Name:      offerPrefix + u.homeClusterID,
+		},
+	}
+
+	op, err := controllerutil.CreateOrUpdate(context.Background(), u.Client, offer, func() error {
+		offer.Labels = map[string]string{
+			discovery.ClusterIDLabel:         request.Spec.ClusterIdentity.ClusterID,
+			crdreplicator.LocalLabelSelector: "true",
+			crdreplicator.DestinationLabel:   request.Spec.ClusterIdentity.ClusterID,
+		}
```
giorio94, line 130: Should not overwrite the entire map, but just add your owns to keep user-added ones.

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,175 @@
...
+		// call createOrUpdate which after some controls generate a new resourceOffer for this clusterID or update it if exists.
+		if requeue, err := u.createOrUpdateOffer(clusterID); err != nil {
+			if requeue {
+				// requeue is true due to a transient error so put the item back on the workqueue.
+				u.queue.AddRateLimited(clusterID)
+			}
+			// requeue == false means that the clusterID is no more valid and so it will be not requeued.
+			u.Remove(clusterID)
+			u.broadcasterInt.RemoveClusterID(clusterID)
```
giorio94, line 92: Should be in the else block, otherwise it is always executed.

internal/resource-request-operator/interfaces/updaterInterface.go
```diff
@@ -0,0 +1,17 @@
...
+
+// UpdaterInterface represents a generic subset of Updater exported methods to be used instead of a direct access to
+// a particular Updater instance.
+type UpdaterInterface interface {
+	// Start runs an instance of an updater which will be stopped when ctx.Done() is closed.
+	Start(ctx context.Context, group *sync.WaitGroup)
+	// Push adds the clusterID to the internal queue to be processed as soon as possible.
+	Push(clusterID string)
+	// Remove removes the clusterID from the internal queue and will be no longer processed.
+	Remove(clusterID string)
```
giorio94, line 16: Seems to me this is no longer needed

internal/resource-request-operator/offerUpdater.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return resourceRequestList, nil
+}
+
+// Push add new clusterID to update queue which will be processes as soon as possible.
+func (u *OfferUpdater) Push(clusterID string) {
+	u.queue.Add(clusterID)
+}
+
+// Remove remove a specified clusterID from the update queue and it will be no more processed.
```
palexster, line 168:
```suggestion
// Remove a specified clusterID from the update queue and it will be no more processed.
```

palexster, line 168:
```suggestion
// Remove removes a specified clusterID from the update queue and it will be no more processed.
```

internal/resource-request-operator/resourceRequest_operator_test.go
```diff
@@ -362,12 +386,55 @@ var _ = Describe("ResourceRequest Operator", func() {
 				}
 				return true
 			}, timeout, interval).Should(BeTrue())
+			By("Checking ResourceOffer update")
```
palexster, line 389: The By clauses in the test should be more explicit about the expected outcome of the test.

internal/resource-request-operator/resourceRequest_operator_test.go
```diff
@@ -402,6 +470,28 @@ var _ = Describe("ResourceRequest Operator", func() {
 				}
 				return true
 			}, timeout, interval).Should(BeTrue())
+			By("Checking in ResourceOffer")
```
palexster, line 473: Ditto.

internal/resource-request-operator/resourceRequest_operator_test.go
```diff
@@ -417,6 +507,27 @@ var _ = Describe("ResourceRequest Operator", func() {
...
```
palexster, line 510: The same code is repeated three times and should be factorized.

giorio94 (approved):

palexster (approved):

PR #735, opened 2021-07-07
Remove Limits on Liqo deployments

palexster:
# Description

This commit removes default values for limits in Liqo components deployments to avoid unwanted performance degradation.

# How Has This Been Tested?

There are no functional changes in this PR, just E2E testing though.

### DIFF
--- a/deployments/liqo/templates/liqo-advertisement-deployment.yaml
+++ b/deployments/liqo/templates/liqo-advertisement-deployment.yaml
@@ -50,9 +50,6 @@ spec:
              fieldRef:
                fieldPath: metadata.namespace
         resources:
-          limits:
-            cpu: 100m
-            memory: 50M
           requests:
             cpu: 100m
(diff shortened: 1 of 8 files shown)


deployments/liqo/templates/liqo-advertisement-deployment.yaml
```diff
@@ -50,9 +50,6 @@ spec:
...
                fieldPath: metadata.namespace
         resources:
-          limits:
-            cpu: 100m
-            memory: 50M
           requests:
             cpu: 100m
-            memory: 50M
\ No newline at end of file
+            memory: 50M
```
giorio94, line 55: Would increase the memory request, given the number of controllers it is currently running.

giorio94 (approved):
/lgmt

Limits should also be dropped from the virtual kubelet deployment.

aleoli (approved):

PR #736, opened 2021-07-07
Integration tests for IPAM module

davidefalcone1:
This PR introduces a set of integration tests for the IPAM module. This important Liqo component is involved in the remapping of IPs of remote endpoints to be reflected in foreign clusters: in particular the IPAM is asked to map an IP by the Virtual Kubelet, responsible of service reflections. However, the VK is not itested in this PR. When the IPAM maps an IP with ExternalCIDR network, it notifies the Gateway via the NatMapping resource, that in turn adds the appropriate DNAT rules. This PR consists of tests on the communication between the IPAM and the Gateway, as well as tests on the correct update of NAT rules by the Gateway.

### DIFF
--- a/pkg/liqonet/ipam/ipam.go
+++ b/pkg/liqonet/ipam/ipam.go
@@ -975,12 +975,13 @@ If the received IP belongs to local PodCIDR, then it maps the address in the tra
 i.e. using the network used in the remote cluster for local PodCIDR.
 If the received IP does not belong to local PodCIDR, then it maps the address using the ExternalCIDR.*/
 func (liqoIPAM *IPAM) mapEndpointIPInternal(clusterID, ip string) (string, error) {
+	const emptyPodCIDR = ""
 	var subnets netv1alpha1.Subnets
 	var exists bool
 
-	// Parse IP
-	if netIP := net.ParseIP(ip); netIP == nil {
-		return "", fmt.Errorf("cannot parse endpointIP %s", ip)
+	err := validateEndpointMappingInputs(clusterID, ip)
+	if err != nil {
+		return "", err
 	}
 
 	// Get cluster subnets
@@ -989,10 +990,13 @@ func (liqoIPAM *IPAM) mapEndpointIPInternal(clusterID, ip string) (string, error
 		return "", fmt.Errorf("cannot get cluster subnets:%w", err)
 	}
 	subnets, exists = clusterSubnets[clusterID]
+	if !exists {
+		return "", fmt.Errorf("cluster %s has not a network configuration", clusterID)
+	}
 
 	// Get PodCIDR
 	podCIDR, err := liqoIPAM.ipamStorage.getPodCIDR()
-	if err != nil {
+	if err != nil || podCIDR == emptyPodCIDR {
 		return "", fmt.Errorf("cannot get cluster PodCIDR: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/liqonet/ipam/ipam.go
```diff
@@ -989,10 +989,13 @@ func (liqoIPAM *IPAM) mapEndpointIPInternal(clusterID, ip string) (string, error
...
 	}
 	subnets, exists = clusterSubnets[clusterID]
+	if !exists {
+		return "", fmt.Errorf("cluster %s has not a network configuration", clusterID)
+	}
 
 	// Get PodCIDR
 	podCIDR, err := liqoIPAM.ipamStorage.getPodCIDR()
-	if err != nil {
+	if err != nil || podCIDR == "" {
```
palexster, line 998: Would be nice to have "" mapped to a constant which explains better the semantic of the value. Something like "EmptyPODCIDR" or similar stuff.

pkg/liqonet/ipam/ipam.go
```diff
@@ -1016,11 +1015,6 @@ func (liqoIPAM *IPAM) mapEndpointIPInternal(clusterID, ip string) (string, error
 	}
 	// IP does not belong to cluster PodCIDR: Pod is a reflected Pod
 
-	// Check existence of Local NAT ExternalCIDR
```
palexster, line 1019: Ditto.

pkg/liqonet/ipam/ipam_integration_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/containernetworking/plugins/pkg/ns"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/kubernetes/pkg/util/slice"
+
+	"github.com/liqotech/liqo/pkg/consts"
+	"github.com/liqotech/liqo/pkg/liqonet/errors"
+	liqonetIpam "github.com/liqotech/liqo/pkg/liqonet/ipam"
+)
+
```
palexster, line 16: Since the test leverage multiple components, I would add a comment to describe which is the test scenario, before the real assertions.

pkg/liqonet/ipam/ipam_integration_test.go
```diff
@@ -0,0 +1,165 @@
...
+		})
+		Context("Ask to map the IP of a local endpoint", func() {
+			It("IPAM module should return the input IP mapped on Local NAT PodCIDR network", func() {
+				response, err := ipamInt.MapEndpointIP(context.Background(), &liqonetIpam.MapRequest{
+					ClusterID: clusterID4,
+					Ip:        localEndpointIP,
+				})
+				Expect(err).To(BeNil())
+				Expect(response.GetIp()).To(Equal("10.70.0.20")) // Local NAT PodCIDR is 10.70.0.0/24
+			})
```
palexster, line 54: Those first assertions (lines 18-54) look like a unit test for the IPAM service validating the correct input, I don't know if they deserve to be put in the corresponding unit tests.

pkg/liqonet/ipam/ipam_suite_test.go
```diff
@@ -1,13 +1,203 @@
...
+		err = ipt.Terminate()
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	return err
+}
+
+func initIPTables() error {
```
palexster, line 183: I would refer to the semantic component in Liqo instead of the technology used under the hoods.

pkg/liqonet/ipam/ipam.go
```diff
@@ -1042,6 +1037,22 @@ func (liqoIPAM *IPAM) MapEndpointIP(ctx context.Context, mapRequest *MapRequest)
 	return &MapResponse{Ip: mappedIP}, nil
 }
 
+func validateEndpointMappingInputs(clusterID, ip string) error {
+	const emptyClusterID = ""
+	// Parse IP
+	if netIP := net.ParseIP(ip); netIP == nil {
+		return fmt.Errorf("cannot parse endpointIP %s", ip)
```
alacuku, line 1044: We have some errors defined in the `liqonet` package. Why not using them? As you have done for the `clusterID`.

alacuku (approved):

palexster (approved):

PR #739, opened 2021-07-08
E2E tests for the new namespacemapping logic

Andreagit97:
# Description

1 - cluster_label_test: this test aims to verify that the labels inserted at installation time are actually put on the virtual nodes.


test/e2e/testutils/tester/tester.go
```diff
@@ -42,65 +58,151 @@ func GetTester(ctx context.Context) *Tester {
...
+		klog.Fatal("KUBECONFIG_2 not set")
 	}
+	kubeconfig3 := os.Getenv("KUBECONFIG_3")
+	if kubeconfig3 == "" {
+		klog.Fatal("KUBECONFIG_3 not set")
+	}
+	kubeconfig4 := os.Getenv("KUBECONFIG_4")
+	if kubeconfig4 == "" {
+		klog.Fatal("KUBECONFIG_4 not set")
+	}
```
giorio94, line 74: IIUC (@palexster please double check), some tests may still be executed with two clusters only, so this should be configurable (and possibly code duplication should be reduced)

test/e2e/cluster_labels_e2e/cluster_labels_test.go
```diff
@@ -0,0 +1,222 @@
...
+				match, err := k8shelper.MatchNodeSelectorTerms(&virtualNodesList.Items[i], testutils.GetClusterSelector())
+				Expect(err).To(Succeed())
+				remoteClusterID := virtualNodesList.Items[i].Annotations[liqoconst.RemoteClusterID]
+				if match {
+					// Check if the remote namespace is correctly created.
+					By(fmt.Sprintf(" 5 - Checking if a remote namespace is correctly created inside cluster '%s'", remoteClusterID))
+					Eventually(func() error {
+						namespace := &corev1.Namespace{}
+						if err = testContext.ClustersClients[remoteClusterID].Get(ctx,
+							types.NamespacedName{Name: testNamespaceName}, namespace); err != nil {
```
giorio94, line 173: IIUC, the remote namespace name is different from the local one.

test/e2e/remote_namespaces_creation_e2e/remote_namespaces_creation_test.go
```diff
@@ -0,0 +1,190 @@
...
+				return testutils.CreateNamespaceWithNamespaceOffloading(ctx, testContext.Clusters[localIndex].ControllerClient, testNamespaceName)
+			}, timeout, interval).Should(Succeed())
+
+			By(" 2 - Getting the just created NamespaceOffloading resource")
+			Eventually(func() error {
+				return testContext.Clusters[localIndex].ControllerClient.Get(ctx,
+					types.NamespacedName{Namespace: testNamespaceName, Name: liqoconst.DefaultNamespaceOffloadingName}, namespaceOffloading)
+			}, timeout, interval).Should(Succeed())
+
+			By(" 3 - Getting the NamespaceMaps and check the presence of the entries for that namespace, both in the spec and status")
```
giorio94, line 69:
```suggestion
			By(" 3 - Getting the NamespaceMaps and checking the presence of the entries for that namespace, both in the spec and status")
```

test/e2e/remote_namespaces_creation_e2e/remote_namespaces_creation_test.go
```diff
@@ -0,0 +1,190 @@
...
+			for i := range testContext.Clusters {
+				if i == localIndex {
+					continue
+				}
+				Eventually(func() error {
+					return testContext.Clusters[i].ControllerClient.Get(ctx,
+						types.NamespacedName{Name: remoteTestNamespaceName}, namespace)
+				}, longTimeout, interval).Should(Succeed())
+				value, ok := namespace.Annotations[liqoconst.RemoteNamespaceAnnotationKey]
+				Expect(ok).To(Equal(true))
```
giorio94, line 120:
```suggestion
				Expect(ok).To(BeTrue())
```

test/e2e/remote_namespaces_creation_e2e/remote_namespaces_creation_test.go
```diff
@@ -0,0 +1,190 @@
...
+			}, longTimeout, interval).Should(Equal(metav1.StatusReasonNotFound))
+
+			By(" 3 - Checking if the NamespaceMaps do not have the entries for that test namespace")
+			Eventually(func() error {
+				return testContext.Clusters[localIndex].ControllerClient.List(ctx, namespaceMapsList)
+			}, timeout, interval).Should(Succeed())
+			Expect(len(namespaceMapsList.Items)).To(Equal(testconsts.NumberOfTestClusters - 1))
+			for i := range namespaceMapsList.Items {
+				_, ok := namespaceMapsList.Items[i].Spec.DesiredMapping[testNamespaceName]
+				Expect(ok).To(Equal(false))
```
giorio94, line 166:
```suggestion
				Expect(ok).To(BeFalse())
```

test/e2e/conflict_remote_namespace_e2e/conflict_creation_test.go
```diff
@@ -0,0 +1,122 @@
...
+				return nil
+			}, longTimeout, interval).Should(Succeed())
+
+			By(fmt.Sprintf(" 4 - Deleting the remote namespace inside the cluster '%d'", remoteIndex))
+			Eventually(func() metav1.StatusReason {
+				err := testContext.Clusters[remoteIndex].ControllerClient.Get(ctx,
+					types.NamespacedName{Name: remoteTestNamespaceName}, namespace)
+				_ = testContext.Clusters[remoteIndex].ControllerClient.Delete(ctx, namespace)
+				return apierrors.ReasonForError(err)
+			}, timeout, interval).Should(Equal(metav1.StatusReasonNotFound))
```
giorio94, line 84: Here you risk racing against your controller. You delete the namespace, then check for its actual removal, but it might already have been recreated.

test/e2e/cluster_labels_e2e/cluster_labels_test.go
```diff
@@ -0,0 +1,222 @@
...
+						}
+						value, ok := namespace.Annotations[liqoconst.RemoteNamespaceAnnotationKey]
+						if !ok {
+							return fmt.Errorf("the annotation '%s' is not present", liqoconst.RemoteNamespaceAnnotationKey)
+						}
+						if value != testContext.Clusters[localIndex].ClusterID {
+							return fmt.Errorf("the value of the annotation is not correct: '%s'", value)
+						}
+						return nil
+					}, longTimeout, interval).Should(Succeed())
```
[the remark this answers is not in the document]
Andreagit97: yes you are right, i don't know if 10 second are enough but it sure does not take two minutes, I will change it thank you.

test/e2e/conflict_remote_namespace_e2e/conflict_creation_test.go
Andreagit97: yes you are right, but in the particular scenario of this test the controller should resync every 30 sec, so there shouldn't be a race condition with the controller. I understand that changing the resync timing could be a problem, I will try a different solution even if the behavior to test is just this, thank you

test/e2e/cluster_labels_e2e/cluster_labels_test.go
```diff
@@ -0,0 +1,222 @@
...
+		})
+
+		It("Delete the local namespace and check if the remote namespaces are deleted", func() {
+			By(" 1 - Getting the local namespace and delete it")
+			namespace := &corev1.Namespace{}
+			Eventually(func() metav1.StatusReason {
+				err := testContext.Clusters[localIndex].ControllerClient.Get(ctx, types.NamespacedName{Name: testNamespaceName}, namespace)
+				_ = testContext.Clusters[localIndex].ControllerClient.Delete(ctx, namespace)
+				return apierrors.ReasonForError(err)
+			}, longTimeout, interval).Should(Equal(metav1.StatusReasonNotFound))
```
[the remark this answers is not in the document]
giorio94: IMHO they are mostly equivalent, since deleting the namespace automatically leads to the deletion of the NamespaceOffloading. The other alternative was just a bit more explicit, but the current approach is also perfectly fine to me

test/e2e/cluster_labels_e2e/cluster_labels_test.go
giorio94: Sorry, I missed that. Totally legitimate

test/e2e/conflict_remote_namespace_e2e/conflict_creation_test.go
giorio94: I would propose the following (let me know if sounds ok to you):
1. Get the namespace, and save the `uid`
2. Delete the namespace, without waiting for its removal and retrying only in case of error
3. Assert that eventually the namespace still exists, but has a different `uid` (hence, it must have been recreated)

test/e2e/cluster_labels_e2e/cluster_labels_test.go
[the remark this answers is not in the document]
Andreagit97: Yes you are right it is more explicit

test/e2e/testutils/tester/tester.go
palexster: Totally agree.

test/e2e/conflict_remote_namespace_e2e/conflict_creation_test.go
Andreagit97: Great solution, thank you for the advise :1st_place_medal:

Issue #670, opened 2021-06-09
ForeignCluster Refactoring


aleoli:
# Description

This issue tracks the refactoring of the ForeingCluster Operator, to get It cleaner and readable

# Steps

### Spec
```go
// ForeignClusterSpec defines the desired state of ForeignCluster.
type ForeignClusterSpec struct {
	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
	// Important: Run "make" to regenerate code after modifying this file

	// Foreign Cluster Identity.
	ClusterIdentity ClusterIdentity `json:"clusterIdentity,omitempty"`
	// Enable join process to foreign cluster.
	// +kubebuilder:default=None
	Join Enum `json:"join,omitempty"`
	// +kubebuilder:validation:Enum="LAN";"WAN";"Manual";"IncomingPeering"
	// +kubebuilder:default="Manual"
	// How this ForeignCluster has been discovered.
	DiscoveryType discovery.Type `json:"discoveryType,omitempty"`
	// URL where to contact foreign Auth service.
	AuthURL string `json:"authUrl"`
	// +kubebuilder:validation:Enum="Unknown";"Enabled";"Disabled"
	// +kubebuilder:default="Unknown"
	// Indicates if this remote cluster is trusted or not.
	TLSVerificationEnabled discovery.TrustMode `json:"tlsVerificationEnabled,omitempty"`
	
	DNSGarbageColleciton
		// If discoveryType is LAN or WAN and this indicates the number of seconds after that
		// this ForeignCluster will be removed if no updated has been received.
		// +kubebuilder:validation:Minimum=0
		TTL int `json:"ttl,omitempty"`
		LastUpdate time `json:"lastUpdate,omitempty"`
}
```
```go
// ForeignClusterStatus defines the observed state of ForeignCluster.
type ForeignClusterStatus struct {
	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
	// Important: Run "make" to regenerate code after modifying this file

	// TenantControlNamespaces names in the peered clusters
	// +kubebuilder:validation:Optional
	TenantNamespace TenantNamespace `json:"tenantNamespace"`

	// It stores the most important information about the peering with a remote cluster.
	PeeringConditions Condition `json:"peeringConditions,omitempty"`
}
```

aleoli:
Closing, as completed

PR #744, opened 2021-07-09
ForeignCluster Fields Refactoring

aleoli:
# Description

This pr refactor and removes some fields in the ForeignCluster resource as described in the related issue

In details:
* the `Join` flag has been moved to the `PeeringEnabled` parameter
* the `AuthURL` has been renamed to `ForeignAuthURL`
* the `TrustMode ` has been moved to `InsecureSkipTLSVerify `

Ref #670 

# How Has This Been Tested?

- [x] Locally on KinD
- [x] Add unit test over the logic to determine if the peering is enabled or not


palexster:
Hi @aleoli! Thanks for this wonderful work which was really needed! :rocket: 

I was wondering if you are considering the effect on documentation, in particular the quickstart section. We do not support multi-version documentation so far, but I would at least annotate somewhere the files that should be updated before the 0.3 release.

deployments/liqo/crds/discovery.liqo.io_foreignclusters.yaml
```diff
@@ -68,38 +65,32 @@ spec:
...
-              namespace:
-                description: Namespace where Liqo is deployed. (Deprecated)
+              foreignAuthUrl:
+                description: URL where to contact foreign Auth service.
+                pattern: https:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)
                 type: string
-              trustMode:
-                default: Unknown
-                description: Indicates if this remote cluster is trusted or not.
+              peeringEnabled:
```
palexster, line 72: I was wondering (maybe a silly idea), would it be clearer for the user to specify the "direction" of the peering? In order to easily identify the corresponding field in the status. What do you think? @aleoli @alacuku @giorio94

palexster (approved):

apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -50,28 +49,42 @@ const (
...
-	AuthURL string `json:"authUrl"`
-	// +kubebuilder:validation:Enum="Unknown";"Trusted";"Untrusted"
-	// +kubebuilder:default="Unknown"
-	// Indicates if this remote cluster is trusted or not.
-	TrustMode discovery.TrustMode `json:"trustMode,omitempty"`
+	// +kubebuilder:validation:Pattern=`https:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)`
+	ForeignAuthURL string `json:"foreignAuthUrl"`
+	// +kubebuilder:default=true
+	// Indicates if the local cluster has to skip the tls verification over the remote Authentication Service or not.
+	InsecureSkipTLSVerify *bool `json:"trustMode,omitempty"`
```
giorio94, line 87: Better not use omitempty in combination with default=true, as it may lead to update failure using the controller-runtime client (although the pointer should reduce the issues).

pkg/utils/foreignCluster/discoveryType.go
```diff
@@ -0,0 +1,22 @@
...
+	if l, ok := labels[discovery.DiscoveryTypeLabel]; ok {
+		return discovery.Type(l)
+	}
+	return discovery.ManualDiscovery
+}
+
+// SetDiscoveryType sets the discovery type to the given ForeignCluster.
+func SetDiscoveryType(foreignCluster *discoveryv1alpha1.ForeignCluster, disocveryType discovery.Type) {
+	labels := foreignCluster.GetLabels()
+	labels[discovery.DiscoveryTypeLabel] = string(disocveryType)
```
giorio94, line 20:
```suggestion
	labels[discovery.DiscoveryTypeLabel] = string(discoveryType)
```

pkg/utils/foreignCluster/discoveryType.go
```diff
@@ -0,0 +1,22 @@
...
+	labels := foreignCluster.GetLabels()
+	if l, ok := labels[discovery.DiscoveryTypeLabel]; ok {
+		return discovery.Type(l)
+	}
+	return discovery.ManualDiscovery
+}
+
+// SetDiscoveryType sets the discovery type to the given ForeignCluster.
+func SetDiscoveryType(foreignCluster *discoveryv1alpha1.ForeignCluster, disocveryType discovery.Type) {
+	labels := foreignCluster.GetLabels()
```
giorio94, line 19: Could labels be nil?

deployments/liqo/crds/discovery.liqo.io_foreignclusters.yaml
giorio94: Do you mean renaming the field key or changing the values semantic? I feel the latter would be probably quite a big modification with many consequences. 

Still, I agree outgoingPeeringEnabled would be clearer, with no impact. And, as for the future, I also agree it would probably nice having the possibility to disable the incoming peering through the foreigncluster (or accept it conditionally)

giorio94 (approved):
/lgtm

apis/discovery/v1alpha1/foreigncluster_types.go
aleoli: are you suggesting something like that?

```go
// +kubebuilder:default=true
// +kubebuilder:optional
InsecureSkipTLSVerify *bool `json:"insecureSkipTLSVerify"`
```

giorio94: Exactly. In the part we experienced problems with the other configuration, although it wasn't a pointer

deployments/liqo/crds/discovery.liqo.io_foreignclusters.yaml
palexster: The first one, but the second looks like an interesting future improvement.

My point was to map the spec field easily with the status one, to avoid the reaction "I enabled the peering, but which field should I look to check if everything went well?".

PR #746, opened 2021-07-12
Fix information leakage vulnerability in the Webhook server.

Andreagit97:
# Description

If the body of the request was not correctly read the error provided was directly sent to the user. Now a standard error message is sent every time there is a problem in the processing of the request body, this fix avoid also the information leakage due to the use of specific errors for every different problem.

### DIFF
--- a/pkg/mutate/server.go
+++ b/pkg/mutate/server.go
@@ -62,16 +62,19 @@ func NewMutationServer(ctx context.Context, c *MutationConfig) (*MutationServer,
 func (s *MutationServer) handleMutate(w http.ResponseWriter, r *http.Request) {
 	// read the body / request
 	body, err := ioutil.ReadAll(r.Body)
-
 	if err != nil {
-		s.sendError(err, w)
+		klog.Error(err)
+		standardErrMessage := fmt.Errorf("unable to correctly read the body of the request")
+		s.sendError(standardErrMessage, w)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/mutate/server.go
```diff
@@ -62,15 +62,16 @@ func NewMutationServer(ctx context.Context, c *MutationConfig) (*MutationServer,
 func (s *MutationServer) handleMutate(w http.ResponseWriter, r *http.Request) {
 	// read the body / request
 	body, err := ioutil.ReadAll(r.Body)
-
 	if err != nil {
+		err = fmt.Errorf("unable to correctly read the body of the request")
```
giorio94, line 66: Would log the actual error, for the sake of troubleshooting. Same below

giorio94 (approved):
/lgtm

A bit intimidating the title, and not sure it really reflects the problem (it seems more information leakage), but the fix seems ok to me.

Andreagit97:
> /lgtm
> 
> A bit intimidating the title, and not sure it really reflects the problem (it seems more information leakage), but the fix seems ok to me.

Completely agree with you, it seems a possible information leakage due to the knowledge of the specific error by the possible attacker.

pkg/mutate/server.go
Andreagit97: Sure, thank you

palexster (approved):

PR #749, opened 2021-07-13
Add custom columns to "kubectl get" output for liqo CRDs

alacuku:
# Description
 
This PR adds custom columns to "kubectl get" output for liqo CRDs. Here is an example when a `kubectl get -o wide` is performed:

kubectl get

Fixes #(issue)

Now with a simple `kubectl get` it is clear if a resource has been generated on the current cluster or has been replicated from a remote cluster.

### DIFF
--- a/apis/sharing/v1alpha1/resourceoffer_types.go
+++ b/apis/sharing/v1alpha1/resourceoffer_types.go
@@ -73,8 +73,8 @@ type ResourceOfferStatus struct {
 
 // ResourceOffer is the Schema for the resourceOffers API.
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.phase`
-// +kubebuilder:printcolumn:name="Expiration",type=string,JSONPath=`.spec.timeToLive`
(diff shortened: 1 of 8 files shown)


aleoli (approved):

apis/sharing/v1alpha1/resourceoffer_types.go
```diff
@@ -75,6 +75,7 @@ type ResourceOfferStatus struct {
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.phase`
 // +kubebuilder:printcolumn:name="Expiration",type=string,JSONPath=`.spec.timeToLive`
```
giorio94, line 76: Does this TTL still exist? If not it should be dropped.

giorio94 (approved):

apis/sharing/v1alpha1/resourceoffer_types.go
alacuku: I don't know. Maybe @aleoli and @palexster could help here.

aleoli: In my opinion, we can drop it

PR #751, opened 2021-07-15
Add AWS IAM Identity Provider - AWS 1/3

aleoli:
# Description

This pr adds a new Identity Provider in the Liqo Auth component, this new provider handles the authentication of uses to an EKS cluster using the Amazon IAM service.

# How Has This Been Tested?

- [x] On EKS cluster
- [ ] Add unit test - others will be added in a future pr


pkg/identityManager/identityManager.go
```diff
@@ -8,18 +8,35 @@ import (
...
 type certificateIdentityManager struct {
+	identityProvider
+
 	client           kubernetes.Interface
 	localClusterID   clusterid.ClusterID
 	namespaceManager tenantnamespace.Manager
 }
 
 // NewCertificateIdentityManager gets a new certificate identity manager.
 func NewCertificateIdentityManager(client kubernetes.Interface,
```
giorio94, line 19: Would personally split this function in two (one for certificates and the other for iam), to avoid relying on the awsConfig parameter being null, which is difficult to understand when observing only the function call. Also the structure name should be probably generalized to `IdentityManager`.

pkg/identityManager/responseTypes/signingRequestResponse.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"github.com/aws/aws-sdk-go/service/eks"
+	"github.com/aws/aws-sdk-go/service/iam"
+)
+
+// SigningRequestResponseType indicates the type for a signign request response.
+type SigningRequestResponseType string
+
+const (
+	// SigningRequestResponseCertificate indicates that the signing request reaponse contains a certificate
```
giorio94, line 12:
```suggestion
	// SigningRequestResponseCertificate indicates that the signing request response contains a certificate
```

pkg/identityManager/responseTypes/signingRequestResponse.go
```diff
@@ -0,0 +1,29 @@
...
+// SigningRequestResponse contains the response from an Indentity Provider.
+type SigningRequestResponse struct {
+	ResponseType SigningRequestResponseType
+
+	Certificate []byte
+
+	AwsIamUserArn string
+	AwsAccessKey  *iam.AccessKey
+	AwsEksCluster *eks.Cluster
+	AwsRegion     string
```
giorio94, line 28: Could be worth wrap all of this in a single structure

pkg/auth/identityResponse.go
```diff
@@ -2,51 +2,83 @@ package auth
...
-	Certificate  string `json:"certificate"`
+	Certificate  string `json:"certificate,omitempty"`
 	APIServerURL string `json:"apiServerUrl"`
 	APIServerCA  string `json:"apiServerCA,omitempty"`
+
+	AWSAccessKeyID     string `json:"awsAccessKeyID,omitempty"`
+	AWSSecretAccessKey string `json:"awsSecretAccessKey,omitempty"`
+	AWSRegion          string `json:"awsRegion,omitempty"`
+	AWSEKSClusterID    string `json:"awsEksClusterID,omitempty"`
+	AWSIAMUserArn      string `json:"awsIamUserArn,omitempty"`
```
giorio94, line 28: Same here

pkg/identityManager/iamIdentityProvider.go
```diff
@@ -0,0 +1,189 @@
...
+	// this method has no meaning for this identity provider
+	return response, kerrors.NewNotFound(schema.GroupResource{
+		Group:    "v1",
+		Resource: "secrets",
+	}, remoteCertificateSecret)
+}
+
+func (identityProvider *iamIdentityProvider) ApproveSigningRequest(clusterID,
+	signingRequest string) (response responsetypes.SigningRequestResponse, err error) {
+	sess := session.Must(session.NewSession(&aws.Config{
```
giorio94, line 49: What does `Must` imply. Does it panic in case of failure?

giorio94 (approved):
/lgtm

Just a couple of minor comments inline

pkg/identityManager/iamIdentityProvider.go
aleoli: Yes, exactly. When I wrote that I was sure that was a good idea to not hide authentication problems with amazon and to have a clear restart of the component, but now I'm changing my idea 😅

aleoli: ok, I removed that

pkg/auth/identityResponse.go
```diff
@@ -2,51 +2,90 @@ package auth
...
 	"github.com/liqotech/liqo/pkg/utils"
 )
 
+// AWSIdentityInfo contains the information required by a cluster to get a valied IAM-based identity.
+type AWSIdentityInfo struct {
+	AccessKeyID     string `json:"accessKeyID,omitempty"`
+	SecretAccessKey string `json:"secretAccessKey,omitempty"`
+	Region          string `json:"region,omitempty"`
+	EKSClusterID    string `json:"eksClusterID,omitempty"`
+	IAMUserArn      string `json:"iamUserArn,omitempty"`
```
giorio94, line 23: Just wondering, could these be empty? Or should they all be present if the parent structure is present?

aleoli: You are right, they have to be all present

cmd/auth-service/main.go
```diff
@@ -31,13 +34,18 @@ func main() {
 	flag.StringVar(&keyFile, "keyFile", "/certs/key.pem", "Path to key file")
 	flag.BoolVar(&useTLS, "useTls", false, "Enable HTTPS server")
 
+	flag.StringVar(&awsConfig.AwsAccessKeyID, "awsAccessKeyId", "", "AWS IAM AccessKeyID for the Liqo User")
```
palexster, line 37: Personally, I would the AWS credentials in a dedicated secret object.

palexster (approved):

cmd/auth-service/main.go
aleoli: fixed in the #755

PR #752, opened 2021-07-15
Deletion of remote namespace conditions in the NamespaceOffloading Resource

Andreagit97:
# Description

- When a virtual node is deleted during the termination of the peering, the remote conditions contained in the namespaceOffloading resource must be updated. In particular, when the NamespaceMap corresponding to the virtual node no longer exists, the condition associated with that remote cluster must be removed. So a new check in the Offloading status controller is added: as long as the NamespaceOffloading is present, a periodic check is carried out on the remoteNamespaceConditions that must be present at that moment.
A new unit test is added in order to verify the behavior of this new feature.

- A return statement in the resourceOffer controller is added to avoid race conditions during the peering deletion phase.

### DIFF
--- a/pkg/liqo-controller-manager/offloadingStatus-controller/offloadingstatus_controller_test.go
+++ b/pkg/liqo-controller-manager/offloadingStatus-controller/offloadingstatus_controller_test.go
@@ -399,5 +399,66 @@ var _ = Describe("Namespace controller", func() {
 
 		})
 
+		It(" TEST 6: Delete a NamespaceMap and check if the corresponding remote condition is deleted", func() {
+
+			// The namespace name is associated with the test number
+			namespace6Name := "namespace6"
+			By(fmt.Sprintf(" 1 - Creating the namespace '%s'", namespace6Name))
+			namespace6 := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: namespace6Name,
+				},
+			}
+			Expect(homeClient.Create(context.TODO(), namespace6)).To(BeNil())
+
+			By(" 2 - Creating the associated NamespaceOffloading")
+			namespaceOffloading6 := &offv1alpha1.NamespaceOffloading{
+				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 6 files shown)


pkg/liqo-controller-manager/offloadingStatus-controller/offloadingstatus_controller_test.go
```diff
@@ -399,5 +399,70 @@ var _ = Describe("Namespace controller", func() {
...
+			namespace6Name := "namespace6"
+			By(fmt.Sprintf(" 1 - Creating the namespace '%s'", namespace6Name))
+			namespace6 := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: namespace6Name,
+				},
+			}
+			Eventually(func() error {
+				return homeClient.Create(context.TODO(), namespace6)
+			}, timeout, interval).Should(BeNil())
```
giorio94, line 414: Would avoid the eventually, since the creation should immediately succeed

pkg/liqo-controller-manager/offloadingStatus-controller/offloadingstatus_controller_test.go
```diff
@@ -399,5 +399,70 @@ var _ = Describe("Namespace controller", func() {
...
+				},
+				Spec: offv1alpha1.NamespaceOffloadingSpec{
+					NamespaceMappingStrategy: offv1alpha1.EnforceSameNameMappingStrategyType,
+					PodOffloadingStrategy:    offv1alpha1.LocalAndRemotePodOffloadingStrategyType,
+					ClusterSelector:          corev1.NodeSelector{NodeSelectorTerms: []corev1.NodeSelectorTerm{}},
+				},
+			}
+			Eventually(func() error {
+				return homeClient.Create(context.TODO(), namespaceOffloading6)
+			}, timeout, interval).Should(BeNil())
```
giorio94, line 430: Same here

giorio94 (approved):
/lgtm

palexster (approved):

PR #755, opened 2021-07-15
Add AWS Configuration in the Helm Chart - AWS 2/3

aleoli:
# Description

This pr adds the AWS configuration in the Liqo helm chart.

# How Has This Been Tested?

- [x] On EKS cluster, to test that the new values are correctly used
- [x] Locally on KinD, to test that the non-AWS install is still working

### DIFF
--- a/deployments/liqo/templates/liqo-auth-deployment.yaml
+++ b/deployments/liqo/templates/liqo-auth-deployment.yaml
@@ -1,5 +1,6 @@
 ---
 {{- $authConfig := (merge (dict "name" "auth" "module" "discovery" "containerName" "cert-creator") .) -}}
+{{- $awsConfig := (merge (dict "name" "aws-config" "module" "aws-config") .) -}}
 
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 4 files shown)


deployments/liqo/templates/liqo-auth-deployment.yaml
```diff
@@ -67,6 +67,22 @@ spec:
           - "$(POD_NAMESPACE)"
           - "--resyncSeconds"
           - "30"
+          {{- if .Values.awsConfig.accessKeyId }}
+          - "--awsAccessKeyId"
+          - {{ .Values.awsConfig.accessKeyId | quote }}
```
giorio94, line 72: To me, it seems more standard (and easier to read) if key and value are on the same line

deployments/liqo/values.yaml
```diff
@@ -189,6 +189,18 @@ nameOverride: ""
 # -- full liqo name override
 fullnameOverride: ""
 
+# aws configuration for the local cluster and the Liqo user
```
giorio94, line 192: Would add a couple of information about what capabilities the liqo user should have (with a pointer to the doc when it will be introduced)

giorio94 (approved):

deployments/liqo/templates/liqo-auth-deployment.yaml
```diff
@@ -67,6 +67,22 @@ spec:
...
+          {{- end }}
+          {{- if .Values.awsConfig.secretAccessKey }}
+          - "--awsSecretAccessKey"
```
palexster, line 75: My 2 cents: I would avoid having a secret value exposed as a flag, but more preferably templated in a secret object dedicated to the identity.

palexster (approved):

PR #760, opened 2021-07-19
Rename advertisement to liqo controller manager

alacuku:
# Description

This PR renames the `advertisement-operator `to `liqo-controller-manager`.


build/liqo-controller-manager/Dockerfile
```diff
@@ -6,9 +6,9 @@ COPY go.mod ./go.mod
 COPY go.sum ./go.sum
```
palexster, line 6: I think this file has been dropped.

aleoli (approved):

palexster (approved):

PR #770, opened 2021-07-21
Minor fixes on Liqo network module

davidefalcone1:
This PR fixes a couple of bugs in Liqo networking. The first causes connection problems on environments that do not NAT Pod traffic, therefore it has been fixed by adding an appropriate NAT rule. The second causes problems during endpoint reflection after the Liqo network manager is rescheduled: the NatMappingInflater module do not recover correctly and therefore remote clusters appear to have no configuration. It has been fixed by recovering correctly taking advantage on existing natmappingresources.


pkg/liqonet/natmappinginflater/coverage.txt
```diff
@@ -0,0 +1,115 @@
+mode: atomic
```
alacuku, line 1: This file does not belong here. We should remove it.

aleoli (approved):

alacuku (approved):

PR #771, opened 2021-07-21
Fix Negative Offer resources values

giuse2596:
This PR adds a PodReady check to the broadcaster Pod Informer handler functions in order to avoid negative offers with negative resources values.

### DIFF
--- a/internal/resource-request-operator/broadcaster.go
+++ b/internal/resource-request-operator/broadcaster.go
@@ -2,6 +2,8 @@ package resourcerequestoperator
 
 import (
 	"context"
+	"fmt"
+	"math"
 	"sync"
 	"time"
 
@@ -12,11 +14,13 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	resourcehelper "k8s.io/kubectl/pkg/util/resource"
+	"k8s.io/kubernetes/pkg/api/v1/pod"
 
 	configv1alpha1 "github.com/liqotech/liqo/apis/config/v1alpha1"
 	"github.com/liqotech/liqo/internal/resource-request-operator/interfaces"
 	crdclient "github.com/liqotech/liqo/pkg/crdClient"
 	"github.com/liqotech/liqo/pkg/utils"
+	errorsmanagement "github.com/liqotech/liqo/pkg/utils/errorsManagement"
 	"github.com/liqotech/liqo/pkg/virtualKubelet/forge"
 )
 
@@ -35,7 +39,21 @@ type Broadcaster struct {
 	updateThresholdPercentage uint64
 }
 
-// SetupBroadcaster create the informer e run it to signal node changes updating Offers.
+// PodTransition represents a podReady condition possible transitions.
+type PodTransition uint8
+
+const (
+	// PendingToReady represents a transition from PodReady status = false to PodReady status = true.
+	PendingToReady PodTransition = iota
+	// ReadyToReady represents no change in PodReady status when status = true.
+	ReadyToReady
+	// ReadyToPending represents a transition from PodReady status = true to PodReady status = false.
+	ReadyToPending
+	// PendingToPending represents no change in PodReady status when status = false.
+	PendingToPending
+)
+
+// SetupBroadcaster initializes all Broadcaster parameters.
 func (b *Broadcaster) SetupBroadcaster(clientset kubernetes.Interface, updater interfaces.UpdaterInterface,
 	resyncPeriod time.Duration, offerUpdateThreshold uint64) error {
 	b.allocatable = corev1.ResourceList{}
@@ -54,6 +72,7 @@ func (b *Broadcaster) SetupBroadcaster(clientset kubernetes.Interface, updater i
 	b.podInformer = factory.Core().V1().Pods().Informer()
 	b.podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    b.onPodAdd,
+		UpdateFunc: b.onPodUpdate,
 		DeleteFunc: b.onPodDelete,
 	})
 
@@ -97,6 +116,22 @@ func (b *Broadcaster) GetConfig() *configv1alpha1.ClusterConfig {
 	defer b.configMutex.RUnlock()
 	configCopy := b.clusterConfig.DeepCopy()
 	return configCopy
+}
+
+func (b *Broadcaster) getPodMap() map[string]corev1.ResourceList {
+	b.podMutex.RLock()
+	defer b.podMutex.RUnlock()
+	mapCopy := make(map[string]corev1.ResourceList)
+	for key, value := range b.resourcePodMap {
+		mapCopy[key] = value.DeepCopy()
+	}
+	return mapCopy
+}
+
+func (b *Broadcaster) getLastRead(remoteClusterID string) corev1.ResourceList {
+	b.podMutex.RLock()
+	defer b.podMutex.RUnlock()
+	return b.lastReadResources[remoteClusterID]
 }
 
 // react to a Node Creation/First informer run.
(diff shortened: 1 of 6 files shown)


internal/resource-request-operator/broadcaster.go
```diff
@@ -147,36 +149,86 @@ func (b *Broadcaster) onNodeDelete(obj interface{}) {
...
+	oldResources := extractPodResources(oldPod)
 	currentResources := b.readClusterResources()
-	// this action is the complementary of the onPodAdd() one. For more details see at that function.
-	addResources(currentResources, podResources)
+	if pod.IsPodReady(newPod) {
+		// pod is already Ready, try to update resources.
+		if pod.IsPodReady(oldPod) {
+			updateResources(currentResources, oldResources, newResources)
+			if clusterID := newPod.Labels[forge.LiqoOriginClusterID]; clusterID != "" {
+				klog.V(4).Infof("OnPodUpdate: Pod %s:%s passed ClusterID check. ClusterID = %s\n", newPod.Namespace, newPod.Name, clusterID)
```
palexster, line 183: This code is duplicate.

internal/resource-request-operator/broadcaster.go
```diff
@@ -147,36 +149,86 @@ func (b *Broadcaster) onNodeDelete(obj interface{}) {
...
+				currentPodsResources := b.readPodResources(clusterID)
+				// add the resource of this pod in the map clusterID => resources to be used in ReadResources() function.
+				// this action is done to correct the computation not considering pod offloaded by the cluster with this ClusterID
+				addResources(currentPodsResources, newResources)
+				b.writePodsResources(clusterID, currentPodsResources)
+			}
+		}
+		// pod is becoming not ready, same of onPodDelete case.
+	} else if pod.IsPodReady(oldPod) && !pod.IsPodReady(newPod) {
```
palexster, line 203: This if statement is asymmetric: I would find it clearer having 3 similar conditions instead of nested ifs.

internal/resource-request-operator/broadcaster.go
```diff
@@ -147,36 +149,86 @@ func (b *Broadcaster) onNodeDelete(obj interface{}) {
...
-		b.writePodsResources(clusterID, currentPodsResources)
+}
+
+func (b *Broadcaster) onPodDelete(obj interface{}) {
+	podDeleted := obj.(*corev1.Pod)
+	klog.V(4).Infof("OnPodDelete: Delete for pod %s:%s\n", podDeleted.Namespace, podDeleted.Name)
+	if pod.IsPodReady(podDeleted) {
+		podResources := extractPodResources(podDeleted)
+		currentResources := b.readClusterResources()
+		// this action is the complementary of the onPodAdd() one. For more details see at that function.
```
palexster, line 223: Not really useful. I think you should explain what you mean and why it is important this "complementarity".

internal/resource-request-operator/broadcaster.go
```diff
@@ -147,36 +149,86 @@ func (b *Broadcaster) onNodeDelete(obj interface{}) {
...
-		b.writePodsResources(clusterID, currentPodsResources)
+	podAdded := obj.(*corev1.Pod)
+	klog.V(4).Infof("OnPodAdd: Add for pod %s:%s\n", podAdded.Namespace, podAdded.Name)
+	if pod.IsPodReady(podAdded) {
+		podResources := extractPodResources(podAdded)
+		currentResources := b.readClusterResources()
+		// subtract the pod resource from cluster resources. This action is done for all pods to extract actual available resources.
+		subResources(currentResources, podResources)
+		b.writeClusterResources(currentResources)
+		if clusterID := podAdded.Labels[forge.LiqoOriginClusterID]; clusterID != "" {
```
palexster, line 160: I think this case deserves a test

internal/resource-request-operator/broadcaster.go
```diff
@@ -246,15 +336,25 @@ func (b *Broadcaster) readPodResources(clusterID string) corev1.ResourceList {
...
+	for resourceName, resources := range clusterResources {
+		podValue, exists := podResourceValue[resourceName]
+		if !exists {
+			podValue = *resource.NewQuantity(0, "")
+		}
+		lastReadValue := lastRead[resourceName]
+		diff := (resources.Value() + podValue.Value()) - lastReadValue.Value()
+		if diff < 0 {
+			diff *= -1
+		}
```
aleoli, line 356: Can we use `Abs`?

internal/resource-request-operator/broadcaster.go
```diff
@@ -246,15 +336,25 @@ func (b *Broadcaster) readPodResources(clusterID string) corev1.ResourceList {
 	return corev1.ResourceList{}
 }
 
+func (b *Broadcaster) setThreshold(threshold uint64) {
+	b.updateThresholdPercentage = threshold
+}
+
 func (b *Broadcaster) checkThreshold(clusterID string) bool {
```
aleoli, line 343:
```suggestion
func (b *Broadcaster) isAboveThreshold(clusterID string) bool {
```

The function name/result meaning is not clear to me: is it checking if the value is above or belove the threshold?

internal/resource-request-operator/broadcaster.go
```diff
@@ -147,47 +170,104 @@ func (b *Broadcaster) onNodeDelete(obj interface{}) {
...
+			b.writePodsResources(clusterID, currentPodsResources)
+		}
 	}
 }
 
 // write nodes resources in thread safe mode.
 func (b *Broadcaster) writeClusterResources(newResources corev1.ResourceList) {
+	err := checkSign(newResources)
+	if err != nil {
+		klog.Errorf("Broadcaster -> Writing global resources error: %s", err)
```
aleoli, line 264: I think that we should do something here, or default to 0 the value, or not update the stored value. What to you think @palexster ?

internal/resource-request-operator/broadcaster.go
```diff
@@ -196,9 +276,13 @@ func (b *Broadcaster) writeClusterResources(newResources corev1.ResourceList) {
 
 // write pods resources in thread safe mode.
 func (b *Broadcaster) writePodsResources(clusterID string, newResources corev1.ResourceList) {
+	err := checkSign(newResources)
+	if err != nil {
+		klog.Errorf("Broadcaster -> Writing pod Resources error for cluster %s: %s", clusterID, err)
+	}
```
aleoli, line 282: Ditto

internal/resource-request-operator/broadcaster.go
palexster: Ok, my personal feeling is that we should notice immediately when a negative offer is generated, to avoid problems and bugs we had in the past. Since it is potentially a critical error and impossible to obtain without a bug in our implementation, I would make the broadcaster fail or something that cannot be ignored. What do you think? @aleoli @alacuku @giorio94

palexster (approved):

internal/resource-request-operator/broadcaster.go
```diff
@@ -147,59 +181,119 @@ func (b *Broadcaster) onNodeDelete(obj interface{}) {
...
+		// subtract the pod resource from cluster resources. This action is done for all pods to extract actual available resources.
+		subResources(currentResources, podResources)
+		b.writeClusterResources(currentResources)
+		if clusterID := podAdded.Labels[forge.LiqoOriginClusterID]; clusterID != "" {
+			klog.V(4).Infof("OnPodAdd: Pod %s:%s passed ClusterID check. ClusterID = %s\n", podAdded.Namespace, podAdded.Name, clusterID)
+			currentPodsResources := b.readPodResources(clusterID)
+			// add the resource of this pod in the map clusterID => resources to be used in ReadResources() function.
+			// this action is done to correct the computation not considering pod offloaded by the cluster with this ClusterID
+			addResources(currentPodsResources, podResources)
+			b.writePodsResources(clusterID, currentPodsResources)
```
giorio94, line 198: Please be consistent, one function is readPodResources, while the other is writePod**s**Resources.

internal/resource-request-operator/broadcaster.go
```diff
@@ -314,7 +422,32 @@ func updateResources(currentResources, oldResources, newResources corev1.Resourc
 	}
 }
 
-func extractPodResources(pod *corev1.Pod) corev1.ResourceList {
-	resourcesToExtract, _ := resourcehelper.PodRequestsAndLimits(pod)
+func extractPodResources(podToExtract *corev1.Pod) corev1.ResourceList {
+	resourcesToExtract, _ := resourcehelper.PodRequestsAndLimits(podToExtract)
```
giorio94, line 426: Would personally revert this change.

internal/resource-request-operator/broadcaster.go
giorio94: I am not entirely sure here. Panic'ing would probably be a sensible option to make sure errors do not get unnoticed, but it would cause the restart of the entire `liqo-controller-manager`. Additionally, it could never recover, making `liqo` unusable. I feel it could make sense very short term, but should be probably reverted to an error message (possibly defaulting to 0) before the 0.3 release.

internal/resource-request-operator/broadcaster.go
giuse2596: use pod as variable name shadows the import from pod package.

internal/resource-request-operator/broadcaster.go
```diff
@@ -35,7 +38,21 @@ type Broadcaster struct {
...
+	PendingToReady PodTransition = iota
+	// ReadyToReady represents no change in PodReady status when status = true.
+	ReadyToReady
+	// ReadyToPending represents a transition from PodReady status = true to PodReady status = false.
+	ReadyToPending
+	// PendingToPending represents no change in PodReady status when status = false.
+	PendingToPending
+)
+
+// SetupBroadcaster initialize all Broadcaster parameters .
```
giorio94, line 55:
```suggestion
// SetupBroadcaster initializes all Broadcaster parameters.
```

giorio94 (approved):
Overall, /lgtm

Before merging, it is necessary to agree on the desired behavior in case of negative offers.

PR #774, opened 2021-07-23
Fix bug causing liqo-gateway to crash when restarting

alacuku:
# Description

Issues fixed in this PR:

1. Moved back the labelOperator in the main manager to fix the problems of the leader election process causing the secondary managers to start, being them unrelated with the leader election of the main manager.
2. Some times, based on the load of the operating system, some network interfaces are not garbage collected by the kernel when their related network namespace has been removed. This caused the operator to restart when these network interfaces are still present. Now when the operator starts removes the outdated network interfaces before trying to create them again.
3. The finalizer added to the `tunnelendpoints.net.liqo.io` by the operator are related to the IP address that the operator has when the finalizer is added. When the operator is rescheduled or another replica becomes the `leader` it would add a new finalizer different from the first one. The old one would prevent the deletion of the resources. Now the finalizer of `liqo-gateway` is a constant, and at the same time, when the operator exits, tries to remove all it's finalizers from the existing resources.  

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

### DIFF
--- a/cmd/liqonet/main.go
+++ b/cmd/liqonet/main.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/containernetworking/plugins/pkg/ns"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
@@ -46,6 +47,7 @@ import (
 	liqonetns "github.com/liqotech/liqo/pkg/liqonet/netns"
 	"github.com/liqotech/liqo/pkg/liqonet/overlay"
 	liqorouting "github.com/liqotech/liqo/pkg/liqonet/routing"
+	tunnelwg "github.com/liqotech/liqo/pkg/liqonet/tunnel/wireguard"
 	"github.com/liqotech/liqo/pkg/liqonet/utils"
 	"github.com/liqotech/liqo/pkg/mapperUtils"
 )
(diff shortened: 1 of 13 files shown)


internal/liqonet/tunnel-operator/tunnel-operator.go
```diff
@@ -83,18 +86,20 @@ type TunnelController struct {
 // +kubebuilder:rbac:groups=core,namespace="do-not-care",resources=secrets,verbs=get;list;watch;create;update;patch;delete
 
 // NewTunnelController instantiates and initializes the tunnel controller.
-func NewTunnelController(podIP, namespace string, er record.EventRecorder,
-	k8sClient k8s.Interface, cl client.Client, readyClustersMutex *sync.Mutex,
-	readyClusters map[string]struct{}, gatewayNetns ns.NetNS) (*TunnelController, error) {
+func NewTunnelController(podIP, namespace string, er record.EventRecorder, k8sClient k8s.Interface, cl client.Client,
+	readyClustersMutex *sync.Mutex, readyClusters map[string]struct{}, gatewayNetns, hostNetns ns.NetNS) (*TunnelController, error) {
+	tunnelEndpointFinalizer := strings.Join([]string{liqoconst.LiqoGatewayOperatorName, "net.liqo.io"}, ".")
```
palexster, line 91: Nit. I think net.liqo.io would deserve a constant.

pkg/liqonet/utils/utils_suite_test.go
```diff
@@ -5,9 +5,16 @@ import (
...
 func TestUtils(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Utils Suite")
 }
+
+var _ = BeforeSuite(func() {
+	// Create dummy link.
+	err := netlink.LinkAdd(&netlink.Dummy{LinkAttrs: netlink.LinkAttrs{Name: interfaceName}})
+	Expect(err).ShouldNot(HaveOccurred())
+})
```
giorio94, line 20: Nit. Could be moved as BeforeEach of the "when network interface exists" context if used only there.

pkg/liqonet/utils/utils_test.go
```diff
@@ -317,4 +318,20 @@ var _ = Describe("Liqonet", func() {
 			})
 		})
 	})
+
+	Describe("testing DeleteIfaceByName function", func() {
+		Context("when network interface exists", func() {
+			It("should return nil", func() {
+				err := utils.DeleteIFaceByName(interfaceName)
+				Expect(err).Should(BeNil())
```
giorio94, line 326: Could add a check that now it does no longer exist

cmd/liqonet/main.go
```diff
@@ -308,10 +303,6 @@ func main() {
 			os.Exit(1)
 		}
 
-		if err := mainMgr.Add(labelMgr); err != nil {
-			klog.Errorf("unable to add the label manager to the main manager: %s", err)
-			os.Exit(1)
-		}
 		klog.Info("Starting manager as Tunnel-Operator")
 		if err := mainMgr.Start(tunnelController.SetupSignalHandlerForTunnelOperator()); err != nil {
```
giorio94, line 307: Nit. Could probably revert mainMgr to mgr

internal/liqonet/tunnel-operator/tunnel-operator.go
```diff
@@ -413,9 +417,8 @@ func (tc *TunnelController) SetupSignalHandlerForTunnelOperator() context.Contex
 	go func(tc *TunnelController) {
 		sig := <-c
 		klog.Infof("the operator received signal {%s}: cleaning up", sig.String())
-		if err := tc.CleanUpConfiguration(liqoconst.GatewayNetnsName, liqoconst.HostVethName); err != nil {
-			klog.Errorf("an error occurred while cleaning up namespace {%s} with path {%s}: %v", liqoconst.GatewayNetnsName, tc.gatewayNetns.Path(), err)
-		}
+		// Here, the error is no checked, at exit time is not possible to recover. Errors are just logged.
```
giorio94, line 420:
```suggestion
		// Here, the error is not checked, as at exit time is not possible to recover. Errors are just logged.
```

giorio94 (approved):
/lgtm

palexster (approved):

PR #784, opened 2021-07-28
Fix Uninstaller

aleoli:
# Description

This pr fixes an issue that prevents the Liqo full uninstallation

# How Has This Been Tested?

- [x] Locally on KinD

### DIFF
--- a/pkg/uninstaller/deletion.go
+++ b/pkg/uninstaller/deletion.go
@@ -25,7 +25,7 @@ func UnjoinClusters(ctx context.Context, client dynamic.Interface) error {
 	klog.Infof("Unjoin %v ForeignClusters", len(foreign.Items))
 	r1 := client.Resource(discoveryV1alpha1.ForeignClusterGroupVersionResource)
 	for index := range foreign.Items {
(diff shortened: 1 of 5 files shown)


palexster (approved):

pkg/uninstaller/deletion.go
```diff
@@ -25,7 +25,7 @@ func UnjoinClusters(ctx context.Context, client dynamic.Interface) error {
 	klog.Infof("Unjoin %v ForeignClusters", len(foreign.Items))
 	r1 := client.Resource(discoveryV1alpha1.ForeignClusterGroupVersionResource)
 	for index := range foreign.Items {
-		patch := []byte(`{"spec": {"join": false}}`)
+		patch := []byte(`{"spec": {"outgoingPeeringEnabled": "No"}}`)
```
giorio94, line 28: If possible, it would probably better to avoid manually forging the json patch, as lacking any type check to prevent the occurrence of unnoticed issues in case the API definition gets updated.

giorio94 (approved):

PR #789, opened 2021-07-28
Get ApiServer Address from full URL

aleoli:
# Description

This pr improves the AuthService that now can parse the local API Server address and port from the full URL. It is no more required to set the two different values in the `values.yaml` file

# How Has This Been Tested?

- [x] Locally on KinD

### DIFF
--- a/deployments/liqo/values.yaml
+++ b/deployments/liqo/values.yaml
@@ -9,8 +9,6 @@ pullPolicy: "IfNotPresent"
 apiServer:
   # -- The address that must be used to contact your API server, it needs to be reachable from the clusters that you will peer with (defaults to your master IP)
   address: ""
-  # -- The port that must be used to contact your API server
-  port: "6443"
   # -- Indicates that the API Server is exposing a certificate issued by a trusted Certification Authority
   trustedCA: false
 
--- a/pkg/kubeconfig/create.go
+++ b/pkg/kubeconfig/create.go
@@ -2,8 +2,8 @@ package kubeconfig
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -13,7 +13,6 @@ import (
 	"k8s.io/klog"
 	kubeconfigutil "k8s.io/kubernetes/cmd/kubeadm/app/util/kubeconfig"
 
-	"github.com/liqotech/liqo/pkg/discovery"
 	"github.com/liqotech/liqo/pkg/utils"
 )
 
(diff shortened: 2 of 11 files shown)


deployments/liqo/values.yaml
```diff
@@ -9,8 +9,8 @@ pullPolicy: "IfNotPresent"
 apiServer:
   # -- The address that must be used to contact your API server, it needs to be reachable from the clusters that you will peer with (defaults to your master IP)
   address: ""
-  # -- The port that must be used to contact your API server
-  port: "6443"
+  # -- The port that must be used to contact your API server (if not yet specified in the address field)
+  port: ""
```
giorio94, line 13: From the user perspective, I would personally keep only the address field, defaulting to the master IP address and port 6443 in the other cases. Having both parameters makes it unclear what to put in the port one and might lead to configuration errors. 

That would also simplify the logic below. If the variable is set you just use that (possibly prepending https://), otherwise you gather the master IP and construct the value.

pkg/utils/address_test.go
```diff
@@ -0,0 +1,61 @@
...
+		Entry("no protocol and no port", addressTestcase{
+			address:      "localhost",
+			defaultPort:  "6443",
+			expectedHost: "localhost",
+			expectedPort: "6443",
+		}),
+
+		Entry("protocol and no port", addressTestcase{
+			address:      "https://localhost",
+			defaultPort:  "6443",
```
palexster, line 41: This is a bit tricky to understand for me, since the https default is 443.

pkg/kubeconfig/create.go
```diff
@@ -56,7 +56,9 @@ func CreateKubeConfigFromServiceAccount(apiServerConfigProvider utils.ApiServerC
 // 1. from the ClusterConfig
 // 2. defaults to 6443.
 func GetApiServerURL(apiServerConfigProvider utils.ApiServerConfigProvider, clientset kubernetes.Interface) (string, error) {
-	address := apiServerConfigProvider.GetAPIServerConfig().Address
+	config := apiServerConfigProvider.GetAPIServerConfig()
+	var port string
+	address := utils.GetHost(config.Address)
```
palexster, line 61: It's not part of the PR but maybe the following logic to extract the fallback address can be encapsulated in a dedicated method for sake of clarity.

palexster (approved):

pkg/kubeconfig/create.go
```diff
@@ -56,30 +54,19 @@ func CreateKubeConfigFromServiceAccount(apiServerConfigProvider utils.ApiServerC
...
```
giorio94, line 59: Would personally simplify this logic. If the address is not empty you just use that (without splitting and recomposing address and port), otherwise you call `utils.GetAPIServerAddressFromMasterNode` and use the address returned (already composed of address and port). In the end, you prepend `https://` if not already present (adding port 443 instead should not be needed, as that is the default).
Doing so, you should be also able to drop the GetAddress and GetPort methods.

giorio94 (approved):

PR #790, opened 2021-07-29
Fix bug which set zero all resources.

giuse2596:
This PR fixes a bug that sets all resources to zero value in the wrong way because of a wrong check of Must function return value.

### DIFF
--- a/internal/resource-request-operator/broadcaster.go
+++ b/internal/resource-request-operator/broadcaster.go
@@ -286,7 +286,7 @@ func (b *Broadcaster) writePodResources(clusterID string, newResources corev1.Re
 	if clusterID == "" {
 		return
 	}
-	if errorsmanagement.Must(checkSign(newResources)) {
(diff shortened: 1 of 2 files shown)


internal/resource-request-operator/broadcaster.go
```diff
@@ -286,7 +286,7 @@ func (b *Broadcaster) writePodResources(clusterID string, newResources corev1.Re
 	if clusterID == "" {
 		return
 	}
-	if errorsmanagement.Must(checkSign(newResources)) {
+	if !errorsmanagement.Must(checkSign(newResources)) {
```
giorio94, line 289: Should be covered by a test case.

giorio94 (approved):

internal/resource-request-operator/broadcaster.go
giorio94: Postponing to a subsequent PR

palexster (approved):

PR #796, opened 2021-08-03
delete virtual-kubelet tests

filippoprojetto:
Delete virtual-kubelet tests leftover from virtual-kubelet go module.


pkg/utils/testUtils/kubernetes.go
```diff
@@ -1,4 +1,4 @@
-package util
+package testutils
```
giorio94, line 1: Would change the folder name to `testutils` for consistency. Could also be `testutil` to avoid having to rename the references below.

giorio94 (approved):

palexster (approved):

PR #801, opened 2021-08-05
Persist VirtualKubelet Keys

aleoli:
# Description

This pr includes the logic to store in a secret the key and the certificate created for a virtual node. In the next VirtualKubelet runs this certificate has not to be created again but it will be restored from the secret.

# How Has This Been Tested?

- [x] Add Unit Test
- [x] Locally on KinD

### DIFF
--- a/pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller.go
+++ b/pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller.go
@@ -73,6 +73,7 @@ type ResourceOfferReconciler struct {
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=capsule.clastix.io,resources=tenants,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=get;list;watch;create;update;patch;delete;deletecollection
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;create;update
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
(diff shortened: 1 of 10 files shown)


palexster (approved):

pkg/vkMachinery/csr/secret.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// createCSRSecret creates the CSR secret with the given key and csr.
+func createCSRSecret(ctx context.Context,
+	clientset kubernetes.Interface, privateKey, csr []byte,
+	nodeName, namespace string) error {
+	secret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nodeName,
+			Namespace: namespace,
```
giorio94, line 87: Nit. Would add a label to identify the resource.

pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller.go
```diff
@@ -73,6 +73,7 @@ type ResourceOfferReconciler struct {
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=capsule.clastix.io,resources=tenants,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=get;list;watch;create;update;patch;delete;deletecollection
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;create;update
```
giorio94, line 76: Are these permissions required?

giorio94 (approved):

pkg/liqo-controller-manager/resourceoffer-controller/resourceoffer_controller.go
aleoli: Yes, it needs them to grant permissions to the virtualKubelet init container

giorio94: Forgot that the controller manager is the one giving the permissions to the virtual-kubelet

PR #802, opened 2021-08-06
Authentication Service Error Handling Improvement

aleoli:
# Description

This pr includes improvements for the Liqo Auth Service:
* logging improved
* return different HTTP errors code based on the golang errors
* change `allowEmptyToken` to `enableAuthentication`
* removed old authentication methods and permissions

# How Has This Been Tested?

- [x] locally on KinD
- [x] add unit test

PR #803, opened 2021-08-06
Remove dependency from k8s.io/kubernetes and k8s.io/klog (v1)

giorio94:
# Description

This PR removes the dependency from the k8s.io/kubernetes repository, since it is currently problematic (it requires many rewrites), making it also problematic to import liqo from external projects. Yet, it was only used for a very few utility functions, which have either been reimplemented or copied verbatim (the same approach was adopted in the virtual-kubelet repo for those functions). Finally, it converts all klog dependencies to v2.

# How Has This Been Tested?

- [ ] Unit tests
- [ ] E2E tests


palexster (approved):

palexster:
Maybe I would add a file (maybe a doc.go) in /internal/kubernetes/ to explain the purpose of this package.

alacuku (approved):

pkg/kubeconfig/create.go
```diff
@@ -10,9 +10,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	clientcmdlatest "k8s.io/client-go/tools/clientcmd/api/latest"
-	"k8s.io/klog"
-	kubeconfigutil "k8s.io/kubernetes/cmd/kubeadm/app/util/kubeconfig"
+	"k8s.io/klog/v2"
 
+	kubeconfigutil "github.com/liqotech/liqo/internal/kubernetes/kubeconfig"
```
giorio94, line 15: @aleoli @palexster

Is the `CreateKubeConfigFromServiceAccount` function below still used after #802? It seems to me it is no longer referenced. In that case, it could be dropped, and also the backport of this package could be removed.

palexster: I don't think it is used anymore, but waiting for @aleoli for confirmation.

aleoli: we can drop it

PR #805, opened 2021-08-09
Add status check for component building

palexster:
# Description

This PR enables custom checks to report pipeline status for Unit Testing and Component Builds.

# How Has This Been Tested?

- [x] E2E
- [x] Manual Testing

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -54,6 +54,20 @@ jobs:
            echo "::set-output name=repo-name::${{ github.event.client_payload.github.payload.repository.full_name }}"
         fi
 
+      # Since we are using a repository-dispatch event, we have to explicitly set a run check. We initialize it to a "pending" state.
+    - uses: octokit/request-action@v2.x
+      name: "Initialize run check to 'pending (For PR-only)"
+      with:
+        route: POST /repos/${{ github.repository }}/statuses/${{ steps.configure.outputs.commit-ref }}
+        state: "pending"
+        description: "Component build status"
+        context:  "Components building"
+        target_url: "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+      env:
+        GITHUB_TOKEN: ${{ secrets.CI_TOKEN }}
+      if: ${{ github.event_name == 'repository_dispatch' }}
+
+
   build:
     runs-on: ubuntu-latest
     needs: configure
@@ -73,6 +87,7 @@ jobs:
         - virtual-kubelet
         - webhook-configuration
     steps:
+
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1.0.1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/integration.yml
```diff
@@ -54,6 +54,19 @@ jobs:
            echo "::set-output name=repo-name::${{ github.event.client_payload.github.payload.repository.full_name }}"
         fi
 
+      # Since we are using a repository-dispatch event, we have to explicitly set a run check. We initialize it to a "pending" state.
+    - uses: octokit/request-action@v2.x
+      id: create_check_run
```
giorio94, line 59: The ID could be probably dropped.

.github/workflows/integration.yml
```diff
@@ -136,8 +150,21 @@ jobs:
...
+         name: "Initialize run check to 'pending (For PR-only)"
+         with:
+           route: POST /repos/${{ github.repository }}/statuses/${{ needs.configure.outputs.commit-ref }}
+           state: "${{ needs.build.result }}"
+           description: "Component build status"
+           context: "Components building"
+           target_url: "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+         env:
+           GITHUB_TOKEN: ${{ secrets.CI_TOKEN }}
+         if: always() && ${{ needs.build.result != 'cancelled' }}
```
aleoli, line 166: Why do we need these `always`?

aleoli (approved):

.github/workflows/integration.yml
palexster: Because they should be executed even if the pipeline has failed (it won't without `always()`).

.github/workflows/integration.yml
```diff
@@ -54,6 +54,19 @@ jobs:
...
+      id: create_check_run
+      name: "Initialize run check to 'pending (For PR-only)"
+      with:
+        route: POST /repos/${{ github.repository }}/statuses/${{ steps.configure.outputs.commit-ref }}
+        state: "pending"
+        description: "Component build status"
+        context:  "Components building"
+        target_url: "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+      env:
+        GITHUB_TOKEN: ${{ secrets.CI_TOKEN }}
```
giorio94, line 68: Should be probably executed only if the event is repository-dispatch. Same for the other occurrences.

.github/workflows/integration.yml
```diff
@@ -136,8 +150,21 @@ jobs:
   e2e-test-trigger:
      runs-on: ubuntu-latest
      needs: [build, configure]
-     if: github.event.client_payload.slash_command.command == 'test'
      steps:
+
+       # Since we are using a repository-dispatch event, we have to explicitly set a run check. We initialize it to a "pending" state.
```
giorio94, line 155: The comment should be updated. Would personally move the status update to a separate job (e.g. post-build), instead of adding it here and changing the execution conditions (and semantic).

.github/workflows/test.yml
```diff
@@ -35,6 +35,20 @@ jobs:
         fi
 
 
+    # Since we are using a repository-dispatch event, we have to explicitly set a run check. We initialize it to a "pending" state.
+    - uses: octokit/request-action@v2.x
+      id: create_check_run
```
giorio94, line 40: Could be dropped

.github/workflows/test.yml
```diff
@@ -95,14 +109,28 @@ jobs:
...
+        GITHUB_TOKEN: ${{ secrets.CI_TOKEN }}
+      if: always() && ${{ needs.build.result != 'cancelled' }}
+
     - name: Send coverage
       uses: shogo82148/actions-goveralls@v1.4.3
       with:
-        path-to-profile: coverage.txt
\ No newline at end of file
+        path-to-profile: coverage.txt
+        working-directory: "${{ github.workspace }}/liqo"
```
giorio94, line 136: Is this modification required?

.github/workflows/test.yml
```diff
@@ -95,14 +109,28 @@ jobs:
...
+      name: "Update run check status"
+      with:
+        route: POST /repos/${{ github.repository }}/statuses/${{ needs.configure.outputs.commit-ref }}
+        state: "${{ job.status }}"
+        description: "Unit Test status"
+        context: "Unit Test"
+        target_url: "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+      env:
+        GITHUB_TOKEN: ${{ secrets.CI_TOKEN }}
+      if: always() && ${{ needs.build.result != 'cancelled' }}
```
giorio94, line 130: The repository-dispatch condition is missing.

giorio94 (approved):
/lgtm

PR #808, opened 2021-08-11
Init liqoctl command

palexster:
# Description

This PR adds the first skeleton of the liqoctl tool.

# How Has This Been Tested?

- [x] UnitTests


pkg/liqoctl/install/interface.go
```diff
@@ -0,0 +1,20 @@
...
+	"k8s.io/client-go/rest"
+)
+
+// ProviderInterface defines the methods required to support the Liqo install for a given provider.
+type ProviderInterface interface {
+	// ValidateCommandArguments validates the flags passed as arguments to the install command
+	ValidateCommandArguments(*flag.FlagSet) error
+	// ExtractChartParameters retrieves the install parameters required for a correct installation. This may require to
+	// instantiate extra clients to interact with cloud provider or the target cluster.
+	ExtractChartParameters(context.Context, *rest.Config) (string, error)
```
aleoli, line 16:
```suggestion
	ExtractChartParameters(context.Context, *rest.Config) error
```

Since it is not used

pkg/liqoctl/install/kubeadmprovider/provider.go
```diff
@@ -0,0 +1,85 @@
...
+	flag "github.com/spf13/pflag"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+)
+
+// NewKubeadmProviderConstructor initializes a new KubeadmProvider struct.
+func NewKubeadmProviderConstructor() *KubeadmProvider {
```
aleoli, line 16:
```suggestion
func NewKubeadmProviderConstructor() ProviderInterface {
```

better to return the interface instead of the specific implementation, since it is simply accessible from other packages (it is exported) and we don't need implementation details

my only doubt is that it can cause some kind of circular import, it may be useful to define the interface in a separate package

aleoli (approved):

pkg/liqoctl/install/helm/init.go
```diff
@@ -0,0 +1,47 @@
...
+
+// InitializeHelmClientWithRepo initiliazes an helm client for a given *rest.Config and adds the Liqo repository.
+func InitializeHelmClientWithRepo(config *rest.Config) (*helm.HelmClient, error) {
+	opt := &helm.RestConfClientOptions{
+		Options: &helm.Options{
+			Namespace:        LiqoNamespace,
+			RepositoryConfig: liqoHelmConfigPath,
+			RepositoryCache:  liqoHelmCachePath,
+			Debug:            true,
+			Linting:          true,
```
giorio94, line 17: Is it desired those two are set to true?

pkg/liqoctl/install/helm/init.go
```diff
@@ -0,0 +1,47 @@
...
+
+func initLiqoRepo(helmClient helm.Client) {
+	// Define a public chart repository
+	chartRepo := repo.Entry{
+		Name: "liqo",
+		URL:  liqoRepo,
+	}
+
+	if err := helmClient.AddOrUpdateChartRepo(chartRepo); err != nil {
+		panic(err)
```
giorio94, line 41: As this will be seen by a user it is probably better to present an error message instead of panic'ing.

pkg/liqoctl/install/kubeadm/kubeadm_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+var ns = &corev1.Namespace{
+	ObjectMeta: metav1.ObjectMeta{
+		Name: "kube-system",
+	},
+}
+
+var _ = Describe("Extract elements from apiServer", func() {
+
+	It("Retrieve parameters from kube-controller-manager pod", func() {
```
giorio94, line 55: Would personally suggest to use the fakeclient instead of creating the entire test environment, as much faster and perfectly suitable for this simple use-case.

pkg/liqoctl/install/kubeadm/provider.go
```diff
@@ -0,0 +1,87 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+
+	"github.com/liqotech/liqo/pkg/liqoctl/install/provider"
+)
+
+// NewKubeadmProviderConstructor initializes a new Kubeadm struct.
+func NewKubeadmProviderConstructor() provider.InstallProviderInterface {
```
giorio94, line 18:
```suggestion
func NewProvider() provider.InstallProviderInterface {
```

Just a proposal though (kubeadm is already implied by the package name)

pkg/liqoctl/install/utils.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	helm "github.com/mittwald/go-helm-client"
+	"gopkg.in/yaml.v3"
+	"helm.sh/helm/v3/pkg/action"
+	"k8s.io/client-go/rest"
+
+	helm2 "github.com/liqotech/liqo/pkg/liqoctl/install/helm"
```
giorio94, line 14: Should be assigned a better alias name.

pkg/liqoctl/install/utils.go
```diff
@@ -0,0 +1,75 @@
...
+	kubeconfigPath, ok := os.LookupEnv("KUBECONFIG")
+	if !ok {
+		kubeconfigPath = filepath.Join(os.Getenv("HOME"), ".kube", "config")
+	}
+
+	config, err := utils.UserConfig(kubeconfigPath)
+	if err != nil {
+		fmt.Printf("Unable to create client config: %s", err)
+		return nil, nil, err
+	}
```
giorio94, line 37: Would separate the rest config retrieval from the helm sub-package, as unrelated

pkg/liqoctl/install/utils.go
```diff
@@ -0,0 +1,75 @@
...
+
+	k.UpdateChartValues(output.Values)
+
+	raw, err := yaml.Marshal(&output.Values)
+	if err != nil {
+		return err
+	}
+	chartSpec := helm.ChartSpec{
+		// todo (palexster): Check if it ReleaseName and LiqoNamespace are really configurable
```
giorio94, line 59: Does this TODO still apply?

pkg/liqoctl/install/handler.go
```diff
@@ -0,0 +1,50 @@
...
+
+// HandleInstallCommand implements the install command. It detects which provider has to be used, generates the chart
+// with provider-specific values. Finally, it performs the installation on the target cluster.
+func HandleInstallCommand(cmd *cobra.Command, args []string) {
+	helmClient, config, err := initHelmClient()
+	if err != nil {
+		fmt.Printf("Unable to create a client to the target cluster: %s", err)
+		return
+	}
+	ctx := context.Background()
```
giorio94, line 21: The context should be created in the main, configuring the appropriate exit handlers, and then propagated.

giorio94 (approved):

pkg/liqoctl/install/utils.go
palexster: Yes, unfortunately we never tested Liqo in a namespace that is not `liqo`. Liqo 0.2.X surely did not work.

pkg/liqoctl/install/handler.go
palexster: Since it requires a bit of refactoring, I propose to solve this in a dedicated PR.

PR #810, opened 2021-08-11
Add AWS Provider in Liqoctl

aleoli:
# Description

This pr adds a new provider to allow the liqoctl to retrieve parameters from EKS clusters

# How Has This Been Tested?

- [x] Manually
- [x] Add new unit test

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,11 +3,12 @@ module github.com/liqotech/liqo
 go 1.16
 
 require (
+	cloud.google.com/go v0.84.0 // indirect
 	github.com/aws/aws-sdk-go v1.39.4
 	github.com/clastix/capsule v0.1.0-rc2
 	github.com/containernetworking/plugins v0.8.6
 	github.com/coreos/go-iptables v0.4.5
-	github.com/google/go-cmp v0.5.5
+	github.com/google/go-cmp v0.5.6
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/uuid v1.2.0
 	github.com/gorilla/mux v1.8.0 // indirect
@@ -33,8 +34,10 @@ require (
 	go.opencensus.io v0.23.0
 	golang.org/x/oauth2 v0.0.0-20210628180205-a41e5a781914 // indirect
 	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
+	golang.org/x/tools v0.1.4 // indirect
 	golang.zx2c4.com/wireguard/wgctrl v0.0.0-20200609130330-bd2cb7843e1b
 	gomodules.xyz/jsonpatch/v2 v2.1.0
+	google.golang.org/genproto v0.0.0-20210624195500-8bfb893ecb84 // indirect
 	google.golang.org/grpc v1.38.0
 	google.golang.org/protobuf v1.26.0
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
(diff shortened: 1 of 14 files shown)


go.mod
```diff
@@ -24,17 +26,20 @@ require (
 	github.com/onsi/gomega v1.13.0
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cast v1.4.0 // indirect
-	github.com/spf13/cobra v1.2.1
```
palexster, line 27: Do we need this downgrade?

aleoli: Probably not, probably it has been downgraded when i was chacking gcloud sdk versions. I have to check, thanks

palexster (approved):

pkg/liqoctl/install/eks/provider.go
```diff
@@ -0,0 +1,386 @@
...
+
+	subFlag.String("access-key-id", "", "The IAM accessKeyID for the Liqo user (optional)")
+	subFlag.String("secret-access-key", "", "The IAM secretAccessKey for the Liqo user (optional)")
+
+	flags.AddFlagSet(subFlag)
+}
+
+// getClusterInfo retrieved information from the EKS cluster.
+func (k *eksProvider) getClusterInfo(sess *session.Session) error {
+	eksSvc := eks.New(sess, aws.NewConfig().WithRegion(k.region))
```
palexster, line 166: Maybe the code to scrape the flags can be generalized.

pkg/liqoctl/install/eks/provider.go
```diff
@@ -0,0 +1,386 @@
+package eks
```
palexster, line 1: Nit. Since the file is long, maybe the util functions can be stored.

pkg/liqoctl/install/eks/provider.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	if err = k.parseClusterOutput(describeClusterResult); err != nil {
+		return err
+	}
+
+	if describeClusterResult.Cluster.ResourcesVpcConfig.VpcId == nil {
+		err := fmt.Errorf("the EKS cluster %v in region %v does not have a valid VPC ID", k.clusterName, k.region)
+		return err
+	}
```
giorio94, line 184: Could be moved to the `parseClusterOutput` function for consistency.

giorio94 (approved):
/lgtm

pkg/liqoctl/install/eks/provider.go
```diff
@@ -0,0 +1,386 @@
...
+// ValidateCommandArguments validates specific arguments passed to the install command.
+func (k *eksProvider) ValidateCommandArguments(flags *flag.FlagSet) (err error) {
+	k.region, err = flags.GetString(prefixedName("region"))
+	if err != nil {
+		return err
+	}
+	if k.region == "" {
+		err := fmt.Errorf("--eks.region not provided")
+		return err
+	}
```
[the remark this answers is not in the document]
aleoli: It would be nice, but the problem is that they are required only if the provider is set to `eks`

To make it possible to use the required check we have to move each provider in a different subcommand, for example:
```bash
liqoctl install eks ... [flags]
```
instead of
```bash
liqoctl install --provider eks ... [flags]
```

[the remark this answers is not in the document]
giorio94: Got it. Personally, I would go that way in the long term, as it also allows to better divide the flags.

PR #817, opened 2021-08-13
Add testing option to /rebase

palexster:
# Description

This PR introduces the possibility to automatically trigger the testing pipelines after a rebase by adding a `trigger-test=true` option in the /rebase comment. The PR fixes also a wrong condition in case of failure for the unit test pipeline.

# How Has This Been Tested?

- [x] Manually Tested


aleoli (approved):

.github/workflows/test.yml
```diff
@@ -126,9 +126,9 @@ jobs:
         target_url: "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
       env:
         GITHUB_TOKEN: ${{ secrets.CI_TOKEN }}
-      if: always() && ${{ needs.build.result != 'cancelled' }} && ${{ github.event_name == 'repository_dispatch' }}
+      if: ${{ !cancelled() && github.event_name == 'repository_dispatch' }}
```
giorio94, line 129: Was wandering whether !cancelled works even in case of failures

giorio94 (approved):

.github/workflows/test.yml
palexster: Yes, in my tests the step is executed even in presence of a previously failed step. Ex: https://github.com/palexster/liqo/runs/3328516435?check_suite_focus=true

PR #818, opened 2021-08-13
Use namespace lister in tenant namespace manager

giorio94:
# Description

This PR converts the tenant namespace manager to use a lister instead of performing direct get operations, to reduce the number of API requests and improve the overall performance. Finally, it adds a predicate to the foreign cluster controller to prevent its execution in case only the status changed.

**Warning**: this is a bit opinionated, since it introduces the possibility for race conditions if a namespace if first created and then immediately retrieved (since it might not have already been cached). IMHO, this should not be a problem, since the real controllers either create the namespace or retrieve it (and even in case of race condition, they would return an error and immediately retry). Differently, this creates problems in the tests, which have been fixed adding an `Eventually` clause to ensure the cache was populated. @liqotech/maintainers, if you do not like this modification feel free to close the PR.

# How Has This Been Tested?

- [x] Unit tests (existing)
- [x] E2E tests (existing)

### DIFF
--- a/deployments/liqo/files/liqo-crd-replicator-ClusterRole.yaml
+++ b/deployments/liqo/files/liqo-crd-replicator-ClusterRole.yaml
@@ -10,6 +10,14 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - namespaces
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - secrets
   verbs:
   - get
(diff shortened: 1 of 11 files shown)


palexster:
/lgtm 
IMHO The race condition does not represent a problem, since the reconciliation will eventually converge, and the transitory should always be short.

palexster (approved):

aleoli (approved):

PR #821, opened 2021-08-13
Add AKS Provider in Liqoctl

aleoli:
# Description

This pr adds a new provider to allow the liqoctl to retrieve parameters from AKS clusters

# How Has This Been Tested?

- [x] Manually
- [x] Add new unit test


pkg/liqoctl/install/aks/provider.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	k.podCIDR = *vnet.AddressPrefix
+	k.serviceCIDR = *cluster.ManagedClusterProperties.NetworkProfile.ServiceCidr
+
+	return nil
+}
+
+func parseSubnetID(subnetID string) (vnetName, subnetName string, err error) {
+	strs := strings.Split(subnetID, "/")
```
palexster, line 180: I think this may require some commentary and may be a util method to be less obscure.

palexster (approved):

pkg/liqoctl/install/aks/provider.go
```diff
@@ -0,0 +1,219 @@
...
+
+// parseSubnetID parses an Azure subnetID returning its vnet name and its subnet name.
+// For example, if the subnetID is
+// "/subscriptions/<YOUR_SUBSCRIPTION>/resourceGroups/test/providers/Microsoft.Network/virtualNetworks/testvnet663/subnets/default"
+// it will return "testvnet663" as vnet name and "default" as subnet name.
+func parseSubnetID(subnetID string) (vnetName, subnetName string, err error) {
+	strs := strings.Split(subnetID, "/")
+	l := len(strs)
+
+	if l < 3 {
```
giorio94, line 187: In case the subnetID has always the format of the comment, it could be worth to check that the length is exactly 10 to be on the safe side.

giorio94 (approved):
/lgtm

PR #828, opened 2021-08-16
Add K3S Provider in Liqoctl

aleoli:
# Description

This pr adds a new provider to allow the liqoctl to retrieve parameters from K3S clusters

# How Has This Been Tested?

- [x] Manually
- [x] Add new unit test

### DIFF
--- a/pkg/liqoctl/install/k3s/provider.go
+++ b/pkg/liqoctl/install/k3s/provider.go
@@ -1,0 +1,204 @@
+package k3s
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/url"
+
+	flag "github.com/spf13/pflag"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+
+	"github.com/liqotech/liqo/pkg/consts"
+	"github.com/liqotech/liqo/pkg/liqoctl/install/provider"
+	installutils "github.com/liqotech/liqo/pkg/liqoctl/install/utils"
+)
+
+const (
+	providerPrefix = "k3s"
+
+	defaultPodCIDR     = "10.42.0.0/16"
+	defaultServiceCIDR = "10.43.0.0/16"
+)
+
+var (
+	localhostValues = []string{
+		"localhost",
+		"127.0.0.1",
+	}
+)
+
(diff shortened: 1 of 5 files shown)


pkg/liqoctl/install/k3s/provider.go
```diff
@@ -0,0 +1,192 @@
...
+	})
+
+	subFlag.String("pod-cidr", defaultPodCIDR, "The Pod CIDR for your cluster (optional)")
+	subFlag.String("service-cidr", defaultServiceCIDR, "The Service CIDR for your cluster (optional)")
+	subFlag.String("api-server", "", "Your cluster API Server URL (optional)")
+
+	flags.AddFlagSet(subFlag)
+}
+
+func (k *k3sProvider) validateServiceCIDR(ctx context.Context) error {
```
palexster, line 132: I would add a few lines to specify why are we doing this check.

pkg/liqoctl/install/k3s/provider.go
```diff
@@ -0,0 +1,192 @@
...
+		clusterIP := svc.Spec.ClusterIP
+		if clusterIP != "None" && clusterIP != "" && !svcNet.Contains(net.ParseIP(clusterIP)) {
+			return fmt.Errorf("the CIDR %v does not contain IP %v", k.serviceCIDR, clusterIP)
+		}
+	}
+
+	return nil
+}
+
+func (k *k3sProvider) validatePodCIDR(ctx context.Context) error {
```
palexster, line 154: Ditto.

palexster (approved):

pkg/liqoctl/install/k3s/provider.go
```diff
@@ -0,0 +1,192 @@
...
+	_, svcNet, err := net.ParseCIDR(k.serviceCIDR)
+	if err != nil {
+		return err
+	}
+
+	for i := range svcList.Items {
+		svc := &svcList.Items[i]
+		clusterIP := svc.Spec.ClusterIP
+		if clusterIP != "None" && clusterIP != "" && !svcNet.Contains(net.ParseIP(clusterIP)) {
+			return fmt.Errorf("the CIDR %v does not contain IP %v", k.serviceCIDR, clusterIP)
```
giorio94, line 147: Would smooth a bit this error, since it will be read by the users. Something along the lines of "It seems the specified service CIDR is not correct as it does not match the services in your cluster", with an higher verbosity message about the actual reason (name of service and IP). Just a proposal though (the same for the pod CIDR)

giorio94 (approved):

PR #834, opened 2021-08-18
Incoming Peering Management

aleoli:
# Description

This pr allows a cluster to explicitly allow/deny an incoming peering from a remote cluster

* a new field has been added in the ClusterConfig to set the default behavior with the newly discovered clusters
* a new field has been added in the ForeignCluster to override the default behavior and to set the incoming policy for a specific remote cluster

# How Has This Been Tested?

- [x] locally on KinD
- [x] add unit test


internal/discovery/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -415,12 +418,24 @@ func getPeeringPhase(foreignCluster *discoveryv1alpha1.ForeignCluster,
 		resourceRequest := &resourceRequestList.Items[0]
 		desiredDelete := !resourceRequest.Spec.WithdrawalTimestamp.IsZero()
 		deleted := !resourceRequest.Status.OfferWithdrawalTimestamp.IsZero()
-		if desiredDelete || deleted {
-			return discoveryv1alpha1.PeeringConditionStatusDisconnecting, resourceRequestDeletingReason,
-				fmt.Sprintf(resourceRequestDeletingMessage, foreignCluster.Status.TenantNamespace.Local), nil
+		offerState := resourceRequest.Status.OfferState
+
+		switch offerState {
```
palexster, line 423: Some comments would be helpful here to understand each different case

palexster (approved):

alacuku (approved):

PR #840, opened 2021-08-23
Broker PoC

giuse2596:
This PR add new Broker component based on the standard Broadcaster. This version just replicates the first ResourceOffer it knows to all cluster that start an incoming peering.


CapacitorSet:
The development was moved to https://github.com/CapacitorSet/liqo-broker/.

PR #853, opened 2021-08-30
Add support for vpn endpoint addresses when given as DNS addresses

alacuku:
# Description

This PR adds support for DNS addresses for the vpn endpoints. Now the `tunnelendpoints.spec.endpointIP` could be a literal ipv4/ipv6 IP address or a DNS one.

# How Has This Been Tested?

- [x] Unit tests

### DIFF
--- a/pkg/liqonet/tunnel/wireguard/driver.go
+++ b/pkg/liqonet/tunnel/wireguard/driver.go
@@ -52,6 +52,9 @@ const (
 	KeepAliveInterval = 10 * time.Second
 	// MTU size of mtu for wireguard interface.
 	MTU = 1415
+	// Ports 1-65535 are available.
+	udpMinPort = 1
+	udpMaxPort = 65535
 )
 
 // Registering the driver as available.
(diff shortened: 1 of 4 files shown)


pkg/liqonet/tunnel/wireguard/driver.go
```diff
@@ -52,6 +52,9 @@ const (
 	KeepAliveInterval = 10 * time.Second
 	// MTU size of mtu for wireguard interface.
 	MTU = 1415
+	// Maybe we can remove privileged ones?
```
palexster, line 55: Can we?

palexster (approved):

pkg/liqonet/tunnel/wireguard/driver.go
alacuku: The vpn endpoint is exposed using a `NodePort` or a `LoadBalancer`. By default the port range is in 30000-32767. On the other hand some LoadBalancer implementations could expose a given service using a node port or directly  routing traffic to the backend pods. Being the port range configurable I just check the port to be inside the valid range of possible UDP ports.

aleoli (approved):

pkg/liqonet/tunnel/wireguard/driver.go
palexster: Fine by me :+1: . But a question in a comment is quite misleading.

PR #855, opened 2021-08-30
Fix issue in init-virtual-kubelet

giorio94:
# Description

This PR fixes an issue in the CSR watching logic, which caused the init-virtual-kubelet code to take longer than expected to complete. Additionally, it introduces a flag in the controller manager to disable the creation of the init container, which is set when installed on EKS (as currently not working).

# How Has This Been Tested?

- [x] Unit testing (existing + ad hoc)
- [x] E2E testing
- [X] Manual testing to ensure correct init-container configuration

### DIFF
--- a/cmd/init-virtual-kubelet/main.go
+++ b/cmd/init-virtual-kubelet/main.go
@@ -69,7 +69,7 @@ func main() {
 	_, hasCertificate, err := csr.GetCSRSecret(ctx, client, nodeName, namespace)
 	if !apierrors.IsNotFound(err) && !hasCertificate {
 		if err != nil {
-			klog.Error(err)
+			klog.Fatal(err)
 		} else {
 			klog.Info("Certificate already present for this nodeName. Skipping")
 		}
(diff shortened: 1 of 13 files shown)


cmd/init-virtual-kubelet/main.go
```diff
@@ -88,8 +88,7 @@ func main() {
 	cancel()
 
 	if err != nil {
-		klog.Error("Unable to get certificate: %w", err)
-		return
+		klog.Fatalf("Unable to get certificate: %w", err)
```
palexster, line 91: I'm afraid that this can break init on EKS, since the certificate is not provided in the CSR.

giorio94: I personally changed that to prevent silent failures (as it was happening before the fix), but I can revert if it creates problems on EKS

giorio94: I updated the PR to add the possibility to disable the init-container. Let me know if this is what you meant.

palexster: lgtm

palexster (approved):

aleoli (approved):

PR #856, opened 2021-08-31
Liqoctl Add Cluster Labels

aleoli:
# Description

This pr adds a simple logic to add some cluster labels to the `liqoctl` installed clusters.

It adds the `liqo.io/provider` label with the provider name as value, and (on managed clusters, i.e. AKS, EKS, GKE) it adds the `topology.kubernetes.io/zone` and the `topology.kubernetes.io/region` basing on the cluster zone/region retrieved from the cloud provider APIs. In that way, the virtual node is now compliant with the [topology keys](https://kubernetes.io/docs/concepts/services-networking/service-topology/) and with the [topology aware hints](https://kubernetes.io/docs/concepts/services-networking/topology-aware-hints/) labeling system.

### DIFF
--- a/pkg/liqoctl/install/aks/provider.go
+++ b/pkg/liqoctl/install/aks/provider.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
+	"github.com/liqotech/liqo/pkg/consts"
 	"github.com/liqotech/liqo/pkg/liqoctl/install/provider"
 	installutils "github.com/liqotech/liqo/pkg/liqoctl/install/utils"
 )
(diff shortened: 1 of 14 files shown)


palexster (approved):

pkg/liqoctl/install/aks/provider.go
```diff
@@ -161,6 +172,11 @@ func (k *aksProvider) parseClusterOutput(ctx context.Context, cluster *container
 
 	k.endpoint = *cluster.Fqdn
 
+	if cluster.Location != nil {
+		k.clusterLabels[consts.TopologyRegionClusterLabel] = *cluster.Location
+		k.clusterLabels[consts.TopologyZoneClusterLabel] = *cluster.Location
```
giorio94, line 177: It seems to me region and zone are always assigned the same value. Is it worth to keep both?

giorio94 (approved):

PR #861, opened 2021-09-03
Remove the cluster config parameters about resource replication and authentication

giorio94:
# Description

This PR starts the removal of the cluster config (moving the configuration to appropriate flags) and the deprecated CRD client, targeting resource replication authentication parameters.  

# How Has This Been Tested?

- [x] Unit testing (existing)
- [x] E2E testing (existing)


deployments/liqo/templates/liqo-remote-peering-rbac.yaml
```diff
@@ -12,6 +12,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: {{ include "liqo.prefixedName" $peeringBasic }}
+  labels:
+    {{- include "liqo.labels" $peeringBasic | nindent 4 }}
+    auth.liqo.io/remote-peering-permissions: "basic"
```
alacuku, line 17: Since these labels are used in code, could we add a comment that reminds us to change the label in both places if we ever need to change them?

internal/crdReplicator/crdReplicator-operator.go
```diff
@@ -431,68 +427,55 @@ func (c *Controller) StartWatchers() {
...
 				DeleteFunc: c.DeleteFunc,
 			}, stopCh, SetLabelsForLocalResources)
-			klog.Infof("%s -> starting local watcher for resource: %s", c.ClusterID, gvr.String())
+			klog.Infof("%s -> starting local watcher for resource: %s", c.ClusterID, res.GroupVersionResource.String())
 		}
 	}
 }
 
 // StopWatchers stops all the watchers for the resources that have been unregistered.
 func (c *Controller) StopWatchers() {
```
alacuku, line 473: Do we have any resource replication process that we need to stop after a given peering phase has been completed? I'm asking this since the operator does not support anymore dynamic registering/unregistering of resources.

alacuku (approved):

internal/crdReplicator/crdReplicator-operator.go
aleoli: Yes, we can have to stop some remote watchers when the peering phase changes

aleoli (approved):

internal/crdReplicator/crdReplicator-operator.go
giorio94: Hope I have not missed anything, but that logic should still be present (since it was unrelated from resource registration/unregistration, which concerned only the configuration part).

PR #862, opened 2021-09-03
Add Extra Annotation and Labels on Virtual Kubelet and Virtual Node

aleoli:
# Description

This pr adds the possibility to add additional fields for the Virtual Kubelets and the Virtual Nodes in the helm values file, in particular:

* additional extra annotation on Virtual Kubelet Deployment/Pod
* additional extra labels on Virtual Kubelet Deployment/Pod
* additional extra arguments on Virtual Kubelet Deployment/Pod
* additional extra annotations on Virtual Node
* additional extra labels on Virtual Node

Now `liqoctl` adds the `kubernetes.azure.com/managed=false` label on the Virtual Nodes when deploying on Azure, allowing these nodes to not be reconciled by the Azure controllers and avoiding conflicts.

# How Has This Been Tested?

- [x] add unit test
- [x] locally on KinD
- [x] on AKS clusters

### DIFF
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -45,9 +45,11 @@ import (
 	resourceoffercontroller "github.com/liqotech/liqo/pkg/liqo-controller-manager/resourceoffer-controller"
 	virtualNodectrl "github.com/liqotech/liqo/pkg/liqo-controller-manager/virtualNode-controller"
 	"github.com/liqotech/liqo/pkg/mapperUtils"
+	argsutils "github.com/liqotech/liqo/pkg/utils/args"
 	errorsmanagement "github.com/liqotech/liqo/pkg/utils/errorsManagement"
 	"github.com/liqotech/liqo/pkg/vkMachinery"
 	"github.com/liqotech/liqo/pkg/vkMachinery/csr"
+	"github.com/liqotech/liqo/pkg/vkMachinery/forge"
 )
 
 const (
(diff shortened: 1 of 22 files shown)


palexster (approved):

pkg/utils/parsearguments.go
```diff
@@ -0,0 +1,75 @@
...
+	"strings"
+)
+
+// StringMap implements the flag.Value interface and allows to parse stringified maps
+// in the form: "key1=val1,key2=val2".
+type StringMap map[string]string
+
+// String returns the stringified map.
+func (sm *StringMap) String() string {
+	return fmt.Sprintf("%#v", *sm)
```
giorio94, line 14: AFAIU, the String method should return a representation as expected by the Set() function, as it may be used to represent default values. Additionally, it should work also with a [nil receiver](https://pkg.go.dev/flag#Value).

pkg/utils/parsearguments.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	if str == "" {
+		return nil
+	}
+	chunks := strings.Split(str, ",")
+	for i := range chunks {
+		chunk := chunks[i]
+		if chunk == "" {
+			continue
+		}
```
giorio94, line 30: Could be probably removed trimming the last comma

pkg/utils/parsearguments.go
```diff
@@ -0,0 +1,75 @@
...
+	if *sl == nil {
+		*sl = []string{}
+	}
+	if str == "" {
+		return nil
+	}
+	chunks := strings.Split(str, ",")
+	for i := range chunks {
+		chunk := chunks[i]
+		if chunk != "" {
```
giorio94, line 65: Same here

pkg/utils/parsearguments.go
```diff
@@ -0,0 +1,75 @@
+package utils
+
+import (
+	"fmt"
+	"strings"
+)
+
+// StringMap implements the flag.Value interface and allows to parse stringified maps
+// in the form: "key1=val1,key2=val2".
+type StringMap map[string]string
```
giorio94, line 10: Nit. Could be defined as wrapper instead of alias, to prevent the need for subsequent casts, as in the [example here](https://pkg.go.dev/flag#example-Value)

giorio94 (approved):
/lgtm. Just some very minor comments inline

PR #869, opened 2021-09-06
Improve Liqoctl Install Flags

aleoli:
# Description

This pr includes several addition and improvements in the `liqoctl` flags. In particular:

* add `cluster-name` flag in the install command
* fix `name` in the generate add command
* disable `klog` flags, but `-v`
* extract cluster name from providers
* add reserved subnets flag
* add resource sharing percentage flag

# How Has This Been Tested?

- [x] add unit test
- [x] locally

### DIFF
--- a/cmd/liqoctl/cmd/install.go
+++ b/cmd/liqoctl/cmd/install.go
@@ -31,6 +31,14 @@ func newInstallCommand(ctx context.Context) *cobra.Command {
 		"Disable the check that the current kubeconfig context contains the same endpoint retrieved from the cloud provider (AKS, EKS, GKE)")
 	installCmd.PersistentFlags().String("chart-path", installutils.LiqoChartFullName,
 		"Specify a path to get the Liqo chart, instead of installing the chart from the official repository")
+	installCmd.PersistentFlags().String("cluster-name", "", "Name to assign to the Liqo Cluster")
+	installCmd.PersistentFlags().String("reserved-subnets", "", "In order to prevent IP conflicting between locally used private subnets in your "+
+		"infrastructure and private subnets belonging to remote clusters "+
+		"you need tell liqo the subnets used in your cluster. E.g if your cluster nodes belong to the 192.168.2.0/24 subnet then "+
+		"you should add that subnet to the reservedSubnets. PodCIDR and serviceCIDR used in the local cluster are automatically "+
+		"added to the reserved list. (e.g. --reserved-subnets 192.168.2.0/24,192.168.4.0/24)")
+	installCmd.PersistentFlags().String("resource-sharing-percentage", "90", "It defines the percentage of available cluster resources that "+
+		"you are willing to share with foreign clusters. It accepts [0 - 100] values.")
 
 	for _, providerName := range providers {
 		cmd, err := getCommand(ctx, providerName)
--- a/cmd/liqoctl/cmd/root.go
+++ b/cmd/liqoctl/cmd/root.go
@@ -19,11 +19,19 @@ func NewRootCommand(ctx context.Context) *cobra.Command {
 		Short: common.LiqoctlShortHelp,
 		Long:  common.LiqoctlLongHelp,
 	}
+
+	// since we cannot access internal klog configuration, we create a new flagset, let klog to install
+	// its flags, and we only keep the ones we are intrested in.
 	klogFlagset := flag.NewFlagSet("klog", flag.PanicOnError)
 	klog.InitFlags(klogFlagset)
+	klogFlagset.VisitAll(func(f *flag.Flag) {
+		if f.Name == "v" {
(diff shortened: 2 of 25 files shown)


cmd/liqoctl/cmd/root.go
```diff
@@ -21,9 +21,13 @@ func NewRootCommand(ctx context.Context) *cobra.Command {
 	}
 	klogFlagset := flag.NewFlagSet("klog", flag.PanicOnError)
 	klog.InitFlags(klogFlagset)
+	klogFlagset.VisitAll(func(f *flag.Flag) {
```
palexster, line 24: It may be worth a comment.

pkg/liqoctl/generate/handler.go
```diff
@@ -61,6 +61,10 @@ func processGenerateCommand(ctx context.Context, clientSet client.Client, liqoNa
 }
 
 func generateCommandString(commandName, authEP, clusterID, localToken, clusterName string) string {
+	if clusterName == "" {
```
palexster, line 64: It may be worth a comment, since it is a deliberate choice.

cmd/liqoctl/cmd/install.go
```diff
@@ -31,6 +31,14 @@ func newInstallCommand(ctx context.Context) *cobra.Command {
...
 	installCmd.PersistentFlags().String("chart-path", installutils.LiqoChartFullName,
 		"Specify a path to get the Liqo chart, instead of installing the chart from the official repository")
+	installCmd.PersistentFlags().String("cluster-name", "", "Name to assign to the Liqo Cluster")
+	installCmd.PersistentFlags().String("reserved-subnets", "", "In order to prevent IP conflicting between locally used private subnets in your "+
+		"infrastructure and private subnets belonging to remote clusters "+
+		"you need tell liqo the subnets used in your cluster. E.g if your cluster nodes belong to the 192.168.2.0/24 subnet then "+
+		"you should add that subnet to the reservedSubnets. PodCIDR and serviceCIDR used in the local cluster are automatically "+
+		"added to the reserved list. (e.g. --reserved-subnets 192.168.2.0/24,192.168.4.0/24)")
+	installCmd.PersistentFlags().String("resource-sharing-percentage", "90", "It defines the percentage of available cluster resources that "+
+		"you are willing to share with foreign clusters. It accepts ranges [0.0 - 1.0) or [0 - 100]")
```
palexster, line 41: Since the two ranges are overlapping, if I put 1 I would get 1 or 100? Not sure that it is clear without a measuring unit in one of the two cases.

palexster (approved):

cmd/liqoctl/cmd/install.go
giorio94: My two cents, since the flag name contains percentage, I would just expect the 0-100 range

pkg/utils/args/cidrlist.go
```diff
@@ -0,0 +1,41 @@
...
+
+// CIDRList implements the flag.Value interface and allows to parse stringified lists
+// in the form: "val1,val2".
+type CIDRList struct {
+	StringList StringList
+	CIDRList   []net.IPNet
+}
+
+// String returns the stringified list.
+func (cl CIDRList) String() string {
```
giorio94, line 15: Wondering whether this should also have pointer receiver

pkg/utils/args/percentage.go
```diff
@@ -0,0 +1,48 @@
+package args
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// Percentage implements the flag.Value interface and allows to parse stringified percentages.
+type Percentage struct {
+	val int64
```
giorio94, line 10:
```suggestion
	val uint64
```

Could also be exported to allow setting a default value, and avoiding the Getter

pkg/utils/args/percentage.go
```diff
@@ -0,0 +1,48 @@
...
+	return p.val
+}
+
+// String returns the stringified percentage.
+func (p Percentage) String() string {
+	return fmt.Sprintf("%v", p.val)
+}
+
+// Set parses the provided string into the percentage.
+func (p *Percentage) Set(str string) error {
```
giorio94, line 24: Would simplify the logic to accept only values between 0 and 100

pkg/utils/args/args_test.go
```diff
@@ -1,18 +1,25 @@
...
+	"github.com/onsi/gomega/types"
 )
 
 func TestParseArguments(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "ParseArguments Suite")
 }
 
+var parseCIDR = func(s string) net.IPNet {
+	_, cidr, _ := net.ParseCIDR(s)
```
giorio94, line 19: Would asset that the error is nil, to prevent silent failures.

giorio94 (approved):

PR #872, opened 2021-09-07
Set Liqoctl Install Verbosity and Handle Signals

aleoli:
# Description

This pr includes some improvements to the liqoctl tool:

* reduce the verbosity of the install command
* handle SIGINT and SIGTERM signals
* add the enable-ha flag to enable liqo-gateway multiple replicas

### DIFF
--- a/cmd/liqoctl/cmd/install.go
+++ b/cmd/liqoctl/cmd/install.go
@@ -39,6 +39,7 @@ func newInstallCommand(ctx context.Context) *cobra.Command {
 		"added to the reserved list. (e.g. --reserved-subnets 192.168.2.0/24,192.168.4.0/24)")
 	installCmd.PersistentFlags().String("resource-sharing-percentage", "90", "It defines the percentage of available cluster resources that "+
 		"you are willing to share with foreign clusters. It accepts [0 - 100] values.")
+	installCmd.PersistentFlags().Bool("enable-ha", false, "Enable the gateway component support active/passive high availability.")
 
 	for _, providerName := range providers {
 		cmd, err := getCommand(ctx, providerName)
--- a/cmd/liqoctl/main.go
+++ b/cmd/liqoctl/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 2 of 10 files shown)


cmd/liqoctl/main.go
```diff
@@ -5,21 +5,32 @@ import (
...
+
 func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
 	go func() {
 		<-sig
 		cancel()
+		select {
+		case <-time.NewTimer(terminationTimeout).C:
```
giorio94, line 28:
```suggestion
		case <-time.After(terminationTimeout):
```

cmd/liqoctl/main.go
```diff
@@ -5,21 +5,32 @@ import (
...
+			os.Exit(1)
+		case <-ctx.Done():
+			return
```
giorio94, line 31: This should return immediately, since you canceled the context. One of the alternatives I saw (in the controller-runtime library IIRC) was close the signal and cause the exit if invoked a second time.

cmd/liqoctl/main.go
```diff
@@ -5,21 +5,32 @@ import (
...
 )
 
+const (
+	terminationTimeout = 5 * time.Second
+)
+
 func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
```
giorio94, line 23: There should be a signal.NotifyChan() function

pkg/liqoctl/install/utils.go
```diff
@@ -89,9 +89,21 @@ func installOrUpdate(ctx context.Context, helmClient *helm.HelmClient, k provide
...
 		}
 
-		_, err = helmClient.InstallOrUpgradeChart(ctx, &chartSpec)
-		if err != nil {
-			return err
+		// provide the possibility to exit installation on context cancellation
+		errCh := make(chan error)
+		defer close(errCh)
+		go func() {
+			_, err = helmClient.InstallOrUpgradeChart(ctx, &chartSpec)
```
giorio94, line 96: Hoped the helmClient automatically terminated when the context is canceled, to simplify the entire logic.

cmd/liqoctl/cmd/install.go
```diff
@@ -39,6 +39,7 @@ func newInstallCommand(ctx context.Context) *cobra.Command {
 		"added to the reserved list. (e.g. --reserved-subnets 192.168.2.0/24,192.168.4.0/24)")
 	installCmd.PersistentFlags().String("resource-sharing-percentage", "90", "It defines the percentage of available cluster resources that "+
 		"you are willing to share with foreign clusters. It accepts [0 - 100] values.")
+	installCmd.PersistentFlags().Bool("enable-ha", false, "Enable the gateway component support active/passive high availability.")
```
giorio94, line 42:
```suggestion
	installCmd.PersistentFlags().Bool("enable-gateway-ha", false, "Execute the gateway component in active/passive high availability mode.")
```

Just a bit more explicit

giorio94 (approved):
Some minor comments inline

pkg/liqoctl/install/utils.go
aleoli: Unfortunately not, it doesn't propagate the context in the install function. And the helm library does the same thing

cmd/liqoctl/cmd/install.go
aleoli: we decided to use a general flag to allow us to add more ha settings In the future

palexster (approved):

PR #876, opened 2021-09-09
IPAM documentation

davidefalcone1:
This PR documents how addresses of reflected endpoints are translated by the IPAM module.


docs/pages/concepts/networking/components/network-manager.md
```diff
@@ -111,27 +113,39 @@ The IPAM module achieves this task by keeping track of how foreign clusters have
 
 
 #### Reflection
```
palexster, line 115: I feel that part is a bit misplaced here, I was wondering if it should move to https://doc.liqo.io/concepts/offloading/features/api-reflection/ which seems to me its natural place.

davidefalcone1: This section describes how the IPAM is involved in the reflection of endpoints; it is not supposed to describe the whole reflection of API. Therefore I think it should be here and maybe a link in the API reflection page should point to this Section.

PR #890, opened 2021-09-14
Enable Default K3s Ingress in Liqoctl

aleoli:
# Description

This pr add the possibility to create an ingress resource with the `liqoctl` install command on providers:

* `kubeadm`
* `k3s` (enabled by default)

# How Has This Been Tested?

- [x] locally
- [x] on k3s installation

### DIFF
--- a/deployments/liqo/templates/liqo-auth-service.yaml
+++ b/deployments/liqo/templates/liqo-auth-service.yaml
@@ -18,7 +18,11 @@ spec:
     {{- include "liqo.selectorLabels" $authConfig | nindent 4 }}
     {{- include "liqo.authServiceLabels" . | nindent 4 }}
   ports:
+    {{- if not .Values.auth.tls }}
+    - name: http
+    {{- else }}
     - name: https
+    {{- end }}
       protocol: TCP
       {{- if not .Values.auth.tls }}
       port: 5000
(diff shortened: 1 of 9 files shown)


palexster (approved):

deployments/liqo/templates/liqo-auth-service.yaml
```diff
@@ -18,7 +18,11 @@ spec:
     {{- include "liqo.selectorLabels" $authConfig | nindent 4 }}
     {{- include "liqo.authServiceLabels" . | nindent 4 }}
   ports:
+    {{- if not .Values.auth.tls }}
     - name: https
+    {{- else }}
+    - name: http
+    {{- end }}
```
giorio94, line 25: Seems reversed to me

giorio94 (approved):
Just a couple of thought:
1. Is it worth to offer the possibility to create the ingress through liqoctl? Would personally avoid to increase the complexity duplicating all helm parameters in the command line tool if not necessary.
2. Unclear to me why the ingress is enabled by default in k3s

aleoli:
1. it gives you the possibility to create a basic ingress with the default class, setting all the other required (and not trivial) parameters for the service. This can be a good starting point for a user that can then template the values file created by the `--only-output-values` flag
2. The ingress is enabled by default on k3s, because k3s deploys by default a traefik ingress controller that binds the port 443 on every node. This prevents the Liqo Auth LoadBalancer Service to get an external IP.

PR #898, opened 2021-09-16
Add a new Liqo API to replicate deployments inside remote clusters.

Andreagit97:
# Description

The new API `LiqoDeployment` allows users to create deployment replicas inside desired remote clusters in a straightforward and automatic way. The resource offers the possibility to specify which are the target clusters of the selection and the granularity for the deployments replication.


giorio94:
Closing, as included in #901

PR #902, opened 2021-09-20
Move Operators from Discovery to Liqo Controller Manager

aleoli:
# Description

This pr moves the ForeignCluster Operator and the SearchDomain operator from the Discovery component to the Liqo Controller Manager

# How Has This Been Tested?

- [x] existing unit test
- [x] locally on KinD


cmd/liqo-controller-manager/main.go
```diff
@@ -147,10 +168,66 @@ func main() {
...
+		MapperProvider:     mapperUtils.LiqoMapperProvider(scheme),
+		Scheme:             scheme,
+		Namespace:          *liqoNamespace,
+		MetricsBindAddress: "0", // Disable the metrics of the auxiliary manager to prevent conflicts.
+	})
+	if err != nil {
+		klog.Fatalf("Unable to create auxiliary (namespaced) manager: %w", err)
+	}
+
+	namespacedClient := client.NewNamespacedClient(auxmgr.GetClient(), *liqoNamespace)
```
giorio94, line 183: Could be removed and replaced by a single manager with per-resource namespace constraints (if not already given cluster-wide permissions for that resource).

cmd/liqo-controller-manager/main.go
```diff
@@ -131,6 +150,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
```
giorio94, line 153:
```suggestion
	ctx = ctrl.SetupSignalHandler()
```

Then, it is simply given this context to the manager when started

pkg/discoverymanager/discovery.go
```diff
@@ -25,6 +25,14 @@ import (
 	"github.com/liqotech/liqo/pkg/clusterid"
 )
 
+// clusterRole
+// +kubebuilder:rbac:groups=discovery.liqo.io,resources=foreignclusters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=discovery.liqo.io,resources=foreignclusters/status,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=core,resources=nodes,verbs=list;watch
```
giorio94, line 31:
```suggestion
// +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch
```

Just to avoid possible unwanted issues.

pkg/discoverymanager/foreign.go
```diff
@@ -136,8 +141,8 @@ func (discovery *Controller) createOrUpdate(data *discoveryData,
 	return nil
 }
 
-func (discovery *Controller) createForeign(
-	ctx context.Context, data *discoveryData,
+func createForeign(
+	ctx context.Context, cl client.Client, scheme *runtime.Scheme, data *discoveryData,
```
giorio94, line 145: You should be able to get the scheme from the client, to avoid passing both as parameters (same above).

pkg/liqo-controller-manager/foreign-cluster-operator/setup.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2019-2021 The Liqo Authors
```
giorio94, line 1: AFAIU this file could be dropped, as already done in the liqo-controller-manager main

giorio94 (approved):

PR #908, opened 2021-09-22
Generate ClusterID During Helm Installation

aleoli:
# Description

This pr moves the logic of clusterID creation in the helm chart

# How Has This Been Tested?

- [x] locally on KinD
- [x] existent E2E test


palexster (approved):

deployments/liqo/templates/liqo-auth-deployment.yaml
```diff
@@ -2,6 +2,9 @@
 {{- $authConfig := (merge (dict "name" "auth" "module" "discovery" "containerName" "cert-creator") .) -}}
 {{- $awsConfig := (merge (dict "name" "aws-config" "module" "aws-config") .) -}}
 
+{{- $clusterIdConfig := (merge (dict "name" "clusterid-configmap" "module" "clusterid-configmap") .) -}}
+{{- $clusterIdConfigName := include "liqo.prefixedName" $clusterIdConfig -}}
```
giorio94, line 6: Would define the name in the helpers.tpl file, to avoid duplicating this in all files.

deployments/liqo/templates/liqo-crd-replicator-deployment.yaml
```diff
@@ -30,11 +33,17 @@ spec:
...
             {{- toYaml .Values.crdReplicator.pod.extraArgs | nindent 12 }}
-          {{- end }}
+            {{- end }}
           env:
+            - name: CLUSTER_ID
+              valueFrom:
+                configMapKeyRef:
+                  name: {{ $clusterIdConfigName }}
+                  key: CLUSTER_ID
             - name: NAMESPACE
```
giorio94, line 47: AFAIK this is no longer required

pkg/consts/clusterid.go
```diff
@@ -16,13 +16,13 @@ package consts
 
 const (
 	// ClusterIDConfigMapName is the name of the configmap where the cluster-id is stored.
-	ClusterIDConfigMapName = "cluster-id"
+	ClusterIDConfigMapName = "liqo-clusterid-configmap"
```
giorio94, line 19: Should not be relying on the name, since it is not guaranteed to remain that one.

deployments/liqo/templates/liqo-clusterid-configmap.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: {{ include "liqo.prefixedName" $clusterIdConfig }}
+data:
+  {{- if (not $oldObject) }}
+  CLUSTER_ID: {{ uuidv4 }}
+  {{- else }}
+  CLUSTER_ID: {{ $oldObject.data.CLUSTER_ID }}
+  {{- end }}
+  {{- if .Values.discovery.config.clusterName }}
+  CLUSTER_NAME: {{ .Values.discovery.config.clusterName }}
+  {{- end }}
```
giorio94, line 19: It seems to me this is currently not used (don't know if intentionally).
Still, would remove the `if` part, as AFAIK it would "break" the pods if mounted as value and not present

giorio94 (approved):

deployments/liqo/templates/liqo-clusterid-configmap.yaml
aleoli: This value is not intended to be mounted by any pod. Pods that need the ClusterName values take it directly from `.Values.discovery.config.clusterName` to ensure they restart on name updates.

This value is to be easily readable by a user that does not have to read it from containers arguments.

Another solution can be to completely drop this value from this ConfigMap and set the whole ConfigMap as immutable, I really don't know which is a better solution.

giorio94: Got it. The current solution seems perfectly reasonable to me then.

PR #910, opened 2021-09-23
Disable Discovery Deployment

aleoli:
# Description

This pr disables the discovery component by default. It will be enabled if one of the following is true:

* the mDNS discovery is enabled
* the mDNS advertisement is enabled
* we are installing on a KinD cluster

# How Has This Been Tested?

- [x] locally


giorio94 (approved):
/lgtm

palexster (approved):

palexster (approved):

PR #913, opened 2021-09-27
Implement basic caching in IPAM storage

giorio94:
# Description

This PR introduces a basic cache in the IPAM storage component, to reduce the number of "get" requests to the API server and improve the overall performance. The cache enforces no guarantees (as the current implemented version), with concurrent reads allowed while performing a patch operation, which only at the end updates the cached version. 

Fixes #(issue)

# How Has This Been Tested?

- [x] Existing

### DIFF
--- a/pkg/liqonet/ipam/ipamStorage.go
+++ b/pkg/liqonet/ipam/ipamStorage.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sync"
 
 	goipam "github.com/metal-stack/go-ipam"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/liqonet/ipam/ipamStorage.go
```diff
@@ -54,20 +56,22 @@ type IpamStorage interface {
...
+	getServiceCIDR() string
+	getNatMappingsConfigured() map[string]netv1alpha1.ConfiguredCluster
 	goipam.Storage
 }
 
 // IPAMStorage is an implementation of IpamStorage that takes advantage of the CRD IpamStorage.
 type IPAMStorage struct {
-	dynClient    dynamic.Interface
-	resourceName string
+	sync.RWMutex
```
alacuku, line 71: In #889 we introduced a mutex to prevent to solve concurrency problems. The mutex introduced in this PR narrows the critical section to less code than the previous one. We should remove the old one.

alacuku (approved):

pkg/liqonet/ipam/ipamStorage.go
giorio94: I'm not that confident that this mutex would solve the problem, since no additional guarantee is introduced with respect to the previous implementation (it just ensures the cached values is not modified while being read).

palexster (approved):

Issue #912, opened 2021-09-27
The liqo IPAM fails to restart with certain reserved pools configured.


giorio94:
#### What happened:

Restarting the liqo-network-manager with reserved pools configured leads to an error during the initialization, and the consequent CrashLoopBackoff of the liqo-network-manager pod. 

#### What you expected to happen:

The IPAM initialization should be idempotent and allow the correct restart of the network manager.

#### How to reproduce it (as minimally and precisely as possible):

* Start a kind cluster
* Install liqo configuring the reserved subnet: `liqoctl install kind --reserved-subnets="192.168.0.0/16"`
* Restart the network manager pod

#### Anything else we need to know?:

liqo-network-manager logs (reserving 192.168.0.0/16):
```
I0927 08:39:20.231413       1 driver.go:36] driver for wireguard VPN successfully registered
I0927 08:39:20.260041       1 ipamStorage.go:76] Init IPAM storage..
I0927 08:39:20.260056       1 ipamStorage.go:79] Looking for Ipam resource..
I0927 08:39:20.265304       1 ipamStorage.go:119] Resource ipamstorage-lmgpn of type net.liqo.io/v1alpha1, Resource=ipamstorages has been found
I0927 08:39:20.265398       1 ipamStorage.go:121] Ipam storage successfully configured
I0927 08:39:20.273757       1 natMappingInflater.go:431] In memory structure of NatMappingInflater has been successfully recovered from resources.
I0927 08:39:20.276397       1 natMappingInflater.go:440] Consistency check between IPAM configuration and NatMapping resources..
I0927 08:39:20.276423       1 natMappingInflater.go:477] IPAM configuration and NatMapping resources are now aligned.
I0927 08:39:20.281551       1 ipam.go:188] Request to reserve network 192.168.0.0/16 has been received
I0927 08:39:20.287289       1 ipam.go:170] Network 192.168.0.0/16 is equal to a network pool, acquiring first half..
E0927 08:39:20.289954       1 endpoint-creator-operator.go:101] Failed to initialize IPAM: %!w(*errors.errorString=&{cannot acquire first half of pool 192.168.0.0/16})
```

liqo-network-manager logs (reserving 192.168.0.0/16):
```
I0927 08:49:10.944286       1 driver.go:36] driver for wireguard VPN successfully registered
I0927 08:49:10.978060       1 ipamStorage.go:76] Init IPAM storage..
[...]

PR #920, opened 2021-09-29
Network manager reserved networks fix

alacuku:
# Description

Subnets in CIDR notation can be passed to the network manager as parameters in order to prevent them to be used for remote clusters. A bug prevented the network manager to restart when networks to be reserved are passed as arguments.  The bug introduced in PR #868 has been fixed. If a subnet has to be removed from the reserved networks we just remove the parameter from the network manager and restart the operator.

Fixes #(issue)
#912 

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B


pkg/liqonet/ipam/ipam.go
```diff
@@ -458,22 +457,24 @@ func (liqoIPAM *IPAM) freePoolInHalves(pool string) error {
...
+				return err
+			}
+		}
+		klog.Infof("freeing half %d...", i)
+		if prefix := liqoIPAM.ipam.PrefixFrom(halfCidr); prefix != nil {
+			err = liqoIPAM.ipam.ReleaseChildPrefix(prefix)
+			if err != nil {
+				return fmt.Errorf("cannot free first half of pool %s: %w", pool, err)
+			}
+		}
```
giorio94, line 475: Here we are completely in the field of opinions, but I personally felt it clearer and more intuitive the previous unrolled version. Also in terms of overall length they seem similar to me. 
BTW, if the new version is maintained, the error message should be generalized.

pkg/liqonet/ipam/ipam.go
```diff
@@ -1181,11 +1182,43 @@ func (liqoIPAM *IPAM) SetServiceCIDR(serviceCIDR string) error {
...
+// SetReservedSubnets acquires and/or frees the reserved networks.
+func (liqoIPAM *IPAM) SetReservedSubnets(subnets []string) error {
+	reserved := liqoIPAM.ipamStorage.getReservedSubnets()
+
+	// Free all the reserved networks not needed anymore.
+	for _, r := range reserved {
+		if !slice.ContainsString(subnets, r) {
+			klog.Infof("freeing old reserved subnet %s", r)
+			if err := liqoIPAM.FreeReservedSubnet(r); err != nil {
+				return fmt.Errorf("una error occurred while freeing reserved subnet {%s}: %w", r, err)
```
giorio94, line 1203:
```suggestion
				return fmt.Errorf("an error occurred while freeing reserved subnet {%s}: %w", r, err)
```

pkg/liqonet/ipam/ipam.go
```diff
@@ -1181,11 +1182,43 @@ func (liqoIPAM *IPAM) SetServiceCIDR(serviceCIDR string) error {
...
+
+	// Reserve the newly added subnets.
+	for _, s := range subnets {
+		if !slice.ContainsString(reserved, s) {
+			klog.Infof("acquiring reserved subnet %s", s)
+			if err := liqoIPAM.AcquireReservedSubnet(s); err != nil {
+				return fmt.Errorf("an error occurred while reserving subnet {%s}: %w", s, err)
+			}
+			if err := liqoIPAM.ipamStorage.updateReservedSubnets(s, updateOpAdd); err != nil {
+				return err
```
giorio94, line 1219: IIUC, here (and less likely above), there is still the possibility of ending up with a broken scenario if this update failed, since during the next attempt the subnet would already be acquired and would then trigger a failure. Do not know if there is any solution for this problem (might free, but then also that could fail).

pkg/liqonet/ipam/ipam_test.go
```diff
@@ -1559,8 +1559,82 @@ var _ = Describe("Ipam", func() {
...
+
+		BeforeEach(func() {
+			toBeReservedSubnets1 = []string{"192.168.0.0/16", "100.200.0.0/16", "10.200.250.0/24"}
+			toBeReservedSubnets2 = []string{"192.168.1.0/24", "100.200.0.0/16", "172.16.34.0/24"}
+			toBeReservedSubnetsIncorrect1 = []string{"192.168.0.0/16", "100.200.0/16", "10.200.250.0/24"}
+		})
+		Context("Reserving subnets", func() {
+			When("Reserving subnets for the first time", func() {
+				It("should reserve all the the subnets and return nil", func() {
+					Expect(ipam.SetReservedSubnets(toBeReservedSubnets1)).To(BeNil())
```
giorio94, line 1578:
```suggestion
					Expect(ipam.SetReservedSubnets(toBeReservedSubnets1)).To(Succeed())
```

Nit. Just a bit more idiomatic when asserting something succeeded.

pkg/liqonet/ipam/ipam_test.go
```diff
@@ -1559,8 +1559,82 @@ var _ = Describe("Ipam", func() {
...
+				})
+			})
+		})
+	})
 })
 
+func checkForPrefixes(subnets []string) {
+	for _, s := range subnets {
+		prefix, err := ipam.ipamStorage.ReadPrefix(s)
+		Expect(err).To(BeNil())
```
giorio94, line 1633:
```suggestion
		Expect(err).ToNot(HaveOccurred())
```

Nit. AFAIU a bit more idiomatic

cmd/liqonet/network-manager.go
```diff
@@ -55,7 +55,8 @@ func addNetworkManagerFlags(managerFlags *networkManagerFlags) {
 }
 
 func validateNetworkManagerFlags(managerFlags *networkManagerFlags) error {
-	cidrRegex := regexp.MustCompile(`^(\d{1,3}.){3}\d{1,3}(/(\d|[12]\d|3[012]))$`)
+	cidrRegex := regexp.MustCompile(`^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]` +
+		`|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(3[0-2]|[1-2][0-9]|[0-9]))$`)
```
giorio94, line 59: IIRC, @aleoli wrote a custom argument parser for network CIDRs. Don't know whether it might be worth to switch using that one instead of the regex.

giorio94 (approved):
Thanks @alacuku for the fix

pkg/liqonet/ipam/ipam.go
alacuku: A solution could be:

1. Check if the subnet to be reserved could be reserved by checking the existing acquired prefixes. If it fails return;
2. Add the subnet to the reserved list; 
3. Acquire subnet. If it fails return;
4. For al the existing subnets in the reserved list make sure that it has been acquired by checking if the prefix exists. If it does not exist then reserve the subnet.

What do you think?

cmd/liqonet/network-manager.go
alacuku: Do you have a pointer to the function?

pkg/liqonet/ipam/ipam.go
giorio94: Seems fine to me

cmd/liqonet/network-manager.go
aleoli: Yes, that struct implements the flag.Value interface https://github.com/liqotech/liqo/blob/master/pkg/utils/args/cidrlist.go#L23

pkg/utils/args/cidr.go
```diff
@@ -0,0 +1,43 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package args
+
+import "net"
+
+// Cidr implements the flag.Value interface and allows to parse strings in CIDR format lists
+// in the form: "x.x.x.x/y".
+type Cidr struct {
```
giorio94, line 21: Acronyms should have consistent case (CIDR in this case): https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/liqonet/ipam/ipam.go
```diff
@@ -1181,11 +1246,121 @@ func (liqoIPAM *IPAM) SetServiceCIDR(serviceCIDR string) error {
...
+	// Enforce the reserved subnets. Being the reservation a two-step process,
+	// it could happen that a subnet is added to the reserved list but not
+	// reserved due to an error. So we make sure that all the subnets in the
+	// reserved list have been acquired.
+	// We are sure that if a reserved network has been added to the reserved list
+	// the prefix for that network is free or has been already acquired on behalf
+	// of the current reserved network.
+	for _, rSubnet := range reserved {
+		if err := liqoIPAM.MarkAsAcquiredReservedSubnet(rSubnet); err != nil {
+			return fmt.Errorf("an error occurred while  enforcing reserved subnet {%s}: %w", rSubnet, err)
```
giorio94, line 1286:
```suggestion
			return fmt.Errorf("an error occurred while enforcing reserved subnet {%s}: %w", rSubnet, err)
```

pkg/liqonet/ipam/ipam.go
```diff
@@ -440,40 +501,44 @@ func (liqoIPAM *IPAM) getNetworkFromPool(mask uint8) (string, error) {
 	// For each pool, try to get a network with mask length mask
 	for _, pool := range pools {
 		if mappedNetwork, err := liqoIPAM.ipam.AcquireChildPrefix(pool, mask); err == nil {
-			klog.Infof("Acquired network %s", mappedNetwork)
+			klog.Infof("isAcquired network %s", mappedNetwork)
```
giorio94, line 504: Is this change intentional?

giorio94 (approved):

pkg/liqonet/ipam/ipam_test.go
```diff
@@ -65,18 +65,15 @@ func fillNetworkPool(pool string, ipam *IPAM) error {
...
-	err = ipam.AcquireReservedSubnet(halfCidr)
+	err := ipam.AcquireReservedSubnet(halfCidr)
 	if err != nil {
 		return err
 	}
 
 	// Get second half CIDR
-	halfCidr, err = utils.Next(halfCidr)
+	halfCidr = utils.Next(halfCidr)
 	if err != nil {
```
palexster, line 77: Can we drop the following error check?

palexster (approved):

PR #934, opened 2021-10-07
Fix --offer-disable-auto-accept for controller manager

CapacitorSet:
# Description

Fix --offer-disable-auto-accept for controller manager (it is currently ignored due to a bug).

# How Has This Been Tested?

Tested by hand.


giorio94:
Would suggest to add a unit test to ensure the function correctness.

CapacitorSet:
I'm not sure how to fix this either. When debugging the tests I can see `resourceOffer.Status.Phase` reverting to "ManualActionRequired" immediately after the update which should set it to "Accepted".

I suppose it's something in the Kubernetes API that I'm missing?

Issue #956, opened 2021-10-18
Uniform the virtual-kubelet parameters


giorio94:
Currently, part of the virtual-kubelet parameters are retrieved from the command-line flags, while others are retrieved using environment variables. The parameters retrieval should be uniformed, moving to command-line flags unless strong reasons require the usage of environment variables for a given parameter.

PR #963, opened 2021-10-22
Virtual kubelet: reflection-based pod offloading

### LINKED ISSUE #604: Pending pods are not deleted by the virtual kubelet
**Describe the bug**

It is impossible to delete pods which have been scheduled on a liqo node but remained in pending status (e.g. due to existing taints). The pod transitions into terminating status, but the remote replicaset is not deleted, as well as the finalizer on the local pod.

**To Reproduce**
1. Peer two clusters using liqo
2. Create a new pod, forcing it to be scheduled in the remote cluster and to remain pending (e.g. tainting all remote nodes as NoSchedule)
3. Check that the pod is Pending 
4. Delete the pod
5. Check that the pod is now terminating, the finalizer has not been removed and that the replicaset is still present 

**Expected behavior**
The remote replicaset should be deleted, and the local pod should terminate correctly.

**Additional context**
This issue is different from #598, since in that case the remote replicaset was correctly deleted

### LINKED ISSUE #678: [Feature] Improve Pod/EndpointSlices IP translation performance
In the newer implementation, the remapping of Pod/EndpointSlice IP addresses is delegated from the Virtual Kubelet to the IPAM module (executed by the network-manager), to achieve separation of concerns. Still, this arises some criticality from the performance point of view, as those operations are relatively time intensive (due to the gRPC interaction, and the subsequent API server requests performed by the IPAM), hence slowing down the Virtual Kubelet itself (as the remapping is performed in a blocking path).

Different actions could be taken to address/mitigate this issue:
* Raise or revove the pod resource limits from the default configuration to prevent throttling;
* Refactor the Virtual Kubelet reflection to introduce working queues (avoiding blocking retries) and remove time expensive operations from non-parallelized paths;
* Introduce a caching layer in the Virtual Kubelet, to avoid requesting the remapping of IPs already mapped (e.g. when an endpointslice is updated);
* Introduce a gRPC method to remap multiple IPs at the same time, hence reducing the overhead with endpointslices with multiple backends.

### LINKED ISSUE #721: Home pod container statuses not (gradually) updated on termination

giorio94:
# Description

This PR implements pod offloading through the reflection logic, instead of leveraging the virtual kubelet provider abstraction. This allows to unify the outgoing (i.e., creation of the remote pod) and incoming (i.e., the status realignment) flows, to reduce code duplication and improve the overall performance. At the same time it moves from a remote resilience mechanism based on replicasets to one leveraging a custom ShadowPod resource, for increased control, better performance and naming consistency.

Caveats:
* pod translation still has most of the limitations of the previous version (in terms of translated fields)

Fixes #721
Fixes #678 
Fixes #604

# How Has This Been Tested?

- [x] Unit testing (new + existing)
- [x] E2E testing
- [x] Manual


cmd/virtual-kubelet/root/root.go
```diff
@@ -114,10 +85,12 @@ func runRootCommand(ctx context.Context, c *Opts) error {
 
 		Namespace: c.KubeletNamespace,
 		NodeName:  c.NodeName,
+		NodeIP:    os.Getenv("VKUBELET_POD_IP"),
```
fproj, line 88: Should we make sure `VKUBELET_POD_IP` is set?

pkg/virtualKubelet/forge/forge.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ReflectionFieldManager -> The name associated with the fields modified by virtual kubelet reflection.
+const ReflectionFieldManager = "reflection.liqo.io"
+
+var (
+	// LocalClusterID -> the cluster ID associated with the local cluster.
```
fproj, line 27: Cosmetics 💄  Do we use this style for comments, with the `->`,  anywhere else?

pkg/virtualKubelet/forge/forge.go
```diff
@@ -0,0 +1,54 @@
...
+package forge
+
+import (
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ReflectionFieldManager -> The name associated with the fields modified by virtual kubelet reflection.
+const ReflectionFieldManager = "reflection.liqo.io"
```
fproj, line 24: Wandering if the name should be `ReflectionManagerField`. It sounds weird 👂🏻

pkg/virtualKubelet/forge/meta_test.go
```diff
@@ -76,6 +71,48 @@ var _ = Describe("Meta forging", func() {
 		}))
 	})
 
+	Describe("the RemoteObjectMeta function", func() {
```
fproj, line 74: Do you think having a test that checks for the local object to be unchanged would be useful?

pkg/virtualKubelet/forge/pods.go
```diff
@@ -15,145 +15,92 @@
...
 
-	f.forgeHomeMeta(&foreignPod.ObjectMeta, &homePod.ObjectMeta, foreignNamespace, reflectionType)
-	delete(homePod.Labels, virtualKubelet.ReflectedpodKey)
+// PodIPTranslator defines the function to translate between remote and local IP addresses.
+type PodIPTranslator func(string) string
 
-	if isNewObject {
-		homePod.Spec = f.forgePodSpec(foreignPod.Spec)
+// LocalPod forges the status of the local pod, given the remote one.
+func LocalPod(local, remote *corev1.Pod, translator PodIPTranslator, restarts int32) *corev1.Pod {
```
fproj, line 39: I think the comment for this function is off 😕

pkg/virtualKubelet/forge/pods.go
```diff
@@ -15,145 +15,92 @@
...
-			klog.Error(err)
-		}
-		homePod.Status.PodIP = response.GetHomeIP()
-		homePod.Status.PodIPs[0].IP = response.GetHomeIP()
+// LocalPodStatus forges the status of the local pod, given the remote one.
+func LocalPodStatus(remote *corev1.PodStatus, translator PodIPTranslator, restarts int32) corev1.PodStatus {
+	// Translate the relevant IPs
+	if remote.PodIP != "" {
+		remote.PodIP = translator(remote.PodIP)
+		remote.PodIPs = []corev1.PodIP{{IP: remote.PodIP}}
```
fproj, line 51: Should this IP be translated too?

fproj (commented):
Review part1: everything except workload folder and forge/pods. 

Just a few comments, nothing major. Well done 💯

cmd/virtual-kubelet/root/root.go
```diff
@@ -70,8 +62,8 @@ func runRootCommand(ctx context.Context, c *Opts) error {
 		return errors.New("cluster id is mandatory")
 	}
 
-	if c.PodSyncWorkers == 0 {
-		return errdefs.InvalidInput("pod sync workers must be greater than 0")
+	if c.PodWorkers == 0 || c.ServiceWorkers == 0 || c.EndpointSliceWorkers == 0 {
```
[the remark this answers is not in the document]
giorio94: Probably that part could be refactored a bit to avoid having to check this twice, but for the moment it makes sense to add the sanity check.

cmd/virtual-kubelet/root/root.go
giorio94: Would postpone to a subsequent refactor, addressing #956

pkg/virtualKubelet/forge/forge.go
giorio94: Sometimes yes, because the linter complains if using colons

pkg/virtualKubelet/forge/forge.go
giorio94: The API field is `FieldManager`, hence the current naming. Couldn't come up with anything better

pkg/virtualKubelet/forge/pods.go
giorio94: It is, since it is assigned the one already translated the line above.

palexster (approved):

pkg/virtualKubelet/reflection/workload/podns.go
```diff
@@ -0,0 +1,446 @@
...
+
+		// The remote object is not yet terminating, trigger its deletion.
+		if shadowExists && shadow.DeletionTimestamp.IsZero() {
+			defer tracer.Step("Ensured the absence of the remote object")
+			klog.V(4).Infof("Deleting remote shadowpod %q, since local pod %q is terminating", npr.RemoteRef(name), npr.LocalRef(name))
+			return npr.DeleteRemote(ctx, npr.remoteShadowPodsClient, "ShadowPod", name, shadow.GetUID())
+		}
+
+		// The remote pod has already terminated.
+		if !remoteExists {
```
fproj, line 163: Can we move this check in the HandleStatus method? 

It sounds like a precondition for the method to run. Indeed, the check is repeated https://github.com/liqotech/liqo/blob/e71366c6e7467a1a3ac13544d7259cec5cad6be8/pkg/virtualKubelet/reflection/workload/podns.go#L226-L233

Also, I don't think this path is actually tested. Checking for the remote pod existence inside the handlestatus method would also make testing it easier.

aleoli (approved):

alacuku (approved):

fproj (commented):
LGTM 🚀

PR #965, opened 2021-10-24
liqoctl install: accept provided parameters + tests (Kind and Kubeadm)

lucafrancescato:
# Description

This PR aims to let users set the following options when using `liqoctl install` command with:
- Kind provider
  - `--enable-lan-discovery`
  - `--cluster-labels="liqo.io/provider=some-value"`
- Kubeadm provider
  - `--cluster-name`
  - `--cluster-labels="liqo.io/provider=some-value"`

Prior to this, those options, even when specified by the user, would always return, respectively:
- Kind provider
  - true
  - liqo.io/provider=kind
- Kubeadm provider
  - ""
  - liqo.io/provider=kubeadm

# How Has This Been Tested?

Verified its functioning on a local Kind cluster.
The code that has been changed is also used for installing Liqo on other cluster providers, but no tests have been performed with different cluster providers.

### DIFF
--- a/pkg/liqoctl/install/install_test.go
+++ b/pkg/liqoctl/install/install_test.go
@@ -1,0 +1,219 @@
+// Copyright 2019-2022 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package install
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+
+	installutils "github.com/liqotech/liqo/pkg/liqoctl/install/utils"
+	argsutils "github.com/liqotech/liqo/pkg/utils/args"
+)
+
+// This recursive function takes a map and a list of keys and visits a tree of nested maps
+// using the keys in the order provided. At each iteration, if the number of non-visited keys
+// is 1, the function returns the value associated to the last key, else if it is greater
+// than 1, the function expects the value to be a map and a new recursive iteration happens.
+// In case the key is not found, an empty string is returned.
+// In case no keys are provided, an error is returned.
+// Example:
+// 		m := map[string]interface{}{
+//			"first": map[string]interface{}{
+// 				"second": map[string]interface{}{
+// 					"third": "value",
+// 				},
+// 			},
+// 		}
+// 		ValueFor(m, "first", "second", "third") // returns "value", nil
+// 		ValueFor(m, "first", "second") // returns map[string]interface{}{ "third": "value" }, nil
+// 		ValueFor(m, "first", "third") // returns "", nil
+// 		ValueFor(m) // returns nil, "At least one key is required"
+func ValueFor(m map[string]interface{}, keys ...string) (val interface{}, err error) {
+	var ok bool
+	if len(keys) == 0 {
+		return nil, fmt.Errorf("At least one key is required")
+	} else if val, ok = m[keys[0]]; !ok {
+		return "", nil
+	} else if len(keys) == 1 {
+		return val, nil
+	} else if m, ok = val.(map[string]interface{}); !ok {
+		return nil, fmt.Errorf("The value for key %s is not map (expected to be a map)", keys[0])
+	} else {
+		return ValueFor(m, keys[1:]...)
+	}
+}
+
+func TestInstallCommand(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Test Install Command")
+}
+
+var _ = Describe("Test the install command works as expected", func() {
+	type testCase struct {
+		provider                 string
+		parameters               []string
+		providerValue            string
+		regionValue              string
+		customLabelValue         string
+		clusterNameValue         string
+		enableAdvertisementValue bool
+		enableDiscoveryValue     bool
+		reservedSubnetsValue     []interface{}
+	}
+
+	DescribeTable("An install command is issued",
+		func(tc testCase) {
+			cmd := &cobra.Command{}
+			cmd.PersistentFlags().Bool("enable-lan-discovery", false, "")
+			cmd.PersistentFlags().String("cluster-labels", "", "")
+			cmd.PersistentFlags().String("cluster-name", "default-cluster-name", "")
+			cmd.PersistentFlags().String("reserved-subnets", "", "")
+			cmd.PersistentFlags().Bool("generate-name", false, "")
+			cmd.SetArgs(tc.parameters)
+			Expect(cmd.Execute()).To(Succeed())
+
+			providerInstance := getProviderInstance(tc.provider)
+			Expect(providerInstance).NotTo(BeNil())
+			err := providerInstance.PreValidateGenericCommandArguments(cmd.Flags())
+			Expect(err).ToNot(HaveOccurred())
+			err = providerInstance.ValidateCommandArguments(cmd.Flags())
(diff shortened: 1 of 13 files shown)


pkg/liqoctl/install/provider/generic.go
```diff
@@ -37,6 +38,16 @@ func (p *GenericProvider) ValidateGenericCommandArguments(flags *flag.FlagSet) (
 		return err
 	}
 
+	if clusterLabelsFlag := flags.Lookup("cluster-labels"); clusterLabelsFlag != nil {
```
palexster, line 41: the label name should be a constant (should be already available somewhere)

pkg/liqoctl/install/kind/provider.go
```diff
@@ -34,6 +44,42 @@ func NewProvider() provider.InstallProviderInterface {
...
 }
 
+// ValidateCommandArguments validates specific arguments passed to the install command.
+func (k *Kind) ValidateCommandArguments(flags *flag.FlagSet) (err error) {
+	err = k.Kubeadm.ValidateCommandArguments(flags)
+	if err != nil {
+		return err
+	}
+	
+	if lanDiscoveryFlag := flags.Lookup(enableLanDiscoveryFlag); lanDiscoveryFlag != nil {
```
palexster, line 54: I think this logic would deserve some comments to be properly explained.

pkg/liqoctl/install/utils.go
```diff
@@ -128,11 +128,11 @@ func installOrUpdate(ctx context.Context, helmClient *helm.HelmClient, k provide
 }
 
 func generateValues(chartValues, commonValues, providerValues map[string]interface{}) (map[string]interface{}, error) {
```
palexster, line 130: A great addition would be a test that taking different values maps as input, verifies that the output is the expected one.

pkg/liqoctl/install/provider/generic.go
lucafrancescato: I didn't find it anywhere, however I added it at the top of the file. Other flags are specified as hard-coded values, but I could also find a constant for "cluster-name" already specified in pkg/const/parameters.go. Should I add the new constants in that file, instead of specifying them at the top of this file?

pkg/liqoctl/install/provider/generic.go
```diff
@@ -52,7 +52,7 @@ func (p *GenericProvider) ValidateGenericCommandArguments(flags *flag.FlagSet) (
 		}
 	}
 
-	subnetString, err := flags.GetString("reserved-subnets")
+	subnetString, err := flags.GetString(consts.ReservedSubnetsParameter)
```
lucafrancescato, line 55: As per https://github.com/liqotech/liqo/pull/965#discussion_r741445658

pkg/liqoctl/install/provider/generic.go
```diff
@@ -33,13 +33,13 @@ type GenericProvider struct {
 
 // ValidateGenericCommandArguments validates the flags required by every install provider.
 func (p *GenericProvider) ValidateGenericCommandArguments(flags *flag.FlagSet) (err error) {
-	p.ClusterName, err = flags.GetString("cluster-name")
+	p.ClusterName, err = flags.GetString(consts.ClusterNameParameter)
```
lucafrancescato, line 36: As per https://github.com/liqotech/liqo/pull/965#discussion_r741445658

pkg/liqoctl/install/provider/generic.go
```diff
@@ -32,21 +33,31 @@ type GenericProvider struct {
...
 	if err != nil {
 		return err
 	}
+	clusterLabelsVar := argsutils.StringMap{}
+	if err := clusterLabelsVar.Set(clusterLabels); err != nil {
+		return err
+	}
+	if clusterLabelsVar.StringMap[consts.ProviderClusterLabel] != "" {
+		p.ClusterLabels[consts.ProviderClusterLabel] = clusterLabelsVar.StringMap[consts.ProviderClusterLabel]
+	}
```
giorio94, line 51: Wondering why a custom logic is required for this specific label.

pkg/liqoctl/install/install_test.go
```diff
@@ -0,0 +1,254 @@
...
+			cmd.PersistentFlags().Bool("dry-run", false, "")
+			cmd.PersistentFlags().Bool("enable-lan-discovery", false, "")
+			cmd.PersistentFlags().String("cluster-labels", "", "")
+			cmd.PersistentFlags().Bool("disable-endpoint-check", false, "")
+			cmd.PersistentFlags().String("chart-path", installutils.LiqoChartFullName, "")
+			cmd.PersistentFlags().String("cluster-name", "", "")
+			cmd.PersistentFlags().String("reserved-subnets", "", "")
+			cmd.PersistentFlags().String("resource-sharing-percentage", "90", "")
+			cmd.PersistentFlags().Bool("enable-ha", false, "")
+			cmd.PersistentFlags().Int("mtu", 0, "")
```
giorio94, line 111: Wondering why it is necessary to redefine all arguments here.

pkg/liqoctl/install/install_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+	liqoconst "github.com/liqotech/liqo/pkg/consts"
+	installprovider "github.com/liqotech/liqo/pkg/liqoctl/install/provider"
+	installutils "github.com/liqotech/liqo/pkg/liqoctl/install/utils"
+)
+
+type StringInterfaceMap map[string]interface{}
+
+// IsContainedIn checks whether specMap's key-value pairs are equal to testMap's ones (not the other way around).
+func (specMap StringInterfaceMap) IsContainedIn(testMap map[string]interface{}) bool {
```
giorio94, line 35: Is this custom logic required? Isn't it possible to express the checks through standard Gomega helpers?

The risk of custom functions is to introduce bugs which make the tests unreliable.

pkg/liqoctl/install/install_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+	for key := range specMap {
+		testValue, ok := testMap[key]
+		if !ok {
+			break
+		}
+		specValue, ok := specMap[key]
+		if !ok {
+			break
+		}
```
giorio94, line 46: Since you are iterating over the keys, this will always succeed.

pkg/liqoctl/install/install_test.go
```diff
@@ -0,0 +1,254 @@
...
+type StringInterfaceMap map[string]interface{}
+
+// IsContainedIn checks whether specMap's key-value pairs are equal to testMap's ones (not the other way around).
+func (specMap StringInterfaceMap) IsContainedIn(testMap map[string]interface{}) bool {
+	isContained := false
+
+	for key := range specMap {
+		testValue, ok := testMap[key]
+		if !ok {
+			break
```
giorio94, line 41: Would `return false`, to make it clear the intent. Do not rely on side-effects, since they might no longer be true after future code changes. Same below.

giorio94 (commented):
A few comments inline.

pkg/liqoctl/install/aks/aks_test.go
```diff
@@ -53,6 +53,7 @@ var _ = Describe("Extract elements from AKS", func() {
 		cmd := &cobra.Command{}
 
 		GenerateFlags(cmd)
+		cmd.Flags().String("cluster-labels", "", "Cluster Labels to append to Liqo Cluster, supports '='.(e.g. --cluster-labels key1=value1,key2=value2)")
 		cmd.Flags().String("cluster-name", "", "Name to assign to the Liqo Cluster")
 		cmd.Flags().String("reserved-subnets", "", "")
```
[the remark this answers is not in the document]
lucafrancescato: Do you mean the description of the flag? If so, it's because I saw that the "cluster-name" flag's decription was given, and I just wanted to stick to the same format.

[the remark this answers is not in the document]
giorio94: I mean the whole flags definition which is copied in the tests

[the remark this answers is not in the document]
lucafrancescato: IIRC, I added them since tests would break up, however I ran again the tests and I don't see any test failing, not to mention I don't see such flags used anywhere in the tests in the first place—I just rebased this PR to master, maybe the old code worked differently.

aleoli (approved):

pkg/liqoctl/install/install_test.go
lucafrancescato: I removed the unnecessary calls to external functions and therefore the definition of many of those arguments was no more required.

Issue #944, opened 2021-10-13
[Feature] Liqo Persistent Storage Support


palexster:
**Is your feature request related to a problem? Please describe.**
Introduce a virtual storage class to allow support for stateful workloads. 

**Describe the solution you'd like**
Implement a virtual storage class that maps virtual PV on actual PV of cluster's storage classes.

### Out of Scope

- [x] Inter-cluster persistent storage replication

## Steps

- [x] Define a Liqo StorageClass - #973 
- [x] Create a provider for the StorageClass - #979 
- [x] Implement the advertising logic of storage classes in the ResourceOffer - #998 
- [x] Implement the logic for local pods - #974
- [x] Implement the logic for remote pods - #991 
- [x] ~Inform scheduler about the binded volumes (nodes attached) handled in the VK~ (handled by PV's nodeSelector)
- [x] update the capsule subchart - #999
- [x] e2e tests - #1004

PR #979, opened 2021-11-02
Add Local Storage Provisioner in the Liqo Controller Manager

aleoli:
# Description

This pr adds the Liqo storage provisioner in the controller manager in order to provide local virtual volumes

Ref #944 

# How Has This Been Tested?

- [x] locally on KinD

### DIFF
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/sig-storage-lib-external-provisioner/v7/controller"
 
 	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
 	netv1alpha1 "github.com/liqotech/liqo/apis/net/v1alpha1"
@@ -46,6 +47,7 @@ import (
 	resourceoffercontroller "github.com/liqotech/liqo/pkg/liqo-controller-manager/resourceoffer-controller"
 	searchdomainoperator "github.com/liqotech/liqo/pkg/liqo-controller-manager/search-domain-operator"
 	shadowpodctrl "github.com/liqotech/liqo/pkg/liqo-controller-manager/shadowpod-controller"
+	liqostorageprovisioner "github.com/liqotech/liqo/pkg/liqo-controller-manager/storageprovisioner"
 	virtualNodectrl "github.com/liqotech/liqo/pkg/liqo-controller-manager/virtualNode-controller"
 	"github.com/liqotech/liqo/pkg/mapperUtils"
 	peeringroles "github.com/liqotech/liqo/pkg/peering-roles"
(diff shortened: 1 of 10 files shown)


cmd/liqo-controller-manager/main.go
```diff
@@ -146,6 +150,12 @@ func main() {
 	flag.Var(&nodeExtraAnnotations, "node-extra-annotations", "Extra annotations to add to the Virtual Node")
 	flag.Var(&nodeExtraLabels, "node-extra-labels", "Extra labels to add to the Virtual Node")
 
+	// Storage Provisioner parameters
+	flag.BoolVar(&enableStorage, "enable-storage", false, "enable the liqo virtual storage class")
+	flag.StringVar(&virtualStorageClassName, "virtual-storage-class-name", "liqo", "Name of the virtual storage class")
+	flag.StringVar(&realStorageClassName, "real-storage-class-name", "", "Name of the real storage class to use for the actual volumes")
+	flag.StringVar(&storageNamespace, "storage-namespace", "liqo-storage", "Namespace where the liqo storage-related resources are stored")
```
giorio94, line 157: Would personally use the flag version without the `Var` suffix, to avoid declaring and assigning the variables separately.

cmd/liqo-controller-manager/main.go
```diff
@@ -339,10 +349,25 @@ func main() {
...
+			klog.Error(err)
+			os.Exit(1)
+		}
+	}()
+
+	if enableStorage {
+		liqoProvisioner := liqostorageprovisioner.NewLiqoLocalStorageProvisioner(mgr.GetClient(),
+			virtualStorageClassName, storageNamespace, realStorageClassName)
+
+		provisionerName := "liqo.io/storage"
```
giorio94, line 366: Should be probably defined as constant

cmd/liqo-controller-manager/main.go
```diff
@@ -339,10 +349,25 @@ func main() {
...
+		provisionController := controller.NewProvisionController(clientset, provisionerName, liqoProvisioner,
+			controller.LeaderElection(false),
+		)
+		provisionController.Run(ctx)
```
giorio94, line 370: Would suggest to add this controller to the manager if possible, and keep the previous structure.

giorio94 (approved):

cmd/liqo-controller-manager/main.go
```diff
@@ -339,10 +349,25 @@ func main() {
 	var wg = &sync.WaitGroup{}
 	broadcaster.StartBroadcaster(ctx, wg)
 
-	klog.Info("starting manager as controller manager")
-	if err := mgr.Start(ctx); err != nil {
-		klog.Error(err)
-		os.Exit(1)
+	wg.Add(1)
```
palexster, line 352: Out of curiosity: Why do we need to add the following logic?

palexster (approved):

cmd/liqo-controller-manager/main.go
aleoli: The provision controller does not implement the Runnable interface, I can wrap it in order to implement this interface (Start(ctx) instead of Run(ctx) method is required)

Do you think this is a good option?

cmd/liqo-controller-manager/main.go
aleoli: We need to start the manager in a separate goroutine, and then take the client and forward it to the storage provisioner

cmd/liqo-controller-manager/main.go
giorio94: Would personally go that route instead of running the manager in a separate goroutine, but that's a matter of taste

PR #980, opened 2021-11-02
liqoctl offload namespace

palexster:
# Description

This PR introduces the liqoctl offload command to configure the NamespaceOffloading resource for a given namespace.

# How Has This Been Tested?

- [x] Unit Tests

### DIFF
--- a/cmd/liqoctl/cmd/offload.go
+++ b/cmd/liqoctl/cmd/offload.go
@@ -1,0 +1,56 @@
+// Copyright 2019-2021 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+
+	offloadingv1alpha1 "github.com/liqotech/liqo/apis/offloading/v1alpha1"
+	"github.com/liqotech/liqo/pkg/liqoctl/offload"
+	"github.com/liqotech/liqo/pkg/utils/args"
+)
+
+func newOffloadCommand(ctx context.Context) *cobra.Command {
+	var offloadClusterCmd = &cobra.Command{
+		Use:          offload.UseCommand,
+		SilenceUsage: true,
+		Short:        offload.LiqoctlOffloadShortHelp,
+		Long:         offload.LiqoctlOffloadLongHelp,
+		Args:         cobra.MinimumNArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return offload.HandleOffloadCommand(ctx, cmd, args)
+		},
+	}
+	podOffloadingStrategy := args.NewEnum([]string{string(offloadingv1alpha1.LocalAndRemotePodOffloadingStrategyType),
+		string(offloadingv1alpha1.RemotePodOffloadingStrategyType),
+		string(offloadingv1alpha1.LocalPodOffloadingStrategyType)},
+		string(offloadingv1alpha1.LocalAndRemotePodOffloadingStrategyType))
+
+	offloadClusterCmd.PersistentFlags().Var(podOffloadingStrategy,
+		offload.PodOffloadingStrategyFlag, offload.PodOffloadingStrategyHelp)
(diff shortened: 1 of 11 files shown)


pkg/liqoctl/offload/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+	for k, v := range clusterLabels.StringMap {
+		l = append(l, corev1.NodeSelectorTerm{
+			MatchExpressions: []corev1.NodeSelectorRequirement{{
+				Key:      k,
+				Operator: corev1.NodeSelectorOpNotIn,
+				Values:   []string{v},
+			},
+			},
+		})
+	}
```
Andreagit97, line 123: Using a different NodeSelectorTerm for each label specified limits the power of the logic; if I want a namespace replica on all clusters that have both provider = A, and region = B, I couldn't get this result with a solution like this. However, I understand that specifying if labels must be considered in "AND"/"OR" from the command line would perhaps complicate things too much. Overall, it seems like a good compromise to only allow the "OR" of labels with liqoctl.

Andreagit97 (approved):
The logic seems good to me, perhaps I would specify in the "long command help"  that labels (both with whitelisting and blacklisting policies) are all combined in logical OR, using an example to make it easier.

cmd/liqoctl/cmd/offload.go
```diff
@@ -0,0 +1,52 @@
...
+	podOffloadingStrategy := args.NewEnum([]string{string(offloadingv1alpha1.LocalAndRemotePodOffloadingStrategyType),
+		string(offloadingv1alpha1.RemotePodOffloadingStrategyType),
+		string(offloadingv1alpha1.LocalPodOffloadingStrategyType)}, string(offloadingv1alpha1.LocalAndRemotePodOffloadingStrategyType))
+
+	offloadClusterCmd.PersistentFlags().Var(podOffloadingStrategy, offload.PodOffloadingStrategyFlag, "Select the desired pod offloading strategy")
+	namespaceMappingStrategy := args.NewEnum([]string{string(offloadingv1alpha1.EnforceSameNameMappingStrategyType), string(offloadingv1alpha1.DefaultNameMappingStrategyType)},
+		string(offloadingv1alpha1.DefaultNameMappingStrategyType))
+
+	offloadClusterCmd.PersistentFlags().Var(namespaceMappingStrategy, offload.NamespaceMappingFlag, "Select the desired namespace mapping strategy")
+	offloadClusterCmd.PersistentFlags().String(offload.AcceptedLabelsFlag,
```
Andreagit97, line 47: I would use a more meaningful description than "The set of labels which characterizes the local cluster when exposed remotely as a virtual node". The same thing  below.

pkg/liqoctl/offload/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+func HandleOffloadCommand(ctx context.Context, command *cobra.Command, args []string) error {
+	if !klog.V(4).Enabled() {
+		klog.SetLogFilter(logsutils.LogFilter{})
+	}
+
+	config, err := common.GetLiqoctlRestConf()
+	if err != nil {
+		return err
```
Andreagit97, line 41: I would log something in case of errors. Same thing for the cases below.

pkg/liqoctl/offload/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func forgeNamespaceOffloading(command *cobra.Command, args []string, labels, clusterLabels argsutils.StringMap) *offloadingv1alpha1.NamespaceOffloading {
```
Andreagit97, line 73: I would keep `whitelistClusterLabels` and `blacklistClusterLabels` for function parameters.

pkg/liqoctl/offload/namespace.go
palexster: Yes, the idea is to make the command very easy to understand sacrificing a bit the expressiveness.

pkg/liqoctl/offload/namespace.go
palexster: The returned error will be always logged, similarly to other logs. We should maybe add a test.

pkg/liqoctl/offload/namespace_test.go
```diff
@@ -0,0 +1,226 @@
...
+	args           []string
+	parameters     []string
+	acceptedLabels args.StringMap
+	deniedLabels   args.StringMap
+	expected       *v1alpha1.NamespaceOffloading
+}
+
+var _ = Describe("Test the generate command works as expected", func() {
+
+	DescribeTable("A generate command is performed",
```
aleoli, line 47:
```suggestion

var _ = Describe("Test the generate command works as expected", func() {

	type testCase struct {
		args           []string
		parameters     []string
		acceptedLabels args.StringMap
		deniedLabels   args.StringMap
		expected       *v1alpha1.NamespaceOffloading
	}

	DescribeTable("A generate command is performed",
```

aleoli (approved):

PR #982, opened 2021-11-04
Fix liqoctl install of a local chart

palexster:
# Description

This PR fixes a panic occurring when installing liqo without triggering `helm dependency update`.

# How Has This Been Tested?

- [x] Manually


pkg/liqoctl/install/helm.go
```diff
@@ -61,6 +61,15 @@ func initLiqoRepo(helmClient helm.Client) error {
 		return err
 	}
 
+	capsuleRepo := repo.Entry{
+		URL:  "https://clastix.github.io/charts",
+		Name: "capsule",
+	}
```
giorio94, line 67: Given these values are hard-coded both here and in the `charts.yaml` file, I would add a comment to link them and warn to update both if it will be required in the future.

giorio94 (approved):

alacuku (approved):

PR #986, opened 2021-11-04
controller-manager: refactor Updater, Broadcaster

CapacitorSet:
# Description

This PR refactors Broadcaster (renamed to Accountant) so that it only reads resources; the ResourceOffer creation logic is moved to OfferUpdater. In turn, OfferUpdater now only implements ResourceOffer creation logic for a generic ResourceReader; the update loop is moved to OfferQueue.

This is part of a deduplication effort towards #840.

Diagram of the proposed structure: refactor-3-1

# How Has This Been Tested?

- [x] Manual tests
- [x] Automated tests


CapacitorSet:
Tests fail because the Broadcaster is no longer made aware of requesting clusters (see diff), so when it enters `writeClusterResources` it has nowhere to send the updated resource. How do we fix this?

I think the problem lies in having Accountant take an active role in sending ResourceOffers, which in turns comes from the idea that we should update ResourceOffers in case resources change "drastically" (more than `updateThresholdPercentage` percent). Maybe the ResourceReader API should also feature something like onChange to let the reader tell OfferUpdater that it saw a change in resources.

```diff
-func (b *Broadcaster) enqueueForCreationOrUpdate(clusterID string) {
-       b.podMutex.Lock()
-       // No offloaded pod case. Enforce clusterID in resourcePodMap with empty resources to process ResourceOffer update.
-       if _, ok := b.resourcePodMap[clusterID]; !ok {
-               b.resourcePodMap[clusterID] = corev1.ResourceList{}
-       }
-       b.podMutex.Unlock()
-       b.updater.Push(clusterID)
-}
```

CapacitorSet:
This is the final scheme: the threshold detection logic is also moved into the OfferUpdater, which is notified whenever resources change and then decides if it should update the ResourceOffers.

refactor-2

CapacitorSet:
Current scheme:

refactor-3-1

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
```diff
@@ -69,69 +63,45 @@ const (
...
+func NewLocalMonitor(ctx context.Context, clientset kubernetes.Interface, resyncPeriod time.Duration,
+	updater *OfferUpdater) *LocalResourceMonitor {
+	nodeInformer := informers.NewSharedInformerFactoryWithOptions(
+		clientset, resyncPeriod, informers.WithTweakListOptions(noVirtualNodesFilter),
+	).Core().V1().Nodes().Informer()
+	podInformer := informers.NewSharedInformerFactoryWithOptions(
+		clientset, resyncPeriod, informers.WithTweakListOptions(noShadowPodsFilter),
+	).Core().V1().Pods().Informer()
+
+	accountant := LocalResourceMonitor{
```
giorio94, line 76:
```suggestion
	lrm := LocalResourceMonitor{
```

Nit.

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
```diff
@@ -69,69 +63,45 @@ const (
...
+		allocatable:    corev1.ResourceList{},
+		resourcePodMap: map[string]corev1.ResourceList{},
+		nodeMutex:      sync.RWMutex{},
+		podMutex:       sync.RWMutex{},
```
giorio94, line 80: AFAIR, mutexes do not need to be explicitly initialized.

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
```diff
@@ -69,69 +63,45 @@ const (
...
-	b.podMutex.RLock()
-	defer b.podMutex.RUnlock()
-	mapCopy := make(map[string]corev1.ResourceList)
-	for key, value := range b.resourcePodMap {
-		mapCopy[key] = value.DeepCopy()
-	}
-	return mapCopy
-}
+	go accountant.nodeInformer.Run(ctx.Done())
+	go accountant.podInformer.Run(ctx.Done())
```
giorio94, line 98: Typically, informers are started through the factory, and not directly. Additionally, it is usually appropriate to wait for the caches to sync before considering a component to be started.

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
```diff
@@ -69,69 +63,45 @@ const (
...
-	b.lastReadResources = map[string]corev1.ResourceList{}
-	nodesFactory := informers.NewSharedInformerFactoryWithOptions(clientset, resyncPeriod, informers.WithTweakListOptions(virtualNodeFilter))
-	b.nodeInformer = nodesFactory.Core().V1().Nodes().Informer()
-	b.nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    b.onNodeAdd,
-		UpdateFunc: b.onNodeUpdate,
-		DeleteFunc: b.onNodeDelete,
+// NewLocalMonitor creates a new LocalResourceMonitor.
+func NewLocalMonitor(ctx context.Context, clientset kubernetes.Interface, resyncPeriod time.Duration,
+	updater *OfferUpdater) *LocalResourceMonitor {
```
giorio94, line 68: Would personally avoid the double link monitor <-> updater. The updater could register itself (a Notifier interface) through an appropriate method exposed by the ResourceReaderInterface.

pkg/liqo-controller-manager/resource-request-controller/resourceRequest_controller.go
```diff
@@ -122,17 +123,28 @@ func (r *ResourceRequestReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-		r.Broadcaster.enqueueForCreationOrUpdate(remoteClusterID)
+		requeue, err = r.OfferUpdater.CreateOrUpdateOffer(remoteCluster)
+		if err != nil {
+			klog.Errorf("Error creating a ResourceOffer: %s", err)
+			if requeue {
+				return ctrl.Result{
+					Requeue:      true,
+					RequeueAfter: 1 * time.Second,
+				}, err
+			}
```
giorio94, line 139: A requeue is automatically performed in case of errors. Hence, this is not required.

pkg/liqo-controller-manager/resource-request-controller/resourceRequest_controller.go
```diff
@@ -143,6 +155,12 @@ func (r *ResourceRequestReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	if requeue {
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 1 * time.Second,
+		}, nil
+	}
```
giorio94, line 163: Should be dropped.

pkg/liqo-controller-manager/resource-request-controller/testutils_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package testutils
+package resourcerequestoperator
```
giorio94, line 15: Wondering why this change is required. If desired, would postpone to a later PR, to keep this one simple (given its criticality).

pkg/liqo-controller-manager/resource-request-controller/offerUpdater.go
```diff
@@ -175,12 +136,42 @@ func (u *OfferUpdater) createOrUpdateOffer(cluster discoveryv1alpha1.ClusterIden
...
 }
 
+// NotifyChange is used by the ResourceReader to notify that resources were added or removed.
+// It checks if any resources have changed by at least a set percentage since we last updated a ResourceOffer, and if so
+// it triggers a new update.
+func (u *OfferUpdater) NotifyChange() {
+	for clusterID := range u.currentResources {
+		if u.isAboveThreshold(clusterID) {
+			var clusterIdentity discoveryv1alpha1.ClusterIdentity
+			cluster, err := foreigncluster.GetForeignClusterByID(context.Background(), u.client, clusterID)
```
giorio94, line 149: Would avoid retrieving this here, to avoid performing potentially useless operations.

pkg/liqo-controller-manager/resource-request-controller/offerUpdater.go
```diff
@@ -175,12 +136,42 @@ func (u *OfferUpdater) createOrUpdateOffer(cluster discoveryv1alpha1.ClusterIden
...
+
+// RemoveClusterID stops tracking the resources to be offered to a given cluster.
+func (u *OfferUpdater) RemoveClusterID(clusterID string) {
+	delete(u.currentResources, clusterID)
+	u.ResourceReader.RemoveClusterID(clusterID)
+	u.OfferQueue.RemoveClusterID(clusterID)
+}
+
+// getResourceRequest returns the list of ResourceRequests for the given cluster.
 func (u *OfferUpdater) getResourceRequest(ctx context.Context, clusterID string) (*discoveryv1alpha1.ResourceRequestList, error) {
```
giorio94, line 172: It seems to me a similar function already exists. Would merge them.

pkg/liqo-controller-manager/resource-request-controller/offerUpdater.go
```diff
@@ -35,118 +33,81 @@ import (
...
+
+	client                    client.Client
 	homeCluster               discoveryv1alpha1.ClusterIdentity
 	clusterLabels             map[string]string
 	scheme                    *runtime.Scheme
 	localRealStorageClassName string
 	enableStorage             bool
+	// currentResources maps the clusters that we intend to offer resources to, to the resource list that we last used
+	// when issuing them a ResourceOffer.
+	currentResources map[string]corev1.ResourceList
```
giorio94, line 59: It seems to me this is never assigned.

pkg/liqo-controller-manager/resource-request-controller/offerUpdater.go
```diff
@@ -175,12 +136,42 @@ func (u *OfferUpdater) createOrUpdateOffer(cluster discoveryv1alpha1.ClusterIden
...
 	klog.Infof("%s -> %s Offer: %s/%s", u.homeCluster.ClusterName, op, offer.Namespace, offer.Name)
-	return true, nil
+	return false, nil
 }
 
+// NotifyChange is used by the ResourceReader to notify that resources were added or removed.
+// It checks if any resources have changed by at least a set percentage since we last updated a ResourceOffer, and if so
+// it triggers a new update.
+func (u *OfferUpdater) NotifyChange() {
+	for clusterID := range u.currentResources {
```
giorio94, line 146: What about clusterIDs not yet in the map?

pkg/liqo-controller-manager/resource-request-controller/offer_queue.go
```diff
@@ -0,0 +1,138 @@
...
+		offerUpdater: offerUpdater,
+		identities: map[string]discoveryv1alpha1.ClusterIdentity{},
+	}
+}
+
+// Start starts the update loop.
+func (u *OfferQueue) Start(ctx context.Context, group *sync.WaitGroup) {
+	group.Add(1)
+	go func() {
+		// Every two seconds, check if there are new items in the queue and process them
```
giorio94, line 62: Update the comment, as incorrect.

pkg/liqo-controller-manager/resource-request-controller/offer_queue.go
```diff
@@ -0,0 +1,138 @@
...
+func NewOfferQueue(offerUpdater *OfferUpdater) OfferQueue {
+	return OfferQueue{
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Offer update queue"),
+		offerUpdater: offerUpdater,
+		identities: map[string]discoveryv1alpha1.ClusterIdentity{},
+	}
+}
+
+// Start starts the update loop.
+func (u *OfferQueue) Start(ctx context.Context, group *sync.WaitGroup) {
```
giorio94, line 59: Would personally drop the waitgroup and make this block on ctx.Done. Then, it can be added to the manager of controller-runtime.

pkg/liqo-controller-manager/resource-request-controller/offer_queue.go
```diff
@@ -0,0 +1,138 @@
...
+			return
+		}
+		clusterID, ok := obj.(string)
+		if !ok {
+			klog.Errorf("Unexpected object in OfferQueue: %v", obj)
+			// Drop the item from the queue so that we don't keep trying to process it
+			u.queue.Forget(obj)
+			u.queue.Done(obj)
+			continue
+		}
```
giorio94, line 95: You are adding elements into the queue, hence this cannot happen.

pkg/liqo-controller-manager/resource-request-controller/offer_queue.go
```diff
@@ -0,0 +1,138 @@
...
+	// identities maps cluster IDs (used by the queue) to ClusterIdentities
+	identities map[string]discoveryv1alpha1.ClusterIdentity
+}
+
+// NewOfferQueue constructs an OfferQueue.
+func NewOfferQueue(offerUpdater *OfferUpdater) OfferQueue {
+	return OfferQueue{
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Offer update queue"),
+		offerUpdater: offerUpdater,
+		identities: map[string]discoveryv1alpha1.ClusterIdentity{},
```
giorio94, line 54: Cluster identities seem useless to me here, and introduce additional complexity.

giorio94 (commented):
Some preliminary comments.

pkg/liqo-controller-manager/resource-request-controller/testutils_test.go
CapacitorSet: Because the testutils now depend on the "parent" package: `checkResourceOfferUpdate` calls `ScaleResources`. Of course, one could just move `checkResourceOfferUpdate` to `resourceRequest_operator_test` and keep the testutils package - let me know which option would be better.

pkg/liqo-controller-manager/resource-request-controller/offer_queue.go
CapacitorSet: Cluster identities will be necessary to name the ResourceOffer in `CreateOrUpdateOffer`, so they can't be removed. What can be done is to use only cluster IDs in the queue and let `CreateOrUpdateOffer` lookup and cache identities, but I think having less informative log messages for the offer queue is not worth the small decrease in complexity.

pkg/liqo-controller-manager/resource-request-controller/offer_queue.go
CapacitorSet: What is incorrect about it?

pkg/liqo-controller-manager/resource-request-controller/offerUpdater.go
CapacitorSet: The loop is meant to iterate over all clusters that we are offering resources to. The map is updated by `CreateOrUpdateOffer`, so there is no risk of "missing" cluster IDs I believe.

pkg/liqo-controller-manager/resource-request-controller/testutils_test.go
giorio94: OK, got it. Overall the entire test suite of the resource request controller seems strange to me (these methods included), but that is a separate matter.

pkg/liqo-controller-manager/resource-request-controller/offer_queue.go
giorio94: `consumeQueue` is an infinite loop which processes elements whenever they are pushed into the queue (BTW, the typical approach is to have a wrapping function which loops, and another which processes the elements, as it was before). `wait.Until` takes action only if that function returns.

pkg/liqo-controller-manager/resource-request-controller/offer_queue.go
giorio94: I would personally follow the approach you are talking about, with the queue processing cluster IDs and the `CreateOrUpdateOffer` dealing with the identities if necessary. But that is a matter of opinion.

pkg/liqo-controller-manager/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -395,20 +396,14 @@ func (r *ForeignClusterReconciler) checkIncomingPeeringStatus(ctx context.Contex
...
+		return fmt.Errorf("reading resource requests for %v: %w", remoteClusterID, err)
 	}
 
-	status, reason, message, err := getPeeringPhaseList(foreignCluster, &incomingResourceRequestList)
+	status, reason, message, err := getPeeringPhaseList(foreignCluster, incomingResourceRequestList)
 	if err != nil {
-		err = fmt.Errorf("[%v] %w in namespace %v", remoteClusterID, err, localNamespace)
-		klog.Error(err)
-		return err
+		return fmt.Errorf("[%v] %w in namespace %v", remoteClusterID, err, localNamespace)
```
aleoli, line 406: we should be consistent in logs

I think that is nice to have each log starting with `[remoteClusterID] ...` (as in the last error, line 406)

pkg/liqo-controller-manager/resource-request-controller/eventHandler.go
```diff
@@ -52,12 +51,9 @@ func getForeignClusterEventHandler(c client.Client) handler.EventHandler {
...
 			if oldForeignCluster.Spec.IncomingPeeringEnabled != newForeignCluster.Spec.IncomingPeeringEnabled {
-				var resourceRequestList discoveryv1alpha1.ResourceRequestList
-				if err := c.List(ctx, &resourceRequestList, client.HasLabels{
-					consts.ReplicationStatusLabel}, client.MatchingLabels{
-					consts.ReplicationOriginLabel: remoteClusterID,
-				}); err != nil {
-					klog.Error(err)
+				resourceRequestList, err := GetResourceRequests(ctx, c, remoteClusterID)
+				if err != nil {
+					klog.Errorf("Failed to list resource requests: %s\n", err)
```
aleoli, line 56: I would follow the logging convention described above

pkg/liqo-controller-manager/resource-request-controller/resourceRequest_operator_test.go
```diff
@@ -583,7 +582,8 @@ var _ = Describe("ResourceRequest Operator", func() {
 					Namespace: ResourcesNamespace,
 				}, offer)
 				Expect(err).ToNot(HaveOccurred())
-				return testutils.IsAllZero(&offer.Spec.ResourceQuota.Hard)
+				fmt.Printf("%#v\n", offer.Spec.ResourceQuota.Hard)
```
aleoli, line 585: is this change intentional?

aleoli (approved):

pkg/liqo-controller-manager/foreign-cluster-operator/foreign-cluster-controller.go
CapacitorSet: Fair point, but I think that should be handled in logs specifically while errors are more general and should follow the usual conventions - I'll add the `[remoteClusterID]` prefix in `klog.Error` in the caller.

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
giorio94: The variable name should still be updated.

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
```diff
@@ -32,27 +31,22 @@ import (
...
-	resourceSharingPercentage uint64
-	updateThresholdPercentage uint64
+// LocalResourceMonitor is an object that keeps track of the cluster's resources.
+type LocalResourceMonitor struct {
+	allocatable    corev1.ResourceList
+	resourcePodMap map[string]corev1.ResourceList
+	nodeMutex      sync.RWMutex
+	podMutex       sync.RWMutex
+	nodeInformer   cache.SharedIndexInformer
+	podInformer    cache.SharedIndexInformer
```
giorio94, line 48: Informers do not need to be stored

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
```diff
@@ -32,27 +31,22 @@ import (
...
+	notifier       ResourceUpdateNotifier
```
giorio94, line 49: Could it make sense to make this a list of notifiers?

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
```diff
@@ -234,150 +210,97 @@ func (b *Broadcaster) onPodUpdate(oldObj, newObj interface{}) {
...
-			b.enqueueForCreationOrUpdate(clusterID)
-		}
+	m.nodeMutex.Lock()
+	m.allocatable = newResources.DeepCopy()
+	m.nodeMutex.Unlock()
+	if m.notifier == nil {
+		klog.Warning("No notifier is configured, an update will be lost")
+	} else {
+		m.notifier.NotifyChange()
 	}
```
giorio94, line 248: Could be extracted to a function to avoid duplication.

pkg/liqo-controller-manager/resource-request-controller/offerUpdater.go
```diff
@@ -35,118 +31,97 @@ import (
...
+// NewOfferUpdater constructs a new OfferUpdater.
+func NewOfferUpdater(k8sClient client.Client, homeCluster discoveryv1alpha1.ClusterIdentity,
+	clusterLabels map[string]string, scheme *runtime.Scheme, reader ResourceReader,
+	updateThresholdPercentage uint, localRealStorageClassName string, enableStorage bool) *OfferUpdater {
+	updater := &OfferUpdater{
+		ResourceReader:            reader,
+		client:                    k8sClient,
+		homeCluster:               homeCluster,
+		clusterLabels:             clusterLabels,
+		scheme:                    scheme,
```
giorio94, line 81: The scheme can be retrieved from the client, it is not necessary to pass it explicitly

pkg/liqo-controller-manager/resource-request-controller/offerUpdater.go
```diff
@@ -35,118 +31,97 @@ import (
...
+func (u *OfferUpdater) CreateOrUpdateOffer(cluster discoveryv1alpha1.ClusterIdentity) (requeue bool, err error) {
 	ctx := context.Background()
-	list, err := u.getResourceRequest(ctx, cluster.ClusterID)
+	list, err := GetResourceRequests(ctx, u.client, cluster.ClusterID)
 	if err != nil {
 		return true, err
 	} else if len(list.Items) != 1 {
-		// invalid clusterID so return requeue = false. The clusterID will be removed from the workqueue and broadacaster maps.
-		return false, fmt.Errorf("ClusterID %s is no more valid. Deleting", cluster.ClusterName)
+		// invalid clusterID so return requeue = false. The clusterID will be removed from the workqueue and broadcaster maps.
```
giorio94, line 105: The comment should be updated (broadcaster no longer exists)

pkg/liqo-controller-manager/resource-request-controller/offerUpdater.go
```diff
@@ -221,23 +202,60 @@ func (u *OfferUpdater) getStorageClasses(ctx context.Context) ([]sharingv1alpha1
...
+	}
+	return true
 }
 
-func getRandomTimeout() time.Duration {
-	max := new(big.Int)
-	max.SetInt64(int64(maxRandom))
-	n, err := rand.Int(rand.Reader, max)
+// GetResourceRequests returns the list of ResourceRequests for the given cluster.
+func GetResourceRequests(ctx context.Context, k8sClient client.Client, clusterID string) (
```
giorio94, line 249: Since you are looking for a single resourcerequest, you could include here the logic to check whether it exists or is duplicated, and return the appropriate error.

giorio94 (approved):
/lgtm Some minor comments inline

pkg/liqo-controller-manager/resource-request-controller/local-resource-monitor.go
CapacitorSet: The need for multiple notifiers in parallel was not mentioned in the meeting. I can see why they could be useful, but it shouldn't be hard to implement them when needed.

giorio94: Yep, it should be trivial to add

PR #995, opened 2021-11-12
Store Bearer Token in a tmp file

aleoli:
# Description

This pr fixes a permission problem in writing token files. The pod user is no more root and it has no permission on writing in the pod filesystem.

# How Has This Been Tested?

- [x] locally on KinD

### DIFF
--- a/pkg/identityManager/identityManager_test.go
+++ b/pkg/identityManager/identityManager_test.go
@@ -45,11 +45,6 @@ import (
 	"github.com/liqotech/liqo/pkg/utils/testutil"
 )
 
-type mockApiServerConfigProvider struct {
-	address   string
-	trustedCA bool
-}
-
 func TestIdentityManager(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "IdentityManager Suite")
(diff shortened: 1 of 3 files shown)


palexster (approved):

pkg/identityManager/identityManager_test.go
```diff
@@ -90,8 +85,10 @@ var _ = Describe("IdentityManager", func() {
 		restConfig = cluster.GetCfg()
 
 		namespaceManager = tenantnamespace.NewTenantNamespaceManager(client)
-		identityMan = NewCertificateIdentityManager(cluster.GetClient(), localClusterID, namespaceManager)
-		identityProvider = NewCertificateIdentityProvider(ctx, cluster.GetClient(), localClusterID, namespaceManager)
+		identityMan = NewCertificateIdentityManager(cluster.GetClient(),
+			localClusterID, namespaceManager)
+		identityProvider = NewCertificateIdentityProvider(ctx, cluster.GetClient(),
+			localClusterID, namespaceManager)
```
giorio94, line 91: Are those changes intentional? Would personally avoid too short lines, but that is definitely opinionated. (Same for the occurrences below)

pkg/identityManager/tokenManager.go
```diff
@@ -135,15 +134,28 @@ func (tokMan *iamTokenManager) addClusterID(remoteClusterID string, secret types
...
+	var err error
+	filename, found := tokMan.tokenFiles[remoteClusterID]
+	if found {
+		_, err = os.Stat(filename)
+	}
+
+	if !found || os.IsNotExist(err) {
+		file, err := os.CreateTemp("", "token")
+		if err != nil {
 			klog.Error(err)
```
giorio94, line 146: Would enrich a bit the message, to avoid meaningless errors.

giorio94 (approved):

PR #1011, opened 2021-11-24
Fix Local Pvc Naming

aleoli:
# Description

This pr fixes an issue that caused names collision if multiple PVC with that same name were created in different namespaces and scheduled on the local cluster.

# How Has This Been Tested?

- [x] locally on KinD
- [x] unit test

### DIFF
--- a/pkg/liqo-controller-manager/storageprovisioner/localprovisioner.go
+++ b/pkg/liqo-controller-manager/storageprovisioner/localprovisioner.go
@@ -31,9 +31,10 @@ import (
 func (p *liqoLocalStorageProvisioner) provisionLocalPVC(ctx context.Context,
 	options controller.ProvisionOptions) (*v1.PersistentVolume, controller.ProvisioningState, error) {
 	virtualPvc := options.PVC
+	realPvcName := virtualPvc.GetUID()
 	realPvc := v1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      virtualPvc.Name,
+			Name:      string(realPvcName),
 			Namespace: p.storageNamespace,
 		},
 	}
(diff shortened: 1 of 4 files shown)


palexster (approved):

pkg/liqo-controller-manager/storageprovisioner/localprovisioner.go
```diff
@@ -134,3 +135,11 @@ func mergeAffinities(vol1, vol2 *v1.PersistentVolumeSpec) *v1.VolumeNodeAffinity
...
 	}
 }
+
+func getLocalRealPvcName(virtualPvcNamespace, virtualPvcName string) string {
+	pvcName := fmt.Sprintf("%s-%s", virtualPvcNamespace, virtualPvcName)
+	if len(pvcName) > 63 {
+		return pvcName[:63]
+	}
+	return pvcName
+}
```
giorio94, line 145: This could nonetheless lead to collisions in case of long namespace and PVC names.
To avoid this, I would personally keep the end of the name (instead of the beginning), to prevent problems with the PVCs originated from statefulsets.
Wondering also whether adding an hash or something similar, instead of the full namespace name, could lower the risk...

aleoli: I used the parent PVC UID, in that way the real PVC name is the same as the virtual PV, and simpler to understand the relation between them (compared to the hash)

giorio94: Makes sense to me. Could it be worth adding a couple of labels identifying the virtual PVC namespace and name?

pkg/liqo-controller-manager/storageprovisioner/storageprovisioner_test.go
```diff
@@ -79,6 +80,10 @@ var _ = Describe("Test Storage Provisioner", func() {
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      name,
 						Namespace: namespace,
+						UID: func() apitypes.UID {
+							uid, _ := uuid.NewUUID()
+							return apitypes.UID(uid.String())
+						}(),
```
giorio94, line 86: Could assign a fake value instead of generating a real UUID, which is even simpler.

giorio94 (approved):

PR #1013, opened 2021-11-26
Cache unit tests in CI

CapacitorSet:
# Description

Adds Go build caching to the CI workflow (see eg. https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds). This allows us to "skip" building and running tests for components that haven't changed.

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,6 +126,7 @@ jobs:
       with:
         options: |
           --mount type=bind,src=${{ github.workspace }},dst=/go/src/liqotech/liqo
+          --mount type=bind,src=/home/runner/.cache,dst=/root/.cache
           --workdir /go/src/liqotech/liqo
           --privileged=true
         image: liqo/liqo-test:${{ needs.configure.outputs.commit-ref }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -103,6 +103,23 @@ jobs:
 
     - name: Fetch External CRDs
       run: make fetch-external-crds
+    
+    - name: Load Go build cache
+      uses: actions/cache@v2
+      with:
+        # In order:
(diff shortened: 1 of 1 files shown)


giorio94:
I am personally against caching the unit-test results, as it would introduce an additional degree of non-determinism. 
This could raise (at least) two possible problems: (a) test coverage reduction, since test suites are slightly randomized, and different runs might uncover different problems; (b) flaky tests: let consider a test that succeeds once (by chance) and fails all the others. If that first run gets cached, it will no longer be executed, and a bug will get unnoticed, regardless of the test existence. This could also potentially cause code modifications to trigger unrelated tests failures, just because they get te-executed, hence making them harder to debug.

CapacitorSet:
I agree with point (b); I'm not sure about point (a), because I don't recall seeing randomized unit tests - a quick `grep -R rand **/*_test.go` reveals that there is effectively one usage of `rand` for the IPAM listening port.

With that said, I believe there is value in "incremental testing" for faster iteration, while keeping uncached tests as a requirement for merging. Of course it is still possible for a flaky test to evade detection, but for a typical PR where one runs a few incremental tests and then an uncached one, each test would run at the very least twice (once when it was cached and once in uncached tests).

giorio94:
As for point a) the order of the tests themself is random (could also randomize the It specs, but that is not done at the moment), and controllers (especially in integration tests) have intrinsic random behaviors.

As for keeping both alternatives, that is definitely possible, although with the increased complexity of correctly handling the mandatory checks. Still, I am missing a bit the advantage, as a subset of tests of interest could be easily executed locally for much faster iteration (only a very few network related require to be executed through the container). That way it is also easier to investigate possible failures.

PR #1018, opened 2021-12-06
Add Controller Manager Probes

aleoli:
# Description

This pr avoids the Liqo controller manager entering in crash loop backup state just after installation/upgrades.

This behavior was due to the capsule controller manager not completely running at the moment when the liqo controller manager is ensuring the liqo-storage namespace.

# How Has This Been Tested?

- [x] locally on KinD

### DIFF
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -16,16 +16,21 @@ package main
 
 import (
 	"flag"
+	"fmt"
+	"net/http"
 	"os"
 	"sync"
 	"time"
 
 	capsulev1beta1 "github.com/clastix/capsule/api/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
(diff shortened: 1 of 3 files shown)


test/e2e/postinstall/basic_test.go
```diff
@@ -64,17 +65,31 @@ var _ = Describe("Liqo E2E", func() {
...
 						return err == nil && len(notReadyPods) == 0 && len(readyPods) > 0
 					}, timeout, interval).Should(BeTrue())
+
+					// Check that the pods were not restarted
+					pods, err := cluster.NativeClient.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					for _, pod := range pods.Items {
+						if strings.Contains(pod.GetName(), "capsule") {
+							// the capsule controller manager is restarted every time liqo is installed
+							continue
```
palexster, line 75: Can we assert that capsule has restarted only once?

cmd/liqo-controller-manager/main.go
```diff
@@ -343,9 +361,18 @@ func main() {
...
-		liqoProvisioner, err := liqostorageprovisioner.NewLiqoLocalStorageProvisioner(ctx, mgr.GetClient(),
-			*virtualStorageClassName, *storageNamespace, *realStorageClassName)
-		if err != nil {
+		var liqoProvisioner controller.Provisioner
+		if err = retry.OnError(wait.Backoff{
+			Duration: 1 * time.Second,
+			Factor:   1,
+			Steps:    180, // 3 minutes
+		}, func(e error) bool {
+			return err != nil
```
giorio94, line 370: Could it be worth retrying only in case of errors caused by the capsule webhook, and fail in all the others? Just to better highlight failures in case of misconfigurations.

test/e2e/postinstall/basic_test.go
```diff
@@ -64,17 +65,31 @@ var _ = Describe("Liqo E2E", func() {
...
+						}
+						Expect(pod.Status.ContainerStatuses).ToNot(BeEmpty())
+						Expect(pod.Status.ContainerStatuses[0].RestartCount).To(Equal(int32(0)))
```
giorio94, line 78:
```suggestion
						Expect(pod.Status.ContainerStatuses[0].RestartCount).To(BeNumerically("==", 0))
```

Nit.

giorio94 (approved):

palexster (approved):

PR #1022, opened 2021-12-07
Authentication: modify transport to leverage persistent connections

giorio94:
# Description

This PR changes the HTTP client generation to always use the same transport, in order to reuse the same underlying connection when possible and reduce the handshake delay in high-latency WANs.

# How Has This Been Tested?

- [X] Existing
- [X] Manual testing

### DIFF
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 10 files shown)


cmd/liqo-controller-manager/main.go
```diff
@@ -212,12 +213,15 @@ func main() {
 	}
 
 	// Setup operators
```
palexster, line 215: The comment is not valid anymore IIUC

palexster (approved):

cmd/liqo-controller-manager/main.go
giorio94: Sorry, I am not getting your point :(

alacuku (approved):

cmd/liqo-controller-manager/main.go
palexster: Nothing important. Just if you can update the comment from that generic one.

PR #1050, opened 2022-01-13
Add lint target to makefile

alacuku:
# Description

Add `lint` target in the Makefile.


giorio94:
Why is go vet required as part of the fmt target. AFAIK, it is more related to linting rather than formatting

aleoli (approved):

giorio94 (approved):

PR #1061, opened 2022-01-24
Enable/disable liqo networking

alacuku:
# Description

This PR adds a new feature that makes possible to disable the liqo networking by just setting the `NetworkingEnabled` field in the `foreigncluster` resource to `No`.

Behavior of the CRDReplicator component when the field `NetworkingEnabled` is set to:

- `Yes`: the network resource is replicated in to the remote cluster;
- `No`: the network resource is not replicated in to the remote cluster;
- `Yes` to `No`: the reflector for the network resources for the given cluster is stopped if running;
- `No` to `Yes`: the reflector for the network resources for the given cluster is started and the network resource is replicated in to the remote cluster.

Behavior of the Network-Manager component when the field `NetworkingEnabled` is set to:

- `Yes`: the local network resource is created for the remote cluster;
- `No`: the  local network resource is not created for the remote cluster;
- `Yes` to `No`: the local network resource previously created for the remote cluster is removed;
- `No` to `Yes`: the local network resource is created for the remote cluster.

Fixes #(issue)

# How Has This Been Tested?

- [x] Unit tests

### DIFF
--- a/apis/discovery/v1alpha1/foreigncluster_types.go
+++ b/apis/discovery/v1alpha1/foreigncluster_types.go
@@ -1,4 +1,4 @@
-// Copyright 2019-2021 The Liqo Authors
+// Copyright 2019-2022 The Liqo Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/apis/discovery/v1alpha1/foreigncluster_types.go
+++ b/apis/discovery/v1alpha1/foreigncluster_types.go
@@ -1,4 +1,4 @@
-// Copyright 2019-2022 The Liqo Authors
+// Copyright 2019-2021 The Liqo Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -66,6 +66,18 @@ const (
 	PeeringEnabledYes PeeringEnabledType = "Yes"
 )
 
+// NetworkingEnabledType indicates the desired state for the network interconnection with this remote cluster.
+type NetworkingEnabledType string
+
+const (
+	// NetworkingEnabledNo indicates to not handle the network interconnection with this remote cluster.
+	NetworkingEnabledNo NetworkingEnabledType = "No"
+	// NetworkingEnabledYes indicates to handle the network interconnection with this remote cluster.
+	NetworkingEnabledYes NetworkingEnabledType = "Yes"
+	// NetworkingEnabledNone is a placeholder to be used when the state of the networking is not known.
+	NetworkingEnabledNone NetworkingEnabledType = "None"
+)
+
 // ForeignClusterSpec defines the desired state of ForeignCluster.
 type ForeignClusterSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
@@ -83,6 +95,11 @@ type ForeignClusterSpec struct {
 	// +kubebuilder:default="Auto"
 	// +kubebuilder:validation:Optional
 	IncomingPeeringEnabled PeeringEnabledType `json:"incomingPeeringEnabled"`
+	// Indicates if Liqo has to handle the network interconnection with the remote cluster.
+	// +kubebuilder:validation:Enum="No";"Yes"
+	// +kubebuilder:default="Yes"
+	// +kubebuilder:validation:Optional
+	NetworkingEnabled NetworkingEnabledType `json:"networkingEnabled,omitempty"`
 	// URL where to contact foreign Auth service.
 	// +kubebuilder:validation:Pattern=`https:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)`
 	ForeignAuthURL string `json:"foreignAuthUrl"`
--- a/internal/crdReplicator/crdReplicator-operator.go
+++ b/internal/crdReplicator/crdReplicator-operator.go
@@ -63,6 +63,9 @@ type Controller struct {
 
 	peeringPhases      map[string]consts.PeeringPhase
 	peeringPhasesMutex sync.RWMutex
+
+	networkingStates     map[string]discoveryv1alpha1.NetworkingEnabledType
+	networkingStateMutex sync.RWMutex
 }
 
 // cluster-role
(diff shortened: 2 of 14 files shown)


internal/liqonet/network-manager/netcfgcreator/networkconfig.go
```diff
@@ -35,11 +35,19 @@ import (
...
 	networkConfigList := &netv1alpha1.NetworkConfigList{}
-	labels := client.MatchingLabels{consts.ReplicationDestinationLabel: clusterID}
+
+	if labels == nil {
+		labels = client.MatchingLabels{
+			consts.ReplicationDestinationLabel: clusterID,
+		}
+	} else {
+		labels[consts.ReplicationDestinationLabel] = clusterID
+	}
```
giorio94, line 50: Could use the `labels.Merge` function (or equivalent) if possible, which is a bit more idiomatic.

internal/liqonet/network-manager/tunnelendpointcreator/tunnelendpointcreator.go
```diff
@@ -256,11 +256,17 @@ func (tec *TunnelEndpointCreator) enforceRemoteNetConfigMeta(ctx context.Context
...
+		if client.IgnoreNotFound(err) != nil {
 			klog.Errorf("Failed to retrieve ForeignCluster associated with NetworkConfig %q: %v", klog.KObj(netcfg), err)
 			return err
 		}
 
+		// It could happen that the networkconfig is created before the foreigncluster.
+		if err != nil {
+			klog.V(4).Infof("Failed to retrieve ForeignCluster associated with NetworkConfig %q: %v", klog.KObj(netcfg), err)
+			return nil
+		}
```
giorio94, line 268: I am missing why this is needed, as the final outcome seems the same to me.

internal/crdReplicator/crdReplicator_operator_test.go
```diff
@@ -210,4 +223,55 @@ var _ = Describe("CRD Replicator Operator Tests", func() {
 			})
 		})
 	})
+
+	Context("replication tests by phase without networking", func() {
+		When("the peering phase is none", func() {
+			JustBeforeEach(func() { DisableNetworking() })
```
giorio94, line 229: Should be moved a level up, and converted to a BeforeEach block which mutates the fc object. Otherwise, you might risk creating flaky tests due to the switching between enabled and not enabled.

internal/crdReplicator/crdReplicator_operator_test.go
```diff
@@ -145,6 +155,9 @@ var _ = Describe("CRD Replicator Operator Tests", func() {
 		foreignCluster.Status = *statusCopy
 		Expect(cl.Status().Update(ctx, &foreignCluster)).To(Succeed())
 
+		foreignCluster.Spec.NetworkingEnabled = discoveryv1alpha1.NetworkingEnabledYes
+		Expect(cl.Update(ctx, &foreignCluster)).To(Succeed())
```
giorio94, line 159: Would add that directly to the foreign cluster definition above, since this is the default value nonetheless.

giorio94 (approved):

internal/crdReplicator/crdReplicator_operator_test.go
alacuku: For sure the JustBeforeEach block could be moved a level up and avoid calling it in the following When blocks. 
But we must stick with the JustBeforeEach since the foreigncluster itself is created in a JustBeforeEach.

giorio94: Yes, but the foreigncluster variable is defined in the BeforeEach block, and here you can just set the network parameter, while leaving to the existing JustBeforeEach block the actual creation (if I understood correctly your goal).

alacuku: Ok, using a BeforeEach we just change the Foreigncluster resource before it is created.  This way we have faster and cleaner way to have a Foreigncluster instance with networking disabled. Thanks!

internal/liqonet/network-manager/netcfgcreator/netcfgcreator_test.go
```diff
@@ -229,4 +231,59 @@ var _ = Describe("NetworkConfigCreator Controller", func() {
 			It("should ensure the absence of the NetworkConfig", AssertNetworkConfigAbsence())
 		})
 	})
+
+	When("the ForeignCluster has not networking enabled", func() {
+		DisableNetworking := func() {
+			fc.Spec.NetworkingEnabled = discoveryv1alpha1.NetworkingEnabledNo
+		}
+
+		BeforeEach(DisableNetworking)
```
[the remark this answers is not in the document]
alacuku: I would leave it for code readability..

internal/liqonet/network-manager/tunnelendpointcreator/tunnelendpointcreator.go
alacuku: If we need to set up the networking before the peering is enabled, i.e foreigncluster does not exist we have to process the networkconfigs even if the fc is not found. So I added for debugging purposes a the log when the fc is not found.

internal/liqonet/network-manager/netcfgcreator/networkconfig.go
alacuku: I don't know a `labels.Merge` equivalent that works with the `client.MatchingLabels`. By using `labels.Merge` would become cumbersome to switch back and forth from the different `implementations` of labels. Any idea?

apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -83,6 +93,11 @@ type ForeignClusterSpec struct {
 	// +kubebuilder:default="Auto"
 	// +kubebuilder:validation:Optional
 	IncomingPeeringEnabled PeeringEnabledType `json:"incomingPeeringEnabled"`
+	// Indicates if Liqo has to handle the network interconnection with the remote cluster.
+	// +kubebuilder:validation:Enum="No";"Yes"
+	// +kubebuilder:default="Yes"
+	// +kubebuilder:validation:Optional
+	NetworkingEnabled NetworkingEnabledType `json:"networkingEnabled,omitempty"`
```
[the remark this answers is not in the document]
alacuku: Removing OmitEmpty flag if the field is not set then the empty string is set as value. Obviously the creation would fail since the `""` is not a valid value.

internal/crdReplicator/crdReplicator-operator.go
```diff
@@ -215,6 +218,14 @@ func (c *Controller) enforceReflectionStatus(ctx context.Context, remoteClusterI
...
 	for i := range c.RegisteredResources {
 		res := &c.RegisteredResources[i]
+		// Do not start the reflector for the networkconfigs if the network has not been enabled.
+		// Stop the reflector if already started.
+		if res.GroupVersionResource == netv1alpha1.NetworkConfigGroupVersionResource && netEnabled == discoveryv1alpha1.NetworkingEnabledNo {
+			if err := reflector.StopForResource(res); err != nil {
+				return err
+			}
+			continue
+		}
```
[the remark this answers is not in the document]
alacuku: I've gone a little bit further and implemented similar logic as the `isReplicationEnabled` for the networking state.

apis/discovery/v1alpha1/foreigncluster_types.go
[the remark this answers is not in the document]
giorio94: You are right, just ignore my comment

giorio94 (approved):

Issue #1071, opened 2022-01-27
[Feature] Move PersistentVolumes between clusters


aleoli:
**Is your feature request related to a problem? Please describe.**
When a stateful pod is scheduled on a cluster and the Liqo volume is created on that cluster, there is no way to change the cluster where the pod is scheduled

**Describe the solution**
A solution to make it possible to copy the content of the volume from one cluster to another

To make it possible we will:
1. create an in-cluster [Restic](https://restic.net/) repository

PR #1073, opened 2022-01-27
Liqoctl: Add Move Volume Command [2/4]

aleoli:
# Description

This pr implements the internal of the new `liqoctl move volume` command.

In particular:
1. ensures that the volume is not mounted
2. creates a new Restic server
3. offloads the liqo-storage namespace
4. creates a snapshotter job to backup the volume in the old cluster
5. creates a restorer job to restore the volume in the new cluster
6. deletes the Restic server

Ref. #1071

### DIFF
--- a/pkg/liqoctl/move/restic.go
+++ b/pkg/liqoctl/move/restic.go
@@ -1,0 +1,234 @@
+// Copyright 2019-2022 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


pkg/liqoctl/move/restic.go
```diff
@@ -0,0 +1,218 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			s.Fail("Failed to wait for restic repository to be up and running: ", ctx.Err())
+			return ctx.Err()
+		case <-ticker.C:
+			if err := cl.Get(ctx, client.ObjectKey{
+				Name:      resticRegistry,
+				Namespace: liqoStorageNamespace,
+			}, &statefulSet); err != nil {
```
alacuku, line 208: We should not ignore the `error` here. By doing so in case of errors that are not temporary, the user will always get `timeout` error.

alacuku (approved):

PR #1083, opened 2022-01-31
Fix Local Volumes Node Affinity

aleoli:
# Description

This ps fixes an issue that makes it possible that some pods mounting local volumes were scheduled in remote clusters.

When the local real PV has no node affinities this bug makes also the virtual PV not have node affinities. Now if the real PV has no affinity only the virtual PV affinities are taken (i.e. node not in virtual nodes).

# How Has This Been Tested?

- [x] locally
- [x] add now unit test


giorio94 (approved):
/lgtm

giorio94:
/host Still to perform some tests on CrownLabs

PR #1106, opened 2022-02-11
fix(gateway): fix arp requests when intra-cluster network traffic is …

alacuku:
# Description

The fix provides a way to route traffic between network namespace without the need of `ARP` requests hence decreasing the latency for new inter-cluster connections.

# How Has This Been Tested?

- [x] Existing Unit and E2E tests


internal/liqonet/tunnel-operator/tunnel-operator.go
```diff
@@ -488,144 +422,76 @@ func (tc *TunnelController) SetUpRouteManager() error {
...
+		tep.Status.VethIFaceIndex == tc.hostVeth.Index && tep.Status.VethIP == liqoconst.GatewayVethIPAddr {
 		return nil
 	}
-	if err = tc.hostNetns.Do(configureHostNetns); err != nil {
+
+	tep.Status.Connection = *con
+	tep.Status.GatewayIP = tc.podIP
+	tep.Status.VethIFaceIndex = tc.hostVeth.Index
+	tep.Status.VethIFaceName = tc.hostVeth.Name
+	tep.Status.VethIP = liqoconst.GatewayVethIPAddr
```
giorio94, line 460: Wondering whether it is necessary to store this as part of the status, given it is always assigned a constant value.

giorio94 (approved):
/lgtm (as far as I understand)

internal/liqonet/tunnel-operator/tunnel-operator.go
alacuku: That's true but there are two reasons while I moved it to the status:

1. I would like to avoid the usage of constants values in functions that configure the routes for a given remote cluster;
2. I would not exclude that in the near future we will support separate VPN tunnel for each remote cluster.

PR #1129, opened 2022-02-23
Add external resource monitor

CapacitorSet:
# Description

Adds a gRPC API and a command-line option for using an external resource monitor. It is currently intended to enable resource brokering without integrating the broker in the Liqo core, but it may be used to implement arbitrary resource manipulation logic.

It translates the ResourceReader interface (added in #986) into gRPC so that an external service may be queried for what resources to offer to a given cluster and may push update notifications (note that the ResourceUpdateNotifier mechanism was replaced with `Subscribe() returns (stream UpdateNotification)`).

# How Has This Been Tested?

- [x] Unit tests
- [x] Manual testing


go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	k8s.io/apimachinery v0.23.3
 	k8s.io/client-go v0.23.3
 	k8s.io/component-helpers v0.23.3
+	k8s.io/klog v1.0.0
```
aleoli, line 50:
```suggestion
	k8s.io/klog/v2 v1.0.0
```

it should already be imported, fix the imports in the files

pkg/liqo-controller-manager/resource-request-controller/resource-reader.proto
```diff
@@ -0,0 +1,29 @@
+syntax="proto3";
+option go_package = "./resourcerequestoperator";
```
aleoli, line 2: shouldn't we move it to a dedicated go package?

pkg/liqo-controller-manager/resource-request-controller/resource-reader.proto
```diff
@@ -0,0 +1,29 @@
...
+
+service broker {
```
aleoli, line 4: I would rename it to "resource_reader" or something like that

pkg/liqo-controller-manager/resource-request-controller/resource-reader.proto
```diff
@@ -0,0 +1,29 @@
...
+  rpc ReadResources (ReadRequest) returns (ReadResponse);
+  rpc RemoveCluster (RemoveRequest) returns (RemoveResponse);
+  rpc Subscribe (SubscribeRequest) returns (stream UpdateNotification);
+}
+
+message ReadRequest {
+  string originator = 1;
```
aleoli, line 11: can you add comments on this field and/or use a more self-explanatory name?

pkg/liqo-controller-manager/resource-request-controller/external-monitor.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	client := NewBrokerClient(conn)
+	return &ExternalResourceMonitor{
+		BrokerClient: client,
+	}, nil
+}
+
+// Register sets an update notifier.
+func (m *ExternalResourceMonitor) Register(notifier ResourceUpdateNotifier) {
+	stream, err := m.BrokerClient.Subscribe(context.Background(), &SubscribeRequest{})
```
aleoli, line 48:
```suggestion
func (m *ExternalResourceMonitor) Register(ctx context.Context, notifier ResourceUpdateNotifier) {
	stream, err := m.BrokerClient.Subscribe(ctx, &SubscribeRequest{})
```

the same for the other functions

pkg/liqo-controller-manager/resource-request-controller/external-monitor.go
```diff
@@ -0,0 +1,91 @@
...
+func (m *ExternalResourceMonitor) ReadResources(clusterID string) corev1.ResourceList {
+	response, err := m.BrokerClient.ReadResources(context.Background(), &ReadRequest{Originator: clusterID})
+	if err != nil {
+		klog.Errorf("grpc error: %s", err)
+		return corev1.ResourceList{}
+	}
+	ret := corev1.ResourceList{}
+	for key, value := range response.Resources {
+		quantity := ret[corev1.ResourceName(key)]
+		err = quantity.UnmarshalJSON([]byte(value))
```
aleoli, line 77: can we use the string description of the resource amount (eg. 100m or 2Gi, ...) instead of json? It seems to be more readable to me

pkg/liqo-controller-manager/resource-request-controller/resource-monitors/external-monitor.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// ExternalResourceMonitor is an object that keeps track of the cluster's resources.
+type ExternalResourceMonitor struct {
+	ResourceReaderClient
+}
+
+// NewExternalMonitor creates a new ExternalResourceMonitor.
+func NewExternalMonitor(address string) (*ExternalResourceMonitor, error) {
+	conn, err := grpc.Dial(address, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
aleoli, line 36: You should have a timeout, print an error, and possibly crash the pod if it is not able to connect to the external service.

You can do it with something like:
```go
ctx = context.WithTimeout(ctx, timeout)
grpc.DialWithContext(ctx, address,
  grpc.WithTransportCredentials(...),
  grpc.WithBlock())
```

pkg/liqo-controller-manager/resource-request-controller/resource-monitors/external-monitor.go
```diff
@@ -0,0 +1,93 @@
...
+func (m *ExternalResourceMonitor) Register(ctx context.Context, notifier ResourceUpdateNotifier) {
+	stream, err := m.ResourceReaderClient.Subscribe(ctx, &SubscribeRequest{})
+	if err != nil {
+		klog.Errorf("grpc error while subscribing: %s", err)
+	}
+	go func() {
+		for {
+			_, err := stream.Recv()
+			if errors.Is(err, io.EOF) {
+				break
```
aleoli, line 57: Can we log something when the function exists? It can be at higher verbosity, but it may be useful for debugging.

And, second, what will happen if the remote monitor pod restarts? The controller manager will reconnect to it? Will it restart? It is not clear to me.

pkg/liqo-controller-manager/resource-request-controller/resource-monitors/external-monitor.go
```diff
@@ -0,0 +1,93 @@
...
+				continue
+			}
+			notifier.NotifyChange()
+		}
+	}()
+}
+
+// ReadResources reads the resources from the upstream API.
+func (m *ExternalResourceMonitor) ReadResources(clusterID string) corev1.ResourceList {
+	response, err := m.ResourceReaderClient.ReadResources(context.Background(), &ReadRequest{Originator: clusterID})
```
aleoli, line 70: Do we have a context to forward? Similar for other functions

pkg/liqo-controller-manager/resource-request-controller/resource-monitors/external-monitor_test.go
```diff
@@ -0,0 +1,108 @@
...
+	Server *grpc.Server
+	ResourceReaderServer
+}
+
+var grpcSemaphore sync.WaitGroup
+
+func (b *FakeGRPCServer) Start(ctx context.Context) error {
+	address := fmt.Sprintf("%s%d", "127.0.0.1:", 7000)
+	lis, err := net.Listen("tcp", address)
+	Expect(err).ToNot(HaveOccurred())
```
aleoli, line 42: You should return the error here, and then assert it in the test statement. Look also at the other functions

pkg/liqo-controller-manager/resource-request-controller/resource-monitors/suite_test.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func destroyCluster() {
+	By("tearing down the test environment")
+	cancel()
+	Expect(testEnv.Stop()).To(Succeed())
+}
+
+var _ = BeforeSuite(func() {
+	createCluster()
```
aleoli, line 93: Do we need the env test cluster, or can we refactor it to use the fake client?

aleoli (approved):

pkg/liqo-controller-manager/resource-request-controller/resource-monitors/external-monitor.go
CapacitorSet: It reconnects with exponential backoff per gRPC docs: https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md, https://stackoverflow.com/a/39357099

pkg/liqo-controller-manager/resource-request-controller/resource-monitors/external-monitor.go
```diff
@@ -0,0 +1,98 @@
...
+
+// ExternalResourceMonitor is an object that keeps track of the cluster's resources.
+type ExternalResourceMonitor struct {
+	ResourceReaderClient
+}
+
+// NewExternalMonitor creates a new ExternalResourceMonitor.
+func NewExternalMonitor(address string) (*ExternalResourceMonitor, error) {
+	klog.Infof("Connecting to %s", address)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
aleoli, line 38: We should forward the context from the caller

```suggestion
func NewExternalMonitor(ctx context.Context, address string) (*ExternalResourceMonitor, error) {
	klog.Infof("Connecting to %s", address)
	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```

aleoli (approved):

PR #1131, opened 2022-02-24
Refactor remote namespaces creation

giorio94:
# Description

This PR refactors the creation of remote namespaces, moving the enforcement logic to the remote cluster in order to prevent issues in case of connectivity issues, as well as removing the need for the integration with capsule, which is dropped.

# How Has This Been Tested?

- [x] Manual testing
- [x] Unit testing (new + existing)
- [x] E2E testing (existing)


internal/crdReplicator/reflection/reflector.go
```diff
@@ -179,16 +179,18 @@ func (r *Reflector) stopForResource(gvr schema.GroupVersionResource) error {
 
 	klog.Infof("[%v] Stopping reflection of %v", r.remoteClusterID, gvr)
 
-	// Check if any reflected object is still present in the remote cluster
-	objects, err := rs.remote.List(labels.Everything())
-	if err != nil {
-		klog.Errorf("[%v] Failed to stop reflection of %v: %v", r.remoteClusterID, gvr, err)
-		return err
-	}
+	// Check if any object is still present in the local or in the remote cluster
```
alacuku, line 182: Being the local objects owned by components other than the CRDReplicator shouldn't be them the ones that handle lifecycle for those resources?

alacuku (approved):

pkg/liqo-controller-manager/foreign-cluster-operator/permission.go
```diff
@@ -47,12 +47,6 @@ func (r *ForeignClusterReconciler) ensurePermission(ctx context.Context, foreign
 			klog.Error(err)
 			return err
 		}
-		if r.OwnerReferencesPermissionEnforcement {
-			if err = r.NamespaceManager.UnbindIncomingClusterWideRole(ctx, remoteCluster); err != nil {
-				klog.Error(err)
-				return err
-			}
-		}
```
aleoli, line 55: Is this change necessary? (the same for the binding)

They are managing the cluster role binding that I think is not related to the tenant resource

giorio94: It seems to me the ClusterRole is related to the tenant finalizer only, if I haven't misinterpreted something. 
https://github.com/liqotech/liqo/blob/fadcbcd767b7d60959d3a53ee311d1cf91bdd4a2/pkg/tenantNamespace/clusterwide.go#L49-L56

giorio94: BTW, do the E2E tests run with the strict permissions mode?

aleoli: ok, yes you are right 👍

no, the e2e are running with the default mode (if I remember correctly). We should change it

aleoli (approved):

PR #1140, opened 2022-03-03
Set max fetchable Issues for the Changelog

aleoli:
# Description

This pr sets the maximum number of issues fetcher by the changelog generator avoiding reaching the GitHub limit


build/liqonet/Dockerfile
```diff
@@ -3,7 +3,7 @@ WORKDIR /home/rust/src
 RUN git clone https://github.com/cloudflare/boringtun.git
 WORKDIR /home/rust/src/boringtun
 RUN rustup target add x86_64-unknown-linux-musl
-RUN cargo build --bin boringtun --release
+RUN cargo build --bin boringtun-cli --release
```
giorio94, line 6: Is this change desired?

giorio94 (approved):

build/liqonet/Dockerfile
aleoli: Yes, it is a fix in the liqonet building. I can move it to another pr

giorio94: The overhead of a new PR is probably not worth for this small change. It just seemed strange to me since previously the change was not reflected in the copy operation.

PR #1146, opened 2022-03-07
Force Remote NodePort

aleoli:
# Description

This pr adds the possibility to force a remote service to use the same NodePort used in the local cluster by adding the `liqo.io/force-remote-node-port=true` annotation on that service.

NOTE: if the specified port is already taken in the remote cluster, the virtual kubelet will not be able to reflect the service to the remote cluster.

# How Has This Been Tested?

- [x] add unit test
- [x] locally

### DIFF
--- a/pkg/virtualKubelet/forge/services_test.go
+++ b/pkg/virtualKubelet/forge/services_test.go
@@ -98,7 +98,7 @@ var _ = Describe("Services Forging", func() {
 		}
 
 		DescribeTable("RemoteServiceSpec table", func(c remoteServiceTestcase) {
-			output := forge.RemoteServiceSpec(c.input.DeepCopy())
+			output := forge.RemoteServiceSpec(c.input.DeepCopy(), false)
 
 			By("should correctly replicate the core fields", func() {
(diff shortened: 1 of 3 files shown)


pkg/virtualKubelet/forge/services_test.go
```diff
@@ -121,6 +121,10 @@ var _ = Describe("Services Forging", func() {
 			input:               getService(corev1.ServiceTypeNodePort, ""),
 			expectedClusterIP:   BeNil(),
 			expectedServiceType: Equal(corev1.ServiceTypeNodePort),
+		}), Entry("NodePort Service (forcing remote NodePort)", remoteServiceTestcase{
+			input:               getService(corev1.ServiceTypeNodePort, ""),
+			expectedClusterIP:   BeNil(),
+			expectedServiceType: Equal(corev1.ServiceTypeNodePort),
```
giorio94, line 127: It seems to me this is a leftover, since it does not test anything in addition to the entry above.

giorio94 (approved):

PR #1150, opened 2022-03-09
liqoctl disconnect 

alacuku:
# Description

`liqoctl disconnect `is used to unpeer two k8s clusters that were previously peered with the 'liqoctl connect' command.

For more info check #1132

### DIFF
--- a/pkg/liqoctl/common/cluster.go
+++ b/pkg/liqoctl/common/cluster.go
@@ -25,6 +25,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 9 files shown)


pkg/liqoctl/common/cluster.go
```diff
@@ -344,6 +345,47 @@ func (c *Cluster) SetUpTenantNamespace(ctx context.Context, remoteClusterID *dis
...
+			err := c.locK8sClient.CoreV1().Namespaces().Delete(ctx, c.locTenantNamespace, metav1.DeleteOptions{})
+			if client.IgnoreNotFound(err) != nil {
+				s.Fail(fmt.Sprintf("an error occurred while deleting tenant namespace {%s} for remote cluster {%s}: %v", c.locTenantNamespace, remName, err))
+				return err
+			}
+
+			if err != nil {
+				s.Success(fmt.Sprintf("tenant namespace {%s} correctly removed for remote cluster {%s}", c.locTenantNamespace, remName))
+				return nil
+			}
```
aleoli, line 382: I would refactor these if to make them more readable. Something like:
```
switch
case k8serrors.IsNotFound(err)
  return
case err != nil
  fail
default
  sleep
```

pkg/liqoctl/common/cluster.go
```diff
@@ -620,22 +662,82 @@ func (c *Cluster) MapProxyIPForCluster(ctx context.Context, ipamClient ipam.Ipam
...
+	}
+	svc, err := liqogetters.GetServiceByLabel(ctx, c.locCtrlRunClient, c.namespace, selector)
+	if client.IgnoreNotFound(err) != nil {
+		s.Fail(fmt.Sprintf("an error occurred while retrieving proxy endpoint: %v", err))
+		return err
+	}
+	if k8serrors.IsNotFound(err) {
+		s.Warning(fmt.Sprintf("service for proxy not found, unable to unmap proxy ip for cluster {%s}", clusterName))
+		return nil
+	}
```
aleoli, line 688: the same here, I would use a switch to make clear that the conditions are mutually exclusive

aleoli (approved):

PR #1160, opened 2022-03-16
Liqoctl: Improve Install/Uninstall Outputs

aleoli:
# Description

This pr improves the output of the liqoctl install and uninstall command using the pterm Spinners instead of fmt/klog

# How Has This Been Tested?

- [x] locally


pkg/liqoctl/install/utils.go
```diff
@@ -62,8 +61,7 @@ func getProviderInstance(providerType string) provider.InstallProviderInterface
 func initHelmClient(config *rest.Config, arguments *provider.CommonArguments) (helm.Client, error) {
 	helmClient, err := InitializeHelmClientWithRepo(config, arguments)
 	if err != nil {
-		fmt.Printf("Unable to create helmClient: %s", err)
-		return nil, err
+		return nil, fmt.Errorf("unable to create helmClient: %w", err)
```
giorio94, line 64:
```suggestion
		return nil, fmt.Errorf("unable to create helm client: %w", err)
```

pkg/liqoctl/install/aks/provider.go
```diff
@@ -78,27 +77,23 @@ func (k *aksProvider) ValidateCommandArguments(flags *flag.FlagSet) (err error)
 	if err != nil {
 		return err
 	}
-	klog.V(3).Infof("AKS SubscriptionID: %v", k.subscriptionID)
```
giorio94, line 81: I would personally keep the possibility to show debug information in case the `--debug` flag is set (could use an helper function, to make it unrelated from klog).

pkg/liqoctl/install/helm.go
```diff
@@ -33,7 +32,7 @@ func InitializeHelmClientWithRepo(config *rest.Config, commonArgs *provider.Comm
 			RepositoryCache:  liqoHelmCachePath,
 			Debug:            commonArgs.Debug,
 			Linting:          false,
-			DebugLog:         klog.V(4).Infof,
+			DebugLog:         func(format string, v ...interface{}) {},
```
giorio94, line 35: Same here

pkg/liqoctl/install/handler.go
```diff
@@ -59,53 +72,66 @@ func HandleInstallCommand(ctx context.Context, cmd *cobra.Command, baseCommand,
...
 	err = installOrUpdate(ctx, helmClient, providerInstance, commonArgs)
 	if err != nil {
+		s.Fail("Error installing or upgrading Liqo: ", err)
 		return err
 	}
 
 	switch {
 	case !commonArgs.DumpValues && !commonArgs.DryRun:
-		fmt.Printf("* All Set! You can use Liqo now! 🚀\n")
+		s.Success("* All Set! You can use Liqo now! 🚀")
```
giorio94, line 129:
```suggestion
		s.Success("All Set! You can use Liqo now! 🚀")
```

In my terminal, the emoji is also not displayed correctly.

giorio94 (commented):
Just to let you know, the helm error is still scrambling the output:

```
 [SUCCESS]  Configuration loaded                                                                                                     
 [SUCCESS]  Installer initialized                                                                                                    
 [SUCCESS]  Chart parameters extracted                                                                                               
⢀⡱ Installing or Upgrading Liqo... (this may take few minutes) (0s)E0317 10:29:09.798766 3389384 memcache.go:196] continuing
⢆⡱ Installing or Upgrading Liqo... (this may take few minutes) (1s)E0317 10:29:10.143795 3389384 memcache.go:196] continuing
E0317 10:29:10.143877 3389384 memcache.go:196] continuing
 [SUCCESS]  * All Set! You can use Liqo now! 🚀     
```

aleoli:
Yes, I know... 😓 we need further improvements, but at least the spinner is running and, I think, the user can understand that the installation is not completely broken

giorio94:
> Yes, I know... sweat we need further improvements, but at least the spinner is running and, I think, the user can understand that the installation is not completely broken

In case klog is no longer used it should be possible to configure a different output stream as in the tests to hide that

cmd/liqoctl/cmd/root.go
```diff
@@ -39,15 +41,20 @@ func NewRootCommand(ctx context.Context) *cobra.Command {
 	klogFlagset := flag.NewFlagSet("klog", flag.PanicOnError)
 	klog.InitFlags(klogFlagset)
 	klogFlagset.VisitAll(func(f *flag.Flag) {
-		if f.Name == "v" {
-			rootCmd.PersistentFlags().AddGoFlag(f)
+		if f.Name == "stderrthreshold" {
+			utilruntime.Must(f.Value.Set("FATAL"))
```
giorio94, line 45: Could be worth to add a comment and specify this is required to silence the helm library messages.

pkg/liqoctl/install/aks/provider.go
```diff
@@ -78,27 +78,26 @@ func (k *aksProvider) ValidateCommandArguments(flags *flag.FlagSet) (err error)
...
 		return err
 	}
-	klog.V(3).Infof("AKS ResourceGroupName: %v", k.resourceGroupName)
+	logsutils.Infof("AKS ResourceGroupName: %v", k.resourceGroupName)
 
 	k.resourceName, err = flags.GetString(resourceNameFlag)
 	if err != nil {
 		return err
 	}
-	klog.V(3).Infof("AKS ResourceName: %v", k.resourceName)
```
giorio94, line 101: Should this be restored?

pkg/liqoctl/install/handler.go
```diff
@@ -33,23 +33,35 @@ import (
...
 // with provider-specific values. Finally, it performs the installation on the target cluster.
 func HandleInstallCommand(ctx context.Context, cmd *cobra.Command, baseCommand, providerName string) error {
-	if !klog.V(4).Enabled() {
-		klog.SetLogFilter(logsutils.LogFilter{})
-	}
+	verbose, err := cmd.Flags().GetBool("verbose")
+	utilruntime.Must(err)
+
+	printer := common.NewPrinter("", common.Cluster1Color)
+	logsutils.SetupLogger(printer, verbose)
```
giorio94, line 40: Could it make sense to move this part to the root, to avoid having to call it for all commands?

giorio94 (approved):
/lgtm

PR #1197, opened 2022-04-13
fix liqoctl disconnect race condition

alacuku:
# Description

Fix race condition when disconnect two clusters using `liqoctl disconnect`. Removing the networking caused the `natmapping` resource to be deleted. When `unmapping` auth and proxy services an error like 
"_error occurred while unmapping proxy address {10.43.4.204} for cluster {dawn-butterfly}: rpc error: code = Unknown desc = cannot unmap the IP of endpoint 10.43.4.204: NatMapping for cluster 7efeb93b-2bf6-4a0a-ad0b-745848bba366 must be initialized first_" 
is returned.

### DIFF
--- a/pkg/liqoctl/common/cluster.go
+++ b/pkg/liqoctl/common/cluster.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pterm/pterm"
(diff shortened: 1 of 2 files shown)


pkg/liqoctl/common/cluster.go
```diff
@@ -765,12 +766,16 @@ func mapServiceForCluster(ctx context.Context, ipamClient ipam.IpamClient, ipToB
...
 	remoteCluster *discoveryv1alpha1.ClusterIdentity) error {
+	var err error
+
 	unMapRequest := &ipam.UnmapRequest{
 		ClusterID: remoteCluster.ClusterID,
 		Ip:        ipToBeUnmapped,
 	}
 
-	_, err := ipamClient.UnmapEndpointIP(ctx, unMapRequest)
+	if _, err = ipamClient.UnmapEndpointIP(ctx, unMapRequest); err != nil && strings.Contains(err.Error(), "must be initialized first") {
```
giorio94, line 776: Would personally add a comment to point out why this might happen and an example of the full error message (as in the PR description), for easier retrieval in the future.

giorio94 (approved):

PR #1220, opened 2022-04-28
Feedback on new VPN connections to remote peers

alacuku:
# Description

This PR handles three issues:

- when connected to remote peer the VPN connection status is correctly showed in the status of the related `tunnelendpoint`;
- the status of the VPN connections is now correctly reported in the conditions of the related `foreigncluster`;
- `liqoctl connect` waits for the VPN connection to be established; it also emits a warning if the endpoint addresse used for the VPN connection belongs to a private subnet (RFC 1918).

# How Has This Been Tested?

- [x] Existing tests


pkg/liqo-controller-manager/foreign-cluster-operator/auth.go
```diff
@@ -216,3 +219,26 @@ func (r *ForeignClusterReconciler) transport(insecureSkipTLSVerify bool) *http.T
...
+			{
+				Key:      discovery.AuthTokenLabel,
+				Operator: metav1.LabelSelectorOpExists,
+			},
+		},
+	})
+
+	if err != nil {
+		klog.Fatal(err)
+	}
```
giorio94, line 241: Would replace with `utilsruntime.Must(err)`

pkg/liqo-controller-manager/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -106,6 +98,8 @@ type ForeignClusterReconciler struct {
 
 	InsecureTransport *http.Transport
 	SecureTransport   *http.Transport
+
+	ForeignClusters sync.Map
```
giorio94, line 102: Would suggest to add a comment about the content of the map (i.e., key and value), for future reference.

pkg/liqo-controller-manager/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -197,6 +191,16 @@ func (r *ForeignClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	// When the networking is not managed by liqo than the networking check is done before
+	// we do the other checks that could rely on networking.
+	if foreignCluster.Spec.NetworkingEnabled == discoveryv1alpha1.NetworkingEnabledNo {
+		// check for TunnelEndpoints
+		if err = r.checkTEP(ctx, &foreignCluster); err != nil {
+			klog.Error(err)
+			return ctrl.Result{}, err
+		}
+		tracer.Step("Checked the TunnelEndpoint status")
+	}
```
giorio94, line 203: Would personally avoid duplicating the tep status check, doing it always here regardless of the peering type (AFAIU it should cause no problems).

pkg/liqo-controller-manager/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -615,11 +541,21 @@ func (r *ForeignClusterReconciler) checkTEP(ctx context.Context,
...
-	case tunnelEndpointNotFoundReason, tunnelEndpointAvailableReason, tunnelEndpointConnectingReason, tunnelEndpointErrorReason:
-		return true
-	default:
-		return false
+func (r *ForeignClusterReconciler) foreignclusterEnqueuer(obj client.Object) []ctrl.Request {
+	gvks, _, err := r.Scheme.ObjectKinds(obj)
+	// Should never happen, but if it happens we panic.
+	utilruntime.Must(err)
+
+	klog.V(4).Infof("handling resource %q of type %q", klog.KObj(obj), gvks[0].String())
```
giorio94, line 549: This risks to panic in case no GVKs are found (although it should never happen IIUC).

pkg/liqo-controller-manager/foreign-cluster-operator/foreign-cluster-controller.go
```diff
@@ -288,6 +285,7 @@ func (r *ForeignClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			klog.Error(err)
 			return ctrl.Result{}, err
 		}
+		r.ForeignClusters.Delete(foreignCluster.Status.TenantNamespace.Local)
```
giorio94, line 288: I am missing where the element is added to the map. Additionally, I would move the deletion to the check at the beginning of the reconcile, to catch all events (here it is called only with auto-discovery deletions).

giorio94 (approved):
/lgtm

PR #1223, opened 2022-05-02
Add Global Ingress Example

aleoli:
# Description

Fixes #(issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B


examples/quick-start/manifests/demo-application.yaml
```diff
@@ -0,0 +1,694 @@
+# Copyright 2018 Google LLC
```
giorio94, line 1: Would personally try to use kustomize to leverage the upstream manifests and reduce/highlight the changes. Should also drop the custom image, which seems to still be there.

giorio94, line 1: Would personally try to use kustomize to leverage the upstream manifests and reduce/highlight the changes. Should also drop the custom image, which seems to still be there.

examples/offloading-with-policies/manifests/testing-deploy.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: apps/v1
```
giorio94, line 1: Would rename this file to `deploy.yaml`, as testing-deploy is longer, ugly (at least IMHO) and adds no value.

examples/offloading-with-policies/manifests/testing-deploy.yaml
```diff
@@ -0,0 +1,65 @@
...
+kind: Deployment
+metadata:
+  name: pod-south
```
giorio94, line 4: Should be probably modified to `app-south`, which is more generic (as this is no longer a single pod). Same for all occurrences.

PR #1224, opened 2022-05-03
Liqo Controller Manager: ShadowPod validating webhook

cannarelladev:
# Description

In a "two-cluster" peering scenario, where one Cluster plays the role of "Customer" and the other one of "Cloud Provider", it is important for the two entities that the Resource plan agreed is as much as possible guaranteed and controlled.

- The Cloud Provider has the need to control what happen in own cluster, that  resource usage limits agreed are respected (also with a toleration) and that nothing exceeds the Offer provided

in the other hand

- The Customer wants that the ResourceOffer Plan "buyed" is always guaranteed (also with a toleration) respecting what has been agreed at the beginning, it has the need to be protected about what is defined in kind of "resource contract"

The goal of a Resource Validator system is to manage all this requirements, defining a new optional way of peering between clusters

# What's needed

This PR introduces some new elements to allow the management of "Customer" offloaded resources on host cluster (that ideally represents a Cloud Provider)

- [x] Validating Webhook for incoming ShadowPods based on existing ResourceOffer
- [x] Used/Free resources calculator after each Offloading Request / Deletion
- [x] Cached information to decrease the overhead generated by each validation check after an Offloading Request
- [x] Cache mutex for concurrent access
- [x] Scheduled cache consistency refresh

## TODO
- [x] improve cache data structure
- [x] Improve logging using klog


pkg/resourcevalidator/webhook.go
```diff
@@ -0,0 +1,241 @@
...
+	// Check existence and get shadow pod origin Cluster ID label
+	spClusterID, found := shadowpod.Labels[forge.LiqoOriginClusterIDKey]
+	if !found {
+		return admission.Denied("missing origin Cluster ID label")
+	}
+
+	// Get ShadowPod Namespace
+	namespace := &corev1.Namespace{}
+	if err := spv.Client.Get(ctx, client.ObjectKey{Name: shadowpod.GetNamespace()}, namespace); err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
```
giorio94, line 102: This should be an InternalServerError, and not a client error. The error message shall be enriched before being returned to the caller.

pkg/resourcevalidator/webhook.go
```diff
@@ -0,0 +1,241 @@
...
+	case admissionv1.Delete:
+		return spv.HandleDelete(ctx, req, shadowpod, nsClusterID, pointer.BoolDeref(req.DryRun, false))
+	default:
+		return admission.Denied("Unsupported operation")
+	}
+}
+
+// HandleCreate is the function in charge of handling Creation requests.
+//nolint:gocritic // the signature of this method is imposed by controller runtime.
+func (spv *ShadowPodValidator) HandleCreate(ctx context.Context, req admission.Request, shadowpod *vkv1alpha1.ShadowPod,
```
giorio94, line 128: Since you are passing the entire req object, dryrun could be extracted internally.

pkg/resourcevalidator/webhook.go
```diff
@@ -0,0 +1,241 @@
...
+		resourceofferlog.Info(error.Error())
+		return admission.Errored(http.StatusInternalServerError, error)
+	}
+
+	// Get ResourceOffer Quota
+	roQuota := getQuotaFromResourceOffer(resourceoffer)
+
+	resourceofferlog.Info(fmt.Sprintf("ResourceOffer found for clusterID %s with %s ", clusterID, quotaFormatter(roQuota, "Quota")))
+
+	peeringInfo := getOrCreatePeeringInfo(spv.PeeringCache, clusterID, resourceoffer.OwnerReferences[0].Name, roQuota)
```
giorio94, line 150: It is risky to extract the cluster name this way, since it might panic if the ownerreference is not present. Since it appears to me it is only used for logging, I would drop the cluster name part for the sake of simplicity (it would require some more refactoring to be integrated cleanly).

pkg/resourcevalidator/webhook.go
```diff
@@ -0,0 +1,241 @@
...
+	obj := req.Object
+	oldObj := req.OldObject
+	shadowpod, err := spv.DecodeShadowPod(obj)
+	oldShadowpod, oldErr := spv.DecodeShadowPod(oldObj)
+	if err != nil || oldErr != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	if pod.CheckShadowPodUpdate(&shadowpod.Spec.Pod, &oldShadowpod.Spec.Pod) {
+		return admission.Allowed("")
+	}
```
giorio94, line 182: You should ensure that the cluster ID label does not get changed (you should be able to reuse the common function extracted from the Handle logic above).

pkg/resourcevalidator/webhook.go
```diff
@@ -0,0 +1,241 @@
...
+	return admission.Denied("")
+}
+
+// HandleDelete is the function in charge of handling Deletion requests.
+//nolint:gocritic // the signature of this method is imposed by controller runtime.
+func (spv *ShadowPodValidator) HandleDelete(ctx context.Context, req admission.Request, shadowpod *vkv1alpha1.ShadowPod,
+	clusterID string, dryRun bool) admission.Response {
+	peeringInfo, foundPI := spv.PeeringCache.getPeeringFromCache(clusterID)
+	if !foundPI {
```
giorio94, line 191: Would add a comment stating why a peeringinfo structure needs to be created.

pkg/resourcevalidator/webhook.go
```diff
@@ -0,0 +1,241 @@
...
+			return admission.Errored(http.StatusInternalServerError, error)
+		}
+		roQuota := getQuotaFromResourceOffer(resourceoffer)
+		resourceofferlog.Info(fmt.Sprintf("ResourceOffer found for clusterID %s with %s ", clusterID, quotaFormatter(roQuota, "Quota")))
+
+		peeringInfo = createPeeringInfo(clusterID, resourceoffer.OwnerReferences[0].Name, roQuota)
+		spv.PeeringCache.addPeeringToCache(clusterID, peeringInfo)
+	}
+
+	spd, foundSPD := peeringInfo.getShadowPodDescription(shadowpod.GetName())
```
giorio94, line 209: Same here about possible race conditions since the mutex is not locked early.

pkg/resourcevalidator/webhook.go
```diff
@@ -0,0 +1,241 @@
...
+	if !foundSPD {
+		spd = createShadowPodDescription(shadowpod.GetName(), string(shadowpod.GetUID()), getQuotaFromShadowPod(shadowpod))
+		if foundPI {
+			spd.terminate()
+			peeringInfo.Lock()
+			peeringInfo.updateShadowPod(spd)
+			peeringInfo.Unlock()
+			error := fmt.Errorf("cannot delete: ShadowPod %s not found (Maybe Cache problem)", shadowpod.GetName())
+			webhooklog.Info(error.Error())
+			return admission.Errored(http.StatusBadRequest, error)
```
giorio94, line 221: Deletions shall never be forbidden.

pkg/resourcevalidator/sp_description.go
```diff
@@ -0,0 +1,61 @@
...
+	return spd.Quota
+}
+
+func (spd ShadowPodDescription) getUID() string {
+	return spd.UID
+}
+
+func (spd ShadowPodDescription) getName() string {
+	return spd.Name
+}
```
giorio94, line 61: Do not create getters for already available fields, as redundant.

pkg/resourcevalidator/sp_description.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// ShadowPodDescription is a struct that contains the main informations about a shadow pod.
+type ShadowPodDescription struct {
+	Name      string
+	UID       string
```
giorio94, line 26: Would use the appropriate type, instead of string.

pkg/resourcevalidator/sp_description.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// ShadowPodDescription is a struct that contains the main informations about a shadow pod.
+type ShadowPodDescription struct {
+	Name      string
```
giorio94, line 25: Name shall be the combination of both namespace and name, otherwise it is not guaranteed to be unique.

pkg/resourcevalidator/common.go
```diff
@@ -0,0 +1,117 @@
...
+	return result
+} */
+
+func getQuotaFromResourceOffer(resourceoffer *sharing.ResourceOffer) v1.ResourceList {
+	resources := v1.ResourceList{}
+
+	for key, value := range resourceoffer.Spec.ResourceQuota.Hard {
+		resources[key] = value
+	}
+	return resources
```
giorio94, line 72: Why not accessing the resources directly (`resourceoffer.Spec.ResourceQuota.Hard.DeepCopy()`)? Then, the function could also be dropped altogether.

pkg/resourcevalidator/common.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func getQuotaFromShadowPod(shadowpod *vkv1alpha1.ShadowPod) v1.ResourceList {
+	resources := v1.ResourceList{}
+	for key, value := range shadowpod.Spec.Pod.Containers[0].Resources.Limits {
+		resources[key] = value
+	}
```
giorio94, line 79: There is no guarantee a pod has a single container. BTW, admission shall be prevented if at least one container has no resource limits specified.

pkg/resourcevalidator/common.go
```diff
@@ -0,0 +1,117 @@
...
+	return resources
+}
+
+func (spv *ShadowPodValidator) getShadowPodListByClusterID(ctx context.Context, shadowPodList *vkv1alpha1.ShadowPodList, clusterID string) error {
```
giorio94, line 83: 1. Do not alias variables if not necessary. That only makes the logic harder to understand
2. Do not pass parameters if you are only using them as return values. Just return them
3. Do not use hard-coded strings that refer to defined constants.

pkg/resourcevalidator/info.go
```diff
@@ -0,0 +1,246 @@
...
+ * PeeringInfo methods
+ */
+
+func (pi *peeringInfo) Lock() {
+	pi.piMutex.Lock()
+}
+
+func (pi *peeringInfo) Unlock() {
+	pi.piMutex.Unlock()
+}
```
giorio94, line 47: Do not create trivial aliases for already existing methods.

pkg/resourcevalidator/info.go
```diff
@@ -0,0 +1,246 @@
...
+	if !found {
+		webhooklog.Info(fmt.Sprintf("\t\tPeeringInfo not found for clusterID %s", clusterID))
+		peeringInfo = createPeeringInfo(clusterID, clusterName, roQuota)
+		webhooklog.Info(fmt.Sprintf("\t\tPeeringInfo created for clusterID %s", clusterID))
+		webhooklog.Info(quotaFormatter(peeringInfo.getQuota(), "\t\tNew PeeringInfo Quota limits"))
+		webhooklog.Info(quotaFormatter(peeringInfo.getUsedQuota(), "\t\tNew PeeringInfo UsedQuota limits"))
+		webhooklog.Info(quotaFormatter(peeringInfo.getFreeQuota(), "\t\tNew PeeringInfo FreeQuota limits"))
+		cache.addPeeringToCache(clusterID, peeringInfo)
+	}
+	return peeringInfo
```
giorio94, line 76: This is not guaranteed to be atomic. You should use the `LoadOrStore` method provided by the SyncMap.

pkg/resourcevalidator/info.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func (pi *peeringInfo) subtractResources(resources v1.ResourceList) {
+	for key, val := range resources {
+		if prevFree, ok := pi.FreePeeringQuota[key]; ok {
+			prevFree.Sub(val)
+			pi.FreePeeringQuota[key] = prevFree
+		} else {
+			pi.FreePeeringQuota[key] = val.DeepCopy()
+		}
```
giorio94, line 98: Shouldn't pods demanding for resources not offered by the resource offer be forbidden?

pkg/resourcevalidator/info.go
```diff
@@ -0,0 +1,246 @@
...
+	webhookrefreshlog.Info(quotaFormatter(pi.PeeringQuota, "\t\tOld Peering Quota %s"))
+	pi.PeeringQuota = resources.DeepCopy()
+	webhookrefreshlog.Info(quotaFormatter(pi.PeeringQuota, "\t\tNew Peering Quota %s"))
+}
+
+func (pi *peeringInfo) updateFreePeeringQuota(resources v1.ResourceList) {
+	webhookrefreshlog.Info(quotaFormatter(pi.FreePeeringQuota, "\t\tOld Free Quota %s"))
+	pi.FreePeeringQuota = resources.DeepCopy()
+	webhookrefreshlog.Info(quotaFormatter(pi.FreePeeringQuota, "\t\tNew Free Quota %s"))
+}
```
giorio94, line 134: Seem trivial methods to me.

pkg/resourcevalidator/info.go
```diff
@@ -0,0 +1,246 @@
...
+func (pi *peeringInfo) getShadowPodDescription(uid string) (ShadowPodDescription, bool) {
+	spd, ok := pi.SPList[uid]
+	return spd, ok
+}
+
+func (pi *peeringInfo) getAllShadowPodDescription() map[string]ShadowPodDescription {
+	return pi.SPList
+}
+
+func (pi *peeringInfo) testAndUpdatePeeringInfo(spd ShadowPodDescription, operation admissionv1.Operation, dryRun bool) error {
```
giorio94, line 184: Would personally create two separate funcitons for increased clarity, as practically no code is reused.

pkg/resourcevalidator/info.go
```diff
@@ -0,0 +1,246 @@
...
+	cachelog.Info(fmt.Sprintf("\tOperation: %s", operation))
+	pi.Lock()
```
giorio94, line 186:
```suggestion
	pi.Lock()
	defer pi.Unlock()
```

pkg/resourcevalidator/cache.go
```diff
@@ -0,0 +1,285 @@
...
+// TODO: refresh has to be an update and not a new cache generation.
+func (spv *ShadowPodValidator) refreshCache(ctx context.Context) (done bool, err error) {
+	webhookrefreshlog.Info("[ REFRESH ] Refreshing peering cache")
+	c := spv.Client
+	shadowPodList := vkv1alpha1.ShadowPodList{}
+	resourceOfferList := sharing.ResourceOfferList{}
+	if err := c.List(ctx, &resourceOfferList, &client.ListOptions{}); err != nil {
+		return true, err
+	}
+	if !spv.PeeringCache.ready {
```
giorio94, line 70: In case the first iteration is completely different it is better to create two separate functions for increased clarity.

pkg/resourcevalidator/webhook.go
cannarelladev: internally it is used also to eventually create the PeeringInfo

pkg/resourcevalidator/common.go
cannarelladev: so in this case it could be better to do a sum of all the container resources?

giorio94: Yes, checking also that the ones mandatory are present in all containers.

pkg/resourcevalidator/sp_description.go
cannarelladev: so i have to use NamspacedName?

giorio94: Yes, that could be a solution

pkg/resourcevalidator/info.go
```diff
@@ -0,0 +1,197 @@
...
+		SPList:           map[string]ShadowPodDescription{},
+		PeeringQuota:     resources.DeepCopy(),
+		FreePeeringQuota: resources.DeepCopy(),
+		piMutex:          sync.RWMutex{},
+		LastUpdateTime:   time.Now().Format(time.RFC3339),
+	}
+}
+
+func (pi *peeringInfo) addShadowPod(spd ShadowPodDescription) {
+	pi.SPList[spd.NamespacedName.Name] = spd
```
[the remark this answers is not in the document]
cannarelladev: So what you suggest?

pkg/resourcevalidator/common.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, initContainer := range shadowpod.Spec.Pod.InitContainers {
+		if initContainer.Resources.Limits != nil || !validate {
+			for key, value := range initContainer.Resources.Limits {
+				if prevResources, ok := resources[key]; ok {
+					prevResources.Add(value)
+					resources[key] = prevResources
+				} else {
+					resources[key] = value.DeepCopy()
+				}
+			}
```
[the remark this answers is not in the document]
cannarelladev: so can i remove the entire "for" cycle for the initContainers?

pkg/resourcevalidator/info.go
```diff
@@ -0,0 +1,197 @@
...
+	admissionv1 "k8s.io/api/admission/v1"
+	v1 "k8s.io/api/core/v1"
+)
+
+type peeringInfo struct {
+	clusterID        string
+	clusterName      string
+	SPList           map[string]ShadowPodDescription
+	PeeringQuota     v1.ResourceList
+	FreePeeringQuota v1.ResourceList
```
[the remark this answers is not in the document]
cannarelladev: maybe i didn't understand, sorry

cannarelladev:
@giorio94 when you are available i leaved 3 question on your review comments

pkg/resourcevalidator/info.go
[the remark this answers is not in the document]
giorio94: I would personally use either the namespacedname as a key. Alternatively, you could use the string returned by `cache.MetaNamespaceKeyFunc(nm)` (essentially `namespace/name`)

pkg/resourcevalidator/common.go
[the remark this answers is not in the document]
giorio94: No, you should to adapt the logic as for the document above. IIRC, it should be something along the lines of max(sum(container_resources), max(init_container_resources)).

pkg/resourcevalidator/info.go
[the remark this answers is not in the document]
giorio94: Here, you are storing `PeeringQuota` (i.e., the total amount of resources offered), and `FreePeeringQuota` (i.e., total - used). I would personally feel easier to consider `PeeringQuota` and `UsedPeeringQuota` (i.e., used), as the latter is unrelated from the total amount of resources offered. Otherwise, in case the total changes, also the free amount needs to be updated. Just personal feeling though.

pkg/resourcevalidator/info.go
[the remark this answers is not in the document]
cannarelladev: if i use a value of type NamespacedName as key, will it match?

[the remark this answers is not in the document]
giorio94: In which sense?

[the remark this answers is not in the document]
giorio94: Structs can be map keys in golang, as long as they are comparable. BTW, if you prefer keeping it a string, the NamespacedName type has a string method which does exactly like `cache.MetaNamespaceKeyFunc(nm)`

pkg/resourcevalidator/common.go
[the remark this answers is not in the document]
cannarelladev: 1. In case of the init containers doesn't define any limit for CPU or Memory i will return an error as defined for the containers?
2. max(init_container_resources) means thats i have to select the maximum between initContainers and not the sum of them? if yes, which criteria i have to use to select one between InitContainer 1 (CPU: 1, Memory: 2) and InitContainer 2 (CPU: 2, Memory: 1), who "wins"? 
maybe it is related to the maximum of each type of resource type, right (in this case the result is CPU: 2, Memory: 2)?

pkg/resourcevalidator/info.go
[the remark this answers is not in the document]
cannarelladev: Yeah, now i understand. This choice was made to speed up the resource check of each ShadowPod. in this way if `ShadowPod resources` fit the `FreePeeringQuota` it is immediately accepted. Even if in the way you suggest i can also firstly do a subtraction to obtain the `FreePeeringQuota`. Tell me which strategy do you prefer, for me is the same and change does not produce a particular effort

pkg/resourcevalidator/info.go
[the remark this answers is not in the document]
cannarelladev: ok thanks

pkg/resourcevalidator/common.go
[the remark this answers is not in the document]
giorio94: I would try to match as much as possible the behavior of standard resource quotas. You should double check, but I think the correct approach is:
1. returning an error
2. considering the max for each resource type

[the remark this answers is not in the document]
cannarelladev: perfect! Thank you

pkg/resourcevalidator/info.go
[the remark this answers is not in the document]
giorio94: I think that comparing against 0, or a value (i.e., the total) should introduce no different overhead. Yet, the latter approach seems less prone to possible drifts, at least to me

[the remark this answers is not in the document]
cannarelladev: ok

pkg/resourcevalidator/sp_description.go
```diff
@@ -0,0 +1,105 @@
...
+	defer pi.piMutex.Unlock()
+	key, _ := cache.MetaNamespaceKeyFunc(types.NamespacedName{Name: sp.Name, Namespace: sp.Namespace})
+	spd, found := pi.getShadowPodDescription(key)
+	check := spd.UID == sp.GetUID()
+	if !found {
+		spQuota, _ := getQuotaFromShadowPod(sp, false)
+		// Create a new ShadowPodDescription for caching purposes
+		spd = createShadowPodDescription(sp.GetName(), sp.GetNamespace(), sp.GetUID(), spQuota)
+		spd.terminate()
+		pi.updateShadowPod(spd)
```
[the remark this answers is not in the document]
cannarelladev: No because in this case there is a mismatch between the request and the cache status, so it update the cache in order to allow the "refresher" to align it successively without touching the available/used resources that are already aligned.

cannarelladev:
waiting for another pipe of comments 😄

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+		return err
+	}
+	cacherefreshlog.Info("Cache initialization started")
+	for i := range resourceOfferList.Items {
+		ro := &resourceOfferList.Items[i]
+		clusterID := ro.Labels[discovery.ClusterIDLabel]
+		clusterName := retrieveClusterName(ctx, spv.client, clusterID)
+		if clusterID == "" {
+			return fmt.Errorf("ResourceOffer %s has no cluster id", clusterName)
+		}
```
giorio94, line 61: Should probably avoid blocking completely and just skip this entry, to prevent a malformed request from breaking the creation of shadow pods in the entire cluster.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+		// Check if the ShadowPod is in terminating phase in cache and has been already terminated/deleted from the cluster
+		// if true ShadowPodDescription can be also deleted from the cache
+		if _, stillPresent := spMap[shadowPodDescription.namespacedName.String()]; !shadowPodDescription.running && !stillPresent {
+			pi.removeShadowPod(shadowPodDescription)
+			cacherefreshlog.Info(fmt.Sprintf("ShadowPodDescription %s removed from cache", shadowPodDescription.namespacedName.String()))
+		}
+	}
+}
+
+func alignExistingShadowPods(shadowPodList *vkv1alpha1.ShadowPodList, pi *peeringInfo) {
```
giorio94, line 143: Would expect this be a method of the peering info structure (same for the others).

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+	spv.peeringCache.peeringInfo.Range(
+		func(key, value interface{}) bool {
+			pi := value.(*peeringInfo)
+			clusterID := key.(string)
+			// Get the List of shadow pods running on the cluster
+			shadowPodList, err := spv.getShadowPodListByClusterID(ctx, clusterID)
+			if err != nil {
+				cacherefreshlog.Info(err.Error())
+				return false
+			}
```
giorio94, line 104: Also in this case I would continue with the next element, instead of blocking completely.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+			// TODO: check if it is the correct logic
+			// Errors are intentionally ignored here. Pod is already running on cluster.
+			spQuota, _ := getQuotaFromShadowPod(shadowPod, false)
+			pi.addShadowPod(createShadowPodDescription(shadowPod.GetName(), shadowPod.GetNamespace(), shadowPod.GetUID(), spQuota))
+			cacherefreshlog.Info(fmt.Sprintf("ShadowPod %s added in cache", nsname.String()))
+		}
+	}
+}
+
+func checkShadowPodsAlignment(shadowPodList *vkv1alpha1.ShadowPodList, pi *peeringInfo, spMap map[string]string) {
```
giorio94, line 161: The map should be returned, and not passed as parameter.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+	for i := range shadowPodList.Items {
+		shadowPod := &shadowPodList.Items[i]
+		nsname := types.NamespacedName{Name: shadowPod.Name, Namespace: shadowPod.Namespace}
+		_, found := pi.spDescriptionList[nsname.String()]
+		if !found {
+			cacherefreshlog.Info(fmt.Sprintf("Warning: ShadowPod %s not found in cache", nsname.String()))
+			// TODO: this should never happen, but maybe we need to manage also this case better
```
giorio94, line 168: This might happen in case the webhook is not configured in fail mode. It would be probably appropriate to add that sp to the cache and align the resources.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+			continue
+		}
+		// Saving a list all the ShadowPods in the cluster that are also in cache
+		spMap[nsname.String()] = nsname.String()
```
giorio94, line 172: In case the value is not needed it is suggested to use a map[string]struct{} for increased efficiency.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func alignTerminatingShadowPodDescriptions(pi *peeringInfo, spMap map[string]string) {
+	for _, shadowPodDescription := range pi.spDescriptionList {
+		// Check if the ShadowPod is in terminating phase in cache and has been already terminated/deleted from the cluster
+		// if true ShadowPodDescription can be also deleted from the cache
+		if _, stillPresent := spMap[shadowPodDescription.namespacedName.String()]; !shadowPodDescription.running && !stillPresent {
+			pi.removeShadowPod(shadowPodDescription)
+			cacherefreshlog.Info(fmt.Sprintf("ShadowPodDescription %s removed from cache", shadowPodDescription.namespacedName.String()))
+		}
```
giorio94, line 139: Should also handle the case in which the sp in cache is still running (i.e., the deletion event was missed), but the object does no longer exist. Also this can occur if the webhook is not in fail mode.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func checkAlignmentResourceOfferPeeringInfo(ctx context.Context, spv *Validator) error {
+	resourceOfferList := sharing.ResourceOfferList{}
+	c := spv.client
+
+	// Get the List of resource offers
+	if err := c.List(ctx, &resourceOfferList); err != nil {
+		return err
+	}
```
giorio94, line 183: Why is this performed again?

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+func (spv *Validator) refreshCache(ctx context.Context) (done bool, err error) {
+	if !spv.peeringCache.ready {
+		err = spv.initializeCache(ctx)
+		if err != nil {
+			cacherefreshlog.Info(err.Error())
+		}
+		return false, nil
+	}
+	resourceOfferList := sharing.ResourceOfferList{}
+	if err := spv.client.List(ctx, &resourceOfferList, &client.ListOptions{}); err != nil {
```
giorio94, line 89: Should add the appropriate label selector to filter only local objects. Could also be moved above to avoid duplicating it also for cache initialization.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,262 @@
...
+			// Get the List of ShadowPods running on the cluster
+			shadowPodList, err := spv.getShadowPodListByClusterID(ctx, clusterID)
+			if err != nil {
+				return err
+			}
+			alignExistingShadowPods(shadowPodList, newPI)
+			// Add the PeeringInfo to the cache
+			spv.peeringCache.peeringInfo.Store(clusterID, newPI)
+		}
+	}
```
giorio94, line 208: Could lead to race conditions if also added by a parallel validating thread.

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo.go
```diff
@@ -0,0 +1,168 @@
...
+	"fmt"
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type peeringInfo struct {
+	clusterID         string
+	clusterName       string
+	spDescriptionList map[string]Description
```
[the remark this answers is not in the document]
cannarelladev: if it works with a struct i switch the type string with ShadowPodDescription type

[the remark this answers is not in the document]
giorio94: I think there was a misunderstanding here. I just meant renaming `spDescriptionList` to `shadowPods`, not changing the map definition.

pkg/liqo-controller-manager/webhooks/shadowpod/spdescription.go
```diff
@@ -0,0 +1,194 @@
...
+	result := corev1.ResourceList{}
+	for k := range quota {
+		result[k] = quantity.DeepCopy()
+	}
+	return result
+}
+
+func quotaFormatter(quota corev1.ResourceList, str string) string {
+	result := fmt.Sprintf("%s [ ", str)
+	r := regexp.MustCompile("^hugepages-")
```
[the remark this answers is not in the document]
cannarelladev: Is has only a logging purpose, to have a simplified print

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo.go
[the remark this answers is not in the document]
cannarelladev: yes, Sorry

pkg/liqo-controller-manager/webhooks/shadowpod/webhook.go
```diff
@@ -0,0 +1,301 @@
...
+		return admission.Denied(err.Error())
+	}
+
+	shadowpodlog.Info(quotaFormatter(spd.quota, "ShadowPod resource limits"))
+
+	err = peeringInfo.testAndUpdateCreation(*spd, *req.DryRun)
+	if err != nil {
+		webhooklog.Info(err.Error())
+		return admission.Denied(err.Error())
+	}
```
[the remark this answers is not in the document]
cannarelladev: how do you suggest to solve this? the first way that comes in my mind is to wrap the `getOrCreateShadowPodDescription` inside the `testAndUpdateCreation`. to avoid race conditions i think we need to do the two operation atomically in one step

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
cannarelladev: and also validate it before?

giorio94: No, given the shadowpod already exists.

pkg/liqo-controller-manager/webhooks/shadowpod/webhook.go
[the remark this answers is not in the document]
giorio94: Yes, you should perform them atomically without releasing the lock.

cannarelladev:
PTAL

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,269 @@
...
+			// Flush terminating ShadowPods and check the correct alignment between cluster and cache
+			klog.V(5).Infof("Aligning ShadowPods for cluster %q", pi.clusterIdentity.String())
+			pi.alignTerminatingOrNotExistingShadowPods(shadowPodList)
+			return true
+		},
+	)
+
+	klog.V(5).Infof("Aligning ResourceOffers - PeeringInfo")
+	if err := spv.checkAlignmentResourceOfferPeeringInfo(ctx); err != nil {
+		return true, err
```
giorio94, line 117: Should not return `true/err`, otherwise polling is aborted.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,269 @@
...
+			klog.Warningf("Warning: ShadowPod %s not found in cache", nsname.String())
+			spQuota, _ := getQuotaFromShadowPod(shadowPod, false)
+			pi.addShadowPod(createShadowPodDescription(shadowPod.GetName(), shadowPod.GetNamespace(), shadowPod.GetUID(), spQuota))
+			continue
+		}
+		// Saving a list all the ShadowPods in the cluster that are also in cache
+		spMap[nsname.String()] = struct{}{}
+	}
+	return spMap
+}
```
giorio94, line 183: This two functions appear to be mostly overlapping. Would consider whether they could be refactored to reduce code duplication.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// Check if there are new ResourceOffers in the system snapshot
+	for i := range resourceOfferList.Items {
+		ro := &resourceOfferList.Items[i]
+		clusterID := ro.Labels[discovery.ClusterIDLabel]
+		clusterName := retrieveClusterName(ctx, spv.client, clusterID)
+
+		if clusterID == "" {
+			return fmt.Errorf("ResourceOffer for cluster %q has no cluster id", clusterName)
+		}
```
giorio94, line 201: If the cluster ID is not set, also the cluster name will be empty.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,269 @@
...
+					break
+				}
+			}
+			// If the corresponding ResourceOffer is not present anymore, remove the PeeringInfo from the cache
+			if !foundRO {
+				klog.V(4).Infof("ResourceOffer %q not found in system snapshot, removing it from cache", peeringInfo.clusterIdentity.String())
+				spv.peeringCache.peeringInfo.Delete(clusterID)
+			}
+			return true
+		},
```
giorio94, line 240: Given that resource offer quotas are already updated when a new shadowpod is created, this part could be simplified, keeping only the deletion of non-existing peeringInfos.

pkg/liqo-controller-manager/webhooks/shadowpod/cache.go
```diff
@@ -0,0 +1,269 @@
...
+	for key, value := range resources {
+		if !value.Equal(quota[key]) {
+			isUpdated = true
+			break
+		}
+	}
+	if isUpdated {
+		klog.V(4).Infof("Quota of PeeringInfo for cluster %q has been updated", pi.clusterIdentity.String())
+		pi.updateQuotas(resources)
+	}
```
giorio94, line 259: Could be simplified to always perform the update, since it is idempotent.

giorio94 (commented):
Some (hopefully) final comments. Quite a few, but mostly nits.

pkg/liqo-controller-manager/webhooks/shadowpod/webhook.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Check existence and get shadow pod origin Cluster ID label
+	clusterID, found := shadowpod.Labels[forge.LiqoOriginClusterIDKey]
+	if !found {
+		klog.Warningf("Missing origin Cluster ID label on ShadowPod %q", shadowpod.Name)
+		return admission.Denied("missing origin Cluster ID label")
+	}
+
+	klog.Infof("ShadowPod %s decoded: UID: %s - clusterID %s", shadowpod.Name, shadowpod.GetUID(), clusterID)
```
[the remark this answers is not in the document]
cannarelladev: referred to the admission message?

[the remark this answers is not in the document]
giorio94: I mean, it should be something like `klog.V(5).Infof(...`

pkg/liqo-controller-manager/webhooks/shadowpod/cache_test.go
```diff
@@ -0,0 +1,197 @@
...
+				sp5 = createShadowPodDescription(testShadowPodName+"-5", testNamespace+"-3", testShadowPodUID+"-5", *resourceQuota4)
+				sp6 = createShadowPodDescription(testShadowPodName+"-6", testNamespace+"-3", testShadowPodUID+"-6", *resourceQuota4)
+				peering = createPeeringInfo(*clusterIdentity, *resourceQuota)
+				peering3 = createPeeringInfo(*forgeClusterIdentity(clusterName3, clusterID3), *resourceQuota)
+				peering.addShadowPod(sp1)
+				peering.addShadowPod(sp2)
+				peering3.addShadowPod(sp5)
+				peering3.addShadowPod(sp6)
+				fakeCache.peeringInfo.Delete(clusterID)
+				fakeCache.peeringInfo.Delete(clusterID2)
```
giorio94, line 167: Why are those deleted? The cache is already empty AFAIU.

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil, found
+	}
+	pi.(*peeringInfo).alignResourceOfferUpdates(roQuota)
+	return pi.(*peeringInfo), found
+}
+
+func (pi *peeringInfo) getFreeQuota() corev1.ResourceList {
+	freeQuota := corev1.ResourceList{}
+	quota := pi.totalQuota
+	usedQuota := pi.usedQuota
```
giorio94, line 79: Nit. Would personally drop these aliases.

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+		spValidator = webhook.Admission{Handler: NewValidator(fakeClient, false)}.Handler.(*Validator)
+
+		cache = spValidator.PeeringCache
+
+		peeringInfoTest = createPeeringInfo(*clusterIdentity, *resourceQuota)
+
+		containers = nil
+
+		containers = append(containers, containerResource{cpu: int64(resourceCPU), memory: int64(resourceMemory)})
```
giorio94, line 60:
```suggestion
		containers = []containerResource{{cpu: int64(resourceCPU), memory: int64(resourceMemory)}}
```

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo_test.go
```diff
@@ -0,0 +1,251 @@
...
+	})
+
+	Describe("Get or Create a PeeringInfo", func() {
+		JustBeforeEach(func() {
+			peeringInfo = cache.getOrCreatePeeringInfo(*clusterIdentity, *resourceQuota)
+		})
+
+		When("The Peering Info exists", func() {
+			BeforeEach(func() {
+				peeringInfo = nil
```
giorio94, line 74:
```suggestion
```

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo_test.go
```diff
@@ -0,0 +1,251 @@
...
+				cache.peeringInfo.Store(clusterID, createPeeringInfo(*clusterIdentity, *resourceQuota))
+			})
+			It("should return a peering info", func() {
+				Expect(peeringInfo).ToNot(BeNil())
```
giorio94, line 78:
```suggestion
```

Nit. This is automatically checked by the equality test.

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo_test.go
```diff
@@ -0,0 +1,251 @@
...
+				Expect(peeringInfo).To(Equal(peeringInfoTest))
+			})
+		})
+		When("The Peering Info does not exist", func() {
+			BeforeEach(func() {
+				cache.peeringInfo.Delete(clusterID)
+				peeringInfo = nil
+			})
```
giorio94, line 86: This is not required, as already reset at the beginning of the test (same for the next test).

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo_test.go
```diff
@@ -0,0 +1,251 @@
...
+		})
+
+		When("The Peering Info exists", func() {
+			BeforeEach(func() {
+				peeringInfo = nil
+				cache.peeringInfo.Store(clusterID, createPeeringInfo(*clusterIdentity, *resourceQuota))
+			})
+			It("should return a peering info", func() {
+				Expect(peeringInfo).ToNot(BeNil())
+				Expect(peeringInfo).To(Equal(peeringInfoTest))
```
giorio94, line 79: Nit. Could check that the resource quota is correctly updated.

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo_test.go
```diff
@@ -0,0 +1,251 @@
...
+				resourceQuotaLower := forgeResourceList(int64(resourceCPU/2), int64(resourceMemory))
+				peeringInfo = nil
+				peeringInfo = createPeeringInfo(*clusterIdentity, *resourceQuotaLower)
+				spd = createShadowPodDescription(testShadowPodName, testNamespace, testShadowPodUID, *resourceQuota)
+				freeQuota := peeringInfo.getFreeQuota()
+				errTest = fmt.Errorf("peering %s quota usage exceeded - free %s / requested %s",
+					corev1.ResourceCPU, freeQuota.Cpu().String(), resourceQuota.Cpu().String())
+			})
+			It("should return an error", func() {
+				Expect(err).ToNot(BeNil())
```
giorio94, line 148:
```suggestion
```

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo_test.go
```diff
@@ -0,0 +1,251 @@
...
+				peeringInfo = createPeeringInfo(*clusterIdentity, *resourceQuota)
+			})
+			It("should not return any error and available resources will not be decremented", func() {
+				Expect(err).To(BeNil())
+				Expect(peeringInfo.usedQuota.Cpu().Value()).To(Equal(freeQuotaZero.Cpu().Value()))
+				freeQuota := peeringInfo.getFreeQuota()
+				Expect(freeQuota.Cpu().Value()).To(Equal(int64(resourceCPU)))
+			})
+		})
+		When("resources are not available", func() {
```
giorio94, line 197: Would add the case the dry run flag is set.

pkg/liqo-controller-manager/webhooks/shadowpod/webhook.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	// Check existence and get resource offer by Cluster ID label
+	resourceoffer, err := liqogetters.GetResourceOfferByLabel(ctx, spv.client, corev1.NamespaceAll,
+		liqolabels.LocalLabelSelectorForCluster(clusterID))
+	if err != nil {
+		newErr := fmt.Errorf("error getting resource offer by label: %w", err)
+		klog.Error(newErr)
+		return admission.Allowed(newErr.Error())
+	}
```
giorio94, line 205: Should not prevent deletion in case this is not found. I would personally drop completely this, and just retrieve the peeringInfo from the cache (without updating the resources),

pkg/liqo-controller-manager/webhooks/shadowpod/webhook_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+var _ = Describe("Validating webhook", func() {
+
+	var (
+		spValidator              *Validator
+		spValidatorWithResources *Validator
+		request                  admission.Request
+		response                 admission.Response
+		fakeNewShadowPod         *vkv1alpha1.ShadowPod
+		// fakeOldShadowPod *vkv1alpha1.ShadowPod
```
giorio94, line 43:
```suggestion
```

pkg/liqo-controller-manager/webhooks/shadowpod/spdescription.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, fmt.Errorf("ShadowPod %s: UID mismatch", sp.GetName())
+	}
+	return &spd, nil
+}
+
+func (spd *Description) terminate() {
+	spd.running = false
+}
+
+func checkShadowPodExistence(ctx context.Context, spvclient client.Client, namespacedName types.NamespacedName) (bool, error) {
```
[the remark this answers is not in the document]
giorio94: Would personally remove the bool return value, as unused.

pkg/liqo-controller-manager/webhooks/shadowpod/spdescription_test.go
```diff
@@ -0,0 +1,250 @@
...
+		When("The ShadowPod Description does not exist", func() {
+			BeforeEach(func() {
+				spDescription = nil
+				err = nil
+				delete(peeringInfo.shadowPods, testShadowPodName)
+			})
+			It("should return a new one without errors", func() {
+				Expect(spDescription).ToNot(BeNil())
+				Expect(spDescription.running).To(BeTrue())
+				Expect(err).To(BeNil())
```
giorio94, line 79:
```suggestion
				Expect(err).ToNot(HaveOccurred())
```

Nit. (same for all occurrences)

pkg/liqo-controller-manager/webhooks/shadowpod/spdescription_test.go
```diff
@@ -0,0 +1,250 @@
...
+				containers = append(containers, containerResource{cpu: 100, memory: 100})
+				_ = fakeClient.Create(ctx, forgeShadowPodWithResourceLimits(containers, nil))
+			})
+			It("should return an error", func() {
+				Expect(spDescription).To(BeNil())
+				Expect(err).ToNot(BeNil())
+				Expect(err.Error()).To(Equal(errTest.Error()))
+			})
+		})
+		When("The ShadowPod Description exists in cache but is not running and the Shadow Pod does not exist in system snapshot yet", func() {
```
giorio94, line 114:
```suggestion
		When("The ShadowPod Description exists in cache but is not running and the Shadow Pod does not exist in system snapshot anymore", func() {
```

IIUC

cmd/liqo-controller-manager/main.go
```diff
@@ -146,11 +154,16 @@ func main() {
...
+
 	liqoerrors.InitFlags(nil)
 	restcfg.InitFlags(nil)
-	klog.InitFlags(nil)
+	klog.InitFlags(klogFlags)
 	flag.Parse()
 
+	// Set up logger flags
+	utilruntime.Must(klogFlags.Set("v", "5"))
+
```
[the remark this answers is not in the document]
giorio94: Still applies.

pkg/liqo-controller-manager/webhooks/shadowpod/peeringinfo_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+		fakeClient = fake.NewClientBuilder().WithScheme(scheme).WithObjects(fakeNamespace).Build()
+
+		spValidator = webhook.Admission{Handler: NewValidator(fakeClient, false)}.Handler.(*Validator)
+
+		cache = spValidator.PeeringCache
+
+		peeringInfoTest = createPeeringInfo(*clusterIdentity, *resourceQuota)
+
+		containers = nil
```
giorio94, line 58:
```suggestion
```

pkg/liqo-controller-manager/webhooks/shadowpod/webhook_test.go
```diff
@@ -0,0 +1,203 @@
...
+			})
+		})
+		When("The PeeringInfo does not exist", func() {
+			BeforeEach(func() {
+				containers = nil
+				containers = append(containers, containerResource{cpu: int64(resourceCPU), memory: int64(resourceMemory)})
+				fakeNewShadowPod = forgeShadowPodWithResourceLimits(containers, nil)
+				request = forgeRequest(admissionv1.Delete, nil, fakeNewShadowPod)
+			})
+			It("request is allowed with error (PeerinfInfo not found)", func() {
```
giorio94, line 197:
```suggestion
			It("request is allowed with error (PeeringInfo not found)", func() {
```

pkg/liqo-controller-manager/webhooks/shadowpod/spdescription_test.go
```diff
@@ -0,0 +1,250 @@
...
+		JustBeforeEach(func() {
+			containers = nil
+			containers = append(containers, containerResource{cpu: 100, memory: 100})
+			spDescription, err = peeringInfo.getOrCreateShadowPodDescription(ctx, spValidator.client, forgeShadowPodWithResourceLimits(containers, nil))
+		})
+
+		When("The ShadowPod Description does not exist", func() {
+			BeforeEach(func() {
+				spDescription = nil
+				err = nil
```
giorio94, line 74:
```suggestion
```

These are useless. Same for all occurrences.

pkg/liqo-controller-manager/webhooks/shadowpod/spdescription_test.go
```diff
@@ -0,0 +1,250 @@
...
+				spDescriptionTest = createShadowPodDescription(testShadowPodName, testNamespace, testShadowPodUID, *forgeResourceList(100, 100))
+				spDescriptionTest.terminate()
+				spDescription = nil
+				errTest = fmt.Errorf("ShadowPod still exists in the system")
+				err = nil
+				peeringInfo.shadowPods[spNamespacedName.String()] = spDescriptionTest
+				containers = nil
+				containers = append(containers, containerResource{cpu: 100, memory: 100})
+				errClient = fakeClient.Create(ctx, forgeShadowPodWithResourceLimits(containers, nil))
+				Expect(errClient).ToNot(HaveOccurred())
```
giorio94, line 107:
```suggestion
				Expect(fakeClient.Create(ctx, forgeShadowPodWithResourceLimits(containers, nil)).ToNot(HaveOccurred())
```

Keep it simple, and try not to use global variables if not necessary.

giorio94 (approved):
A couple final comments, then squash the commits and from my side this is good to go.

cmd/liqo-controller-manager/main.go
```diff
@@ -96,6 +99,12 @@ func main() {
 	metricsAddr := flag.String("metrics-address", ":8080", "The address the metric endpoint binds to")
 	probeAddr := flag.String("health-probe-address", ":8081", "The address the health probe endpoint binds to")
 
+	// ShadowPods webhook
+	enableResourceValidation := flag.Bool("enable-resource-enforcement", false,
+		"Enforce offerer-side that offloaded pods do not exceed offered resources (based on container limts)")
```
giorio94, line 104:
```suggestion
		"Enforce offerer-side that offloaded pods do not exceed offered resources (based on container limits)")
```

deployments/liqo/values.yaml
```diff
@@ -27,6 +27,10 @@ controllerManager:
   config:
     # -- It defines the percentage of available cluster resources that you are willing to share with foreign clusters.
     resourceSharingPercentage: 30
+    # -- It enforces offerer-side that offloaded pods do not exceed offered resources (based on container limts).
```
giorio94, line 30:
```suggestion
    # -- It enforces offerer-side that offloaded pods do not exceed offered resources (based on container limits).
```

giorio94 (approved):

PR #1231, opened 2022-05-09
README: fix broken image

frisso:
Replace a broken image in the main README file with a markdown emoji.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ kubectl port-forward -n liqo-demo service/frontend 8080:80
 * Your application is transparently deployed across two different clusters:
 
 ```bash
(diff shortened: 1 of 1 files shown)


giorio94:
Oops, my fault I believe.

Honestly, I was wondering whether it is worth keeping that icon, as (a) most likely the problem will happen again in the future, (b) it adds practically no value and, by the way, (c) it is completely invisible with the dark theme.

Alternatively, we could use the markup emoji `:bell:` 🔔, that, although a bit different, would make the README self-contained (and works regardless of the theme).

frisso:
Completely fine with your proposal, Marco.
Please, discard my changes and go with the :bell:

aleoli (approved):

PR #1243, opened 2022-05-26
Liqoctl: Improve Offload Command

aleoli:
# Description

This pr improves the `liqoctl offload / unoffload` commands. They now wait for the condition to be verified.

# How Has This Been Tested?

- [x] locally

### DIFF
--- a/cmd/liqoctl/cmd/offload.go
+++ b/cmd/liqoctl/cmd/offload.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"context"
+	"time"
 
 	"github.com/spf13/cobra"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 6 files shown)


cmd/liqoctl/cmd/offload.go
```diff
@@ -110,6 +111,7 @@ func newOffloadNamespaceCommand(ctx context.Context, f *factory.Factory) *cobra.
 		"The constraints regarding pods scheduling in this namespace, among Local, Remote and LocalAndRemote")
 	cmd.Flags().Var(namespaceMappingStrategy, "namespace-mapping-strategy",
 		"The naming strategy adopted for the creation of remote namespaces, among DefaultName and EnforceSameName")
+	cmd.Flags().DurationVar(&options.Timeout, "timeout", 120*time.Second, "The timeout for the offloading process")
```
giorio94, line 114: Would lower the timeout to something in the range 10-30 seconds, as more than enough to complete the process. Unoffloading, insteat, might take longer, since namespace deletion is typically slow.

PR #1288, opened 2022-06-15
Fix Wireguard Endpoint DNS Resolution

aleoli:
# Description

This pr fixes the DNS resolver for the remote Wireguard endpoint.

If the endpoint is a DNS name, it caches the resolved IP (preferring IPv4 over IPv6) and returns it for the other resolutions. This prevents the Wireguard interface from periodically reconfiguring when multiple IPs were resolved in random order; causing connection issues.

# How Has This Been Tested?

- [x] on EKS + AKS peering (where EKS LB has 3 IPv4 addresses)

### DIFF
--- a/pkg/liqonet/tunnel/resolver/resolver_test.go
+++ b/pkg/liqonet/tunnel/resolver/resolver_test.go
@@ -1,0 +1,184 @@
+// Copyright 2019-2022 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


pkg/liqonet/tunnel/resolver/resolver_test.go
```diff
@@ -0,0 +1,148 @@
...
+				}, nil
+			},
+			cache: map[string]*net.IPAddr{
+				"example.com": {IP: net.ParseIP("2001:db8:a0b:12f0::1")},
+			},
+			expectedIP:    "1.2.3.4",
+			expectedCache: HaveKeyWithValue("example.com", &net.IPAddr{IP: net.IPv4(1, 2, 3, 4)}),
+		}),
+	)
+
```
giorio94, line 147: Could add an entry to check that everything works as expected in case the entry previously in cache is no longer returned during resolution.

giorio94 (commented):
A couple of comments inline.

giorio94 (approved):

PR #1291, opened 2022-06-15
Network config: fix incorrect log level

giorio94:
# Description

This PR fixes an incorrect log instruction in the network config controller.

### DIFF
--- a/internal/liqonet/network-manager/netcfgcreator/networkconfig.go
+++ b/internal/liqonet/network-manager/netcfgcreator/networkconfig.go
@@ -229,6 +229,6 @@ func (ncc *NetworkConfigCreator) EnforceNetworkConfigAbsence(ctx context.Context
 		return err
 	}
 
-	klog.Errorf("Correctly ensured no NetworkConfigs associated with cluster %s are present", clusterIdentity)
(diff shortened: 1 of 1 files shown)


internal/liqonet/network-manager/netcfgcreator/networkconfig.go
```diff
@@ -229,6 +229,6 @@ func (ncc *NetworkConfigCreator) EnforceNetworkConfigAbsence(ctx context.Context
 		return err
 	}
 
-	klog.Errorf("Correctly ensured no NetworkConfigs associated with cluster %s are present", clusterIdentity)
+	klog.Infof("Correctly ensured no NetworkConfigs associated with cluster %s are present", clusterIdentity)
```
aleoli, line 232: How often is it printed? Should we have a verbosity level set?

aleoli (approved):

internal/liqonet/network-manager/netcfgcreator/networkconfig.go
giorio94: This is printed only in case an object was present and it has been requested to be deleted. For this reason, if you agree, I would keep the current verbosity level.

PR #1303, opened 2022-06-22
Docs: add example in "install development version" section

lucafrancescato:
# Description

This PR adds an example in the "install development version" section in the "Install" page of the documentation.

### DIFF
--- a/docs/installation/install.md
+++ b/docs/installation/install.md
@@ -387,8 +387,17 @@ Development versions include:
 * The commits of *pull requests* to the Liqo repository, whose images have been built through the appropriate bot command.
 
 The installation of a development version of Liqo can be triggered specifying a **commit *SHA*** through the `--version` flag.
-In this case, *liqoctl* proceeds to **clone the repository** (either from the official repository, or from a fork configured through the `--repo-url` flag) at the given revision, and to leverage the Helm chart therein contained.
-Alternatively, the Helm chart can be retrieved from a **local path**, as configured through the `--local-chart-path` flag.
+In this case, *liqoctl* proceeds to **clone the repository** (either from the official repository, or from a fork configured through the `--repo-url` flag) at the given revision, and to leverage the Helm chart therein contained:
+
+```bash
+liqoctl install <provider> --version <commit-sha> --repo-url <forked-repo-url>
+```
+
+Alternatively, the Helm chart can be retrieved from a **local path**, as configured through the `--local-chart-path` flag:
+
+```bash
+liqoctl install <provider> --version <commit-sha> --local-chart-path <path-to-local-chart>
+```
 
 (InstallationCalicoConfiguration)=
 


docs/installation/install.md
```diff
@@ -390,6 +390,13 @@ The installation of a development version of Liqo can be triggered specifying a
 In this case, *liqoctl* proceeds to **clone the repository** (either from the official repository, or from a fork configured through the `--repo-url` flag) at the given revision, and to leverage the Helm chart therein contained.
 Alternatively, the Helm chart can be retrieved from a **local path**, as configured through the `--local-chart-path` flag.
 
+Example:
+```bash
+liqoctl install <provider> --version <commit-sha> \
+    --repo-url <forked-repo-url> \
+    --local-chart-path <path-to-local-chart>
```
giorio94, line 397: The `--repo-url` and `--local-chart-path` are alternative. I would suggest to add two examples to show both possibilities (i.e., either automatically downloading the chart from a repository, or using an already existing local copy of the chart).

docs/installation/install.md
```diff
@@ -387,13 +387,14 @@ Development versions include:
...
 
 The installation of a development version of Liqo can be triggered specifying a **commit *SHA*** through the `--version` flag.
-In this case, *liqoctl* proceeds to **clone the repository** (either from the official repository, or from a fork configured through the `--repo-url` flag) at the given revision, and to leverage the Helm chart therein contained.
-Alternatively, the Helm chart can be retrieved from a **local path**, as configured through the `--local-chart-path` flag.
-
-Example:
+In this case, *liqoctl* proceeds to **clone the repository** (either from the official repository, or from a fork configured through the `--repo-url` flag) at the given revision, and to leverage the Helm chart therein contained:
+```bash
+liqoctl install <provider> --version <commit-sha> \
+    --repo-url <forked-repo-url>
```
lucafrancescato, line 393: My only concern is that these examples don't show that the command could be used only with `--version`, however adding another example would be overkill for such a simple case.

docs/installation/install.md
```diff
@@ -387,13 +387,14 @@ Development versions include:
...
+```
+Alternatively, the Helm chart can be retrieved from a **local path**, as configured through the `--local-chart-path` flag:
 ```bash
 liqoctl install <provider> --version <commit-sha> \
-    --repo-url <forked-repo-url> \
     --local-chart-path <path-to-local-chart>
 ```
```
lucafrancescato, line 399: My only concern is that these examples don't show that the command could be used only with `--version`, however adding another example would be overkill for such a simple case.

lucafrancescato: Also, maybe they can just take one line instead of going on the next line with `\`

giorio94: Yes, I think it should already be implicit that it is necessary to specify the version, as talking about development versions. As for using a single line, I completely agree.

giorio94 (approved):

PR #1328, opened 2022-07-11
Simplified requirements.

frisso:
Additional notes (to be discussed in person):
- not clear why we should select in-band vs outofband (is there any advantage?)
- it would be nice to create a picture showing the workflow
- not clear the meaning of the command "--context=consumer" in the peering

### DIFF
--- a/docs/installation/requirements.md
+++ b/docs/installation/requirements.md
@@ -4,12 +4,11 @@ This page presents an overview of the main requirements, both in terms of **reso
 
 ## Overview
 
-Typically, Liqo requires very **limited resources** (i.e., in terms of CPU, RAM, and network bandwidth) for the control plane execution, and it is compatible with both standard clusters and more **resource constrained devices** (e.g., Raspberry Pi), leveraging K3s as Kubernetes distribution.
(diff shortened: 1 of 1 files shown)


frisso:
@giorio94 Feel free to add comments. If it's ok, I would merge this and leave the open points for a future PR.

docs/installation/requirements.md
```diff
@@ -4,25 +4,27 @@ This page presents an overview of the main requirements, both in terms of **reso
...
+
+* [**Out-of-band control plane peering**](FeaturesPeeringOutOfBandControlPlane): it requires **three separated traffic flows** (hence, three exposed endpoints).
+* [**in-band control plane peering**](FeaturesPeeringInBandControlPlane): it requires a **single endpoint**, as all control plane traffic is tunneled inside the cross-cluster VPN.
+
+More details available in the [peering section](/features/peering).
 
 ```{admonition} Note
-The two peering approaches are **non-mutually exclusive**.
-In other words, a single cluster can leverage different approaches toward different remote clusters, in case all connectivity requirements are fullfilled.
+The two peering approaches are **non-mutually exclusive**: a cluster can leverage different approaches toward different remote clusters, provided that the connectivity requirements are verified.
```
giorio94, line 27: To me, *verified* conveys a different meaning compared to the original *fulfilled*. Is this intentional?

giorio94 (approved):

giorio94:
/hold

Holding, as a remainder to squash the commits before merging.

PR #1342, opened 2022-07-21
Liqoctl: Improve Validation Feedback

aleoli:
# Description

This pr improves the liqoctl install parameters feedback to tell the user to use the generic provider or to install with helm when the pod/service CIDRs are not the ones expected by the provider. This can be useful when using different CNIs that the ones expected by cloud providers.

# How Has This Been Tested?

- [x] locally

### DIFF
--- a/pkg/liqoctl/install/validation.go
+++ b/pkg/liqoctl/install/validation.go
@@ -43,12 +43,16 @@ func (o *Options) validate(ctx context.Context) error {
 	o.Printer.Verbosef("Kubernetes API Server: %s\n", o.PodCIDR)
 
 	if err := o.validatePodCIDR(ctx); err != nil {
-		return fmt.Errorf("failed validating Pod CIDR %q: %w", o.PodCIDR, err)
+		return fmt.Errorf("failed validating Pod CIDR %q: %w. "+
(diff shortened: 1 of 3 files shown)


pkg/liqoctl/install/validation.go
```diff
@@ -43,12 +43,14 @@ func (o *Options) validate(ctx context.Context) error {
 	o.Printer.Verbosef("Kubernetes API Server: %s\n", o.PodCIDR)
 
 	if err := o.validatePodCIDR(ctx); err != nil {
-		return fmt.Errorf("failed validating Pod CIDR %q: %w", o.PodCIDR, err)
+		return fmt.Errorf("failed validating Pod CIDR %q: %w. "+
+			"Try setting the correct PodCIDR by using the generic provider or by installing with Helm", o.PodCIDR, err)
```
giorio94, line 47:
```suggestion
			"Try setting the correct Pod CIDR using the vanilla *liqoctl install* command, or installing Liqo with Helm", o.PodCIDR, err)
```

Definitely just a proposal (and same for the one below). I fear the term *generic provider* might generate a bit of confusion, since there is no *liqoctl install generic* command. Alternatively, we might take that route...

giorio94 (approved):
/lgtm

PR #1445, opened 2022-09-19
Helm fixes to ease usage with Helm-related tools

### LINKED ISSUE #1437: [Helm templates] duplicate webhooks job-patch name
#### What happened:

Command used to generate helm files:

```
helm template liqo liqo/liqo --namespace liqo -f vendor/liqo//helm_values.yaml --version v0.5.4 --output-dir=vendor/liqo//base --include-crds
```

The jobs `webhooks/job-patch/job-patch-webhook.yaml` and `webhooks/job-patch/job-create-secret.yaml` have the same name. This is causing kustomize build fail.

Job names should be different and therefore should not block kustomize build.

Use the helm command with default helm values.

#### Environment:
- Liqo version: v0.5.4
- Kubernetes version (use `kubectl version`): 1.22 (server)
- Cloud provider or hardware configuration: gke+k3s
- Network plugin and version:
- Install tools:
- Others:

QcFe:
# Description

This PR 
- changes the name of webhook related jobs in order to differentiate them;
- imposes a CLUSTER_ID taken from values if exists (certain tools like ArgoCD cannot retrieve current value while templating)

Fixes #1437

# How Has This Been Tested?

- [x] Local Helm templating
- [x] Cluster deployment through Helm

### DIFF
--- a/deployments/liqo/templates/liqo-clusterid-configmap.yaml
+++ b/deployments/liqo/templates/liqo-clusterid-configmap.yaml
@@ -9,10 +9,18 @@ metadata:
     {{- include "liqo.labels" $clusterIdConfig | nindent 4 }}
   name: {{ include "liqo.prefixedName" $clusterIdConfig }}
 data:
-  {{- if (not $oldObject) }}
-  CLUSTER_ID: {{ uuidv4 }}
-  {{- else }}
+  {{- if $oldObject }}
+  {{- if and (.Values.discovery.config.clusterIDOverride) (ne $oldObject.data.CLUSTER_ID .Values.discovery.config.clusterIDOverride) }}
+    {{ fail "Changing the cluster ID by providing a different .discovery.config.clusterIDOverride value is not allowed." }}
+  {{- end}}
   CLUSTER_ID: {{ $oldObject.data.CLUSTER_ID }}
+  {{- else if .Values.discovery.config.clusterIDOverride }}
+    {{- if not (regexMatch "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$" .Values.discovery.config.clusterIDOverride) }}
+      {{ fail "Provided cluster ID must be a valid UUID"}}
+    {{- end}}
+  CLUSTER_ID: {{ .Values.discovery.config.clusterIDOverride }}
+  {{- else }}
+  CLUSTER_ID: {{ uuidv4 }}
   {{- end }}
   {{- if .Values.discovery.config.clusterName }}
   CLUSTER_NAME: {{ .Values.discovery.config.clusterName }}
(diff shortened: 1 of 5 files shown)


deployments/liqo/templates/liqo-clusterid-configmap.yaml
```diff
@@ -9,10 +9,12 @@ metadata:
     {{- include "liqo.labels" $clusterIdConfig | nindent 4 }}
   name: {{ include "liqo.prefixedName" $clusterIdConfig }}
 data:
-  {{- if (not $oldObject) }}
-  CLUSTER_ID: {{ uuidv4 }}
-  {{- else }}
+  {{- if $oldObject }}
```
giorio94, line 12: I would suggest to emit a warning in case a clusterIDOverride is set, but a different value is being retrieved from the old object. Silently ignoring it might fool the users.

deployments/liqo/templates/webhooks/job-patch/job-create-secret.yaml
```diff
@@ -1,4 +1,5 @@
-{{- $cfg := (merge (dict "name" "webhook-certificate-patch" "module" "webhook-certificate-patch") .) -}}
+{{- $cfg := (merge (dict "name" "webhook-certificate-patch-pre" "module" "webhook-certificate-patch") .) -}}
+{{- $rbacCfg := (merge (dict "name" "webhook-certificate-patch") .) -}}
```
giorio94, line 2:
```suggestion
{{- $rbacConfig := (merge (dict "name" "webhook-certificate-patch") .) -}}
```

Nit. Looks more consistent to me with the other variables below (same for the other file)

giorio94 (approved):
Thanks @QcFe, just a couple of minor suggestions inline,

giorio94:
@QcFe Could you please run `make docs`?

deployments/liqo/templates/liqo-clusterid-configmap.yaml
```diff
@@ -9,10 +9,15 @@ metadata:
...
+  {{- if and (.Values.discovery.config.clusterIDOverride) (ne $oldObject.data.CLUSTER_ID .Values.discovery.config.clusterIDOverride) }}
+    {{ fail "Changing the cluster ID by providing a different .discovery.config.clusterIDOverride value is not allowed." }}
+  {{- end}}
   CLUSTER_ID: {{ $oldObject.data.CLUSTER_ID }}
+  {{- else if .Values.discovery.config.clusterIDOverride }}
```
giorio94, line 17: Before merging, I was wondering whether it might be appropriate to add a regex check to ensure this actually looks like a UUID. @QcFe, @aleoli WDYT?

QcFe: Not a problem for me

aleoli: 👍

PR #1454, opened 2022-09-22
Foreign cluster resource: enforce peering type

giorio94:
# Description

This PR introduces an "immutable" PeeringType field as part of the ForeignCluster API, in order to explicit the type of peering and prevent switching from one to another, which would cause potential issues. This also enables a better UX through liqoctl.

Fixes #(issue)

# How Has This Been Tested?

- [x] Existing tests
- [x] Manual testing

### DIFF
--- a/deployments/liqo/templates/webhooks/liqo-validating-webhook.yaml
+++ b/deployments/liqo/templates/webhooks/liqo-validating-webhook.yaml
@@ -8,6 +8,23 @@ metadata:
   labels:
     {{- include "liqo.labels" $webhookConfig | nindent 4 }}
 webhooks:
+  - name: fc.validate.liqo.io
+    admissionReviewVersions:
+      - v1
+      - v1beta1
+    clientConfig:
+      service:
+        name: {{ include "liqo.prefixedName" $ctrlManagerConfig }}
+        namespace: {{ .Release.Namespace }}
+        path: "/validate/foreign-cluster"
+        port: {{ .Values.webhook.port }}
+    rules:
+      - operations: ["UPDATE"]
+        apiGroups: ["discovery.liqo.io"]
+        apiVersions: ["v1alpha1"]
+        resources: ["foreignclusters"]
+    sideEffects: None
+    failurePolicy: {{ .Values.webhook.failurePolicy }}
   - name: nsoff.validate.liqo.io
     admissionReviewVersions:
       - v1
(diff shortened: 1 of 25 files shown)


deployments/liqo/templates/webhooks/liqo-validating-webhook.yaml
```diff
@@ -8,6 +8,23 @@ metadata:
...
+      - v1
+      - v1beta1
+    clientConfig:
+      service:
+        name: {{ include "liqo.prefixedName" $ctrlManagerConfig }}
+        namespace: {{ .Release.Namespace }}
+        path: "/validate/foreign-cluster"
+        port: {{ .Values.webhook.port }}
+    rules:
+      - operations: ["CREATE", "UPDATE"]
```
aleoli, line 22: If I understand correctly we are only handling update events

pkg/liqoctl/inband/cluster.go
```diff
@@ -833,13 +860,18 @@ func (c *Cluster) DisablePeering(ctx context.Context, remoteClusterID *discovery
...
 
+	// Do not proceed if the peering is not in-band.
+	if fc.Spec.PeeringType != discoveryv1alpha1.PeeringTypeInBand {
+		return fmt.Errorf("the peering type towards remote cluster %q is %s, expected %s",
+			remName, fc.Spec.PeeringType, discoveryv1alpha1.PeeringTypeInBand)
+	}
+
 	// Set outgoing peering to no.
-	if _, err := controllerutil.CreateOrPatch(ctx, c.local.CRClient, fc, func() error {
+	if _, err := controllerutil.CreateOrUpdate(ctx, c.local.CRClient, fc, func() error {
```
aleoli, line 870:
```suggestion
	if _, err = controllerutil.CreateOrUpdate(ctx, c.local.CRClient, fc, func() error {
```

We can avoid the re-definition of `err`

aleoli (approved):

deployments/liqo/templates/webhooks/liqo-validating-webhook.yaml
giorio94: Definitely. I currently kept both operations since it is likely we will add checks that apply also to creations in the future. Yet, I'm open to remove the "CREATE" one for the moment if you feel it is better.

aleoli: Yes, I think that we can remove it for now

PR #1460, opened 2022-09-27
Collect Liqo Telemetry

aleoli:
# Description

This pr adds a client to collect anonymous and aggregated Liqo installation telemetries.

The collection is enabled by default, but it can be disabled by adding the `--set telemetry.enable=false` flag during installations and/or updates.

### DIFF
--- a/cmd/telemetry/main.go
+++ b/cmd/telemetry/main.go
@@ -1,0 +1,109 @@
+// Copyright 2019-2022 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	discoveryv1alpha1 "github.com/liqotech/liqo/apis/discovery/v1alpha1"
+	netv1alpha1 "github.com/liqotech/liqo/apis/net/v1alpha1"
+	offloadingv1alpha1 "github.com/liqotech/liqo/apis/offloading/v1alpha1"
+	sharingv1alpha1 "github.com/liqotech/liqo/apis/sharing/v1alpha1"
+	"github.com/liqotech/liqo/pkg/consts"
+	"github.com/liqotech/liqo/pkg/telemetry"
+	argsutils "github.com/liqotech/liqo/pkg/utils/args"
+	"github.com/liqotech/liqo/pkg/utils/mapper"
+	"github.com/liqotech/liqo/pkg/utils/restcfg"
+)
+
+var scheme = runtime.NewScheme()
+
+func init() {
+	_ = clientgoscheme.AddToScheme(scheme)
+	_ = discoveryv1alpha1.AddToScheme(scheme)
+	_ = offloadingv1alpha1.AddToScheme(scheme)
+	_ = sharingv1alpha1.AddToScheme(scheme)
+	_ = netv1alpha1.AddToScheme(scheme)
+}
+
+// cluster-role
+// +kubebuilder:rbac:groups=core,resources=configmaps;nodes;pods,verbs=get;list;watch
+// +kubebuilder:rbac:groups=discovery.liqo.io,resources=foreignclusters,verbs=get;list;watch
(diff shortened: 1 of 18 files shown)


pkg/telemetry/builder.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	offloadedPods, err := liqogetters.ListOffloadedPods(ctx, c.Client, namespaceOffloading.Namespace)
+	runtime.Must(err)
+
+	var nodePodBucket = make(map[string]int64)
+	for i := range offloadedPods.Items {
+		pod := &offloadedPods.Items[i]
+		if pod.Spec.NodeName == "" {
+			klog.Warningf("Pod %s/%s has no node assigned", pod.Namespace, pod.Name)
```
giorio94, line 100:
```suggestion
			klog.Warningf("Pod %s has no node assigned", klog.KObj(pod))
```

Nit.

pkg/telemetry/builder.go
```diff
@@ -0,0 +1,212 @@
...
+			continue
+		}
+
+		if _, ok := nodePodBucket[pod.Spec.NodeName]; !ok {
+			nodePodBucket[pod.Spec.NodeName] = 1
+		} else {
+			nodePodBucket[pod.Spec.NodeName]++
+		}
```
giorio94, line 108:
```suggestion
		nodePodBucket[pod.Spec.NodeName]++
```

Nit. It is enough.

pkg/telemetry/builder.go
```diff
@@ -0,0 +1,212 @@
...
+		if err != nil {
+			klog.Errorf("unable to get cluster id from node %s: %v", node.Name, err)
+			continue
+		}
+		if _, ok := nodePodBucket[node.Name]; !ok {
+			namespaceInfo.NumOffloadedPods[clusterID] = 0
+			continue
+		}
+		namespaceInfo.NumOffloadedPods[clusterID] = nodePodBucket[node.Name]
+	}
```
giorio94, line 126: Could perform this operation in advance to build a node name -> clusterID map, and avoid repeating it for every offloaded namespace (as the client is also not cached).

pkg/telemetry/builder.go
```diff
@@ -0,0 +1,212 @@
...
+		clusterID, err := liqogetters.RetrieveClusterIDFromVirtualNode(node)
+		if err != nil {
+			klog.Errorf("unable to get cluster id from node %s: %v", node.Name, err)
+			continue
+		}
+		if _, ok := nodePodBucket[node.Name]; !ok {
+			namespaceInfo.NumOffloadedPods[clusterID] = 0
+			continue
+		}
+		namespaceInfo.NumOffloadedPods[clusterID] = nodePodBucket[node.Name]
```
giorio94, line 125:
```suggestion
		namespaceInfo.NumOffloadedPods[clusterID] = nodePodBucket[node.Name]
```

Accessing a non existing key returns the default value (0 in this case).

pkg/telemetry/builder.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	var latency time.Duration
+	tunnel, err := liqogetters.GetTunnelEndpoint(ctx, c.Client,
+		foreignCluster.Spec.ClusterIdentity.ClusterID, foreignCluster.Status.TenantNamespace.Local)
+	switch {
+	case apierrors.IsNotFound(err):
+		// do nothing
+	case err != nil:
+		klog.Errorf("unable to get tunnel endpoint for cluster %s: %v", foreignCluster.Spec.ClusterIdentity.ClusterID, err)
```
giorio94, line 163: Would print the cluster name rather than the cluster ID in error messages.

pkg/telemetry/send.go
```diff
@@ -0,0 +1,64 @@
...
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to send telemetry item: %w", err)
+	}
+	if err = resp.Body.Close(); err != nil {
+		return fmt.Errorf("failed to close response body: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		message, err2 := io.ReadAll(resp.Body)
```
giorio94, line 56:
```suggestion
		message, err := io.ReadAll(resp.Body)
```

cmd/telemetry/main.go
```diff
@@ -0,0 +1,108 @@
...
+
+var scheme = runtime.NewScheme()
+
+func init() {
+	_ = clientgoscheme.AddToScheme(scheme)
+	_ = discoveryv1alpha1.AddToScheme(scheme)
+	_ = offloadingv1alpha1.AddToScheme(scheme)
+	_ = sharingv1alpha1.AddToScheme(scheme)
+	_ = netv1alpha1.AddToScheme(scheme)
+	// +kubebuilder:scaffold:scheme
```
giorio94, line 49:
```suggestion
```

giorio94 (commented):
The logic seems OK to me, a few nits inline.

pkg/telemetry/send.go
giorio94, line 56: The body you are attempting to read from has been previously closed, thus this operation always fails.

pkg/telemetry/send.go
```diff
@@ -0,0 +1,64 @@
...
+		if err2 != nil {
+			return fmt.Errorf("failed to send telemetry item: %w", err)
+		}
+		return fmt.Errorf("unexpected status code: %d: %s", resp.StatusCode, message)
+	}
+	klog.Infof("successfully sent telemetry item %+v to %s", item, endpoint)
```
giorio94, line 62: Nit. Rather than printing `item` as a go struct, you could use the `json.MarshalIndent` function to output a pretty-printed json document, which is much more human friendly.

giorio94 (approved):

PR #1467, opened 2022-09-29
Override Services Address/Port

aleoli:
# Description

This pr adds the possibility to override the address and the port for Liqo external services. It can be helpful when using reverse proxies or NATs.

### DIFF
--- a/pkg/utils/foreignCluster/getForeignCluster.go
+++ b/pkg/utils/foreignCluster/getForeignCluster.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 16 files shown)


pkg/utils/foreignCluster/getForeignCluster.go
```diff
@@ -71,11 +72,7 @@ func GetOlderForeignCluster(
 }
 
 // getAuthAddress retrieves the external address where the Authentication Service is reachable from the external world.
-func getAuthAddress(ctx context.Context, cl client.Client, authServiceAddress, namespace string) (string, error) {
-	if authServiceAddress != "" {
-		return authServiceAddress, nil
-	}
-
+func getAuthAddress(ctx context.Context, cl client.Client, namespace string) (string, error) {
```
giorio94, line 75: Nit. These two functions could now be refactored to retrieve the service only once in the `GetHomeAuthURL` function, and propagate it here and in `getAuthPort`.

giorio94 (approved):

PR #1475, opened 2022-10-04
Added suggestion how big should be the cluster

frisso:


giorio94:
Merging without running code tests, as including only documentation changes.

PR #1525, opened 2022-10-25
Draft Releases

aleoli:
# Description

This pr makes new releases to be created in draft state. It makes it possible to add content to the release notes before sending emails and notifications


giorio94 (approved):

giorio94:
Merging without running code tests, as dealing only with the release pipeline.

PR #1526, opened 2022-10-25
Docs: add warning about mismatching versions

giorio94:
# Description

This PR extends the documentation adding the warnings concerning mismatching versions, both between liqo and liqoctl, and for the peering process.


aleoli (approved):

giorio94:
Merging without running the code checks, as involving documentation only.

PR #1536, opened 2022-11-07
Migrate container registry to GitHub

aleoli:
# Description

This pr moves the container registry from DockerHub to GitHub Container Registry in order to avoid the docker pull limits in development environments


giorio94 (commented):
I would also migrate the image used for the unit tests: https://github.com/liqotech/liqo/blob/b7af2b38daa05a1df4adbf3b5b1c215b9e93052e/.github/workflows/test.yml#L75-L88

giorio94 (approved):

PR #1560, opened 2022-11-18
Remote Metrics: Filtering not ready nodes

LucaRocco:
# Description

This PR includes a filter on not-ready nodes to skip them while getting remote metrics and avoid communication errors.

# How Has This Been Tested?

Pre-existing tests have been updated. Specifically, I've added a not-ready node in the node list to check that the filter works as expected.

- [x] Updating pre-existing tests
- [x] Manually

### DIFF
--- a/pkg/remotemetrics/resources.go
+++ b/pkg/remotemetrics/resources.go
@@ -25,6 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/liqotech/liqo/pkg/consts"
+	liqoutils "github.com/liqotech/liqo/pkg/utils"
 	"github.com/liqotech/liqo/pkg/virtualKubelet/forge"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/remotemetrics/resources.go
```diff
@@ -108,12 +108,24 @@ func (m *resourceGetter) GetNodeNames(ctx context.Context) []string {
...
+
+func nodeIsReady(node *corev1.Node) bool {
+	conditions := node.Status.Conditions
+	for i := range conditions {
+		if conditions[i].Type == corev1.NodeReady {
+			return conditions[i].Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
```
giorio94, line 131: This function already exists as liqo utility: https://github.com/liqotech/liqo/blob/3461899762cc889892133ea7cf5b914353a7d6fb/pkg/utils/nodeUtils.go#L24-L31

giorio94 (approved):

PR #1568, opened 2022-11-21
Liqoctl status: removed helm dependency

cheina97:
This PR changes how `liqoctl status` get information about **local cluster name** and **kubernetes API addres**.
Before this PR **helm** was used to get this information, but this can be a problem if **Liqo** is installed without helm (for example **ArgoCD** uses **helm** to create manifests, without making a **helm release**).
Now `liqoctl status` uses other methods avoiding helm dependencies.

### DIFF
--- a/pkg/liqoctl/status/local/localinfo_test.go
+++ b/pkg/liqoctl/status/local/localinfo_test.go
@@ -16,13 +16,21 @@ package statuslocal
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/pterm/pterm"
+	appv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
+	liqoconsts "github.com/liqotech/liqo/pkg/consts"
 	"github.com/liqotech/liqo/pkg/liqoctl/factory"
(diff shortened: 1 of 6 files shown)


pkg/liqoctl/status/localinfo.go
```diff
@@ -100,14 +84,12 @@ func (lic *LocalInfoChecker) Collect(ctx context.Context) error {
 			networkSection.AddEntry("Reserved Subnets", ipamStorage.Spec.ReservedSubnets...)
 		}
 	}
-	apiServerAddress, err := util.ExtractValuesFromNestedMaps(values, "apiServer", "address")
+
+	apiServerAddress, err := apiserver.GetAddressFromMasterNode(ctx, lic.options.KubeClient)
```
giorio94, line 88: You should consider the case in which it is overridden through the arguments: https://github.com/liqotech/liqo/blob/95938948984c871d6a45e7819ee3be5277d9d186/pkg/utils/apiserver/address.go#L31-L40

pkg/liqoctl/status/local/localinfo_test.go
```diff
@@ -42,70 +51,158 @@ var _ = Describe("LocalInfo", func() {
...
+					"node-role.kubernetes.io/control-plane": "",
+				},
+			},
+			Status: corev1.NodeStatus{
+				Addresses: []corev1.NodeAddress{
+					{Type: corev1.NodeInternalIP, Address: testutil.APIAddress},
+				},
+			},
+		}, metav1.CreateOptions{})
+		utilruntime.Must(err)
```
giorio94, line 103: Would convert to an Expect assertion.

pkg/utils/testutil/consts.go
```diff
@@ -16,7 +16,7 @@ package testutil
 
 const (
 	// APIAddress is the address of the API server used for testing.
-	APIAddress = "fake API address"
+	APIAddress = "1.0.0.1"
```
giorio94, line 19: Not referring to this explicitly, but the FakeHelmValues function (in testutil/kubernetes.go) is no longer used and could be dropped.

giorio94 (approved):
/lgtm, just a couple of nits inline.

PR #1570, opened 2022-11-25
Add metrics for virtual-kubelet

damianot98:
# Description

This PR aims to extend the virtual kubelet component of the Liqo open-source project, to expose a set of metrics in the Prometheus format describing the most relevant characteristics of the amount of offloaded and replicated objects.

Metrics were implemented using the CounterVec Prometheus resource which is a Collector that bundles a set of Counters that all share the same Description but have different values for their variable labels. When an instance of CounterVec is created it is empty and it does not show up when requesting the metric. Only when the metric is incremented the content can be seen.

Prometheus metrics developed for each peered remote cluster are the **liqo_virtual_kubelet_reflection_item_counter** and the **liqo_virtual_kubelet_reflection_error_counter**. Both metrics expose information about the namespace and the type of item extracted from the workqueue. 
 

- **liqo_virtual_kubelet_reflection_item_counter**: is the counter of the reflected resources successfully processed. A fast increase of this metric can indicate a race condition between local and remote operators.

- **liqo_virtual_kubelet_reflection_error_counter**: is the counter of the transient errors.

**Prometheus Operator** is required to scrape metrics from the **PodMonitor** resource. This can be achieved by setting the `—enable-metrics` flag during the installation, which will expose and create the PodMonitor resource.  

# How Has This Been Tested?

- [x] Locally on kind
- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/cmd/virtual-kubelet/root/root.go
+++ b/cmd/virtual-kubelet/root/root.go
@@ -36,6 +36,7 @@ import (
 	"github.com/liqotech/liqo/pkg/utils"
 	"github.com/liqotech/liqo/pkg/utils/restcfg"
 	nodeprovider "github.com/liqotech/liqo/pkg/virtualKubelet/liqoNodeProvider"
+	metrics "github.com/liqotech/liqo/pkg/virtualKubelet/metrics"
 	podprovider "github.com/liqotech/liqo/pkg/virtualKubelet/provider"
 )
 
@@ -110,6 +111,7 @@ func runRootCommand(ctx context.Context, c *Opts) error {
 		EnableStorage:              c.EnableStorage,
 		VirtualStorageClassName:    c.VirtualStorageClassName,
 		RemoteRealStorageClassName: c.RemoteRealStorageClassName,
+		EnableMetrics:              c.EnableMetrics,
 
 		HomeAPIServerHost: c.HomeAPIServerHost,
 		HomeAPIServerPort: c.HomeAPIServerPort,
@@ -192,6 +194,10 @@ func runRootCommand(ctx context.Context, c *Opts) error {
 	err = setupHTTPServer(ctx, podProvider.PodHandler(), localClient, remoteConfig, c)
 	if err != nil {
 		return errors.Wrap(err, "error while setting up HTTPS server")
+	}
+
+	if c.EnableMetrics {
+		metrics.SetupMetricHandler()
 	}
 
 	go func() {
--- a/deployments/liqo/templates/liqo-virtualkubelet-podmonitor.yaml
+++ b/deployments/liqo/templates/liqo-virtualkubelet-podmonitor.yaml
@@ -1,0 +1,22 @@
+{{- $kubeletMetricsConfig := (merge (dict "name" "virtual-kubelet" "module" "virtual-kubelet") .) -}}
+{{- if .Values.virtualKubelet.metrics.podMonitor.enabled }}
+
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: {{ include "liqo.prefixedName" $kubeletMetricsConfig }}
+  labels:
+    {{- include "liqo.labels" $kubeletMetricsConfig | nindent 4 }}
+spec:
+  namespaceSelector:
+    any: true
+  selector:
+    matchLabels: 
+      app.kubernetes.io/name: "virtual-kubelet"
+      app.kubernetes.io/component: "virtual-kubelet"
+  podMetricsEndpoints:
+  - port: metrics
+    interval: {{ .Values.virtualKubelet.metrics.podMonitor.interval }}
+    scrapeTimeout: {{ .Values.virtualKubelet.metrics.podMonitor.scrapeTimeout }}
+{{- end }}
+
--- a/pkg/virtualKubelet/metrics/metrics.go
+++ b/pkg/virtualKubelet/metrics/metrics.go
@@ -1,0 +1,84 @@
+// Copyright 2019-2023 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"k8s.io/klog/v2"
+)
+
+const (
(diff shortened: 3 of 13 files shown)


pkg/virtualKubelet/metrics/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	MetricsPort = ":9090"
+)
+
+var (
+	// ErrorsCounter is the counter of the errors occurred during the reflection.
+	ErrorsCounter *prometheus.CounterVec
+	// ItemsCounter is the counter of the items extracted from the workqueue and correctly processed.
+	ItemsCounter *prometheus.CounterVec
+	// MetricsLabels is the labels used for the metrics.
+	MetricsLabels []string
```
aleoli, line 23: Do we need this variable to be global and exported from the package? Can we use a local-scoped variable inside the function?

pkg/virtualKubelet/metrics/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+
+const (
+	// Metrics port
+	MetricsPort = ":9090"
+)
+
+var (
+	// ErrorsCounter is the counter of the errors occurred during the reflection.
+	ErrorsCounter *prometheus.CounterVec
+	// ItemsCounter is the counter of the items extracted from the workqueue and correctly processed.
```
aleoli, line 20:
```suggestion
	// ItemsCounter is the counter of the reflected resources. A fast increase of this metric can indicate a race condition between local and remote operators.
```

pkg/virtualKubelet/metrics/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+			Name: "liqo_virtual_kubelet_reflection_error_counter",
+			Help: "The counter of the transient errors.",
+		},
+		MetricsLabels,
+	)
+
+	ItemsCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "liqo_virtual_kubelet_reflection_item_counter",
+			Help: "The counter of the items extracted from the workqueue and correctly processed.",
```
aleoli, line 41: Change here accordingly to the variable comment

pkg/virtualKubelet/metrics/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+		MetricsLabels,
+	)
+
+	// Register the metrics to the prometheus registry.
+	prometheus.MustRegister(ItemsCounter)
+	prometheus.MustRegister(ErrorsCounter)
```
aleoli, line 48: Move these lines in the setup function to avoid accidental registration of these metrics on package imports

cmd/virtual-kubelet/root/root.go
```diff
@@ -189,6 +191,8 @@ func runRootCommand(ctx context.Context, c *Opts) error {
 		return errors.Wrap(err, "error while setting up HTTPS server")
 	}
 
+	metrics.SetupMetricHandler()
```
aleoli, line 194: It should only be set up if the flag is enabled

pkg/virtualKubelet/provider/provider.go
```diff
@@ -73,6 +73,7 @@ type InitConfig struct {
 	EnableStorage              bool
 	VirtualStorageClassName    string
 	RemoteRealStorageClassName string
+	EnabledMetrics             bool
```
aleoli, line 76: It seems to be never used

pkg/virtualKubelet/reflection/generic/reflector.go
```diff
@@ -35,6 +35,9 @@ import (
 	traceutils "github.com/liqotech/liqo/pkg/utils/trace"
 	"github.com/liqotech/liqo/pkg/virtualKubelet/reflection/manager"
 	"github.com/liqotech/liqo/pkg/virtualKubelet/reflection/options"
+
+	"github.com/liqotech/liqo/pkg/virtualKubelet/metrics"
+	"github.com/prometheus/client_golang/prometheus"
```
aleoli, line 40: They should not be in another import group

pkg/virtualKubelet/reflection/generic/reflector.go
```diff
@@ -199,9 +202,18 @@ func (gr *reflector) processNextWorkItem() bool {
 
 		// Put the item back on the workqueue to handle any transient errors.
 		gr.workqueue.AddRateLimited(key)
+
+		// Increase the error counter metric.
+		metrics.ErrorsCounter.With(prometheus.Labels{"namespace": key.(types.NamespacedName).Namespace,
+			"reflector_type": gr.name}).Inc()
```
aleoli, line 208: We should increment this counter also for enqueueAfterErrors

pkg/virtualKubelet/metrics/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	ErrorsCounter *prometheus.CounterVec
+	// ItemsCounter is the counter of the items extracted from the workqueue and correctly processed.
+	ItemsCounter *prometheus.CounterVec
+	// MetricsLabels is the labels used for the metrics.
+	MetricsLabels []string
+)
+
+// Init initializes the metrics.
+func init() {
+	MetricsLabels = []string{"namespace", "reflector_type"}
```
aleoli, line 28: In my opinion "reflector_type" meaning is hard to understand, I suggest something like "reflected_resource" or similar

deployments/liqo/templates/liqo-virtualkubelet-podmonitor.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: {{ include "liqo.prefixedName" $kubeletMetricsConfig }}
+  labels:
+    {{- include "liqo.labels" $kubeletMetricsConfig | nindent 4 }}
+spec:
+  namespaceSelector:
+    any: true
```
aleoli, line 12:
```suggestion
    matchLabels:
      discovery.liqo.io/tenant-namespace: "true"
```

deployments/liqo/templates/liqo-virtualkubelet-podmonitor.yaml
```diff
@@ -0,0 +1,21 @@
...
+  selector:
+    matchLabels: 
+      app.kubernetes.io/name: "virtual-kubelet"
```
aleoli, line 15:
```suggestion
      app.kubernetes.io/name: "virtual-kubelet"
      app.kubernetes.io/component: "virtual-kubelet"
```

Sharathmk99:
@damianot98 @aleoli This feature is required for us as well. Can I open fresh PR with the changes as this PR has no activity?
Thanks

frisso:
> @damianot98 @aleoli This feature is required for us as well. Can I open fresh PR with the changes as this PR has no activity? Thanks

Let me check if this guy would like to finalize this PR, or we take it over.

frisso:
@damianot98 (cc: @aleoli )
1) Why is this PR still in Draft?
2) Is there any documentation required for this feature to be used/known by users?
3) The description of this PR says: "This can be achieved by setting the —enable-metrics flag during the installation, which will expose and create the PodMonitor resource." Is there any way to turn on/off this feature *after* the installation?

cheina97 (approved):

PR #1606, opened 2022-12-19
VK: enable retrieval of SA tokens through API (k8s >= 1.24)

### LINKED ISSUE #1185: [Feature] Introduce the support for offloaded applications interacting with the API server

This feature aims to introduce the possibility for offloaded applications to interact with the home API server, thus supporting scenarios such as controllers, high-availability database solutions, and in general all the workloads that leverage service accounts for authentication purposes. 

This feature involves three main aspects:
1. The reflection of the secrets associated with the service accounts, and the appropriate mutation of the pod specifications (#1184).
2. The appropriate remapping of the API server address, to ensure the communication with the home cluster instead of the remote one (#1196).
3. The implementation of a logic to retrieve tokens from the TokenRequest API and store them in remote secrets, to support for token rotation and configuration of specific audiences. This is similar to the [token_manager](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/token/token_manager.go) functionality implemented by the standard Kubelet.

**Describe alternatives you've considered**
As for point 2., two main alternatives are possible:
1. Leverage the http proxy already used by `liqoctl connect`, appropriately configuring the HTTP_PROXY variables;
2. Configure a static DNS host alias, to associate the home API server IP address to `kubernetes.default`. 

In the end, #1196 follows strategy 2., as using the http proxy would mean forcing all traffic from offloaded pods to flow through it, which might introduce an excessive overhead.

giorio94:
# Description

This PR modifies the virtual kubelet to introduce the support for the retrieval of service account tokens from the dedicated API, in addition to the standard secrets-based approach. This enables the support for offloaded applications interacting with the local Kubernetes API server, when hosted on Kubernetes 1.24 and above. Moreover, it includes a few related changes, allowing to filter reflection events per type, and enqueue a given event after a certain interval elapsed. 

Fixes #1185 

# How Has This Been Tested?

- [x] Existing E2E tests (now enabled also on k8s 1.24+)
- [x] Unit tests (new + existing)
- [x] Manual testing on Kind


pkg/virtualKubelet/forge/pods.go
```diff
@@ -346,18 +359,23 @@ func RemoteVolumes(volumes []corev1.Volume, enableAPIServerSupport bool, saSecre
...
 						continue
 					}
 
 					// Replace the ServiceAccountToken entry with the corresponding secret one, only in case it is enabled.
+					secretKey := corev1.ServiceAccountTokenKey
+					if apiServerSupport == APIServerSupportTokenAPI {
+						secretKey = ServiceAccountTokenKey(volumes[i].Name, source.ServiceAccountToken.Path)
+					}
+
 					source.ServiceAccountToken = nil
```
liggitt, line 375: this looks like it is dropping audience information from `source.ServiceAccountToken`, which is critical to inject the correct token... without that, this will default to injecting tokens valid for use against the kubernetes api-server, which could enable a privilege escalation if the user had requested a token scoped to a different audience

giorio94: Thanks @liggitt for your comment.

Still, I feel that I might be missing something, as the current behavior seems correct to me. 

Let me give a bit of context. The idea here is to enable pods which are offloaded to remote clusters to contact back the origin API server, as well as support scenarios in which offloaded pods use ServiceAccount tokens with specific audiences to authenticate among each others (e.g., service mesh scenarios).

The solution proposed involves a two step process:

1. A controller recognizes that a new pod mounting one or more service account tokens has been offloaded to a remote cluster. It requests the creation of the token(s) from the origin API server (for the requested audience), and saves them within a secret in the remote cluster. Additionally, it takes care of the refresh/deletion process when necessary. This reflection logic is implemented [here](https://github.com/giorio94/liqo/blob/mio%2Fsa-reflection-token-api/pkg/virtualKubelet/reflection/configuration/serviceaccount.go), and the *TokenRequest* is forged [here](https://github.com/giorio94/liqo/blob/a9b6d49c3b42f5fceba9b0dc457ac6690a65ea9f/pkg/virtualKubelet/forge/secrets.go#L230-L244).
2. The pod specification (and in particular that of the volumes) propagated to the remote cluster is mutated to account for this, changing the ServiceAccountToken references (since the kubelet on the target node would request the tokens from the remote API server) to the Secret entries created by the logic part of step 1 (and containing the tokens valid for the origin cluster). This logic is the one you mentioned, and available [here](https://github.com/giorio94/liqo/blob/a9b6d49c3b42f5fceba9b0dc457ac6690a65ea9f/pkg/virtualKubelet/forge/pods.go#L371-L382).

The `ServiceAccountTokenKey` function you mentioned is not responsible for retrieving the actual service account token (I see that the name might be a bit misleading), but only the key identifying the entry within the secret containing the actual token. This key is the concatenation of the name of the volume itself, and an hash of the path where that token will be mounted (cf. [here](https://github.com/giorio94/liqo/blob/mio%2Fsa-reflection-token-api/pkg/virtualKubelet/forge/secrets.go#L117-L123)). It does not include any audience information (although it would be possible), since two ServiceAccountToken volume entries cannot share the same volume name and path AFAIU. Yet, the token itself is scoped with the requested audience, thus preventing privilege escalation.

aleoli (approved):

pkg/virtualKubelet/forge/pods.go
liggitt: > This key is the concatenation of the name of the volume itself, and an hash of the path where that token will be mounted (cf. [here](https://github.com/giorio94/liqo/blob/mio%2Fsa-reflection-token-api/pkg/virtualKubelet/forge/secrets.go#L117-L123)). It does not include any audience information (although it would be possible), since two ServiceAccountToken volume entries cannot share the same volume name and path AFAIU

Is the secret unique per pod? If not, two different pods with the same service account, volume name, and mount path could specify two different audiences, which would collide, right?

giorio94: Yes, the secret is unique per pod.

liggitt: > Yes, the secret is unique per pod.

ok, I missed that... tracing the secret name getter looked like it was unique per service account, but not per pod

giorio94: Probably that getter name is a bit misleading since it takes the pod name as parameter...

PR #1633, opened 2023-01-18
Add optional feature to guarantee better service continuity in case of worker node failure in the remote cluster

Fra98:
# Description

This PR add the possibility to ensure that offloaded pods running on a failed node are evicted and rescheduled on a healthy node, preventing them to remain in a terminating state indefinitely. 

Use case: this feature can be useful in case of remote worker node failure to guarantee better service continuity and to have the expected pods workload running on the remote cluster.

Downside: it could produce zombies in the worker node, in case the node returns *Ready* again without a restart.

Implementation: a controller in the *liqo-controller-manager* checks for all offloaded pods (i.e. managed by a *shadowpod*) that are *Terminating* **and** running on a node marked as *NotReady*. These pods are then force deleted. The *shadowpod* reconciler will then take care of rescheduling a new pod on a healthy node.

# How Has This Been Tested?

- Locally, with 2 running clusters. 
  * Offload an application
  * On the remote cluster, stop one worker node (hosting offloaded pods) to make it *NotReady*


pkg/liqo-controller-manager/nodefailure-controller/nodefailure_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+// +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;delete
+
+// Reconcile nodes objects.
+func (r *NodeFailureReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Fetch the Node instance
+	var node corev1.Node
+	if err := r.Get(ctx, client.ObjectKey{Name: req.Name}, &node); err != nil {
+		klog.Errorf("an error occurred while getting node %s: %v", req.Name, err)
```
aleoli, line 54: I would ignore the not found error also for the message (or at least put it in higher verbosity in that case)

pkg/liqo-controller-manager/nodefailure-controller/nodefailure_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// If node is Ready exit without doing anything
+	if utils.IsNodeReady(&node) {
+		return ctrl.Result{}, nil
+	}
+
+	// Node NotReady: delete all terminating pods that are managed by shadowpods
+	var pods corev1.PodList
+	labelSelector := client.MatchingLabelsSelector{Selector: labels.Set{consts.ManagedByLabelKey: consts.ManagedByShadowPodValue}.AsSelector()}
+	fieldSelector := client.MatchingFieldsSelector{Selector: fields.OneTermEqualSelector(nodeNameField, node.Name)}
```
aleoli, line 66: Use more meaningfull names, such as offloadedPodSelector and nodePodSelector

pkg/liqo-controller-manager/nodefailure-controller/nodefailure_controller.go
```diff
@@ -0,0 +1,127 @@
...
+// getPodTerminatingEventHandler returns an event handler that reacts on Pod updates.
+// In particular, it reacts on changes over pods that are terminating and managed by a ShadowPod,
+// triggering the reconciliation of the related node hosting the pod.
+func getPodTerminatingEventHandler(c client.Client) handler.EventHandler {
+	return &handler.Funcs{
+		UpdateFunc: func(ue event.UpdateEvent, rli workqueue.RateLimitingInterface) {
+			ownedByShadowPod := ue.ObjectNew.GetLabels()[consts.ManagedByLabelKey] == consts.ManagedByShadowPodValue
+			isTerminating := !ue.ObjectNew.GetDeletionTimestamp().IsZero()
+			if ownedByShadowPod && isTerminating {
+				ctx := context.Background()
```
aleoli, line 94: We can forward the context from the caller

pkg/liqo-controller-manager/nodefailure-controller/nodefailure_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+func extractNodeNameFromPod(rawObj client.Object) []string {
+	pod := rawObj.(*corev1.Pod)
+	return []string{pod.Spec.NodeName}
+}
+
+// SetupWithManager monitors updates on nodes and watch for pods that are terminating and managed by a ShadowPod.
+func (r *NodeFailureReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Add field containing node Name to the Field Indexer
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Pod{}, nodeNameField, extractNodeNameFromPod); err != nil {
```
aleoli, line 119: Again, we can forward the context from the caller

pkg/liqo-controller-manager/nodefailure-controller/nodefailure_controller.go
```diff
@@ -0,0 +1,127 @@
...
+				if !ok {
+					klog.Errorf("object %v is not a pod", ue.ObjectNew)
+					return
+				}
+				nodeName := pod.Spec.NodeName
+				var node corev1.Node
+				if err := c.Get(ctx, client.ObjectKey{Name: nodeName}, &node); err != nil {
+					klog.Errorf("unable to get node %s (through pod %q): %v", nodeName, klog.KObj(pod), err)
+					return
+				}
```
aleoli, line 105: This is useless. You should not use the client in the handler function, and you don't need to

You may check that the `nodeName != ""` before add it

pkg/liqo-controller-manager/nodefailure-controller/suite_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+var _ = BeforeSuite(func() {
+	testutil.LogsToGinkgoWriter()
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "deployments", "liqo", "crds")},
+		ErrorIfCRDPathMissing: true,
+	}
```
aleoli, line 47: Do you need envtest or you can do it with the fakeClient? It will simplify a lot, makeing it possible to re-create a new cleint per test and do not clean it each time

frisso:
@aleoli 
1) Where should we add a better description of what exactly happens in the system? Here in the PR description?
2) Should we add a patch to the documentation here in this PR, or in a separate one, dedicated to the documentation?

aleoli:
Hi @frisso,

> 1. Where should we add a better description of what exactly happens in the system? Here in the PR description?

Yes, we can add it here. @Fra98 can you add to the description some lines about how you addressed the problem _"ensure that offloaded pods running on a failed node are evicted and rescheduled on a healthy node, preventing them to remain in a terminating state indefinitely"_

> 2. Should we add a patch to the documentation here in this PR, or in a separate one, dedicated to the documentation?

We also need a second PR on endpoints reflection to correctly handle service continuity, I would wait for it to have a full example for this scenario

Fra98:
@frisso, @aleoli 

> Yes, we can add it here. @Fra98 can you add to the description some lines about how you addressed the problem _"ensure that offloaded pods running on a failed node are evicted and rescheduled on a healthy node, preventing them to remain in a terminating state indefinitely"_
> 

PR description expanded

pkg/liqo-controller-manager/nodefailure-controller/nodefailure_controller_test.go
```diff
@@ -0,0 +1,194 @@
...
+	})
+
+	JustBeforeEach(func() {
+		r := &nodefailurectrl.NodeFailureReconciler{
+			Client: fakeClient,
+			Scheme: scheme.Scheme,
+		}
+		res, err = r.Reconcile(ctx, reqNode)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(res).To(BeZero())
```
aleoli, line 145: Why it should be zero?

aleoli (approved):

PR #1655, opened 2023-02-01
Liqoctl: status cluster-labels bug-fix

cheina97:
# Description

This PR fix a bug which cause an error in **liqoctl status** output when there are no cluster labels associated to a liqo installation (controller-manager have not the **--cluster-labels** argument). In order to avoid this in the future, tests have been extended to cover this case.

### DIFF
--- a/pkg/liqoctl/status/local/localinfo_test.go
+++ b/pkg/liqoctl/status/local/localinfo_test.go
@@ -45,43 +45,24 @@ var _ = Describe("LocalInfo", func() {
 	)
 
 	var (
-		clientBuilder fake.ClientBuilder
-		lic           *LocalInfoChecker
-		ctx           context.Context
-		text          string
-		options       status.Options
-		baseObjects   []client.Object
+		clientBuilder     fake.ClientBuilder
+		lic               *LocalInfoChecker
+		ctx               context.Context
+		text              string
+		options           status.Options
+		baseObjects       []client.Object
+		argsClusterLabels []string
 	)
 
 	BeforeEach(func() {
 		ctx = context.Background()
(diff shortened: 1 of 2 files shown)


pkg/liqoctl/status/local/localinfo_test.go
```diff
@@ -110,82 +91,76 @@ var _ = Describe("LocalInfo", func() {
...
+		clientBuilder.WithObjects(objects...)
+		options.CRClient = clientBuilder.Build()
+		lic = NewLocalInfoChecker(&options)
+		lic.Collect(ctx)
+
+		text = lic.Format()
+		text = pterm.RemoveColorFromString(text)
+		text = testutil.SqueezeWhitespaces(text)
+
+		fmt.Fprint(GinkgoWriter, text)
```
aleoli, line 117: Do we need it?

pkg/liqoctl/status/local/localinfo_test.go
```diff
@@ -110,82 +91,76 @@ var _ = Describe("LocalInfo", func() {
...
-				text = testutil.SqueezeWhitespaces(text)
-			})
-			It("should format a valid text", func() {
-				Expect(text).To(ContainSubstring(
-					pterm.Sprintf("Address: %s", fmt.Sprintf("https://%v", testutil.OverrideAPIAddress)),
-				))
-			})
-		})
-	})
+	DescribeTable("Collecting and Formatting LocalInfoChecker", func(clusterLabels, apiServerOverride bool) {
```
aleoli, line 95: You can encapsulate the test case arguments in a struct, which makes the entries' definitions more readable.

aleoli (approved):

PR #1660, opened 2023-02-07
Liqonet: wg userspace metric

cheina97:
# Description

This PR add the possibility to force **wireguard userspace** implementation using helm **values.yaml** file.
In addition the information about what implementation Liqo is using has been added in **metrics** and in **grafana**. 

image

image

### DIFF
--- a/deployments/liqo/values.yaml
+++ b/deployments/liqo/values.yaml
@@ -90,6 +90,8 @@ gateway:
     portOverride: ""
     # -- port used by the vpn tunnel.
     listeningPort: 5871
+    # -- implementation used by wireguard. Possible values are "userspace" and "kernel". Do not use "userspace" unless strictly necessary.
+    wireguardImplementation: "kernel"
   metrics:
     # -- expose metrics about network traffic towards cluster peers.
     enabled: false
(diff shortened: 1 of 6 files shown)


deployments/liqo/values.yaml
```diff
@@ -90,6 +90,8 @@ gateway:
     portOverride: ""
     # -- port used by the vpn tunnel.
     listeningPort: 5871
+    # -- implementation used by wireguard. Possible values are "userspace" and "kernel".
```
aleoli, line 93: I would add a warning to the user to not use the userspace version unless necessary

pkg/liqonet/tunnel/wireguard/driver.go
```diff
@@ -518,6 +546,17 @@ func (w *Wireguard) Collect(ch chan<- prometheus.Metric) {
...
+	var wgUsrActive float64
+	if w.Implementation == Userspace {
+		wgUsrActive = 1
+	}
+	ch <- prometheus.MustNewConstMetric(
+		WgUserImpl,
+		prometheus.GaugeValue,
+		wgUsrActive,
+		[]string{liqoconst.DriverName}...,
+	)
```
aleoli, line 558: I would use the metrics differently. We can define a GaugeVect and then use a different label value for each implementation type (similar to how it is done with k8s pod phases). In this way is clearer to read the metric, and we can also support multiple values for that metric

Screenshot 2023-02-09 at 09 21 56

aleoli (approved):

Issue #1671, opened 2023-02-14
[Feature] Decouple Offloading and Networking


aleoli:
### Describe the solution you'd like

The possibility to disable the networking module of Liqo at install time. This will allow offloading pods without enabling the pod-to-pod connectivity between the two peered clusters if not required, enforcing more robust security requirements among them.

At the same time, this feature will allow Liqo users to use external networking tools that better suit their specific use cases.

### Required Steps

* [ ] Disable VirtualKubelet IPAM connection and EPS/IPs translations
* [ ] Disable the Node controller network condition check if this feature is enabled
* [ ] Add a flag to the VirtualKubelet to make it aware of this configuration
* [ ] Add a flag to disable Liqo network deployments, services, and daemon sets
* [ ] Fix telemetry forwarding APIs that cause some external tools to be unable to process APIs

aleoli:
Closing as completed

PR #1673, opened 2023-02-16
Make it possible to configure IPAM module

aleoli:
# Description

This pr allows the Liqo users to use different IPAM modules (implementing the Liqo APIs) or to disable the IPAM completely, also disabling the IP translation between the two clusters (this may be useful in case of not overlapping pod CIDRs with an external network layer)

Ref #1671

# How Has This Been Tested?

- [x] locally on KinD with the Liqo network module enabled
- [x] locally on KinD without the IP translation

### DIFF
--- a/cmd/virtual-kubelet/root/opts.go
+++ b/cmd/virtual-kubelet/root/opts.go
@@ -15,7 +15,6 @@
 package root
 
 import (
-	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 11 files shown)


cmd/virtual-kubelet/root/opts.go
```diff
@@ -105,7 +103,7 @@ func NewOpts() *Opts {
 		TenantNamespace:      corev1.NamespaceDefault,
 		InformerResyncPeriod: DefaultInformerResyncPeriod,
 
-		LiqoIpamServer: fmt.Sprintf("%v:%v", consts.NetworkManagerServiceName, consts.NetworkManagerIpamPort),
+		LiqoIpamServer: "",
```
cheina97, line 106:
```suggestion
		
```
Could be avoided as it is not the only value not set in this struct.

cheina97 (approved):

PR #1677, opened 2023-02-17
Resiliency to network problems in the remote (provider) cluster through Shadow Endpointslices

Fra98:
# Description

This PR implements a resliency mechanism to guarantee service continuity in the remote (provider) cluster when the local (consumer) cluster is unhealhty. This PR handles network problems in the Liqo-gateway tunnel, but the implementation allows to extend it to other scenarios (i.e., local API server unavailable) in future. 

## The issue
When an application or client in the remote cluster contacts a service that has some endpoints in the local cluster, all requests to that cluster will fail if the Liqo-gateway network is unavailable.
## Implementation
Resiliency is achieved by implementing a new custom resource, the ShadowEndpointSlice.
This resource is created by the virtual kubelet in the local cluster and is a template for the desired endpointslice in the remote cluster. A custom controller in the remote cluster whatches for ShadowEndpointSlices resources and creates/updates the actual remote endpointslice required by the Liqo reflection logic. When a network problem occurs in the Liqo-gateway, it reacts by invalidating the status of all endpoints in the remote endpointslice. In this way, services do not redirect anymore the traffic through the unavailable tunnel, and service continuity is guaranteed. Subsequently, the endpoints are restored when the tunnel becomes available again.

# How Has This Been Tested?

- Not yet tested


pkg/liqo-controller-manager/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+		klog.Errorf("an error occurred while getting shadowendpointslice %s: %v", nsName, err)
+		return ctrl.Result{}, err
+	}
+
+	// Get ForeignCluster associated with the shadowendpointslice
+	clusterID, ok := shadowEps.Labels[forge.LiqoOriginClusterIDKey]
+	if !ok {
+		klog.Errorf("shadowendpointslice %s has no label with origin clusterID", nsName)
+		return ctrl.Result{}, fmt.Errorf("shadowendpointslice %s has no label with origin clusterID", nsName)
```
aleoli, line 80:
```suggestion
		return ctrl.Result{}, fmt.Errorf("shadowendpointslice %s has no label %s", nsName, forge.LiqoOriginClusterIDKey)
```

pkg/liqo-controller-manager/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	fc, err := foreigncluster.GetForeignClusterByID(ctx, r.Client, clusterID)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.Error("foreigncluster associated with shadowendpointslice %s not found", nsName)
```
aleoli, line 85: I would show somewhere the clsuterID that is not found

pkg/liqo-controller-manager/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -0,0 +1,244 @@
...
+			return ctrl.Result{}, err
+		}
+		klog.Errorf("an error occurred while getting foreigncluster %s: %v", clusterID, err)
+		return ctrl.Result{}, err
+	}
+
+	// Check network status of the foreigncluster
+	networkStatus := foreigncluster.IsNetworkingEstablished(fc)
```
aleoli, line 93: Open question: what we should do when the networking is extrernal? Should we use the more generic function (not merged yet)? https://github.com/liqotech/liqo/blob/877d0b92babbcedc90f0265700d4f53408a5fe69/pkg/utils/foreignCluster/peeringStatus.go#L80

pkg/liqo-controller-manager/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -0,0 +1,244 @@
...
+		if err := r.Patch(ctx, &existingEps, clientutils.Patch(epsApply),
+			client.ForceOwnership, client.FieldOwner(ctrlFieldManager)); err != nil {
+			klog.Errorf("unable to update endpointslice %q: %v", klog.KObj(&existingEps), err)
+			return ctrl.Result{}, err
+		}
+
+		klog.Infof("updated endpointslice %q with success", klog.KObj(&existingEps))
+
+		return ctrl.Result{}, nil
+	}
```
aleoli, line 130:
```suggestion
    err = r.Get(ctx, nsName, &existingEps)
    switch {
    case errors.IsNotFound(err):
        break
    case err != nil:
        ...
    default:
        ...
    }
```

pkg/liqo-controller-manager/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// Else, create a new EndpointSlice
+	newEps := discoveryv1.EndpointSlice{
```
aleoli, line 133: It's probably better (for readability reasons) to create this resource at the beginning and then operate over it

pkg/liqo-controller-manager/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	newForeignCluster, ok := newObj.(*discoveryv1alpha1.ForeignCluster)
+	if !ok {
+		klog.Errorf("object %v is not a ForeignCluster", newObj)
+		return false
+	}
+
+	oldFcNetworkReady := foreigncluster.IsNetworkingEstablished(oldForeignCluster)
+	newFcNetworkReady := foreigncluster.IsNetworkingEstablished(newForeignCluster)
```
aleoli, line 208: The same open question here

pkg/liqo-controller-manager/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -0,0 +1,244 @@
...
+		DeleteFunc:  func(e event.DeleteEvent) bool { return false },
+		CreateFunc:  func(e event.CreateEvent) bool { return false },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return r.endpointsShouldBeUpdated(e.ObjectNew, e.ObjectOld) },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	}
+
+	epsPredicates := predicate.Funcs{
+		DeleteFunc:  func(e event.DeleteEvent) bool { return true },
+		CreateFunc:  func(e event.CreateEvent) bool { return false },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return r.endpointSliceMetaChanged(e.ObjectNew, e.ObjectOld) },
```
aleoli, line 233: Why this?

pkg/virtualKubelet/roles/remote/role.go
```diff
@@ -26,4 +26,5 @@ package remote
 // +kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses,verbs=get;list;watch;create;update;patch;delete
 
 // +kubebuilder:rbac:groups=virtualkubelet.liqo.io,resources=shadowpods,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=virtualkubelet.liqo.io,resources=shadowendpointslices,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=discovery.k8s.io,resources=endpointslices,verbs=get;list;watch;create;update;patch;delete
```
aleoli, line 30: I think that this should be removed

aleoli (approved):

PR #1691, opened 2023-02-25
E2E fix + liqoctl bump golang version to 1.19

cheina97:
# Description

This PR mainly reverts the golang version used to compile **liqoctl** to solve some issues with the e2e tests.

the main changes introduced are:

- Liqoctl building in CI using a script instead of github actions (reason why explained below)
- Added K3S E2E tests using K3D
- Bug fix in KIND E2E tests
- Removed local containers registry in E2E tests

## Why build liqoctl in a script ?
In our PR we use a **slash command dispatcher** (https://github.com/peter-evans/slash-command-dispatch) to run the test and build the pipeline. This **github action** is based on **repository dispatch** which can be received only from the **main/master** branch. So when the golang version (used to compile liqoctl) changes, we cannot leverage on values contained inside the pipeline to test the new golang version.

PR #1697, opened 2023-03-03
Bump Liqoctl Golang version to 1.19

cheina97:
# Description
This pr bumps the Liqoctl Golang version to v1.19 because Liqoctl is not able to pass the e2e tests if compiled with version 1.20 .


cheina97:
This PR is going to be solved in #1691

PR #1703, opened 2023-03-08
Add readiness status of the foreign API server to the ForeignCluster CR

Fra98:
# Description

This PR add to the ForeignCluster a new condition status tracking the readiness of the API server of a given ForeignCluster.
Ref #1705

## Implementation
The ForeignCluster controller launchs a routine which periodically pings the foreign API server to check for its readiness. Whenever the status change, it updates the *ForeignAPIServerStatus* condition in the FC accordingly.

## Motivation
The new *ForeignAPIServerStatus* status condition can be used by the ShadowEndpointSlices controller to check the current status of the foreign cluster API server. If it is unhealthy or unreachable, it will invalidate all endpoints of the pods scheduled on that foreign cluster, so that Services do not redirect the traffic through them. Note: this PR does not include any modifications to the ShadowEndpointSlices controller logic, which will be updated in another PR.

# How Has This Been Tested?

- [x] Locally on Kind
- [x] Unit tests
- [x] e2e tests

### DIFF
--- a/apis/discovery/v1alpha1/foreigncluster_types.go
+++ b/apis/discovery/v1alpha1/foreigncluster_types.go
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package v1alpha1 contains API Schema definitions for the discovery v1alpha1 API group
+//
+//nolint:lll // ignore long lines given by Kubebuilder marker annotations
 package v1alpha1
 
 import (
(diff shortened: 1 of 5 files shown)


apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -155,16 +155,18 @@ const (
 	IncomingPeeringCondition PeeringConditionType = "IncomingPeering"
 	// NetworkStatusCondition informs users about the network status.
 	NetworkStatusCondition PeeringConditionType = "NetworkStatus"
+	// ForeignAPIServerStatusCondition informs users about the foreign API server status.
+	ForeignAPIServerStatusCondition PeeringConditionType = "ForeignAPIServerStatus"
```
aleoli, line 159:
```suggestion
	APIServerStatusCondition PeeringConditionType = "APIServerStatus"
```

apis/discovery/v1alpha1/foreigncluster_types.go
```diff
@@ -198,6 +200,7 @@ type TenantNamespaceType struct {
 // +kubebuilder:printcolumn:name="Outgoing peering",type=string,JSONPath=`.status.peeringConditions[?(@.type == 'OutgoingPeering')].status`
 // +kubebuilder:printcolumn:name="Incoming peering",type=string,JSONPath=`.status.peeringConditions[?(@.type == 'IncomingPeering')].status`
 // +kubebuilder:printcolumn:name="Networking",type=string,JSONPath=`.status.peeringConditions[?(@.type == 'NetworkStatus')].status`
+// +kubebuilder:printcolumn:name="API Server",type=string,JSONPath=`.status.peeringConditions[?(@.type == 'ForeignAPIServerStatus')].status`
```
aleoli, line 203: I would add a "priority=1" annotation (as for the cluster id and cluster name) to avoid too wide outputs

pkg/liqo-controller-manager/foreign-cluster-operator/apiServerChecker.go
```diff
@@ -0,0 +1,200 @@
...
+	clusterID := foreignCluster.Spec.ClusterIdentity.ClusterID
+
+	// If the ping interval is 0, we consider the foreign API server as always ready.
+	if r.APIServerCheckers.pingInterval == 0 {
+		peeringconditionsutils.EnsureStatus(foreignCluster,
+			discoveryv1alpha1.ForeignAPIServerStatusCondition,
+			discoveryv1alpha1.PeeringConditionStatusEstablished,
+			apiServerReadyReason,
+			apiServerReadyMessage)
+		return true, ctrl.Result{}, nil
```
aleoli, line 69: You also need to ensure that the finalizer is not set in this case

pkg/liqo-controller-manager/foreign-cluster-operator/apiServerChecker.go
aleoli, line 1: The file name should be lowercase without -/_ (`apiserverchecker.go`).

I know it is not true for all liqo files, but we should try using this convention for new files.

aleoli (commented):
It looks good to me, please squash your commits, and then we can proceed with merging this pr

aleoli (approved):

PR #1727, opened 2023-03-21
Liqoctl: status supports internal network disabled

cheina97:
# Description

This PR adds support for Liqo with **disabled internal network**.

Ref #1671

## Screenshot

image

# How Has This Been Tested?

- [x] Locally with Kind

### DIFF
--- a/pkg/liqoctl/status/local/pods.go
+++ b/pkg/liqoctl/status/local/pods.go
@@ -23,6 +23,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/utils/strings/slices"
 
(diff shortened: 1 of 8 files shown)


pkg/liqoctl/status/local/pods.go
```diff
@@ -199,6 +200,13 @@ func (pc *PodChecker) Silent() bool {
 // it collects the status of the components of Liqo. The status is
 // collected at the pod level.
 func (pc *PodChecker) Collect(ctx context.Context) {
+	if pc.options.InternalNetworkDisabled {
+		pc.deployments = slice.RemoveString(pc.deployments, liqconsts.LiqoNetworkManagerName)
```
aleoli, line 204: It could probably be easier to define them in two separate lists and merge them if the internal network is enabled

aleoli (approved):
It looks ok to me, only two things, 1 check that it is running correctly also with peered clusters, and 2 add a step in the e2e test to check that the status command is returning a 0 exit code

PR #1742, opened 2023-03-27
Run e2e with cluster-api provided clusters

aleoli:
# Description

This pr adds a new e2e test provider that creates clusters with the kubevirt cluster-api implementation

### DIFF
--- a/pkg/liqoctl/move/handler.go
+++ b/pkg/liqoctl/move/handler.go
@@ -165,7 +165,7 @@ func getResticRepositoryURL(ctx context.Context, cl client.Client, isLocal bool)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/liqoctl/move/handler.go
```diff
@@ -165,7 +165,7 @@ func getResticRepositoryURL(ctx context.Context, cl client.Client, isLocal bool)
 		}
 	}
 
-	return fmt.Sprintf("rest:http://%s.%s.svc.cluster.local:%d/", resticRegistry, namespace, resticPort), nil
+	return fmt.Sprintf("rest:http://%s.%s.svc:%d/", resticRegistry, namespace, resticPort), nil
```
cheina97, line 168: Why is it necessary?

aleoli: Yes, it creates clusters with a different domain to avoid conflicts with the main cluster's DNS resolution

PR #1770, opened 2023-04-07
Docs: EKS minimum permissions required

cheina97:
# Description

This PR adds to the documentation, the minimum permissions required to install Liqo on EKS.

## Screenshot

image

### DIFF
--- a/docs/installation/install.md
+++ b/docs/installation/install.md
@@ -135,6 +135,43 @@ In a nutshell, after having installed the CLI, you have to set up your identity:
 aws configure
 ```
 
+You can install Liqo even if you are not an EKS administrator.
+The minimum **IAM** permissions required by a user to install Liqo are the following:
+
+```json
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "eks:DescribeCluster",
+                "iam:CreateUser",
+                "iam:CreateAccessKey",
+                "ec2:DescribeVpcs"
+            ],
+            "Resource": "*"
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "iam:CreatePolicy",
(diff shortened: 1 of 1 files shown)


aleoli (approved):

docs/installation/install.md
```diff
@@ -135,6 +135,41 @@ In a nutshell, after having installed the CLI, you have to set up your identity:
 aws configure
 ```
 
+The minimum permissions required by the user (in addition to the ones required to create and operate the EKS cluster) are the following:
```
frisso, line 138: This is not clear to me. If the user must already have the permissions to create and operate the EKS cluster, what are those permissions needed?
Probably a more verbose explanation would help here.

cheina97: I spent a few hours understanding how eks manages to integrate iam users with k8s users. The IAM user is only used to authenticate with the k8s cluster, and the cluster uses that authentication to remap the IAM user to a k8s user. This means that the management of permissions to "operate" on the cluster is not managed via IAM, but via k8s. So what I have written in parentheses loses its meanings.

cheina97: If an admin were to create a cluster and then delegate the liqo installation to another user (who has no create or destroy permissions on EKS, CloudFormation and EC2) the only additional permission needed would be **DescribeVPC**

frisso: This last comment looks interesting to me. It is worth mentioning in the documentation?

cheina97: I explained badly. There are 2 possible scenarios.
- A user who has the permissions to create and manage EKS clusters.
- A user who does not have permissions to create or manage EKS, but wants to install Liqo.
In both cases the permissions required to install Liqo are the ones that I'm adding to the documentation.

cheina97: I would add something like `Even if you are not an EKS administrator you can install Liqo`.

PR #1805, opened 2023-05-04
ShadowPod Status

### LINKED ISSUE #1804: Shadow Pod Controller recreating Completed/Failed pods
#### What happened:
Home cluster in on-premises with liqo installed and peered with private AKS cluster(remote cluster) with auto scaling enabled. Everything was fine till now.
When I create a pod A in home cluster with affinity to AKS cluster, liqo creates a pod in remote cluster which is good. Pod A will complete after X minutes and after 30min AKS auto scaler will scale down the node due to no activity in that node, when AKS scaling down the node it deleted all pod spec(completed status) now liqo recreates the completed status pod again to AKS cluster.

Pod should not be created if status is Completed/Failed.

Create Pod A in remote cluster and wait for completion. Delete the pod in remote cluster.

#### Environment:
- Liqo version: 0.8.0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Network plugin and version:
- Install tools:
- Others:

Sharathmk99:
# Description

Fixes #1804

# How Has This Been Tested?

- [x] Locally


frisso:
@Sharathmk99 Just a silly question: is the status of the pod (complete/failed) still visible when interacting with the home cluster, isn't it?

Sharathmk99:
> @Sharathmk99 Just a silly question: is the status of the pod (complete/failed) still visible when interacting with the home cluster, isn't it?

Yes. Home cluster Pod has the right status from remote cluster Pod.

apis/virtualkubelet/v1alpha1/shadowpod_types.go
```diff
@@ -24,15 +24,43 @@ type ShadowPodSpec struct {
...
+	// +kubebuilder:default="Unknown"
+	Phase PodPhase `json:"phase"`
+}
+
 // +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
 // +genclient
 
 // ShadowPod is the Schema for the Shadowpods API.
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.phase`
```
aleoli, line 57: Is the `Age` automatically printed, or you need to add a `// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`` annotation?

pkg/liqo-controller-manager/shadowpod-controller/shadowpod_controller.go
```diff
@@ -113,6 +118,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
+	// Update ShadowPod status same as Pod status
+	shadowPod.Status.Phase = vkv1alpha1.ShadowPodUnknown
+	if newErr := r.Client.Status().Update(ctx, &shadowPod); newErr != nil {
+		klog.Error(newErr)
+		return ctrl.Result{}, newErr
+	}
```
aleoli, line 126: Is this update required? I mean, tecnically this Reconcile should be triggered very soon by pod creation/updates and we will set the value when this happens

apis/virtualkubelet/v1alpha1/shadowpod_types.go
```diff
@@ -24,15 +24,43 @@ type ShadowPodSpec struct {
...
+	ShadowPodPending PodPhase = "Pending"
+	// ShadowPodRunning indicates running phase.
+	ShadowPodRunning PodPhase = "Running"
+	// ShadowPodSucceeded indicates succeeded phase.
+	ShadowPodSucceeded PodPhase = "Succeeded"
+	// ShadowPodFailed indicates failed phase.
+	ShadowPodFailed PodPhase = "Failed"
+	// ShadowPodUnknown indicates unknown phase.
+	ShadowPodUnknown PodPhase = "Unknown"
+)
```
aleoli, line 41: Can we re-use the corev1.PodPhase instead of re-define the type with its values?

pkg/liqo-controller-manager/shadowpod-controller/shadowpod_controller.go
```diff
@@ -66,6 +60,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 		return ctrl.Result{}, err
 	}
+	if shadowPod.Status.Phase == vkv1alpha1.ShadowPodSucceeded || shadowPod.Status.Phase == vkv1alpha1.ShadowPodFailed {
```
aleoli, line 63: Can we check the pod restart policy to take this decision?

aleoli (commented):
Hi @Sharathmk99, thanks for your contribution; it seems pretty good.

I left you some small comments and fixes.

pkg/liqo-controller-manager/shadowpod-controller/shadowpod_controller.go
Sharathmk99: Yes technically it's not required to update the status, just in case anything happens so updating. Will remove it. Thanks

pkg/liqo-controller-manager/shadowpod-controller/shadowpod_controller.go
Sharathmk99: I agree with. For Succeeded status we don't have to check restart Policy, but for failed pod we need check restart policy. This is a good point. Thank you for caching it.

aleoli: We should restart also Succeeded pods if the restart policy is Always. I know that the recreation of the pod is very unlikely because Kubernetes should restart the pod, but it is a possibility

apis/virtualkubelet/v1alpha1/shadowpod_types.go
Sharathmk99: Done

apis/virtualkubelet/v1alpha1/shadowpod_types.go
Sharathmk99: Done

pkg/liqo-controller-manager/shadowpod-controller/shadowpod_controller.go
Sharathmk99: I updated code to ignore the recreation only if `RestartPolicy` is set to `Never`. 
But I have a question,

- If pod has `RestartPolicy` set as `Always`, scheduled to remote cluster and it completed with `Success` status. Now `Node` is deleted this doesn't mean pod should be recreated right?
- If pod has `RestartPolicy` set as `OnFailure`, scheduled to remote cluster and it completed with `Success` status. Now `Node` is deleted this doesn't mean pod should be recreated right?
- If pod has `RestartPolicy` ser as 'Always', scheduled to remote cluster and it completed with 'Failed` stauts. Now anyway kube-scheduler will restart this pod to same node or other node. But if the node is deleted(when pod is in running state) it should be recreated as RestartPolicy is set to Always.
- if pod has `RestartPolicy` set as 'OnFailure', scheduled to remote cluster and it completed with `Failed` status. Now anyways kube-scheduler will restart this pod to same node or other node. But if the node is deleted(when pod is in running state) it should be recreated by ShadowPod Controller as RestartPolicy is set to Failed

I think if the pod status is success, irrespective of `RestartPolicy` ShadowPod Controller should not recreate the pod. But if the pod status is other than success then need to check the `RestartPolicy`, if it's set to `Never` don't recreate else recreate the pod. What do you think?

aleoli: I agree with points 2, 3, and 4, but not with the first. If I remember correctly, the only difference between `Success` and `Failed` is the exit code of the container.

If you run a pod that exits with code 0, for instance, an Ubuntu image without any command, it should alternate the `Success` and `CrashLoopBackoff` statuses since the kubelet restarts it. Following this scenario I think that we should restart the pod if it is succeeded with restart always policy set

Sharathmk99: I’ll make the charges.

Sharathmk99: Done. Please review. Thanks

aleoli (approved):

PR #1807, opened 2023-05-05
Add virtual-kubelet Grafana dashboard

damianot98:
This PR adds a Grafana Dashboard to monitor the reflected resources by the virtual-kubelet.
Metrics are exposed using the PodMonitor resource:

- **liqo_virtual_kubelet_reflection_item_counter**
- **liqo_virtual_kubelet_reflection_error_counter**

## Screenshot

Screenshot 2023-05-23 at 16 37 34

# How Has This Been Tested?

Locally with [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus)


damianot98:
@Sharathmk99 Yes it is possible. Right now the dashboard offers four selectors for the data source, the cluster, the namespace and a time range.

docs/installation/install.md
```diff
@@ -380,7 +380,7 @@ Finally, remember that:
 The main global flags, besides those concerning the installation of [development versions](InstallationDevelopmentVersions), include:
 
 * `--enable-ha`: enables the support for **high-availability of the Liqo components**, starting two replicas (in an active/standby configuration) of the **gateway** to ensure no cross-cluster connectivity downtime in case one of the replicas is restarted, as well as of the **controller manager**, which embeds the Liqo control plane logic.
-* `--enable-metrics`: exposes Liqo **metrics** through **prometheus**.
+* `--enable-metrics`: exposes Liqo **metrics** through **prometheus** (see the dedicated [prometheus metrics page](/usage/prometheus-metrics.md) for additional details).
```
frisso, line 383:
```suggestion
* `--enable-metrics`: exposes Liqo **metrics** through **Prometheus** (see the dedicated [Prometheus metrics page](/usage/prometheus-metrics.md) for additional details).
```

docs/usage/prometheus-metrics.md
```diff
@@ -27,3 +27,16 @@ We provide a {download}`sample Grafana dashboard </_downloads/grafana/liqonetwor
 As presented in the screenshot below, it includes an overview section presenting the overall cross-cluster throughput, followed by detailed per-peering throughput and latency information.
 
 ![Grafana Network Dashboard](/_static/images/usage/prometheus-metrics/network-dashboard.png)
+
+## Virtual kubelet metrics
+
+These metrics are available for each peered remote cluster, providing statistics about the reflected resources:
+
+- **liqo_virtual_kubelet_reflection_item_counter**: the number of reflected resources successfully processed.
```
frisso, line 35: Not clear what "resources" means. I would suggest to rephrase such as:
... the number of reflected resources (e.g., pods, ...)
(list, in the parenthesis, all the objects that are counted)

docs/usage/prometheus-metrics.md
```diff
@@ -27,3 +27,16 @@ We provide a {download}`sample Grafana dashboard </_downloads/grafana/liqonetwor
...
+- **liqo_virtual_kubelet_reflection_error_counter**: the number of transient errors during the reflection phase.
```
frisso, line 36: I would suggest to make an example here as well.
E.g.:
.... during the reflection phase, e.g., due to ....
(and list 2-3 reasons we may have errors here)

frisso:
Some additional comments on the screenshot.
1) Are graphs referred to "reflected items", or "reflection events"? This looks very different to me.
2) "Resource reflected" ==> "Reflected resources" (assuming, of course, that we're talking about resources, and not events)
3) "Reflected item rate" is not clear to me. What does "rate" mean? What are the number on the Y axis (missing measurement unit, which is a major error for an engineer)
Probably, more comments once the previous ones have been resolved.

docs/usage/prometheus-metrics.md
damianot98: Ok, I will rephrase that. For resources, I meant pods, configmaps, services and others.

docs/usage/prometheus-metrics.md
damianot98: Ok, I will add some examples.

damianot98:
@frisso about the previous comment:

1. Graphs are referred to "reflected items" those elements that are extracted from the workqueue and processed.
2. I will change that.
3. In this case rate means a per-second average rate of increase according to the specification of PromQL, but I would like to make some changes to the "rate" graphs.

docs/usage/prometheus-metrics.md
damianot98: While discussing it with @aleoli, the main motivation behind the errors is that the issue is caused by temporary race conditions that can be resolved by retrying the synchronization. These conditions mainly occur when some of the requested resources are not yet fully configured. There could be three specific cases to discuss (e.g., reflector not completely initialized or fallback not completely initialized), but we think it would be too detailed to add to the documentation.

docs/usage/prometheus-metrics.md
```diff
@@ -4,11 +4,12 @@ This section presents the metrics exposed by Liqo, using the [Prometheus](https:
 
 ## Scraping metrics
 
-Metrics are scraped depending on how your **Prometheus server(s)** has been configured.
-Liqo assumes that you leverage the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) to run Prometheus, hence providing a **ServiceMonitor** resource for each component.
+How metrics are scraped will depend on how you are operating your **Prometheus server(s)**.
```
frisso, line 7:
```suggestion
Metrics are scraped depending on how your **Prometheus server(s)** has been configured.
```

docs/usage/prometheus-metrics.md
```diff
@@ -4,11 +4,12 @@ This section presents the metrics exposed by Liqo, using the [Prometheus](https:
...
+Liqo presumes that the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) is being used to run Prometheus, providing a **ServiceMonitor** resource for each component and a **PodMonitor** resource for the Virtual kubelet.
```
frisso, line 8:
```suggestion
Liqo assumes that you leverage the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) is being used to run Prometheus, hence providing a **ServiceMonitor** resource for each component and a **PodMonitor** resource for the Virtual kubelet.
```

docs/usage/prometheus-metrics.md
```diff
@@ -4,11 +4,12 @@ This section presents the metrics exposed by Liqo, using the [Prometheus](https:
...
 
-Gathering of Liqo metrics is **disabled** by default.
-To allow Prometheus to scrape metrics from Liqo, you should set the `--enable-metrics` *liqoctl* flag during installation (cf. [installation customization options](InstallCustomization)), which enables Liqo to expose metrics, and it creates the proper **ServiceMonitor** and **PodMonitor** resources.
+Metrics are **disabled** by default.
```
frisso, line 10:
```suggestion
Gathering of Liqo metrics is **disabled** by default.
```

docs/usage/prometheus-metrics.md
```diff
@@ -4,11 +4,12 @@ This section presents the metrics exposed by Liqo, using the [Prometheus](https:
...
+In order to allow Prometheus to scrape metrics from the Liqo components, you should set the `--enable-metrics` *liqoctl* flag during installation (cf. [installation customization options](InstallCustomization)).
```
frisso, line 11:
```suggestion
To allow Prometheus to scrape metrics from Liqo, you should set the `--enable-metrics` *liqoctl* flag during installation (cf. [installation customization options](InstallCustomization)), which configures Liqo to expose metrics, and it creates the proper **ServiceMonitor** and **PodMonitor** resources.
```

docs/usage/prometheus-metrics.md
```diff
@@ -4,11 +4,12 @@ This section presents the metrics exposed by Liqo, using the [Prometheus](https:
...
+This flag enables the metrics exposition, the **PodMonitor** and the **ServiceMonitor** resources creation.
```
frisso, line 12:
```suggestion
```

docs/usage/prometheus-metrics.md
```diff
@@ -5,10 +5,10 @@ This section presents the metrics exposed by Liqo, using the [Prometheus](https:
 ## Scraping metrics
 
 Metrics are scraped depending on how your **Prometheus server(s)** has been configured.
-Liqo assumes that you leverage the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) to run Prometheus, hence providing a **ServiceMonitor** resource for each component.
+Liqo assumes that you leverage the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) is being used to run Prometheus, hence providing a **ServiceMonitor** resource for each component and a **PodMonitor** resource for the Virtual kubelet.
```
frisso, line 8:
```suggestion
Liqo assumes that you leverage the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) to run Prometheus, hence providing a **ServiceMonitor** resource for each component and a **PodMonitor** resource for the Virtual kubelet.
```

docs/usage/prometheus-metrics.md
```diff
@@ -27,3 +27,16 @@ We provide a {download}`sample Grafana dashboard </_downloads/grafana/liqonetwor
 As presented in the screenshot below, it includes an overview section presenting the overall cross-cluster throughput, followed by detailed per-peering throughput and latency information.
 
 ![Grafana Network Dashboard](/_static/images/usage/prometheus-metrics/network-dashboard.png)
+
+## Virtual kubelet metrics
+
+These metrics are available for each peered remote cluster, providing statistics about the reflected resources:
+
+- **liqo_virtual_kubelet_reflection_item_counter**: the number of reflected resources successfully processed (e.g., Pod, ConfigMap, Secret, Service, ServiceAccount, EndpointSlice, Ingress and PersistentVolumeClaim).
+- **liqo_virtual_kubelet_reflection_error_counter**: the number of transient errors during the reflection phase (e.g., due to  temporary race conditions that can be resolved by retrying the synchronization, these conditions mainly occur when some of the requested resources are not yet fully configured).
```
frisso, line 36:
```suggestion
- **liqo_virtual_kubelet_reflection_error_counter**: the number of transient errors during the reflection phase (e.g., due to temporary race conditions that can be resolved by retrying the synchronization; these conditions mainly occur when some of the requested resources are not yet fully configured).
```

docs/usage/prometheus-metrics.md
```diff
@@ -27,3 +27,16 @@ We provide a {download}`sample Grafana dashboard </_downloads/grafana/liqonetwor
 As presented in the screenshot below, it includes an overview section presenting the overall cross-cluster throughput, followed by detailed per-peering throughput and latency information.
 
 ![Grafana Network Dashboard](/_static/images/usage/prometheus-metrics/network-dashboard.png)
+
+## Virtual kubelet metrics
+
+These metrics are available for each peered remote cluster, providing statistics about the reflected resources:
+
+- **liqo_virtual_kubelet_reflection_item_counter**: the number of reflected resources successfully processed (e.g., Pod, ConfigMap, Secret, Service, ServiceAccount, EndpointSlice, Ingress and PersistentVolumeClaim).
```
frisso, line 35:
```suggestion
- **liqo_virtual_kubelet_reflection_item_counter**: the number of resources that are currently successfully reflected (e.g., Pod, ConfigMap, Secret, Service, ServiceAccount, EndpointSlice, Ingress and PersistentVolumeClaim). This number can increase/decrease over time, and it may reach zero when two peered clusters have no reflected resources.
```

docs/usage/prometheus-metrics.md
```diff
@@ -27,3 +27,16 @@ We provide a {download}`sample Grafana dashboard </_downloads/grafana/liqonetwor
...
+- **liqo_virtual_kubelet_reflection_error_counter**: the number of transient errors during the reflection phase (e.g., due to  temporary race conditions that can be resolved by retrying the synchronization, these conditions mainly occur when some of the requested resources are not yet fully configured).
+
+### Grafana dashboard
+
+We offer a {download}`sample Grafana dashboard </_downloads/grafana/liqovirtualkubelet.json>` that allows you to monitor the reflected resources for every componenet of the virtual-kubelet. As shown in the screenshot below, it contains visual representations of the total number of reflected resources and the average rate per second. Additionally, there are detailed tables that provide information on the total number of resources of each type, as well as an overall summary of all reflected items during a certain time period.
```
frisso, line 40:
```suggestion
We offer a {download}`sample Grafana dashboard </_downloads/grafana/liqovirtualkubelet.json>` that allows you to monitor the reflected resources for each component of the virtual-kubelet.
As shown in the screenshot below, it contains visual representations of the total number of reflected resources and the average rate per second.
Additionally, there are detailed tables that provide information on the total number of each type of resource, as well as an overall summary of all reflected items during a certain time period.
```

docs/usage/prometheus-metrics.md
frisso: @aleoli @damianot98 Reporting an error is something that *should* be interesting for the user. For me, as a customer, knowing that are errors and (a) knowing nothing about why they are here, (b) having no way to solve them, means that this counter is completely useless. Even worse, it makes me crazy, as I know that something is not working, and I cannot do anything to solve the problem.
Please consider this and improve the documentation according to the above points.

frisso:
Some more comments about the graph.
1) "Reflected item rate" and "Transient error rate". The "rate", in events/second, looks very hard to understand. Much better to show the actual number of item/errors. Given that Grafana has a given refresh time interval (e.g., 10/30 seconds, if I remember correctly), the user should know that the number shown refers to the number of events in this interval, which is much simpler than now. So, I would suggest to change how the graph is made, and the title (just avoid "rate").
2) "Reflected items for resource type" and "Transient errors for resource type" looks confusing to me. Particularly errors, I assume that they should stay zero for most of the time. Instead, those counters appear to be always growing, such as the graphs are *cumulative* instead of *instantaneous*. I really prefer the second (instantaneous); please make sure that this is the exact behaviour.

frisso (changes requested):
It looks good to me.
Just address the latest comments, and (for me) we're done.

damianot98:
@frisso about the previous comment:

1. I combined the two graphs into a single one that shows the current trend of both items and errors. As shown in the figure. 
Screenshot 2023-05-14 at 19 16 06
2. The graphs "Reflected items for resource type" and "Transient errors for resource type" convey the same information as the tables "Reflected Resources" and "Resource errors," respectively, but in a graphical way. The upper graphs display instantaneous values, I can modify them to behave similarly to the upper ones. Something like this: 
Screenshot 2023-05-14 at 19 36 34

docs/usage/prometheus-metrics.md
damianot98: @frisso I extended the previous example, providing the three main reasons why errors occur in the reflection stage. Although it is highly specific, it would be more helpful from a user point of view.

damianot98:
@frisso I will proceed to push the changes to the dashboard in my next commit provided that the graphs are good.

frisso:
@aleoli LGTM. Can you please just make sure that the definition of "**liqo_virtual_kubelet_reflection_error_counter**" is ok, and possibly adding something "Usually the above errors are not significant for the average user, which can ignore this counter, and are shown to facilitate possible debugging", or, if something can be done, tell the user what can be done here.

frisso:
@damianot98 LGTM. Just rename "Item and error trend" in the dashboard with "Reflected resources and transient errors", and "Total items reflected" with "Total resources reflected".
Finally, I assume the last graph in the picture (the one about errors) is wrong, as it should be "zero" most of the time.

frisso (approved):

cheina97:
@damianot98 just some final comments and then your PR is ready to be merged:

1. The "trick" to set the value of metrics to 0 does not work always (it depends on the prometheus-server requests interval), so could be a good idea to avoid it.  
2. I think it is difficult for someone who does not what your dashboard is showing to understand the meaning of "item". For example "Reflected items for resource type" could become "Reflected resources rate".
3. You never say that your dashboard is showing a "rate", please add it inside the titles.

frisso:
> @damianot98 just some final comments and then your PR is ready to be merged:
> 
> 1. The "trick" to set the value of metrics to 0 does not work always (it depends on the prometheus-server requests interval), so could be a good idea to avoid it.
> 2. I think it is difficult for someone who does not what your dashboard is showing to understand the meaning of "item". For example "Reflected items for resource type" could become "Reflected resources".

What about "Number of reflected resources"?

PR #1812, opened 2023-05-10
Helm: ingress class name

cheina97:
# Description

This PR inserts ingressClassName into the auth ingress resource created by helm. The old annotation has been retained for backward compatibility.


frisso:
@cheina97 ok about retaining compatibility, but we're in a phase in which we still can make breaking changes. Please consider setting a deadline, e.g. 2 versions from now, in which the backward compatibility is removed.

cheina97:
@frisso I talked about it with @aleoli and we agree with you. I removed the ingress annotation from the helm file (it's still possible to add it through the auth.ingress.annotations value)

aleoli (approved):

Issue #1705, opened 2023-03-08
[Feature] Resliency to clusters failures


Fra98:
The remote cluster is not resilient to failures of the local cluster (i.e., the one doing the offloading). Requests to pods running on the local cluster will fail if the local cluster is unhealthy, resulting in service disruption.

Resiliency could be achieved by invalidating endpoints of the local pods if at least one of the following scenarios occurs:
* Local cluster not reachable through the VPN tunnel (TunnelEndpoint)
* Local cluster's API server is unhealthy (unreachable or not ready)

In this way, services do not redirect the traffic towards local pods, ensuring service continuity.

### Required steps
* [x] 1) Implement a controller that, in the remote cluster,sets the status of the local pods endpoints to Ready/NotReady depending on whether the local cluster is reachable via the TunnelEndpoint #1677
* [x] 2) Add a new status condition in the ForeignCluster CR that tracks the readiness status of the foreign API server #1703
* [x] 3) Update the new controller to also check for failures of the local API server using the new status condition added in step 2 #1774 
* [x] 4) Change the API server check of the virtual kubelet: it should simply monitor the new status condition added in step 2 #1820

PR #1820, opened 2023-05-17
Virtual node checks the status of the API server via the ForeignCluster

Fra98:
# Description

This PR changes the virtual node to check the status of the API server by looking at the *APIServerStatusCondition* of the ForeignCluster.

Ref #1705 

# How Has This Been Tested?
- [x] Locally on Kind
- [x] Unit tests
- [x] e2e tests

### DIFF
--- a/pkg/virtualKubelet/liqoNodeProvider/nodeProvider.go
+++ b/pkg/virtualKubelet/liqoNodeProvider/nodeProvider.go
@@ -16,6 +16,7 @@ package liqonodeprovider
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
@@ -24,6 +25,8 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
+
+	foreignclusterutils "github.com/liqotech/liqo/pkg/utils/foreignCluster"
 )
 
 // LiqoNodeProvider is a node provider that manages the Liqo resources.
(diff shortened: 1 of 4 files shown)


pkg/virtualKubelet/liqoNodeProvider/nodeProvider.go
```diff
@@ -58,11 +69,27 @@ func (p *LiqoNodeProvider) Ping(ctx context.Context) error {
...
+		return err
+	}
+	var foreignClusters discoveryv1alpha1.ForeignClusterList
+	utilruntime.Must(runtime.DefaultUnstructuredConverter.FromUnstructured(unstr.UnstructuredContent(), &foreignClusters))
+	if len(foreignClusters.Items) == 0 {
+		err := errors.Errorf("Failed to get foreigncluster %q", p.foreignClusterID)
+		klog.Error(err)
 		return err
 	}
+	fc := &foreignClusters.Items[0] // there should be only one foreigncluster with the given clusterID
```
aleoli, line 87: You should create a second implementation of this function https://github.com/liqotech/liqo/blob/master/pkg/utils/foreignCluster/getForeignCluster.go#L39 and consolidate the common part in a private one, in order to have more maintanable code and common error messages

pkg/virtualKubelet/liqoNodeProvider/nodeProvider.go
```diff
@@ -58,11 +69,27 @@ func (p *LiqoNodeProvider) Ping(ctx context.Context) error {
...
+
+	// Check the foreign API server status
+	if !foreignclusterutils.IsAPIServerReady(fc) {
+		return fmt.Errorf("API server readiness check failed")
```
aleoli, line 91: Can you write which cluster is causing the problem (by adding the clusterID and/or the clusterName in the log)?

PR #1825, opened 2023-05-18
Liqoctl: flag to set service type

cheina97:
# Description

This PR adds the `--service-type` flag to **liqoctl**. It forces **Liqo** to use the specified **service type** instead of **LoadBalancer** for **liqo-gateway** and **liqo-auth**.

This flag does not override the provider's settings, this behavior will be implemented in the next PR.

# How Has This Been Tested?

- [x] Locally, generating helm values files using the `--only-output-values` flag

### DIFF
--- a/cmd/liqoctl/cmd/install.go
+++ b/cmd/liqoctl/cmd/install.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/liqotech/liqo/pkg/liqoctl/completion"
 	"github.com/liqotech/liqo/pkg/liqoctl/factory"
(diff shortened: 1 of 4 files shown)


frisso:
@cheina97 Just 2 cents: should we consider instead a flag like "--set-external-service" with values "loadbalancer | nodeport", in order to enable explicit setting for all the service options we have?

cheina97:
> @cheina97 Just 2 cents: should we consider instead a flag like "--set-external-service" with values "loadbalancer | nodeport", in order to enable explicit setting for all the service options we have?

I agree, maybe it is more compact if we use `--service-type`

frisso (approved):

aleoli (approved):

PR #1829, opened 2023-05-20
Improve error message

frisso:
This PR enriches the error message in case of failed validation during the installation with **liqoctl**.

### DIFF
--- a/pkg/liqoctl/install/validation.go
+++ b/pkg/liqoctl/install/validation.go
@@ -71,7 +71,8 @@ func (o *Options) validateClusterName() (err error) {
 
 	errs := validation.IsDNS1123Label(o.ClusterName)
 	if len(errs) != 0 {
-		return fmt.Errorf("the cluster name may only contain lowercase letters, numbers and hyphens, and must not be no longer than 63 characters. Try using 'liqoctl --cluster-name' to overcome this error")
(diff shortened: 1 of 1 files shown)


pkg/liqoctl/install/validation.go
```diff
@@ -71,7 +71,7 @@ func (o *Options) validateClusterName() (err error) {
 
 	errs := validation.IsDNS1123Label(o.ClusterName)
 	if len(errs) != 0 {
-		return fmt.Errorf("the cluster name may only contain lowercase letters, numbers and hyphens, and must not be no longer than 63 characters")
+		return fmt.Errorf("the cluster name may only contain lowercase letters, numbers and hyphens, and must not be no longer than 63 characters. Try using 'liqoctl --cluster-name' to overcome this error")
```
shrinandj, line 74: One small suggestion, can we also add some text suggesting that the name should be `DNS compatible`? 

Otherwise this looks great. Thanks for fixing this.

frisso, line 74:
```suggestion
		return fmt.Errorf(`the cluster name must be DNS-compatible (e.g., lowercase letters, numbers and hyphens) and
		must be <= 63 characters. Try using 'liqoctl --cluster-name' to set another name and overcome this error`)
```

cheina97 (approved):

PR #1833, opened 2023-05-22
Fix deployment on ArgoCD

QcFe:
# Description

This PR fixes deployment on ArgoCD that ended up being stuck because of Argo wanting to delete hook jobs that had ttl set: the jobs controller deleted the jobs instead of Argo and this stopped the deployment from progressing.

Added a side note on telemetry cronjob schedule helm value.

# How Has This Been Tested?
- [x] On ArgoCD-driven cluster
- [ ] On liqoctl provisioned cluster

### DIFF
--- a/deployments/liqo/templates/webhooks/job-patch/job-create-secret.yaml
+++ b/deployments/liqo/templates/webhooks/job-patch/job-create-secret.yaml
@@ -13,7 +13,7 @@ metadata:
   labels:
     {{- include "liqo.labels" $cfg | nindent 4 }}
 spec:
-  ttlSecondsAfterFinished: 0
+  ttlSecondsAfterFinished: 150
   template:
     metadata:
       name:  {{ include "liqo.prefixedName" $cfg }}
--- a/deployments/liqo/values.yaml
+++ b/deployments/liqo/values.yaml
@@ -292,7 +292,7 @@ telemetry:
   # -- Image repository for the telemetry pod.
   imageName: "ghcr.io/liqotech/telemetry"
   config:
-    # -- Set the schedule of the telemetry collector CronJob
+    # -- Set the schedule of the telemetry collector CronJob. Consider setting this value on ArgoCD deployments to avoid randomization.
     schedule: ""
     # schedule: "0 */12 * * *"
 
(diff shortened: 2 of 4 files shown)


deployments/liqo/templates/webhooks/job-patch/job-create-secret.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
   labels:
     {{- include "liqo.labels" $cfg | nindent 4 }}
 spec:
-  ttlSecondsAfterFinished: 0
+  ttlSecondsAfterFinished: 600
```
cheina97, line 16: Maybe would be better a lower value, like 150 seconds. Same for the next one.

cheina97 (approved):

deployments/liqo/templates/webhooks/job-patch/job-create-secret.yaml
frisso: @aleoli What's your opinion here?

aleoli: 150 will be fine since it seems to be the value generally used

deployments/liqo/values.yaml
```diff
@@ -292,7 +292,7 @@ telemetry:
   # -- Image repository for the telemetry pod.
   imageName: "ghcr.io/liqotech/telemetry"
   config:
-    # -- Set the schedule of the telemetry collector CronJob
+    # -- Set the schedule of the telemetry collector CronJob. Consider setting this value on ArgoCD deployments to avoid randomization.
```
aleoli, line 295: It is not random, but 10 minutes after you create the release. Since ArgoCD is continuously redoing helm templates, it is forever 10 minutes after now (if not set)

QcFe: Thought it was random since mine is set at 22 and I don't think I deployed it around that time 😬
Should I remove this comment or do you have any suggestion to improve it?

QcFe: made no further changes, please resolve if no action is required

PR #1853, opened 2023-06-05
Mark local offloaded pods as NotReady when virtual-node is not ready (i.e. remote cluster failure)  

fra98:
# Description

This PR changes the virtual-kubelet logic to mark local offloaded pods as *NotReady* when the remote cluster has a failure (i.e., the liqo virtual node has the *Ready* condition set to *False* or *Unknown*). 
In summary, it sets the *Ready* pod condition to *False* on all pods scheduled on a *NotReady* virtual node.

### Implementation
- A new controller in the *liqo-controller-manager* enforces the presence/absence of a custom label on
local offloaded pods. The presence of the label indicates that the status of the offloaded pod can be managed/updated by the local cluster and can't currently be reflected/updated from the remote cluster (likely due to a remote cluster failure).
- The virtual-kubelet makes sure that the *Ready* conditions of the local offloaded pods are set to *False* if the label is present.

### Motivation
We mark pods as NotReady because the Kubernetes node controller skips this update (likely due to a bug). This is necessary to prevent services redirecting traffic to not ready pods (since the endpointslice controller keeps in sync the endpointslice *Ready* condition with the associated pod *Ready* condition).

# How Has This Been Tested?

- [x] Locally
- [x] Unit tests
- [x] e2e tests

### DIFF
--- a/pkg/liqo-controller-manager/podstatus-controller/podstatus_controller_test.go
+++ b/pkg/liqo-controller-manager/podstatus-controller/podstatus_controller_test.go
@@ -1,0 +1,229 @@
+// Copyright 2019-2023 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package podstatusctrl
+
+import (
+	"bytes"
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	"k8s.io/kubectl/pkg/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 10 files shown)


pkg/liqo-controller-manager/podstatus-controller/podstatus_controller_test.go
```diff
@@ -0,0 +1,215 @@
...
+			pod.Labels = map[string]string{
+				consts.LocalPodLabelKey: consts.LocalPodLabelValue,
+			}
+			pod.Spec.NodeName = liqoNodeName
+			return pod
+		}
+	)
+
+	BeforeEach(func() {
+		ctx = context.TODO()
```
aleoli, line 124:
```suggestion
		ctx = context.Background()
```

pkg/liqo-controller-manager/podstatus-controller/podstatus_controller_test.go
```diff
@@ -0,0 +1,215 @@
...
+				Expect(localPod.Labels).ToNot(HaveKeyWithValue(consts.PodStatusLocalKey, consts.PodStatusLocalValue))
+			})
+		})
+
+	})
+
+	AfterEach(func() {
+		pod := corev1.Pod{}
+		Expect(fakeClient.Get(ctx, reqPod.NamespacedName, &pod)).To(Succeed())
+		Expect(pod.Labels).ToNot(HaveKeyWithValue(consts.PodStatusLocalKey, consts.PodStatusLocalValue))
```
aleoli, line 212: You are never using this client to manage pods, it is probably better to add a second pod (not selected bt the controller) and do this check on it

pkg/utils/pod/pod.go
```diff
@@ -34,6 +34,20 @@ func IsPodReady(pod *corev1.Pod) (ready bool, reason string) {
 	return false, "no conditions in pod status"
 }
 
+// GetPodCondition extracts the provided condition from the given status and returns that.
+// Returns nil and -1 if the condition is not present, and the index of the located condition.
+func GetPodCondition(status *corev1.PodStatus, conditionType corev1.PodConditionType) (int, *corev1.PodCondition) {
```
aleoli, line 39:
```suggestion
func GetPodCondition(status *corev1.PodStatus, conditionType corev1.PodConditionType) (*corev1.PodCondition, int) {
```

Do we need the index? If not we should remove it, it makes easier to use the function in a single line statement

PR #1868, opened 2023-06-16
Update documentation flag for moving a PVC between clusters.

tom-asmblr:
# Description

Seems to be --target-node, --node is an unknown flag.


fra98 (approved):

fra98:
@tom-asmblr thanks for your contribution

PR #1891, opened 2023-07-03
Liqoctl: added confirm prompt to liqoctl unpeer

cheina97:
# Description

This PR adds the **"confirm"** prompt to `liqoctl unpeer`

# How Has This Been Tested?

- [x] Locally on KinD


frisso:
@cheina97 This looks really great!
Just a minor comment: would it be possible to write a small text explaining why this is a dangerous action? E.g., All the jobs on the remote cluster will be torn down, or such?
Thanks!

aleoli:
> @cheina97 This looks really great! Just a minor comment: would it be possible to write a small text explaining why this is a dangerous action? E.g., All the jobs on the remote cluster will be torn down, or such? Thanks!

sorry, I didn't see the comment... anyway it seems to be a larger discussion also for the other confirm messages. It can be addressed in another pr

Issue #1860, opened 2023-06-13
[Feature] Disable the reflection of specific labels and annotations


fra98:
### Describe the solution you'd like
We should provide a way to optionally disable the reflection of custom labels and annotations (from local cluster to remote and vice-versa). 

### Why this may be useful
This can be useful, for example, to avoid reflecting labels that are tied to the configuration of a particular cluster (e.g.: avoid reflecting a cloud-provided annotation indicating a LoadBalancer IP pool, which leads to inconsistency and/or conflicts between the reflected resource)

### Implementation steps
* [x] During installation, specify (via Helm values) a list of labels/annotations that Liqo should blacklist when reflecting a resource. #1907 
* [x] Add to the default list some well-known labels/annotations that are known to cause problems when reflected. #1948

PR #1907, opened 2023-07-13
Disable the reflection of specific labels and annotations

fra98:
# Description

This PR introduces the possibility of disabling the reflection of custom labels and annotations (from local cluster to remote and vice-versa). The list of labels and annotations that must not be reflected can be specified with the Helm values (respectively `reflection.skip.labels` and `reflection.skip.annotations`).

Ref #1860 

# How Has This Been Tested?

- [x] locally
- [x] e2e tests
- [x] unit tests

### DIFF
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -106,6 +106,8 @@ func main() {
 	var kubeletRAMRequests, kubeletRAMLimits argsutils.Quantity
 	var kubeletMetricsAddress string
 	var kubeletMetricsEnabled bool
+	var labelsNotReflected argsutils.StringList
+	var annotationsNotReflected argsutils.StringList
 
 	webhookPort := flag.Uint("webhook-port", 9443, "The port the webhook server binds to")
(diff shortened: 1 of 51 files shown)


cmd/liqo-controller-manager/main.go
```diff
@@ -166,6 +168,8 @@ func main() {
 	flag.BoolVar(&kubeletMetricsEnabled, "kubelet-metrics-enabled", false, "Enable the kubelet metrics endpoint")
 	flag.Var(&nodeExtraAnnotations, "node-extra-annotations", "Extra annotations to add to the Virtual Node")
 	flag.Var(&nodeExtraLabels, "node-extra-labels", "Extra labels to add to the Virtual Node")
+	flag.Var(&labelsNotReflected, "labels-not-reflected", "List of labels (key) that must not be reflected")
```
aleoli, line 171: What do you think about calling them something like `skip-label-reflection` or `ignore-label-reflection`?

aleoli (approved):

PR #1936, opened 2023-08-02
Liqoctl: install warnings

cheina97:
# Description

This PR introduces the possibility of adding **Warners** objects in **liqoctl install**. Warners check some conditions and print a warning message if necessary.

# How Has This Been Tested?

- [x] Locally with KinD


pkg/liqoctl/install/warning.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+	}
+	return len(sw.warnings) == 0, nil
+}
+
+func (sw *serviceWarner) warn() []string {
+	return sw.warnings
+}
+
+// ValuesWarning checks the values map and returns a list of warningsw.
```
aleoli, line 56:
```suggestion
// ValuesWarning checks the values map and returns a list of warnings.
```

pkg/liqoctl/output/output.go
```diff
@@ -98,6 +98,11 @@ type Printer struct {
 	verbose    bool
 }
 
+func (p *Printer) SpinnerPrintWhileRunning(spinner *pterm.SpinnerPrinter, f func(message ...interface{}), message ...interface{}) *pterm.SpinnerPrinter {
+	f(message...)
+	return p.StartSpinner(spinner.Text)
+}
```
aleoli, line 104:
```suggestion
func (p *Printer) SpinnerPrintWhileRunning(spinner *pterm.SpinnerPrinter, f func()) *pterm.SpinnerPrinter {
	f()
	return p.StartSpinner(spinner.Text)
}
```

and then in the caller

```
s = o.Printer.SpinnerPrintWhileRunning(s, func() {s.Warning(warning)})
```

all these interfaces and the parameter not used but forwarded to the function seams strange and dangerous

pkg/liqoctl/install/warning.go
```diff
@@ -0,0 +1,70 @@
...
+type warner interface {
+	name() string
+	check(values map[string]interface{}) (bool, error)
+	warn() []string
+}
+
+type serviceWarner struct {
+	warnings []string
+}
+
```
aleoli, line 21:
```suggestion
var _ warner = serviceWarner{}
```

if the interface is required, we should check that the object implements it

aleoli (approved):

pkg/liqoctl/output/output.go
cheina97: Otherwise, we can create a dedicated function for  Warning and Success, which take only the spinner and the  message as arguments.

PR #1944, opened 2023-08-07
liqoctl: add create VirtualNode command

aleoli:
# Description

This pr adds a new command to `liqoctl` to manage (create and delete) `VirtualNodes`

# How Has This Been Tested?

- [x] yaml generation, using the `-o yaml` flag
- [x] directly creating a second vn for an established peering

### DIFF
--- a/cmd/liqoctl/cmd/root.go
+++ b/cmd/liqoctl/cmd/root.go
@@ -30,10 +30,18 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/cmd/util"
 
+	"github.com/liqotech/liqo/pkg/liqoctl/create"
+	"github.com/liqotech/liqo/pkg/liqoctl/delete"
 	"github.com/liqotech/liqo/pkg/liqoctl/factory"
+	"github.com/liqotech/liqo/pkg/liqoctl/rest"
+	"github.com/liqotech/liqo/pkg/liqoctl/rest/virtualnode"
 )
 
 var liqoctl string
+
+var liqoResources = []rest.APIProvider{
+	virtualnode.VirtualNode,
+}
 
 func init() {
 	liqoctl = os.Args[0]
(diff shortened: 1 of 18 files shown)


pkg/liqoctl/virtualnode/create.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return cmd
+}
+
+func (o *Options) handleCreate(ctx context.Context) error {
+	opts := o.createOptions
+	if opts.OutputFormat != "" {
+		opts.Printer.CheckErr(o.output())
+		return nil
+	}
```
cheina97, line 108: Sorry, I don't understand what's the purpose of this piece of code. Why you are printing the resource in the create command?

pkg/liqoctl/virtualnode/create.go
```diff
@@ -0,0 +1,214 @@
...
+		// TODO: we cannot use the cluster identity here
+		if err := waiter.ForNode(ctx, virtualNode.Spec.ClusterIdentity); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (o *Options) getVirtualNode() *virtualkubeletv1alpha1.VirtualNode {
```
cheina97, line 133:
```suggestion
func (o *Options) forgeVirtualNode() *virtualkubeletv1alpha1.VirtualNode {
```
Maybe it's more understandable? It seems it gets a VirtualNode which already exists from the name.

pkg/liqoctl/virtualnode/create.go
```diff
@@ -0,0 +1,214 @@
...
+			options.Name = args[0]
+			o.createOptions = options
+		},
+
+		Run: func(cmd *cobra.Command, args []string) {
+			output.ExitOnErr(o.handleCreate(ctx))
+		},
+	}
+
+	cmd.Flags().VarP(outputFormat, "output", "o",
```
cheina97, line 72: Could be a good idea to add the completion?

cheina97:
Last comment about files hierarchy. Could be a good idea moving `virtualnode` folder inside `rest`? I think it would be more tidy.

cheina97 (approved):

PR #1949, opened 2023-08-14
Liqonet: wireguard userspace golang

cheina97:
# Description

This PR replaces [boringtun](https://github.com/cloudflare/boringtun) with [wireguard-go]()
This allows to replace the Ubuntu docker image with Alpine, which was previously replaced (https://github.com/liqotech/liqo/pull/1852)

It also improves the wireguard userspace performance:

### boringtun
image

### wireguard-go
image

# How Has This Been Tested?

- [x] On KinD  arch:x86_64kernel:5.4.0-155-generic
- [x] On AWS EC2 4.14.256-197.484.amzn2.aarch64
- [x] On AWS EKS with AmazonLinux2 ARM

### DIFF
--- a/pkg/liqonet/tunnel/wireguard/driver.go
+++ b/pkg/liqonet/tunnel/wireguard/driver.go
@@ -35,6 +35,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	k8s "k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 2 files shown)


pkg/liqonet/tunnel/wireguard/driver.go
```diff
@@ -351,8 +355,22 @@ func (w *Wireguard) Close() error {
...
 
+// runWgUserCmd runs the wg command with the given arguments.
+func runWgUserCmd(cmd *exec.Cmd) {
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		if err != nil {
+			outStr, errStr := stdout.String(), stderr.String()
+			fmt.Printf("out:\n%s\nerr:\n%s\n", outStr, errStr)
```
aleoli, line 366: Can we use klog here?

aleoli (approved):

pkg/liqonet/tunnel/wireguard/driver.go
cheina97: Since that output can have multiple lines is better to use `fmt`

PR #1969, opened 2023-08-23
AKS vnet resource group

scottboring:
Adding optional flag for vnet resource group for scenarios when the vnet is in a different resource group from the AKS resource

# Description

The virtual network is not always in the same resource group as the AKS resource and subsequently this needs to be exposed as a param to the CLI

Fixes #(issue)

# How Has This Been Tested?

- [x] Executed CLI using param and verified param was used
- [x] Executed CLI without param and verified param defaulted to resource-group-name


aleoli (approved):
Hi @scottboring! Thanks for your contribution!

PR #1970, opened 2023-08-24
Added possibility to switch to iptables nf_tables

cheina97:
# Description

This PR adds the possibility of forcing the usage of iptables legacy/nf_tables using the env variable IPTABLES_MODE which accept 2 values:
- legacy
- nf_tables

If not set the used iptables mode will be the system default (legacy in case of alpine)

This PR uses a fork of the official go-iptables package while waiting for the PR
https://github.com/coreos/go-iptables/pull/110 to be merged.

# How Has This Been Tested?

- [x] Locally on KinD


deployments/liqo/values.yaml
aleoli, line 1: Can we have a unique place to set it instead of three different? If you need a mode in one component, you also need it in the others.

aleoli (approved):

deployments/liqo/values.yaml
cheina97: Yes, it makes sense

PR #1974, opened 2023-08-28
Helm values files for CLI

scottboring:
# Description

Adding optional flag for values file similar to helm values file making CLI easier to use when multiple values need to be overridden. The new flag is --values, the same as helm and functions the same way in that is supports multiple files including files from URLs.

# How Has This Been Tested?

Tested multiple scenarios locally

- [x] Install help message
- [x] Install without values param
- [x] Install with 1 values param
- [x] Install with 2 values params 
- [x] Install with --set overriding --values params
- [x] Install fails with invalid values file
- [x] Install fails values file not found

### DIFF
--- a/docs/installation/install.md
+++ b/docs/installation/install.md
@@ -413,12 +413,12 @@ To change this behavior, check the [network flags](NetworkFlags).
 ## Customization options
 
 This section lists the main **customization parameters** supported by the *liqoctl install* command, along with a brief description.
-Additionally, **all parameters** available in the Helm *values* file (the full list is provided in the dedicated [repository page](https://github.com/liqotech/liqo/tree/master/deployments/liqo)) can be modified through the `liqoctl install --set` flag, which supports the standard Helm syntax.
+Additionally, **all parameters** available in the Helm *values* file (the full list is provided in the dedicated [repository page](https://github.com/liqotech/liqo/tree/master/deployments/liqo)) can be modified through the `liqoctl install --values` and or `liqoctl install --set` flags, which supports the standard Helm syntax.
 
 Finally, remember that:
 
 * You can type `liqoctl install --help` to get the list of available options.
-* Some of the above parameters can be changed after installation by simply updating their value and re-applying the Helm chart, or by re-issuing the proper `liqoctl install --set [param=value]` command. However, given that not all parameters can be updated at run-time, please check that the command triggered the desired effect; a precise list of commands that can be changed at run-time is left for our future work.
+* Some of the above parameters can be changed after installation by simply updating their value and re-applying the Helm chart, or by re-issuing the proper `liqoctl install --values [file] --set [param=value]` command. However, given that not all parameters can be updated at run-time, please check that the command triggered the desired effect; a precise list of commands that can be changed at run-time is left for our future work.
 
 ### Global
 
--- a/pkg/liqoctl/install/handler.go
+++ b/pkg/liqoctl/install/handler.go
@@ -18,7 +18,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -28,6 +31,8 @@ import (
 	"golang.org/x/mod/semver"
 	"gopkg.in/yaml.v3"
 	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/getter"
 	"helm.sh/helm/v3/pkg/repo"
 	"helm.sh/helm/v3/pkg/strvals"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


cheina97:
Hi @scottboring thanks for your PR, that's something we planned to implement and I'm sure it will be really useful in the future. So thanks for your commitment to our project and your time, we really appreciate it.

Just a few comments:
1. I think it would be more correct to implement the behaviour of the `--values` flag here https://github.com/scottboring/liqo/blob/b5c459137f0c43e74048fb83d7b6a1ea416066d8/pkg/liqoctl/install/handler.go#L279.
2. Suggestion: get inspiration from here in order to emulate the helm `--values` flag behaviour
3. I would expect the `--values` flag does not have priority on the rest of the flags, especially `--set` (like in **helm**).
4. PLease adds an explanation of the `--values` flag in the documentation.
5. Please put the PR in **draft** till it is ready (it includes linting check).

scottboring:
Hi @cheina97, thank you for the review and comments.

1. I don't follow this, how would this logic be implemented here and why wouldn't it be next to the logic for the --set flag?
2. I can make some changes to use or copy that code
3. The --values flag overrides is ran before the --set flag so this priority is aligns with your expectations. I have also tested this.
4. I'lll update the documentation 
5. This has been converted to draft

If you could please provide some more detail around the points 1 & 2, that would be helpful to ensure the changes I make align with your expectations.

cheina97:
1. Because it's where the chart values are loaded. You can avoid that line if the --values parameter is provided and load the one specified.
2. I think it's a good idea copy that code to be sure to behave like expected in helm.
3. I wasn't clear in my explanation. If you looks the code on top of yours other values which are configured using liqoctl flags or provider's values. These values do not have to be overrided by --values.

scottboring:
Hi @cheina97, I've made the requested updates. Thank you for reviewing my PR and providing feedback. Let me know if there is any other feedback.

Note that I used the Helm cli env settings here which I'm not sure is correct but aligns with how Helm works: https://github.com/liqotech/liqo/pull/1974/files#diff-a849a3ee27a1eaa15291b34f78d4adee1ef1e4e268d48eb3dd1338acb2f2c685R422

docs/installation/install.md
```diff
@@ -413,12 +413,12 @@ To change this behavior, check the [network flags](NetworkFlags).
 ## Customization options
 
 This section lists the main **customization parameters** supported by the *liqoctl install* command, along with a brief description.
-Additionally, **all parameters** available in the Helm *values* file (the full list is provided in the dedicated [repository page](https://github.com/liqotech/liqo/tree/master/deployments/liqo)) can be modified through the `liqoctl install --set` flag, which supports the standard Helm syntax.
+Additionally, **all parameters** available in the Helm *values* file (the full list is provided in the dedicated [repository page](https://github.com/liqotech/liqo/tree/master/deployments/liqo)) can be modified through the `liqocotl install --values` and or `liqoctl install --set` flags, which supports the standard Helm syntax.
```
cheina97, line 416:
```suggestion
Additionally, **all parameters** available in the Helm *values* file (the full list is provided in the dedicated [repository page](https://github.com/liqotech/liqo/tree/master/deployments/liqo)) can be modified through the `liqoctl install --values` and or `liqoctl install --set` flags, which supports the standard Helm syntax.
```

pkg/liqoctl/install/handler.go
```diff
@@ -398,6 +414,50 @@ func (o *Options) postProviderValues() map[string]interface{} {
 	return values
 }
 
+func (o *Options) valuesFiles() (map[string]interface{}, error) {
+	values := map[string]interface{}{}
+	// Near copy from Helm https://github.com/helm/helm/blob/eb4edc96c581c1978fe935197e20ced4071af5d5/pkg/cli/values/options.go#L48
+	for _, filePath := range o.OverrideValuesFiles {
+		currentMap := map[string]interface{}{}
+		bytes, err := readFile(filePath, getter.All(cli.New()))
```
cheina97, line 422:
```suggestion
		bytes, err := os.ReadFile(filePath)
```
I think it could be enough, @aleoli and @fra98 what do you think about it?

scottboring: Hi @cheina97 , the os.ReadFile only reads files from the local file system where Helm support reading files from URLs. I don't have a need to read files from URLs but the intent was to mimic Helm's functionality.

docs/installation/install.md
scottboring: @cheina97 thank you for catching that

pkg/liqoctl/install/handler.go
cheina97: It's ok, keep it as it is

cheina97:
Please, launch `make generate` to fix the majority of lining errors.

scottboring:
Hi @cheina97, I believe this PR is ready to be merged.

cheina97:
Please, can you squash the commits?

scottboring:
@cheina97 commits have been squashed

cheina97 (approved):
lgtm

PR #1975, opened 2023-09-01
Liqoctl: check --dump-values-path install flag

fra98:
# Description

Liqoctl: added a check that raises an error if `--dump-values-path` is set, while `--only-output-values` is not.


cmd/liqoctl/cmd/install.go
```diff
@@ -138,8 +138,8 @@ func newInstallCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
 
 	cmd.PersistentFlags().BoolVar(&options.OnlyOutputValues, "only-output-values", false,
 		"Generate the pre-configured values file for further customization, instead of installing Liqo (default false)")
-	cmd.PersistentFlags().StringVar(&options.ValuesPath, "dump-values-path", "./values.yaml",
```
aleoli, line 141: I would add a comment to specify that the default value is set after the check if the flag has been set or not

aleoli (approved):

PR #1985, opened 2023-09-06
AKS Private Link

scottboring:
# Description

Adding privateLink flag to liqoctl to better support private AKS clusters. AKS clusters that use private link have a generated fqdn stored on the AKS cluster object as privateFQDN in addition to a public FQDN which is likely blocked through Azure policies.

Fixes #(issue)

This has been tested locally to verify the correct fqdn property is used based on flag presence or absence.

- [x] Tested without using the flag --privateLink
- [x] Tested with using the flag --privateLink

### DIFF
--- a/docs/installation/install.md
+++ b/docs/installation/install.md
@@ -120,6 +120,8 @@ Alternatively, you can manually set a different name with the `--cluster-name` *
 ```{admonition} Note
 If you are running an [AKS private cluster](https://learn.microsoft.com/en-us/azure/aks/private-clusters), you may need to set the `--disable-api-server-sanity-check` *liqoctl* flag, since the API Server in your kubeconfig may be different from the one retrieved from the Azure APIs.
 
+If the private cluster uses private link, you can set the `--private-link` *liqoctl* flag to use the private FQDN for the API server.
+
 Additionally, since your API Server is not accessible from the public Internet, you shall leverage the [in-band peering approach](FeaturesPeeringInBandControlPlane) towards the clusters not attached to the same Azure Virtual Network.
 ```
 
--- a/pkg/liqoctl/install/aks/provider.go
+++ b/pkg/liqoctl/install/aks/provider.go
@@ -42,6 +42,7 @@ type Options struct {
 	resourceGroupName     string
 	resourceName          string
 	vnetResourceGroupName string
+	privateLink           bool
 	fqdn                  string
 
 	authorizer *autorest.Authorizer
@@ -76,6 +77,7 @@ func (o *Options) RegisterFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.vnetResourceGroupName, "vnet-resource-group-name", "",
 		"The Azure ResourceGroup name of the Virtual Network (defaults to --resource-group-name if not provided)")
 	cmd.Flags().StringVar(&o.fqdn, "fqdn", "", "The private AKS cluster fqdn")
+	cmd.Flags().BoolVar(&o.privateLink, "private-link", false, "Use the private FQDN for the API server")
 
 	utilruntime.Must(cmd.MarkFlagRequired("resource-group-name"))
 	utilruntime.Must(cmd.MarkFlagRequired("resource-name"))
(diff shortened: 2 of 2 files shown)


docs/installation/install.md
```diff
@@ -120,6 +120,8 @@ Alternatively, you can manually set a different name with the `--cluster-name` *
 ```{admonition} Note
 If you are running an [AKS private cluster](https://learn.microsoft.com/en-us/azure/aks/private-clusters), you may need to set the `--disable-api-server-sanity-check` *liqoctl* flag, since the API Server in your kubeconfig may be different from the one retrieved from the Azure APIs.
 
+If the private cluster uses private link, you can set the `--privateLink` *liqoctl* flag to use the private FQDN for the API server.
```
Sharathmk99, line 123: Private link is same as fqdn right? What’s the difference between fqdn and private link flag? Thanks

scottboring: Hi @Sharathmk99 , the fqdn flag would be used for cases where customers setup a a custom DNS. https://learn.microsoft.com/en-us/azure/aks/private-clusters?tabs=azure-portal#configure-a-private-dns-zone

Private link with AKS is a feature where a private FQDN is generated for AKS clusters that are not accessible over the public internet. The FQDN is a generated value by Azure and stored on the AKS cluster object. In this scenario, the API server used by Liqo should be the privateFQDN, not the public FQDN. https://learn.microsoft.com/en-us/azure/private-link/private-link-service-overview

Sharathmk99: Understood. We use private AKS cluster as well with private link and we are using fqdn to pass Private fqdn. May be there is confusion about naming conventions whether it should be ‘fqdn’ or ‘privateLink’ but use case is same. We are overwriting ApiServer with what’s present in flag.

scottboring: Thanks for responding @Sharathmk99, I understand that the existing --fqdn param can be used; but, the caller would first need to lookup the privateFQDN and pass that value in; but, that extra step is not necessary given that the value is on the AKS cluster object. Please take note of this line: https://github.com/scottboring/liqo/blob/aks-private-endpoint/pkg/liqoctl/install/aks/provider.go#L165

I considered suggesting that the existing --fqdn param be deprecated as this param is more useful; but, upon reading more it seemed like it is possible to setup a custom FQDN which I have not done with AKS and subsequently I assumed that value is not on the AKS cluster object. 

I feel this param simplifies the usage of liqoctl and is useful for my usecase. The --fqdn param would work; but requires and extra step that I would prefer to avoid.

Sharathmk99: Makes sense. Thank you for the clarification.

pkg/liqoctl/install/aks/provider.go
```diff
@@ -159,6 +161,8 @@ func (o *Options) parseClusterOutput(ctx context.Context, cluster *containerserv
 	}
 
 	switch {
+	case o.privateLink:
+		o.APIServer = *cluster.PrivateFQDN
```
aleoli, line 165: Can you please check that the value is not nil and return an error in this case?

aleoli (commented):
Hi, @scottboring! The PR looks good to me. Can you please fix this comment? And then I think we can merge.

scottboring:
@aleoli thank you for reviewing. The requested change has been made.

docs/installation/install.md
```diff
@@ -120,6 +120,8 @@ Alternatively, you can manually set a different name with the `--cluster-name` *
 ```{admonition} Note
 If you are running an [AKS private cluster](https://learn.microsoft.com/en-us/azure/aks/private-clusters), you may need to set the `--disable-api-server-sanity-check` *liqoctl* flag, since the API Server in your kubeconfig may be different from the one retrieved from the Azure APIs.
 
+If the private cluster uses private link, you can set the `--private-link` *liqoctl* flag to use the private FQDN for the API server.
```
frisso, line 123: @scottboring Why are we defining a flag named 'private-link' to set a DNS name?
To me, it looks more intuitive something as follows:

If the private clusters use a private link, you can use set the `--private-api-server` *liqoctl* flag to use the private FQDN for the API server.

What do you think?

aleoli: Hi @frisso, this flag is related to the AKS provider and is used to tell liqoctl to get the API Server address not from the public but from the private one (called `private-link` in Azure).

We already have another flag to set a generic different name for the API Server.

frisso: Got the point. However, we have to make sure that who reviews the documentation, even if it is not 100% expert of Azure, feels comfortable with what he's reading. The current sentence looks strange to me. What about:

If the private cluster uses private link, you can set the `--private-link` *liqoctl* flag, which configures this tool to use the private FQDN for the API server instead of the default (public) one.

aleoli: Perfectly fine for me

scottboring:
@aleoli commits have been squashed

PR #1991, opened 2023-09-08
Controller client timeout

cheina97:
# Description

This PR introduces the possibility of setting the timeout of the controllers' client from helm values **extra-args**.
It also inserts the possibility to specify **common extra-args** that will be applied to all the Liqo pods.

### DIFF
--- a/pkg/utils/restcfg/ratelimiting.go
+++ b/pkg/utils/restcfg/ratelimiting.go
@@ -16,11 +16,14 @@ package restcfg
 
 import (
 	"flag"
+	"time"
 
 	"k8s.io/client-go/rest"
 )
 
 const (
+	// DefaultClientTimeout -> The default timeout value assigned to client-go clients.
+	DefaultClientTimeout = 10 * time.Second
 	// DefaultQPS -> The default QPS value assigned to client-go clients.
 	DefaultQPS = uint(100)
 	// DefaultBurst -> The default burst value assigned to client-go clients.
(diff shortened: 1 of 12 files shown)


Sharathmk99:
Thank you so much. Looking forward to test it out.

pkg/utils/restcfg/ratelimiting.go
```diff
@@ -38,18 +42,21 @@ func InitFlags(flagset *flag.FlagSet) {
 		flagset = flag.CommandLine
 	}
 
+	flagset.DurationVar(&timeout, "client-timeout", timeout,
+		"The maximum length of time to wait before giving up on a server request. A value of zero means no timeout.")
 	flagset.UintVar(&qps, "client-qps", qps, "The maximum number of queries per second performed towards the API server.")
-	flagset.UintVar(&burst, "client-max-burst", qps, "The maximum burst of requests in excess of the rate limit towards the API server.")
+	flagset.UintVar(&burst, "client-max-burst", burst, "The maximum burst of requests in excess of the rate limit towards the API server.")
```
aleoli, line 48: I think the default values should be `DefaultClientTimeout`, `DefaultQPS` and `DefaultBurst`

aleoli (approved):

pkg/utils/restcfg/ratelimiting.go
cheina97: I know amd I agree with you, but if you look at the beginning of that file qps and burst are initialized with DefaultBurst and DefaultQPS. I know it's strange but it works. If you want I can change it

frisso: What about changing those values (keeping the suggestion from @aleoli ) and see what happens?

cheina97: Ok let's clean the code

PR #2014, opened 2023-09-13
liqo-route: mac annotation fix

cheina97:
# Description

This PR changes the method used to understand if a reconciled pod is the pod where the controller is running.

Previously we checked that the reconciled pod IP and the pod where the IP is running have the same IP. In some rare cases, this creates problems, in fact, if a node is deleted and another node is created with the same IP, it's possible that the controller gets an event that refer to a zombie pod. 

Using the podName instead of the IP solves this problem.

### DIFF
--- a/internal/liqonet/route-operator/overlayOperator.go
+++ b/internal/liqonet/route-operator/overlayOperator.go
@@ -42,7 +42,7 @@ var (
 type OverlayController struct {
 	client.Client
 	vxlanDev   *overlay.VxlanDevice
-	podIP      string
+	podName    string
 	nodesLock  *sync.RWMutex
 	vxlanPeers map[string]*overlay.Neighbor
 	// For each nodeName contains its IP addr.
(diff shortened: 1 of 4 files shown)


Sharathmk99:
Thank you so much for this fix. Will build and test it out in our cluster.

aleoli (approved):

internal/liqonet/route-operator/overlayOperator.go
```diff
@@ -75,8 +75,14 @@ func (ovc *OverlayController) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 		return ctrl.Result{}, nil
 	}
+
+	// If the pod don't have an IP address yet, then we don't need to do anything.
```
frisso, line 79:
```suggestion
	// If the pod doesn't have an IP address yet, then we don't need to do anything.
```

PR #2043, opened 2023-09-22
Endpointslices reflection fix

### LINKED ISSUE #2036: EndpointSlices Reflection should happen for services I annotated to offload from local cluster
#### What happened:
We are using "AllowList-based and DenyList-based Resource Reflection" feature and facing below problem.

We wanted to select few services from namespace A to offload to remote cluster, so we have set `reflection.service.type=AllowList` and added annotation `liqo.io/allow-reflection` to service which I wanted to offload to remote cluster. But I kept `reflection.endpointslice.type` to `DenyList`. Now I can see only selective services in remote cluster, which is good.

But now I'm creating `service A` in remote cluster which has same service name as local cluster but not selected to offload to remote cluster, now when I try to access `Service A` from remote cluster it's been forwarded to local cluster because `endpointslices` is created for all services under namespace.

I could have set `reflection.endpointslice.type` to `AllowList` but I don't manually create EndpointSlice in local cluster to add annotations. Ideally EndpointSlices should be created for the services I annotated to offload from local cluster. 

EndpointSlices should be created for the services I annotated to offload from local cluster, not for all services under namespace

Create `Service A` and `Service B` under `Namespace A` and annotate `Service A` with  `liqo.io/allow-reflection`. 
Try to create `Service B` in remote cluster and access `Service B` from remote cluster.

#### Environment:
- Liqo version: v0.9.4
- Liqoctl version: v0.9.4
- Kubernetes version (use `kubectl version`): 1.22.x
- Cloud provider or hardware configuration: kubeadm & AKS
- Node image:
- Network plugin and version:
- Install tools:
- Others:

fra98:
# Description

Fixes #2036 
This PR fixes a bug where endpointslices were reflected even if the associated service is marked with skip annotation.
The new logic to decide whether an endpointslice must be reflected or not is the following:
- the endpointslices reflector will inherit the reflection type from the services reflector
- an endpointslice is (not) reflected if the associated service is (not) reflected
- you can bypass the above behavior if you explicitly annotate the endpointslice itself (i.e., reflect the endpointslice using `liqo.io/allow-reflection` annotation, do not reflect using `liqo.io/skip-reflection`)

### DIFF
--- a/cmd/virtual-kubelet/root/root.go
+++ b/cmd/virtual-kubelet/root/root.go
@@ -283,7 +283,12 @@ func getReflectorsConfigs(c *Opts) (map[generic.ResourceReflected]*generic.Refle
 		if isReflectionTypeNotCustomizable(*resource) {
 			reflectionType = DefaultReflectorsTypes[*resource]
 		} else {
-			reflectionType = consts.ReflectionType(*c.ReflectorsType[string(*resource)])
+			if *resource == generic.EndpointSlice {
+				// the endpointslice reflector inherits the reflection type from the service reflector.
+				reflectionType = consts.ReflectionType(*c.ReflectorsType[string(generic.Service)])
+			} else {
+				reflectionType = consts.ReflectionType(*c.ReflectorsType[string(*resource)])
+			}
 			if reflectionType != consts.DenyList && reflectionType != consts.AllowList {
 				return nil, fmt.Errorf("reflection type %q is not valid for resource %s. Ammitted values: %q, %q",
 					reflectionType, *resource, consts.DenyList, consts.AllowList)
--- a/pkg/virtualKubelet/reflection/exposition/endpointslice.go
+++ b/pkg/virtualKubelet/reflection/exposition/endpointslice.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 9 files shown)


cmd/virtual-kubelet/root/root.go
```diff
@@ -283,7 +283,11 @@ func getReflectorsConfigs(c *Opts) (map[generic.ResourceReflected]*generic.Refle
 		if isReflectionTypeNotCustomizable(*resource) {
 			reflectionType = DefaultReflectorsTypes[*resource]
 		} else {
-			reflectionType = consts.ReflectionType(*c.ReflectorsType[string(*resource)])
+			if *resource == generic.EndpointSlice {
+				reflectionType = consts.ReflectionType(*c.ReflectorsType[string(generic.Service)])
+			} else {
```
aleoli, line 288: Can you add a comment here?

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -327,30 +328,65 @@ func (ner *NamespacedEndpointSliceReflector) UnmapEndpointIPs(ctx context.Contex
 	return nil
 }
 
+// forcedAllowOrSkip checks whether the given object is explicitly marked to be allowed or skipped.
+// If so, it returns whether the object should be skipped, or an error if unable to determine it.
+func forcedAllowOrSkip(obj metav1.Object) (forced, shouldSkip bool, err error) {
```
aleoli, line 333: If I understood correctly, you ignore the function's return (the shouldSkip value) if no annotation is set. I would add a comment to the function to point out this fact, since if the forced is false, the should skip should not be used since it is undetermined (here, we are not considering the reflection policy)

It can be better to return shouldSkip as *bool and return nil if no annotation is set

aleoli (approved):

PR #2045, opened 2023-09-24
Fix proxy image tagging

PaoloC68:
# Description

fixes this problem:

Failed to apply default image tag "envoyproxy/envoy:v1.21.0:v0.9.4": couldn't parse image reference "envoyproxy/envoy:v1.21.0:v0.9.4": invalid reference format

Then to change the image in the values this is the correct value:
```
proxy:
  imageName: envoyproxy/envoy
  imageTag: v1.21.0
```

# How Has This Been Tested?

I tried to deploy on my 2 clusters


cheina97:
Hi @PaoloC68, thanks for your PR.
Just a few comments before the review:

1. Rebase the PR with liqotech/liqo master
2. It is not a "fix" but a "feature", please change the title
3. It's not clear to me why you need this change, can you explain your problem?
4. Why you need this change only in liqo-proxy?

aleoli:
Hi @PaoloC68! These changes seem strange to me since the proxy follows the same release lifecycle as all the other components. Are these changes required?

cheina97:
Closed due to inactivity

PR #2047, opened 2023-09-26
Makefile: Improved MacOS compatibility

cannarelladev:
# Description

In this PR I want to improve the `Makefile` compatibility with MacOS:

`sed` command in MacOS has a different behavior from the GNU one, in particular when it is used with the `-i` flag it needs a suffix for the name of the "backup" original file. 

With this little fix it automatically selects the right `sed` syntax checking the running OS


cheina97:
lgtm

cheina97 (approved):

PR #2055, opened 2023-10-02
Gateway Tunnel: wireguard

cheina97:
# Description

Fixes #(issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B


cmd/gateway/tunnel/wireguard/main.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to create client: %w", err)
+	}
+
+	// Create the manager.
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		MapperProvider:         mapper.LiqoMapperProvider(scheme),
+		Scheme:                 scheme,
+		Namespace:              options.Namespace,
+		MetricsBindAddress:     "0", // Metrics are exposed by "connection" container.
```
aleoli, line 108: And the metrics of the controllers?

cmd/gateway/tunnel/wireguard/main.go
```diff
@@ -0,0 +1,165 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("unable to create manager: %w", err)
+	}
+
+	// Setup the controller.
+	pkr, err := wireguard.NewPublicKeysReconciler(
+		mgr.GetClient(),
+		mgr.GetScheme(),
+		mgr.GetEventRecorderFor("wireguard-controller"),
```
aleoli, line 130: Should we set the container name or the controller one? In this case this should be "public-keys-controller"

aleoli (commented):
You have some TODOs in the code. Should they be resolved before the merge, or will they be here?

pkg/gateway/tunnel/wireguard/publickeys-controller.go
```diff
@@ -0,0 +1,117 @@
...
+			return ctrl.Result{}, err
+		}
+	}
+
+	return ctrl.Result{}, EnsureConnection(ctx, r.Client, r.Scheme, r.Options)
+}
+
+// SetupWithManager register the ConfigurationReconciler to the manager.
+func (r *PublicKeysReconciler) SetupWithManager(mgr ctrl.Manager, src <-chan event.GenericEvent) error {
+	watcherDNS := NewWatcherDNS(src, r.Client, r.Options)
```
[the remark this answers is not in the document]
cheina97: Is to avoid to call another function to forge the predicate. But if you think is better i can change it.

cmd/gateway/tunnel/wireguard/main.go
```diff
@@ -0,0 +1,165 @@
...
+	// Set controller-runtime logger.
+	log.SetLogger(klog.NewKlogr())
+
+	// Get the rest config.
+	cfg := config.GetConfigOrDie()
+
+	// Create the client. This cliend should be used only outside the reconciler.
+	cl, err := client.New(cfg, client.Options{
+		Scheme: scheme,
+		Cache:  nil,
```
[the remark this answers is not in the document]
aleoli: No, simply add a comment to specify that we don't wont the cache

pkg/gateway/tunnel/wireguard/publickeys-controller.go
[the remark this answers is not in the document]
aleoli: ok, it's fine 👍 but can we call DNSWatcher instead of watcherDNS? it hurts me 😅

pkg/gateway/tunnel/wireguard/device.go
```diff
@@ -0,0 +1,57 @@
...
+				AllowedIPs: []net.IPNet{{IP: net.IP{0, 0, 0, 0}, Mask: net.CIDRMask(0, 32)}},
+				Endpoint: &net.UDPAddr{
+					IP:   options.EndpointIP,
+					Port: options.EndpointPort,
+				},
+			},
+		},
+		ReplacePeers: true,
+	})
+}
```
aleoli, line 57: can we define only one function and check if the listen port is defined or not?

pkg/gateway/tunnel/wireguard/dns.go
```diff
@@ -0,0 +1,125 @@
...
+				return false, nil
+			}
+		}
+
+		klog.Infof("DNS %q resolved to %q: updating endpoint", opts.EndpointAddress, ips[0])
+
+		// Copies the new IPs to store for the next check
+		opts.EndpointIPMutex.Lock()
+		defer opts.EndpointIPMutex.Unlock()
+		opts.EndpointIP = ips[0]
```
aleoli, line 73: can you check that len(ips) > 0?

aleoli (approved):

PR #2074, opened 2023-10-10
External Network: connection controller

cheina97:
# Description

This PR introduces a container that implements the generic features, like connection check and firewall.

This PR focuses only on Connection resources and connection checking, migrating the ConnChecker already used in the current network.


fra98 (approved):

pkg/gateway/connection/connections-controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	remoteIP, err := common.GetRemoteInterfaceIP(r.Options.Mode)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to get the remote interface IP: %w", err)
+	}
+
+	go r.ConnChecker.AddAndRunSender(
+		ctx, r.Options.RemoteClusterID, remoteIP,
+		ForgeUpdateConnectionCallback(ctx, r.Client, req),
+	)
```
aleoli, line 91: this is not guaranteed to be thread safe; the check if it is running and the set of the running one are safe

pkg/gateway/tunnel/wireguard/dns.go
```diff
@@ -106,10 +106,10 @@ func NewDNSEventHandler(cl client.Client, opts *Options) handler.EventHandler {
 	return handler.EnqueueRequestsFromMapFunc(
 		func(ctx context.Context, _ client.Object) []reconcile.Request {
 			labelSet := labels.Set{
-				string(LabelsMode):             string(opts.Mode),
-				string(consts.RemoteClusterID): opts.RemoteClusterID,
+				string(LabelsMode):             string(opts.GwOptions.Mode),
+				string(consts.RemoteClusterID): string(opts.GwOptions.Mode),
```
aleoli, line 110:
```suggestion
				string(consts.RemoteClusterID): string(opts.GwOptions.RemoteClusterID),
```

aleoli (approved):

aleoli (approved):

PR #2080, opened 2023-10-12
Metric Agent: cache fix

cheina97:
# Description

This PR fixes the cache used in **liqo-metric-agent**.

# How Has This Been Tested?

- [x] Locally on KinD

### DIFF
--- a/deployments/liqo/templates/liqo-metric-agent-deployment.yaml
+++ b/deployments/liqo/templates/liqo-metric-agent-deployment.yaml
@@ -65,6 +65,8 @@ spec:
           args:
           - --key-path=/certs/key.pem
           - --cert-path=/certs/cert.pem
+          - --read-timeout={{ .Values.metricAgent.config.timeout.read }}
+          - --write-timeout={{ .Values.metricAgent.config.timeout.write }}
           resources: {{- toYaml .Values.metricAgent.pod.resources | nindent 12 }}
           volumeMounts:
             - mountPath: '/certs'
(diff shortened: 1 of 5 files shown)


alessandro111297 (approved):

deployments/liqo/templates/liqo-metric-agent-deployment.yaml
```diff
@@ -65,6 +65,8 @@ spec:
           args:
           - --key-path=/certs/key.pem
           - --cert-path=/certs/cert.pem
+          - --read-timeout={{ .Values.metricAgent.config.timeout }}
+          - --write-timeout={{ .Values.metricAgent.config.timeout }}
```
aleoli, line 69: We should use different variables also in the halm values file

pkg/utils/clients/get_cached_client.go
```diff
@@ -18,60 +18,56 @@ package clients
...
-	scheme *runtime.Scheme, conf *rest.Config, cacheOptions *cache.Options) (client.Client, error) {
-	if conf == nil {
-		err := fmt.Errorf("the rest.Config parameter is nil")
-		klog.Error(err)
-		return nil, err
-	}
-
-	liqoMapper, err := (mapper.LiqoMapperProvider(scheme))(conf, nil)
+	scheme *runtime.Scheme, mapper meta.RESTMapper, conf *rest.Config, cacheOptions *cache.Options) (client.Client, error) {
+	podcache, err := cache.New(conf, *cacheOptions)
```
aleoli, line 35: Why `podcache`? It is a generic cache; change in the logs, too

aleoli (approved):

PR #2084, opened 2023-10-13
Security Mode: E2E tests

cheina97:
# Description

This PR adds E2E tests for security modes.

# How Has This Been Tested?

- [x] Liqo CI


test/e2e/pipeline/installer/liqoctl/setup.sh
```diff
@@ -79,6 +80,8 @@ do
   else
     LIQO_PROVIDER="${INFRA}"
   fi
+COMMON_ARGS=("${COMMON_ARGS[@]}" --set "networking.securityMode=${SECURITY_MODE}")
```
aleoli, line 83: Can we set a default in order to avoid to set it each time if running locally?

aleoli (approved):

PR #2105, opened 2023-10-20
External Network: support allocating static LoadBalancerIP and NodePort

fra98:
# Description

This PR adds the support to allocate static LoadBalancerIP and NodePort.


cmd/liqoctl/cmd/network.go
```diff
@@ -160,6 +160,10 @@ func newNetworkConnectCommand(ctx context.Context, options *network.Options) *co
 		fmt.Sprintf("Service type of the Gateway Server. Default: %s", gatewayserver.DefaultServiceType))
 	cmd.Flags().Int32Var(&options.ServerPort, "server-port", gatewayserver.DefaultPort,
 		fmt.Sprintf("Port of the Gateway Server. Default: %d", gatewayserver.DefaultPort))
+	cmd.Flags().Int32Var(&options.ServerNodePort, "node-port", 0,
+		"Force NodePort of Gateway Server")
+	cmd.Flags().StringVar(&options.ServerLoadBalancerIP, "load-balancer-ip", "",
+		"Force LoadBalancer IP of Gateway Server")
```
cheina97, line 166: Can you specify that values 0 and "" makes the port and ip random?

pkg/liqoctl/network/cluster.go
```diff
@@ -211,6 +213,25 @@ func (c *Cluster) GetGatewayClient(ctx context.Context, name string) (*networkin
...
+
+		if endpoint.NodePort != nil && *endpoint.NodePort != service.Spec.Ports[0].NodePort {
+			return true
+		}
+	}
+	if endpoint.LoadBalancerIP != nil && *endpoint.LoadBalancerIP != service.Spec.LoadBalancerIP {
+		return true
+	}
+	return false
+}
```
cheina97, line 233: What about using something like this?

https://github.com/kubernetes-sigs/controller-runtime/blob/5f8d96b043ef8ccadc544535415395be336630e8/pkg/controller/controllerutil/controllerutil.go#L216C1-L218C3

pkg/liqoctl/rest/gatewayserver/create.go
```diff
@@ -78,6 +78,8 @@ func (o *Options) Create(ctx context.Context, options *rest.CreateOptions) *cobr
 	cmd.Flags().Var(o.ServiceType, "service-type", fmt.Sprintf("Service type of Gateway Server. Default: %s", DefaultServiceType))
 	cmd.Flags().IntVar(&o.MTU, "mtu", DefaultMTU, "MTU of Gateway Server")
 	cmd.Flags().Int32Var(&o.Port, "port", DefaultPort, "Port of Gateway Server")
+	cmd.Flags().Int32Var(&o.NodePort, "node-port", 0, "Force NodePort of Gateway Server")
+	cmd.Flags().StringVar(&o.LoadBalancerIP, "load-balancer-ip", "", "Force LoadBalancer IP of Gateway Server")
```
cheina97, line 82: Can you specify that values 0 and "" makes the port and ip random?

cheina97 (approved):

pkg/liqoctl/network/cluster.go
fra98: I don't think we could use it in this case as the two types are different from each other, resulting always in a false equivalency.

PR #2109, opened 2023-10-23
Liqonet: IPSet

cheina97:
# Description

This PR adds the https://pkg.go.dev/k8s.io/kubernetes/pkg/util/ipset inside the liqo repository since it is not included anymore in the new **kubernetes** package release.


aleoli (commented):
We should change the import in order to use this package

aleoli (approved):

PR #2162, opened 2023-11-22
Move IPAM to separate pod

fra98:
# Description

This PR moves the IPAM server (currently started in the *network-manager*) to a separate pod.
It also adds the support to enable multiple replicas of the IPAM pod to provide high availability (implemented using an **active-passive** strategy).


pkg/leaderelection/leaderelection.go
```diff
@@ -40,17 +40,19 @@ var (
...
+	RenewDeadline     time.Duration
+	RetryPeriod       time.Duration
+	InitCallback      func()
+	StopCallback      func()
 }
 
 // InitAndRun initializes and runs the leader election mechanism.
-func InitAndRun(ctx context.Context, opts Opts, rc *rest.Config,
-	eb record.EventBroadcaster, initCallback func()) error {
+func InitAndRun(ctx context.Context, opts *Opts, rc *rest.Config, eb record.EventBroadcaster) error {
```
cheina97, line 55: Can you create a function that return a leaderelector and then run it inside the main instead of using a go function? You can use just one context in this way

cmd/ipam/main.go
```diff
@@ -0,0 +1,185 @@
...
+	eb := record.NewBroadcaster()
+	eb.StartRecordingToSink(&corev1clients.EventSinkImpl{Interface: localClient.CoreV1().Events(corev1.NamespaceAll)})
+
+	// Start IPAM using leader election mechanism.
+	if err := leaderelection.InitAndRun(ctx, leaderelectionOpts, cfg, eb); err != nil {
+		return err
+	}
+
+	// Wait until leader election context is canceled in the stop callback.
+	<-ctxLeaderElection.Done()
```
cheina97, line 149: You can run the leaderelector here instead of waiting on a context

cheina97 (approved):

PR #2193, opened 2023-12-06
Network: firewall controller webhook uniqe table name

cheina97:
# Description

This PR fixes a bug in webhook validation for firewall configuration table names

It depends on #2174

### DIFF
--- a/pkg/liqo-controller-manager/webhooks/firewallconfiguration/table.go
+++ b/pkg/liqo-controller-manager/webhooks/firewallconfiguration/table.go
@@ -17,6 +17,7 @@ package firewallconfiguration
 import (
 	"context"
 	"fmt"
+	"maps"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


pkg/liqo-controller-manager/webhooks/firewallconfiguration/table.go
```diff
@@ -60,17 +62,15 @@ func checkUniqueTableName(ctx context.Context, cl client.Client, currentFwcfg *n
...
 			continue
 		}
-		fwcfg := fwcfglist.Items[i]
-		tableName := fwcfg.Spec.Table.Name
-		if tableName == nil {
-			return fmt.Errorf("table name is nil")
-		}
-		if *tableName == *currentTableName {
-			return fmt.Errorf("table name %v is duplicated", *tableName)
+		if *fwcfglist.Items[i].Spec.Table.Name == *currentFwcfg.Spec.Table.Name &&
```
aleoli, line 70: you should check the pointer is not nil

pkg/liqo-controller-manager/webhooks/firewallconfiguration/table.go
```diff
@@ -41,7 +42,8 @@ func checkUniqueChainName(chains []firewallapi.Chain) error {
 
 // checkImmutableTableName checks if the table name is immutable.
 func checkImmutableTableName(fwcfg, oldFwcfg *networkingv1alpha1.FirewallConfiguration) error {
-	if oldFwcfg.Spec.Table.Name != fwcfg.Spec.Table.Name {
+	if fwcfg.Spec.Table.Name != nil && oldFwcfg.Spec.Table.Name != nil &&
```
aleoli, line 45: do we accept nil names? if not, you should check it in a separate if and return an error

aleoli (approved):

PR #2196, opened 2023-12-07
Network: remapping controller PodCIDR

cheina97:
# Description

This PR adds a controller which reconciles configuration resources and create a firewallconfiguration resource, used to remap overlapped CIDRs

This PR only considers PodCIDR remapping

### DIFF
--- a/pkg/gateway/remapping/const.go
+++ b/pkg/gateway/remapping/const.go
@@ -1,0 +1,25 @@
+// Copyright 2019-2023 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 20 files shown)


apis/networking/v1alpha1/firewall/chain_types.go
```diff
@@ -21,7 +21,7 @@ import "math"
 type ChainType string
 
 // Possible ChainType values.
-const (
+var (
```
aleoli, line 24: Why did you change all these const to var?

pkg/gateway/remapping/const.go
```diff
@@ -0,0 +1,25 @@
...
+package remapping
+
+var (
+	// TableNamePodCIDR is the name of the table for the pod CIDR.
+	TableNamePodCIDR = "remap-podcidr"
+
+	// DNATChain is the name of the chain for the output traffic.
+	DNATChain = "Outgoing packets remapping"
+	// SNATChain is the name of the chain for the input traffic.
+	SNATChain = "Incoming packets remapping"
```
aleoli, line 24:
```suggestion
	DNATChainName = "Outgoing packets remapping"
	// SNATChain is the name of the chain for the input traffic.
	SNATChainName = "Incoming packets remapping"
```

aleoli (approved):

PR #2216, opened 2023-12-13
liqoctl: add identity command

aleoli:
# Description

This pr implements the `generate` and the `update` APIs for certificate rotation. It also restarts pods that are using the certificate.

Now, you can use the `liqoctl generate identity` command on a cluster and apply the output to the other one.

Note: if the certificate is expired you probably need to update certificates in both directions (applying this couple of commands twice, starting from different clusters)

# How Has This Been Tested?

- [x] locally


pkg/liqoctl/rest/identity/create.go
```diff
@@ -0,0 +1,54 @@
...
+	case yamlLabel:
+		printer = &printers.YAMLPrinter{}
+	case jsonLabel:
+		printer = &printers.JSONPrinter{}
+	default:
+		return fmt.Errorf("unsupported output format %q", outputFormat)
+	}
+
+	return printer.PrintObj(conf, os.Stdout)
+}
```
cheina97, line 54: This should be moved in generate.go or types.go

cheina97 (approved):

PR #2259, opened 2024-01-08
add net interfaces monitoring

pucciorenzo:
# Description

This package includes a tool to monitor network interfaces changes and send a channel with a generic event whenever a change happens. It can detect:

- Added/Removed interface
- Interface set to up/down
- Added/Removed IP address to/from an interface
- Added/Removed route to/from an interface's routing table

# How Has This Been Tested?

Manually, adding/removing interfaces, IP addresses, routes

### DIFF
--- a/pkg/utils/netmonitor/netmonitor.go
+++ b/pkg/utils/netmonitor/netmonitor.go
@@ -1,0 +1,170 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package netmonitor
+
+import (
+	"context"
+	"net"
+	"syscall"
+
+	"github.com/vishvananda/netlink"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+// Options defines the options for the interfaces monitoring.
+// If the option is true, the monitoring will be enabled for that type of change.
+// Possible options are: link, address, route.
+type Options struct {
+	Link  bool
+	Addr  bool
+	Route bool
+}
+
+// InterfacesMonitoring starts the monitoring of the network interfaces.
+// If there is a change in the network interfaces, it will send a message to the channel.
+// With the options, you can choose to monitor only the link, address, or route changes (default: all options are true).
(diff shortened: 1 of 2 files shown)


pkg/utils/netmonitor/netmonitor.go
```diff
@@ -0,0 +1,163 @@
...
+// Options defines the options for the monitoring.
+type Options struct {
+	Link  bool
+	Addr  bool
+	Route bool
+}
+
+// StartMonitoring starts the monitoring of the network interfaces.
+// If there is a change in the network interfaces, it will send a message to the channel.
+// With the options, you can choose to monitor only the link, address, or route changes.
```
aleoli, line 36: Can we have a default behaviour if the options are nil? (and describe it in the comment)

It can be:

* return an error
* set all to true
* set all to false
* other

pucciorenzo: I setted all to true as default

pkg/utils/netmonitor/netmonitor.go
```diff
@@ -0,0 +1,171 @@
...
+	if options == nil {
+		options = &Options{
+			Link:  true,
+			Addr:  true,
+			Route: true,
+		}
+	}
+
+	// Subscribe to the address updates
+	if err := netlink.AddrSubscribe(chAddr, doneAddr); err != nil {
```
aleoli, line 62: You should use the ctx.Done() channel for those functions

pkg/utils/netmonitor/netmonitor.go
```diff
@@ -0,0 +1,171 @@
...
+
+func handleLinkUpdate(updateLink *netlink.LinkUpdate, interfaces map[string]bool, eventChannel chan<- event.GenericEvent) {
+	switch {
+	case updateLink.Header.Type == syscall.RTM_DELLINK:
+		// Link has been removed
+		klog.Info("Interface removed: ", updateLink.Link.Attrs().Name)
+		delete(interfaces, updateLink.Link.Attrs().Name)
+	case !interfaces[updateLink.Link.Attrs().Name] && updateLink.Header.Type == syscall.RTM_NEWLINK:
+		// New link has been added
+		klog.Info("Interface added: ", updateLink.Link.Attrs().Name)
```
aleoli, line 122:
```suggestion
		klog.Infof("Interface added: %s", updateLink.Link.Attrs().Name)
```

The same for the others

pkg/utils/netmonitor/netmonitor.go
```diff
@@ -0,0 +1,171 @@
...
+		interfaces[updateLink.Link.Attrs().Name] = true
+	case updateLink.Header.Type == syscall.RTM_NEWLINK:
+		// Link has been modified
+		if updateLink.Link.Attrs().Flags&net.FlagUp != 0 {
+			klog.Info("Interface ", updateLink.Link.Attrs().Name, " is up")
```
aleoli, line 127:
```suggestion
			klog.Info("Interface %s is up", updateLink.Link.Attrs().Name)
```

The same for the others

aleoli (approved):

aleoli:
@cheina97, do you want to review it, too?

cheina97 (approved):

PR #2276, opened 2024-01-16
Added preuninstall checks for new networking resources

fra98:
# Description

This PR adds a series of pre-uninstall checks for new networking resources

### DIFF
--- a/pkg/liqoctl/uninstall/preuninstall.go
+++ b/pkg/liqoctl/uninstall/preuninstall.go
@@ -17,9 +17,15 @@ package uninstall
 import (
 	"context"
 	"fmt"
+	"slices"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


pkg/liqoctl/uninstall/preuninstall.go
```diff
@@ -99,15 +116,93 @@ func (o *Options) preUninstall(ctx context.Context) error {
...
+			addResourceToErrMap(&networks.Items[i], &errMap, &foreignClusterList)
+		}
+	}
+
 	return errMap.getError()
 }
+
+func addResourceToErrMap(obj client.Object, errMap *errorMap, foreignClusters *discoveryv1alpha1.ForeignClusterList) {
+	// Check if object is a networking resource associated with a remote cluster
+	clusterName, found := getRemoteClusterName(obj, foreignClusters)
```
aleoli, line 183: I would use the same method to get the name, independently if it is a network resource or not. It can be another resource with the label, or a network resource where the foreignclsuter is no more available

PR #2288, opened 2024-01-18
Helm: Fix extraArgs

kariya-mitsuru:
# Description

ExtraArgs was not applied in the following two deployments and one cronjob, so fixed it so that it is applied correctly.

- deployments/liqo/templates/liqo-metric-agent-deployment.yaml
- deployments/liqo/templates/liqo-proxy-deployment.yaml
- deployments/liqo/templates/liqo-telemetry-cronjob.yaml


cheina97:
Hi @kariya-mitsuru, thanks for your contribution. In other components we use this syntax to add extraargs https://github.com/kariya-mitsuru/liqo/blob/a3748d493921bc07262461d7277a21fd5d9e633c/deployments/liqo/templates/liqo-crd-replicator-deployment.yaml#L39-L41.

Can you adapt your commit to our syntax or modify all the other components?

fra98:
LGTM! 👍🏻 
Can you please squash all commits into one?

fra98 (approved):

PR #2304, opened 2024-01-26
Virtual node taints

aleoli:
# Description

This pr adds the possibility to specify additional annotations and taint to be applied in the liqo node setting them in the VirtualNode spec

# How Has This Been Tested?

- [x] locally

### DIFF
--- a/pkg/utils/slice/slice.go
+++ b/pkg/utils/slice/slice.go
@@ -14,27 +14,14 @@
 
 package slice
 
-// ContainsString checks if a given slice of strings contains the provided string.
-func ContainsString(slice []string, s string) bool {
-	for _, item := range slice {
-		if item == s {
-			return true
-		}
-	}
-	return false
-}
+import "slices"
 
-// RemoveString returns a newly created []string that contains all items from slice that
+// Remove returns a newly created slice that contains all items from slice that
 // are not equal to s.
-func RemoveString(slice []string, s string) []string {
-	newSlice := make([]string, 0)
-	for _, item := range slice {
-		if item == s {
-			continue
-		}
-		newSlice = append(newSlice, item)
-	}
-	return newSlice
+func Remove[K comparable](slice []K, s K) []K {
+	return slices.DeleteFunc(slice, func(t K) bool {
+		return t == s
+	})
 }
 
 // LongestString returns the longest string in a slice of strings.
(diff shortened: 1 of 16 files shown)


pkg/utils/slice/slice.go
```diff
@@ -14,8 +14,8 @@
 
 package slice
 
-// ContainsString checks if a given slice of strings contains the provided string.
-func ContainsString(slice []string, s string) bool {
+// Contains checks if a given slice contains the provided element.
+func Contains[K comparable](slice []K, s K) bool {
```
fra98, line 18: I think we could delete this function and use the new Contains() from standard library package slices

pkg/utils/slice/slice.go
```diff
@@ -24,10 +24,10 @@ func ContainsString(slice []string, s string) bool {
 	return false
 }
 
-// RemoveString returns a newly created []string that contains all items from slice that
+// Remove returns a newly created slice that contains all items from slice that
 // are not equal to s.
-func RemoveString(slice []string, s string) []string {
-	newSlice := make([]string, 0)
+func Remove[K comparable](slice []K, s K) []K {
+	newSlice := make([]K, 0)
```
fra98, line 30: What about using standard library function `DeleteFunc()` ?
```suggestion
func Remove[K comparable](slice []K, s K) []K {
	return slices.DeleteFunc(slice, func(t K) bool {
		return t == s
	})
}
```

pkg/virtualKubelet/liqoNodeProvider/setup.go
```diff
@@ -101,7 +101,7 @@ func node(cfg *InitConfig) *corev1.Node {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        cfg.NodeName,
 			Labels:      labels.Merge(lbls, cfg.ExtraLabels),
-			Annotations: cfg.ExtraAnnotations,
+			Annotations: cfg.ExtraAnnotations, // TODO: check that they are not overwritten by annotations patch function
```
fra98, line 104: Is this TODO still relevant?

fra98 (approved):

pkg/virtualKubelet/liqoNodeProvider/setup.go
aleoli: No, I checked it, but forgot to remove the comment

PR #2310, opened 2024-01-30
Select remote namespace name when offloading

### LINKED ISSUE #2307: [Feature] Allow setting the name of the remote namespaces
Currently when offloading a local cluster to remote cluster, there are two supported methods. One method is with the `EnforceSameName` strategy, and the other is with the `DefaultName` which would create a randomally generated namespace

Assuming I got several clusters with offloaded namespaces that got the same name, I cannot use the `EnforceSameName` strategy

I'd like to have namespaces on the remote cluster with predetermined names, for several reasons

- Namespaces names got meaning, for example when looking in K8s billing reports that are based on namespace
- Having a "stable" name would allow to more easily offload/unoffload for fixing an issue, without having to change many other things that needs to be aware of the name of the namespace, and it would also make the billing reports stable
- When able to control the names of the namespace, doing things programmatically is much easier

Thanks in advance and love your project ❤️

aleoli:
# Description

This PR allows selecting the name a namespace will assume in the other cluster after offloading.

Fixes #2307 

# How Has This Been Tested?

- [x] locally

### DIFF
--- a/docs/usage/namespace-offloading.md
+++ b/docs/usage/namespace-offloading.md
@@ -44,6 +44,8 @@ The accepted values are:
 * **EnforceSameName**: remote namespaces are named after the local cluster's namespace.
 This approach ensures **naming transparency**, which is required by certain applications, as well as guarantees that **cross-namespace DNS queries** referring to reflected services work out of the box (i.e., without adapting the target namespace name).
 Yet, it can lead to **conflicts** in case a namespace with the same name already exists inside the selected remote clusters, ultimately causing the remote namespace creation request to be rejected.
+* **SelectedName**: you can specify the name of the remote namespace through the `--remote-namespace-name` flag.
+This flag is ignored in case the *namespace mapping strategy* is set to *DefaultName* or *EnforceSameName*.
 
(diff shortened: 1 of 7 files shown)


docs/usage/namespace-offloading.md
```diff
@@ -44,6 +44,8 @@ The accepted values are:
 * **EnforceSameName**: remote namespaces are named after the local cluster's namespace.
 This approach ensures **naming transparency**, which is required by certain applications, as well as guarantees that **cross-namespace DNS queries** referring to reflected services work out of the box (i.e., without adapting the target namespace name).
 Yet, it can lead to **conflicts** in case a namespace with the same name already exists inside the selected remote clusters, ultimately causing the remote namespace creation request to be rejected.
+* **SelectedName**: you can specify the name of the remote namespace through the `--emote-namespace-name` flag.
```
fra98, line 47:
```suggestion
* **SelectedName**: you can specify the name of the remote namespace through the `--remote-namespace-name` flag.
```

fra98 (approved):

PR #2318, opened 2024-02-04
liqoctl: add incoming flag to peer and unpeer commands

### LINKED ISSUE #957: [Feature] Remove Incoming Peering
The possibility to disable an incoming peering with `liqoctl`.

A simple `liqoctl` subcommand to disable the incoming peering for this cluster, to stop the resource sharing and remove the pod scheduled by the remote cluster on the local one.

hamzalsheikh:
# Description

The pull request adds an incoming flag in the peer and unpeer commands. 
User will be able to explicitly set the incoming flag while executing the commands.

Passing --incoming in peer commands sets the ForeignCluster resource .spec.incomingPeeringEnabled to Yes.
Passing --incoming in unpeer commands sets the ForeignCluster resource .spec.incomingPeeringEnabled to No.

Fixes #957 

# How Has This Been Tested?

- locally

### DIFF
--- a/pkg/liqoctl/wait/wait.go
+++ b/pkg/liqoctl/wait/wait.go
@@ -76,6 +76,20 @@ func (w *Waiter) ForOutgoingUnpeering(ctx context.Context, remoteClusterID *disc
 	return nil
 }
 
+// ForIncomingUnpeering waits until the status on the foreiglcusters resource states that the incoming peering has been successfully
+// set to None or the timeout expires.
+func (w *Waiter) ForIncomingUnpeering(ctx context.Context, remoteClusterID *discoveryv1alpha1.ClusterIdentity) error {
+	remName := remoteClusterID.ClusterName
+	s := w.Printer.StartSpinner(fmt.Sprintf("Disabling incoming peering to the remote cluster %q", remName))
+	err := fcutils.PollForEvent(ctx, w.CRClient, remoteClusterID, fcutils.IsIncomingPeeringNo, 1*time.Second)
+	if client.IgnoreNotFound(err) != nil {
+		s.Fail(fmt.Sprintf("Failed disabling incoming peering to the remote cluster %q: %s", remName, output.PrettyErr(err)))
+		return err
+	}
+	s.Success(fmt.Sprintf("Successfully disabled incoming peering to the remote cluster %q", remName))
+	return nil
+}
+
 // ForAuth waits until the authentication has been established with the remote cluster or the timeout expires.
 func (w *Waiter) ForAuth(ctx context.Context, remoteClusterID *discoveryv1alpha1.ClusterIdentity) error {
 	remName := remoteClusterID.ClusterName
@@ -113,6 +127,20 @@ func (w *Waiter) ForOutgoingPeering(ctx context.Context, remoteClusterID *discov
 		return err
 	}
 	s.Success(fmt.Sprintf("Outgoing peering activated to the remote cluster %q", remName))
+	return nil
+}
+
+// ForIncomingPeering waits until the status on the foreiglcusters resource states that the incoming peering has been successfully
+// set to Yes or the timeout expires.
+func (w *Waiter) ForIncomingPeering(ctx context.Context, remoteClusterID *discoveryv1alpha1.ClusterIdentity) error {
+	remName := remoteClusterID.ClusterName
+	s := w.Printer.StartSpinner(fmt.Sprintf("Activating incoming peering to the remote cluster %q", remName))
+	err := fcutils.PollForEvent(ctx, w.CRClient, remoteClusterID, fcutils.IsIncomingPeeringYes, 1*time.Second)
+	if err != nil {
+		s.Fail(fmt.Sprintf("Failed activating outgoing peering to the remote cluster %q: %s", remName, output.PrettyErr(err)))
+		return err
+	}
+	s.Success(fmt.Sprintf("Incoming peering activated to the remote cluster %q", remName))
 	return nil
 }
 
--- a/pkg/utils/foreignCluster/peeringStatus.go
+++ b/pkg/utils/foreignCluster/peeringStatus.go
@@ -55,6 +55,16 @@ func IsIncomingPeeringNone(foreignCluster *discoveryv1alpha1.ForeignCluster) boo
 	return curPhase == discoveryv1alpha1.PeeringConditionStatusNone
 }
 
+// IsIncomingPeeringYes checks if the incoming peering is set to Yes.
+func IsIncomingPeeringYes(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
+	return foreignCluster.Spec.IncomingPeeringEnabled == discoveryv1alpha1.PeeringEnabledYes
+}
+
+// IsIncomingPeeringNo checks if the incoming peering is set to No.
+func IsIncomingPeeringNo(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
+	return foreignCluster.Spec.IncomingPeeringEnabled == discoveryv1alpha1.PeeringEnabledNo
+}
+
 // IsOutgoingPeeringNone checks if the outgoing peering is set to none.
 func IsOutgoingPeeringNone(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
 	curPhase := peeringconditionsutils.GetStatus(foreignCluster, discoveryv1alpha1.OutgoingPeeringCondition)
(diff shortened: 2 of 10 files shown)


hamzalsheikh:
kind/feature

hamzalsheikh:
Both unpeering commands delete the foreign cluster resource.

The unpeer out-of-band have a [boolean](https://github.com/liqotech/liqo/blob/f3f48579db504b543c2bb6ddcda6b5230a825594/pkg/liqoctl/unpeeroob/handler.go#L40) that prevents that, but is [always set to true](https://github.com/liqotech/liqo/blob/f3f48579db504b543c2bb6ddcda6b5230a825594/cmd/liqoctl/cmd/unpeer.go#L131) at the moment.

aleoli:
Thank @hamzalsheikh! When the `--incoming` flag is set, no change has to be made to the outgoing peering (neither for peer nor unpeer).

Can you change the behavior of the flag you are pointing out?

hamzalsheikh:
Sorry, my comment was ambiguous. 

No change is being made to the outgoing peering. The flag only changes incoming peering variable in foreign cluster spec. 

I'm pointing out that the current implementation of the unpeer commands automatically deletes the foreign cluster resource. I haven't changed that, but would suggest actively using the [UnpeerOOBMode ](https://github.com/liqotech/liqo/blob/f3f48579db504b543c2bb6ddcda6b5230a825594/pkg/liqoctl/unpeeroob/handler.go#L40) and adding a similar option to Inband in future enhancements.

aleoli:
Ok, I tested your implementation; if I peer c1 with c2:

```bash
# on c2
liqoctl generate peer-command
```

```bash
# on c1
liqoctl peer ...

k get foreignclusters.discovery.liqo.io
NAME         TYPE        OUTGOING PEERING   INCOMING PEERING   NETWORKING    AUTHENTICATION   AGE
snowy-wood   OutOfBand   Established        None               Established   Established      97s
```

Then disable the incoming on the other cluster
```bash
# on c2
liqoctl unpeer --incoming twilight-shape

k get foreignclusters.discovery.liqo.io
NAME             TYPE        OUTGOING PEERING   INCOMING PEERING   NETWORKING   AUTHENTICATION   AGE
twilight-shape   OutOfBand   None               Pending            None         Established      3m5s
```

and re-enable the incoming on c2
```bash
# on c2
liqoctl peer --incoming twilight-shape

k get foreignclusters.discovery.liqo.io
NAME             TYPE        OUTGOING PEERING   INCOMING PEERING   NETWORKING    AUTHENTICATION   AGE
twilight-shape   OutOfBand   Established        Pending            Established   Established      4m16s
```

It enables the outgoing peering but not the incoming; I expect to do exactly the opposite, leaving the outgoing in the previous state and enabling the incoming.

hamzalsheikh:
The flag should work as expected now. Running peer and unpeer commands with the --incoming flag will only change the foreign cluster resource .spec.incomingPeeringEnabled to yes and no, respectively.

Can you please test and confirm. :) 

However, the PR is **not ready to be merged yet**.

1. My initial changes was keeping the "outgoing peering unchanged," as in keep it running as is, with the --incoming flag adding to that. I think this is what you wanted to achieve in [this](https://github.com/liqotech/liqo/issues/957#issuecomment-1206191279) comment with the --all flag. Should I keep those changes and create a new flag --all for them, or do you prefer merging the simple version at the moment?

2. The flag change is reflected in:
```
:~/liqo$ sudo kubectl describe foreignclusters
   .
   .
   .

Spec:
   .
   .
   .

  Incoming Peering Enabled:  Yes/No

```
Correct me if I'm wrong, I don't think the flag should change the Incoming peering status too (in ```k get foreignclusters.discovery.liqo.io```, that should be done when establishing an incoming peering. If this is not the case, what other changes are needed?

3. (related to the above) For error handling and correctness I am currently waiting on the foreign cluster resource change, not the peering condition. Is that the right thing to wait for / is there other conditions I should consider (e.g. ForAuth, ForNode, etc ...)? I'm also not using the [GetStatus](https://github.com/liqotech/liqo/blob/6c2d035daaddf0105017d4c954ec06aa1c6dc397/pkg/utils/peeringConditions/peeringConditionsUtils.go#L55) method since it expects a peering condition which would be misleading, but I can add a new [condition](https://github.com/liqotech/liqo/blob/6c2d035daaddf0105017d4c954ec06aa1c6dc397/apis/discovery/v1alpha1/foreigncluster_types.go#L30) and use it.

hamzalsheikh:
The peer and unpeer commands should be working as intended. 

I'm not waiting on the peering status change, but the change in the foreign cluster resource:
1- https://github.com/hamzalsheikh/liqo/blob/860d783a992ff06c1a967d9532773875682082a1/pkg/liqoctl/wait/wait.go#L81
2- https://github.com/hamzalsheikh/liqo/blob/860d783a992ff06c1a967d9532773875682082a1/pkg/liqoctl/wait/wait.go#L136

aleoli:
Thanks @hamzalsheikh! It looks ok; I will test it deeper in the next few days.

pkg/liqoctl/wait/wait.go
```diff
@@ -76,6 +76,20 @@ func (w *Waiter) ForOutgoingUnpeering(ctx context.Context, remoteClusterID *disc
 	return nil
 }
 
+// ForIncomingUnpeering waits until the status on the foreiglcusters resource states that the incoming peering has been successfully
+// set to None or the timeout expires.
+func (w *Waiter) ForIncomingUnpeering(ctx context.Context, remoteClusterID *discoveryv1alpha1.ClusterIdentity) error {
+	remName := remoteClusterID.ClusterName
+	s := w.Printer.StartSpinner(fmt.Sprintf("Disabling incoming peering to the remote cluster %q", remName))
+	err := fcutils.PollForEvent(ctx, w.CRClient, remoteClusterID, fcutils.IsIncomingPeeringNone, 1*time.Second)
```
aleoli, line 84:
```suggestion
	err := fcutils.PollForEvent(ctx, w.CRClient, remoteClusterID, fcutils.IsIncomingPeeringNo, 1*time.Second)
```

pkg/utils/foreignCluster/peeringStatus.go
```diff
@@ -55,6 +55,12 @@ func IsIncomingPeeringNone(foreignCluster *discoveryv1alpha1.ForeignCluster) boo
 	return curPhase == discoveryv1alpha1.PeeringConditionStatusNone
 }
 
+// IsIncomingPeeringYes checks if the incoming peering is set to Yes.
+func IsIncomingPeeringYes(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
+	// curPhase := peeringconditionsutils.GetStatus(foreignCluster, discoveryv1alpha1.IncomingPeeringEnabled)
+	return foreignCluster.Spec.IncomingPeeringEnabled == discoveryv1alpha1.PeeringEnabledYes
+}
```
aleoli, line 62:
```suggestion
// IsIncomingPeeringYes checks if the incoming peering is set to Yes.
func IsIncomingPeeringYes(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
	return foreignCluster.Spec.IncomingPeeringEnabled == discoveryv1alpha1.PeeringEnabledYes
}

// IsIncomingPeeringNo checks if the incoming peering is set to No.
func IsIncomingPeeringYes(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
	return foreignCluster.Spec.IncomingPeeringEnabled == discoveryv1alpha1.PeeringEnabledNo
}
```

aleoli (commented):
Hi @hamzalsheikh, there is only a small change; then it's ok and we can merge!

Can you please fix it and squash your commits?

pkg/utils/foreignCluster/peeringStatus.go
```diff
@@ -55,6 +55,16 @@ func IsIncomingPeeringNone(foreignCluster *discoveryv1alpha1.ForeignCluster) boo
...
 
+// IsIncomingPeeringYes checks if the incoming peering is set to Yes.
+func IsIncomingPeeringYes(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
+	return foreignCluster.Spec.IncomingPeeringEnabled == discoveryv1alpha1.PeeringEnabledYes
+}
+
+// IsIncomingPeeringNo checks if the incoming peering is set to No.
+func IsIncomingPeeringYes(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
+	return foreignCluster.Spec.IncomingPeeringEnabled == discoveryv1alpha1.PeeringEnabledNo
+}
```
fra98, line 66:
```suggestion
// IsIncomingPeeringNo checks if the incoming peering is set to No.
func IsIncomingPeeringNo(foreignCluster *discoveryv1alpha1.ForeignCluster) bool {
	return foreignCluster.Spec.IncomingPeeringEnabled == discoveryv1alpha1.PeeringEnabledNo
}
```

fra98:
@hamzalsheikh you need to resolve linting and formatting errors. I think if you run `make generate` it should fix most of them. Also, remember to squash all the commits into one at the end

aleoli:
@hamzalsheikh, you still have a minor linting issue; thanks for fixing the others!

fra98 (approved):

aleoli (approved):

PR #2345, opened 2024-02-21
Support priorityClassName in Helm chart

EladDolev:
# Description

This PR adds support for defining [priorityClassName](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#pod-priority) on some of the pods

It is especially useful for the `liqo-route` as in real case scenarios it is not being schedule to some of the worker nodes and it breaks the network for anything which is running there. When assigning a high (enough) priority to `liqo-route` it is always scheduled to all of the worker nodes

# How Has This Been Tested?

- [x] Test A
Installed with default values -> no priorityClassName exists on any of the pods
- [x] Test B
Assigned priorityClassName values to all relevant workloads and the workloads are running with priorityClassName

### DIFF
--- a/deployments/liqo/README.md
+++ b/deployments/liqo/README.md
@@ -18,6 +18,7 @@
 | auth.pod.annotations | object | `{}` | Annotations for the auth pod. |
 | auth.pod.extraArgs | list | `[]` | Extra arguments for the auth pod. |
 | auth.pod.labels | object | `{}` | Labels for the auth pod. |
+| auth.pod.priorityClassName | string | `""` | PriorityClassName (https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#pod-priority) for the auth pod. |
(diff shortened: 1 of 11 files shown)


fra98:
Hi @EladDolev, thanks for your contribution. It seems like a useful feature.

Btw, keep in mind that the `liqo-route` daemonset is going to be replaced in the next release, due to a major refactoring of the liqo networking.

fra98 (changes requested):
Can you also add the `PriorityClass` value to the `crdReplicator`, `discovery`, and `metric-agent` pods?
Also, you can correctly generate the README.md file with the command `make docs`.

EladDolev:
Done as requested @fra98 
Much appreciated 🙏

fra98 (approved):

fra98:
LGTM!
Can you please rebase to the latest commit on master and squash the commits into one? Then we are ready to merge

EladDolev:
Done @fra98 
Thanks !!

PR #2366, opened 2024-02-29
Add Prometheus scraping in Helm chart

EladDolev:
# Description

Adding scraping using ServiceMonitor to the Controller Manager and PodMonitor to the CRD Replicator

Motivation is having the ability to scrape the controllers metrics in order of creating alerts on failed reconciles etc.

# How Has This Been Tested?

- [x] Test A
  Installed with default values -> only change is the addition of the `metrics` port on the relevant Deployment(s)
- [x] Test B
  Installed with relevant PodMonitor/ServiceMonitor enabled -> metrics are visible in Prometheus


EladDolev:
I'm happy to change the CRD Replicator PodMonitor to ServiceMonitor if requested 😄

PR #2384, opened 2024-03-11
Fix nat address and port override

aleoli:
# Description

This pr fixes overriding gateway services addresses and ports by setting annotations over them

In particular:
* `liqo.io/override-address` will override the address
* `liqo.io/override-port: "51820"` will override the port

This is particularly useful when exposing them behind a NAT or with a LoadBalancer external to Kubernetes

# How Has This Been Tested?

- [x] locally

### DIFF
--- a/pkg/liqo-controller-manager/external-network/wireguard/wggatewayserver_controller.go
+++ b/pkg/liqo-controller-manager/external-network/wireguard/wggatewayserver_controller.go
@@ -384,6 +384,10 @@ func (r *WgGatewayServerReconciler) forgeEndpointStatusNodePort(ctx context.Cont
 		}
 	}
 
+	if err := checkServiceOverrides(service, &addresses, &port); err != nil {
+		return nil, nil, err
+	}
+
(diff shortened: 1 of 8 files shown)


pkg/liqo-controller-manager/external-network/wireguard/wggatewayserver_controller.go
```diff
@@ -384,6 +385,20 @@ func (r *WgGatewayServerReconciler) forgeEndpointStatusNodePort(ctx context.Cont
...
+		}
+		if v, ok := service.Annotations[consts.OverridePortAnnotation]; ok {
+			p, err := strconv.ParseInt(v, 10, 32)
+			if err != nil {
+				klog.Errorf("unable to parse port %q from service %s/%s annotation: %v", v, service.Namespace, service.Name, err)
+				return nil, nil, err
+			}
+			port = int32(p)
+		}
+	}
```
fra98, line 400: What about creating a function in utils.go to avoid duplicate code? (i.e., L452-L464)

fra98 (approved):

PR #2399, opened 2024-03-20
[Auth] Initialize cluster authentication keys

fra98:
# Description

This PR handles the initialization of the cluster authentication keys when liqo is first installed.

It generates a pair of private and public keys using the Ed25519 signature algorithm. The keys are stored in a secret, encoded in PEM format.

### DIFF
--- a/pkg/liqo-controller-manager/authentication/utils.go
+++ b/pkg/liqo-controller-manager/authentication/utils.go
@@ -1,0 +1,46 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pkg/liqo-controller-manager/authentication/utils.go
```diff
@@ -0,0 +1,46 @@
...
+	"crypto/ed25519"
+	"crypto/rand"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+)
+
+// GenerateEd25519Keys returns a new pair of private and public keys in PEM format.
+// Keys are generated using the Ed25519 signature algorithm and encoded in PEM format.
+func GenerateEd25519Keys() (privateKey, publicKey []byte, err error) {
```
aleoli, line 27: Probably we already have a function doing the same thing in the current auth package

fra98: @aleoli you probably refer to [this](https://github.com/liqotech/liqo/blob/f72d0fe7cf40ae9cc5e64e3b683e5b701c7e3532/pkg/utils/csr/create.go#L28) funcion in CSR utils. I did not use that because it also generates and returns a CSR.

aleoli: Ok, perfect

cheina97 (approved):

PR #2411, opened 2024-03-27
In-band Peering - warning concerning the usage of "--api-server-url" …

Andrew64-bit:
…flag during K3s installation.

# Description

Usage of --api-server-url flag when installing K3s with liqoctl can generate problems during the In-band peering phase.

### DIFF
--- a/docs/usage/peer.md
+++ b/docs/usage/peer.md
@@ -144,7 +144,8 @@ To this end, these subcommands feature a parallel set of flags concerning Kubern
 ```
 
 ```{warning}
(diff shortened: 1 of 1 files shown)


frisso:
@Andrew64-bit Thank you very much for the patch!
Two questions, though:
1. from the current sentence I'm unable to understand if you suggest to install k3s with the --api-server-url flag, or you suggest *not to use* this flag. Can you please clarify?
2. Can you add also "why" this flag creates troubles? So that the reader is more aware of what's going on underneath.
Thanks!

frisso:
Closing as duplicate of #2412

PR #2412, opened 2024-03-27
Doc update

Andrew64-bit:
# Description

Update of the document concerning the In-Band peering phase.
Hope that it's clear.

### DIFF
--- a/docs/usage/peer.md
+++ b/docs/usage/peer.md
@@ -144,7 +144,8 @@ To this end, these subcommands feature a parallel set of flags concerning Kubern
 ```
 
 ```{warning}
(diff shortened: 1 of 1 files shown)


frisso (approved):

frisso:
LGTM. Thanks!

docs/usage/peer.md
```diff
@@ -143,6 +143,12 @@ The host used to issue the *liqoctl peer in-band* command must have **concurrent
 To this end, these subcommands feature a parallel set of flags concerning Kubernetes API access to the remote cluster, in the form `--remote-<flag>` (e.g., `--remote-kubeconfig`, `--remote-context`).
 ```
 
+```{warning}
+If you are using K3s as Kubernetes distribution and your local cluster is in a LAN behind a NAT (your home network, for example), be aware that Liqo should have been installed without the `--api-server-url` flag.
```
aleoli, line 147: I think it is not directly related to k3s, so I would not explicitly mention it. Here and in the last line

cheina97, line 147: The problem is not strictly related to the presence of a NAT, you will experience the same issue even with a publicly exposed k8s API.

PR #2414, opened 2024-03-28
[Auth] Refactor values file and controller manager main

fra98:
# Description

Refactor values file and controller-manager main.


cheina97 (approved):

aleoli (approved):
We need to remember to warn users about big breaking changes in the helm values file

PR #2432, opened 2024-04-05
[Auth] Liqoctl generate identity

fra98:
# Description

This PR implements the `liqoctl generate identity` command


pkg/liqoctl/rest/identity/generate.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/liqotech/liqo/pkg/liqo-controller-manager/authentication/forge"
+	"github.com/liqotech/liqo/pkg/liqoctl/completion"
+	"github.com/liqotech/liqo/pkg/liqoctl/output"
+	"github.com/liqotech/liqo/pkg/liqoctl/rest"
+	liqoutils "github.com/liqotech/liqo/pkg/utils"
+	"github.com/liqotech/liqo/pkg/utils/args"
+	"github.com/liqotech/liqo/pkg/utils/getters"
+)
+
+const liqoctlGenerateConfigHelp = `Generate the Identity resource to be applied on the remote consumer cluster.`
```
aleoli, line 34: Can we add an example here as we did with other generate/create commands?

And additionally, I would specify that this command only generates control plane identities

pkg/liqoctl/rest/identity/utils.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"os"
+
+	"k8s.io/cli-runtime/pkg/printers"
+
+	authv1alpha1 "github.com/liqotech/liqo/apis/authentication/v1alpha1"
+)
+
+// output implements the logic to output the generated Configuration resource.
+func (o *Options) output(conf *authv1alpha1.Identity) error {
```
aleoli, line 27: identity and not configuration, both in the comment and in the function argument

aleoli (approved):

PR #2447, opened 2024-04-09
Network: liqoctl pod wait for ready status

cheina97:
# Description

This PR solves a bug in **liqoctl network connect** command. The command doesn't wait for a pod with the correct name ("gw-" is missing).

### DIFF
--- a/pkg/liqoctl/wait/wait.go
+++ b/pkg/liqoctl/wait/wait.go
@@ -21,6 +21,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/liqoctl/wait/wait.go
```diff
@@ -224,10 +225,15 @@ func (w *Waiter) ForConfiguration(ctx context.Context, conf *networkingv1alpha1.
 
 // ForGatewayPodReady waits until the pod of a Gateway resource has been created and is ready.
 func (w *Waiter) ForGatewayPodReady(ctx context.Context, gateway client.Object) error {
-	s := w.Printer.StartSpinner(fmt.Sprintf("Waiting for gateway pod %s to be ready", gateway.GetName()))
-	gatewayDeployment := &appsv1.Deployment{}
+	gatewayDeployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("gw-%s", gateway.GetName()),
```
fra98, line 230: We should use `forge.GatewayResourceName()` for better compatibility

fra98 (approved):

pkg/liqoctl/wait/wait.go
cheina97: Of course, I thought there was a reason not to have used it before

PR #2450, opened 2024-04-11
[Auth] Create Identity from ResourceSlice

fra98:
# Description

This PR adds a controller that creates an Identity from a ResourceSlice resource.

### DIFF
--- a/pkg/liqo-controller-manager/authentication/forge/identity.go
+++ b/pkg/liqo-controller-manager/authentication/forge/identity.go
@@ -22,31 +22,42 @@ import (
 	"github.com/liqotech/liqo/pkg/consts"
 )
 
+// ControlPlaneIdentityName forges the name of a ControlPlane Identity resource given the remote cluster name.
+func ControlPlaneIdentityName(remoteClusterName string) string {
+	return "controlplane-" + remoteClusterName
+}
+
+// ResourceSliceIdentityName forges the name of a ResourceSlice Identity.
+func ResourceSliceIdentityName(resourceSlice *authv1alpha1.ResourceSlice) string {
+	return "resourceslice-" + resourceSlice.Name
+}
+
 // IdentityForRemoteCluster forges a Identity resource to be applied on a remote cluster.
-func IdentityForRemoteCluster(localClusterIdentity discoveryv1alpha1.ClusterIdentity, identityType authv1alpha1.IdentityType,
-	authParams *authv1alpha1.AuthParams, namespace *string) *authv1alpha1.Identity {
-	identity := Identity(localClusterIdentity)
-	MutateIdentity(identity, localClusterIdentity, identityType, authParams, namespace)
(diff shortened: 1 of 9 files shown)


pkg/liqo-controller-manager/authentication/identitycreator-controller/identitycreator_controller.go
```diff
@@ -0,0 +1,136 @@
...
+		klog.Error(err)
+		return err
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&authv1alpha1.ResourceSlice{}, builder.WithPredicates(localResSliceFilter), r.withResourceSliceAccepted()).
+		Complete(r)
+}
+
+func (r *IdentityCreatorReconciler) withResourceSliceAccepted() builder.ForOption {
```
aleoli, line 120: In the predicate, I would only check that the condition is set, and not to check its state letting us to react on denied resources

pkg/liqo-controller-manager/authentication/identitycreator-controller/identitycreator_controller.go
```diff
@@ -0,0 +1,136 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if !r.resourceSliceAccepted(&resourceSlice) {
+		klog.V(4).Infof("Skip creation of identity as resourceslice %q is not valid or denied", req.NamespacedName)
+		r.eventRecorder.Event(&resourceSlice, corev1.EventTypeWarning, "IdentitySkipped", "Identity skipped as ResourceSlices not accepted")
+		return ctrl.Result{}, nil
+	}
+
+	// TODO: should we cancel the Identity if status condition become Denied?
```
aleoli, line 87: Yes, it will probably never become denied, but it is a good idea

pkg/liqo-controller-manager/authentication/forge/identity.go
```diff
@@ -22,24 +22,35 @@ import (
...
+func ResourceSliceIdentityName(resourceSlice *authv1alpha1.ResourceSlice) string {
+	return "resourceslice-" + resourceSlice.Name
+}
+
 // IdentityForRemoteCluster forges a Identity resource to be applied on a remote cluster.
-func IdentityForRemoteCluster(localClusterIdentity discoveryv1alpha1.ClusterIdentity, identityType authv1alpha1.IdentityType,
+func IdentityForRemoteCluster(name string, localClusterIdentity discoveryv1alpha1.ClusterIdentity, identityType authv1alpha1.IdentityType,
 	authParams *authv1alpha1.AuthParams, namespace *string) *authv1alpha1.Identity {
-	identity := Identity(localClusterIdentity)
+	identity := Identity(name, "")
```
aleoli, line 38: The empty parameter here is quite strange. Can we add it as a function parameter?

aleoli (approved):

PR #2477, opened 2024-04-22
[Net] Higher verbosity for internalnodes not found

fra98:
# Description

This PR raises the verbosity for errors where InternalNodes are not found in the pod_controller.


pkg/fabric/source-detector/gateway_controller.go
```diff
@@ -75,7 +75,7 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	internalnode := &networkingv1alpha1.InternalNode{}
 	if err = r.Get(ctx, client.ObjectKey{Name: r.Options.NodeName}, internalnode); err != nil {
 		if apierrors.IsNotFound(err) {
-			klog.Infof("There is no internal node %s", r.Options.NodeName)
+			klog.Infof("There is no internalnode %s", r.Options.NodeName)
 			return ctrl.Result{}, nil
```
aleoli, line 79: It's not related to your PR, but @cheina97, is it okay not to return an error here even if there is no re-enqueue for internal nodes in the setupwithmanager function?

aleoli (approved):

PR #2480, opened 2024-04-23
[Auth] Refactor IAM auth

aleoli:
# Description

This pr makes it possible to manage IAM users with the new authentication

# How Has This Been Tested?

- [x] on EKS

### DIFF
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -120,6 +120,7 @@ func main() {
 	var apiServerAddressOverride string
 	var caOverride string
 	var trustedCA bool
+	var awsConfig identitymanager.LocalAwsConfig
 
 	authenticationEnabled := flag.Bool("authentication-enabled", true, "Enable/disable the authentication module")
 	disableInternalNetwork := flag.Bool("disable-internal-network", false, "Disable the creation of the internal network")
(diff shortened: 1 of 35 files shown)


cmd/liqo-controller-manager/main.go
```diff
@@ -151,6 +152,12 @@ func main() {
 	foreignClusterPingTimeout := flag.Duration("foreign-cluster-ping-timeout", 5*time.Second,
 		"The timeout of the ForeignCluster API server readiness check")
 
+	// AWS configurations
+	flag.StringVar(&awsConfig.AwsAccessKeyID, "aws-access-key-id", "", "AWS IAM AccessKeyID for the Liqo User")
+	flag.StringVar(&awsConfig.AwsSecretAccessKey, "aws-secret-access-key", "", "AWS IAM SecretAccessKey for the Liqo User")
+	flag.StringVar(&awsConfig.AwsRegion, "aws-region", "", "AWS region where the local cluster is running")
+	flag.StringVar(&awsConfig.AwsClusterName, "aws-cluster-name", "", "Name of the local EKS cluster")
```
fra98, line 159: Can you move these flags few lines below with the other authentication flags?

pkg/identityManager/const.go
```diff
@@ -28,6 +26,8 @@ const (
 	remoteClusterIDTagKey = "liqo.io/remote-cluster-id"
 	managedByTagKey       = "liqo.io/managed-by"
 	managedByTagValue     = "liqo"
+	identityTypeTagKey    = "liqo.io/identity-type"
```
fra98, line 29: We could re-use the identical label in `pkg/consts/authentication.go`

fra98 (approved):

pkg/identityManager/const.go
aleoli: I copied the value from this constant to the other, in order to continue using the tagkey suffix, is it ok for you?

PR #2482, opened 2024-04-24
[Auth] Create virtualnodes from resourceslices

fra98:
# Description

This PR:
- automates the creation of virtualnodes from resourceslice resources (it can be disabled  `liqo.io/create-virtual-node=false` annotation)
- refactor `liqoctl create virtualnode`

### DIFF
--- a/cmd/virtual-kubelet/root/root.go
+++ b/cmd/virtual-kubelet/root/root.go
@@ -98,8 +98,7 @@ func runRootCommand(ctx context.Context, c *Opts) error {
 
 	// Retrieve the remote restcfg
 	tenantNamespaceManager := tenantnamespace.NewManager(localClient) // Do not use the cached version, as leveraged only once.
-	identityManager := identitymanager.NewCertificateIdentityReader(ctx, cl, localClient, localConfig,
-		c.HomeCluster, tenantNamespaceManager)
+	identityManager := identitymanager.NewCertificateIdentityReader(ctx, cl, localClient, localConfig, c.HomeCluster, tenantNamespaceManager)
 
 	if c.RemoteKubeconfigSecretName == "" {
 		return fmt.Errorf("remote kubeconfig secret name is mandatory")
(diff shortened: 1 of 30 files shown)


cmd/virtual-kubelet/root/root.go
```diff
@@ -105,18 +100,18 @@ func runRootCommand(ctx context.Context, c *Opts) error {
...
+	restcfg.SetRateLimiter(remoteConfig)
+
 	reflectorsConfigs, err := getReflectorsConfigs(c)
 	if err != nil {
 		return err
 	}
 
-	restcfg.SetRateLimiter(remoteConfig)
-
 	cl, err := client.New(localConfig, client.Options{
```
aleoli, line 115: It's totally fine, but why did you change the order? 😅

pkg/liqo-controller-manager/authentication/virtualnodecreator-controller/virtualnodecreator_controller.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		klog.Errorf("Unable to get the Identity associated to resourceslice %q: %s", req.NamespacedName, err)
+		return ctrl.Result{}, err
+	}
+
+	// Get associated secret
+	if identity.Status.KubeconfigSecretRef == nil || identity.Status.KubeconfigSecretRef.Name == "" {
+		err := fmt.Errorf("identity %q does not contain the kubeconfig secret reference", identity.Name)
+		klog.Error(err)
+		return ctrl.Result{}, err
```
aleoli, line 104: I would avoid returning errors here since it will error forever until other controllers fill the status.

It is more correct to return a nil error here without re-enqueue and set an own in the setupwithmanager function to trigger reconciliations on identity updates

pkg/liqo-controller-manager/offloading/forge/virtualnode.go
```diff
@@ -0,0 +1,88 @@
...
+		Hard: opts.ResourceList,
+	}
+
+	virtualNode.Spec.StorageClasses = make([]sharingv1alpha1.StorageType, len(opts.StorageClasses))
+	for i, storageClass := range opts.StorageClasses {
+		sc := sharingv1alpha1.StorageType{
+			StorageClassName: storageClass,
+		}
+		if i == 0 {
+			sc.Default = true
```
aleoli, line 82: In the old version, we had a flog in the storage class struct to define the default one, since it should be the remote cluster to select the default one

aleoli (approved):

PR #2531, opened 2024-05-24
Network: enable full masquerade

cheina97:
# Description

This PR introduces the **fabric-full-masquerade-enabled** flag. If enabled it masquerade all the traffic using the first IP of the local externalCIDR

### DIFF
--- a/deployments/liqo/values.yaml
+++ b/deployments/liqo/values.yaml
@@ -309,6 +309,13 @@ fabric:
     version: ""
   # -- Extra tolerations for the fabric daemonset.
   tolerations: []
+  config:
+    # -- Enabe/Disable the full masquerade mode for the fabric pod.
+    # It means that all traffic will be masquerade using the first external cidr IP, instead of using the pod IP.
+    # Full masquerade is useful when the cluster nodeports uses a PodCIDR IP to masqerade the incoming traffic.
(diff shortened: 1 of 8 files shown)


deployments/liqo/values.yaml
```diff
@@ -305,10 +305,14 @@ fabric:
   image:
     # -- Image repository for the fabric pod.
     name: "ghcr.io/liqotech/fabric"
-    # -- Custom version for the fabric image. If not specified, the global tag is used.
+    # -- Custom version for the fabric image. If not specified, the global tag |is used.
```
aleoli, line 308:
```suggestion
    # -- Custom version for the fabric image. If not specified, the global tag is used.
```

deployments/liqo/values.yaml
```diff
@@ -305,10 +305,14 @@ fabric:
...
     version: ""
   # -- Extra tolerations for the fabric daemonset.
   tolerations: []
+  config:
+    # -- Enabe/Disable the full masquerade mode for the fabric pod.
+    # It means that all traffic will be masquerade using the first external cidr IP, instead of using the pod IP.
```
aleoli, line 314: I would add a short description indicating the pros and cons of this flag and when it can be helpful to (1-2 lines)

aleoli (approved):

PR #2532, opened 2024-05-24
[Auth] offloading patch node selector

aleoli:
# Description

This pr adds the node selector fields to the resource slice resource


fra98 (commented):
Can you also add to `liqoctl create` the flag to add NodeSelector like we do with NodeLabels?

fra98 (approved):

PR #2540, opened 2024-05-30
Network: aggressive dns resolution + EKS loadbalancer v2

cheina97:
# Description

This PR modifies the DNS resolution for wireguard tunnel. Now THere is a period of **10 minutes** in which the resolver routine tries to resolve the DNS every 5 seconds. This helps with EKS UDP loadbalancers where the first resolution includes IPs that are not working. These IPs are removed by EKS from the DNS only after some time, so the DNS resolution has to be more "reactive" during the first minutes after creating the gateway server

This PR also includes a new wireguard gateway server template  that uses the EKS loadbalancer controller https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html

### DIFF
--- a/pkg/gateway/tunnel/wireguard/dns.go
+++ b/pkg/gateway/tunnel/wireguard/dns.go
@@ -40,23 +40,22 @@ import (
 // The DNS is resolved every 5 minutes.
 // If the DNS changed a new publickkeys-controller reconcile is triggered through a generic event.
 func StartDNSRoutine(ctx context.Context, ch chan event.GenericEvent, opts *Options) {
-	// Try to solve the DNS every 5 seconds until the DNS is resolved.
-	// This is useful to avoid to wait 5 minutes before the first DNS resolution.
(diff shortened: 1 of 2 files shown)


pkg/gateway/tunnel/wireguard/dns.go
```diff
@@ -40,23 +40,23 @@ import (
 // The DNS is resolved every 5 minutes.
 // If the DNS changed a new publickkeys-controller reconcile is triggered through a generic event.
 func StartDNSRoutine(ctx context.Context, ch chan event.GenericEvent, opts *Options) {
-	// Try to solve the DNS every 5 seconds until the DNS is resolved.
+	// Try to solve the DNS every 5 seconds until the DNS is resolved for 30 minutes.
```
fra98, line 43:
```suggestion
	// Try to solve the DNS every 5 seconds until the DNS is resolved for 10 minutes.
```

fra98 (approved):

PR #2554, opened 2024-06-10
[Offloading] VirtualKubelet options as CRD

fra98:
# Description

This PR refactors the VirtualKubelet deployment options:
- remove flags passed to the controller-manager
- make a template of the flags, generated from Helm values (cluster-wide setting)
- users can define custom templates and reference one in the VirtualNode Spec (per-virtualnode setting)


deployments/liqo/templates/liqo-vk-options-template.yaml
```diff
@@ -0,0 +1,82 @@
...
+    secret: {{ .Values.offloading.reflection.secret.workers }}
+    serviceaccount: {{ .Values.offloading.reflection.serviceaccount.workers }}
+    persistentvolumeclaim: {{ .Values.offloading.reflection.persistentvolumeclaim.workers }}
+    event: {{ .Values.offloading.reflection.event.workers }}
+  reflectorsType:
+    service: {{ .Values.offloading.reflection.service.type }}
+    ingress: {{ .Values.offloading.reflection.ingress.type }}
+    configmap: {{ .Values.offloading.reflection.configmap.type }}
+    secret: {{ .Values.offloading.reflection.secret.type }}
+    event: {{ .Values.offloading.reflection.event.type }}
```
aleoli, line 58: I don't know if it's better to have this on this form or change the hierarchy

```
reflection:
  pod:
    workers:
    type:
  service:
    workers:
    type:
...
```

in order to reuse the same small object for each resource and make it easier to add new ones in the future

aleoli (approved):

PR #2583, opened 2024-06-28
[Offloading] Offloading fixes

aleoli:
# Description

This PR fixes some bugs in the offloading module and improves its logs.

In particular, the virtual kubelet can offload pods even if the networking module is disabled.


deployments/liqo/values.yaml
```diff
@@ -292,7 +292,7 @@ controllerManager:
     # -- It enforces offerer-side that offloaded pods do not exceed offered resources (based on container limits).
     # This feature is suggested to be enabled when consumer-side enforcement is not sufficient.
     # It has the same tradeoffs of resource quotas (i.e, it requires all offloaded pods to have resource limits set).
-    enableResourceEnforcement: false
+    enableResourceEnforcement: true
```
fra98, line 295: Are we sure we want to set this as the new default?

fra98 (approved):

PR #2584, opened 2024-07-01
Docs: new peer usage

fra98:
# Description

This PR updates the documentation with the new liqo peer command


aleoli (approved):

frisso:
@fra98 Thanks for this nice PR!
One note: I would suggest to add a sentence somewhere in the documentation saying that in-band and out-of-band peerings are no longer available in Liqo 1.0, replaced by a more flexible mechanism (even better to say how the above mechanisms can be implemented right now). Otherwise, we may have people start asking "how do I do in-band peering" or such. Thanks!

PR #2596, opened 2024-07-03
Liqoctl: test network

cheina97:
# Description

This PR adds in liqoctl a new subsets of commands **"liqoctl test"** used to run E2E tests using liqoctl.

In this PR the network tests command have been added (**"liqoctl test network"**) which tests these scenarios:
- **Pod-to-Pod** connectivity
- **Pod-to-Node** connectivity
- **Pod-to-Service** connectivity (cluster-ip and nodeport)
- **External-to-LoadBalancer** connectivity
- **Pod-to-External** connectivity
- **Pod-To-External** (using IP resource remapping) connectivity

### DIFF
--- a/cmd/liqoctl/cmd/test.go
+++ b/cmd/liqoctl/cmd/test.go
@@ -1,0 +1,93 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/util/runtime"
+
+	"github.com/liqotech/liqo/pkg/liqoctl/completion"
+	"github.com/liqotech/liqo/pkg/liqoctl/factory"
+	"github.com/liqotech/liqo/pkg/liqoctl/output"
+	"github.com/liqotech/liqo/pkg/liqoctl/test"
+	"github.com/liqotech/liqo/pkg/liqoctl/test/network"
+	"github.com/liqotech/liqo/pkg/liqoctl/test/network/flags"
+)
+
+const liqoctlTestNetworkLongHelp = `Launch network E2E tests.
+
+WARNING: to run the tests you need to have kyverno installed on every cluster https://kyverno.io/docs/installation/methods/ .
+
+This command allows to launch E2E tests, which are used to check the network functionalities between the clusters.
+The command needs to be run on the cluster that will act as the consumer,
+and it requires the kubeconfig of the remote cluster that will act as the providers.
+The consumer cluster must be peered with the providers, previously using "{{ .Executable }} peer".
+
+
+Examples:
+  $ {{ .Executable }} test network --remote-kubeconfigs $HOME/.kube/config2,$HOME/.kube/config3
+or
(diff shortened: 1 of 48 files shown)


pkg/liqoctl/test/flags.go
```diff
@@ -0,0 +1,44 @@
...
+	FlagNamesRemoveNamespace FlagNames = "rm"
+	// FlagNamesFailFast is the flag name for the fail-fast option.
+	FlagNamesFailFast FlagNames = "fail-fast"
+	// FlagNamesInfo is the flag name for the information output.
+	FlagNamesInfo FlagNames = "info"
+)
+
+// AddFlags adds the flags to the flag set.
+func (opts *Options) AddFlags(fs *pflag.FlagSet) {
+	fs.StringSliceVarP(&opts.RemoteKubeconfigs, string(FlagNamesProducersKubeconfigs), "p", []string{}, "A list of kubeconfigs for remote clusters")
```
fra98, line 39:
```suggestion
	fs.StringSliceVarP(&opts.RemoteKubeconfigs, string(FlagNamesProducersKubeconfigs), "p", []string{}, "A list of kubeconfigs for remote provider clusters")
```

cmd/liqoctl/cmd/test.go
```diff
@@ -0,0 +1,93 @@
...
+		Short: "Launch E2E tests",
+		Long:  "Launch E2E tests",
+		Args:  cobra.NoArgs,
+	}
+
+	cmd.AddCommand(newTestNetworkCommand(ctx, options))
+
+	options.AddFlags(cmd.PersistentFlags())
+
+	runtime.Must(cmd.MarkPersistentFlagRequired(string(test.FlagNamesProducersKubeconfigs)))
```
fra98, line 66: Shouldn't this be in the sub-command network?

pkg/utils/maps/maps.go
```diff
@@ -197,3 +197,14 @@ func SerializeMap(m map[string]string) string {
...
+// GetIndexFromMap returns the key and the value of the map at the given index.
+func GetIndexFromMap[K comparable, V any](m map[K]V, i int) (k K, v V) {
+	for k, v = range m {
+		if i == 0 {
+			return k, v
+		}
+		i--
+	}
+	return *new(K), *new(V)
+}
```
fra98, line 210: Looking at the official go documentation (https://go.dev/ref/spec#RangeClause) the order of iteration is not guaranteed to be the same every time (it is only when the map printed). Is this safe in our case?

pkg/liqoctl/utils/pod/pod.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	pods := corev1.PodList{}
+	if err := cl.List(ctx, &pods,
+		ctrlclient.InNamespace(setup.NamespaceName),
+		ctrlclient.MatchingLabels{setup.PodLabelAppCluster: deploymentName + "-" + owner},
+	); err != nil {
+		return nil, fmt.Errorf("failed to list pods: %w", err)
+	}
+	return &pods, nil
+}
```
fra98, line 85: This function is placed in a general util pkg but uses names from a specific network test. Is it possible to generalize it? if not I would place it inside the network/test folder

pkg/liqoctl/test/client/k8s.go
```diff
@@ -0,0 +1,125 @@
...
+
+func getClientName(ctx context.Context, cl client.Client) (string, error) {
+	cfgmap := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Name: "liqo-clusterid-configmap", Namespace: "liqo"},
+	}
+	if err := (cl).Get(ctx, client.ObjectKeyFromObject(&cfgmap), &cfgmap); err != nil {
+		return "", fmt.Errorf("failed to get cluster ID: %w", err)
+	}
+	return cfgmap.Data["CLUSTER_ID"], nil
+}
```
fra98, line 125: Why not using `GetClusterIDWithControllerClient()` function?

cheina97: Because I didn't know it was there :)

fra98 (approved):

fra98 (approved):

PR #2624, opened 2024-07-17
[Test] Unoffload

aleoli:
# Description

This pr adds test over the namespace unoffloading

### DIFF
--- a/test/e2e/cruise/unoffload/unoffload_test.go
+++ b/test/e2e/cruise/unoffload/unoffload_test.go
@@ -1,0 +1,261 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


test/e2e/cruise/unoffload/unoffload_test.go
```diff
@@ -0,0 +1,261 @@
...
+		BeforeEach(func() {
+			// ensure the namespace is created
+			for i := range testContext.Clusters {
+				if testContext.Clusters[i].Role == liqov1alpha1.ConsumerRole {
+					// ensure the namespace is created
+					Expect(util.Second(util.EnforceNamespace(ctx, testContext.Clusters[i].NativeClient,
+						testContext.Clusters[i].Cluster, namespaceName))).To(Succeed())
+
+					_ = util.OffloadNamespace(testContext.Clusters[i].KubeconfigPath, namespaceName,
+						"--namespace-mapping-strategy=EnforceSameName")
```
fra98, line 180: Why the error is discarded?

fra98 (approved):

test/e2e/cruise/unoffload/unoffload_test.go
aleoli: Because if one test fails, the namespace may already be offloaded, and the function returns errors if the namespace is already offloaded so that the next test will fail

PR #2625, opened 2024-07-17
`liqoctl create nonce` namespace is not created when output specified

claudiolor:
# Description

Previously when to the `liqoctl create nonce` the output was specified, a namespace was created in the cluster. With this patch, when the output is specified, the command does not perform any change to the cluster, and the Namespace resource is added to the output. If a tenant namespace already esists, the Namespace resource of the output will have the same name of the existing Namespace.

### DIFF
--- a/pkg/liqoctl/rest/nonce/create.go
+++ b/pkg/liqoctl/rest/nonce/create.go
@@ -22,6 +22,7 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/cli-runtime/pkg/printers"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/liqotech/liqo/pkg/liqo-controller-manager/authentication/forge"
 	authutils "github.com/liqotech/liqo/pkg/liqo-controller-manager/authentication/utils"
@@ -81,17 +82,18 @@ func (o *Options) Create(ctx context.Context, options *rest.CreateOptions) *cobr
 
 func (o *Options) handleCreate(ctx context.Context) error {
 	opts := o.createOptions
+
+	if opts.OutputFormat != "" {
+		opts.Printer.CheckErr(o.output(ctx))
+		return nil
+	}
+
 	waiter := wait.NewWaiterFromFactory(opts.Factory)
 
 	tenantNs, err := o.namespaceManager.CreateNamespace(ctx, o.clusterID.GetClusterID())
 	if err != nil {
 		opts.Printer.CheckErr(fmt.Errorf("unable to create tenant namespace: %v", output.PrettyErr(err)))
 		return err
-	}
-
-	if opts.OutputFormat != "" {
-		opts.Printer.CheckErr(o.output(tenantNs.GetName()))
-		return nil
 	}
 
 	// Ensure the presence of the nonce secret.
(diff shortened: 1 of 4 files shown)


pkg/liqoctl/rest/nonce/create.go
```diff
@@ -81,6 +83,23 @@ func (o *Options) Create(ctx context.Context, options *rest.CreateOptions) *cobr
...
+			return err
+		}
+
+		nsname := tenantnamespace.GetNameForNamespace(o.clusterID.GetClusterID())
+		if exitingTenantNs != nil {
+			nsname = exitingTenantNs.Name
+		}
+		tenantNs := o.namespaceManager.GenNamespaceObj(o.clusterID.GetClusterID(), nsname)
+		opts.Printer.CheckErr(o.output(tenantNs))
+		return nil
```
fra98, line 100: I think is better to not use the `GetNameForNamespace()` function since we plan to make it private as can have unintended effects. My suggestion would be to modify the interface of the `GenNamespaceObj()` to accept an optional value `name *string`: if `name` is not set, it uses internally the `GetNameForNamespace()` function to generate the name., if set it uses the provided name (in this case, the tenant ns name if already exists).

Small note, I would move this logic directly to the output() function.

pkg/tenantNamespace/namespaceManager.go
```diff
@@ -123,6 +115,23 @@ func (nm *tenantNamespaceManager) CreateNamespace(ctx context.Context, cluster l
 	return ns, nil
 }
 
+// GenNamespaceObj returns a Tenant Namespace resource object given name and clusterid.
+func (nm *tenantNamespaceManager) GenNamespaceObj(cluster liqov1alpha1.ClusterID, name string) (ns *v1.Namespace) {
+	return &v1.Namespace{
```
fra98, line 120: I would call this function Forge to be more coherent with the terminology used in the rest of the codebase and on liqoctl.

```suggestion
// ForgeNamespace returns a Tenant Namespace resource object given name and remote clusterID.
func (nm *tenantNamespaceManager) ForgeNamespace(clusterID liqov1alpha1.ClusterID, name string) *v1.Namespace {
```

pkg/liqoctl/rest/nonce/create.go
```diff
@@ -132,10 +134,30 @@ func (o *Options) output(tenantNamespace string) error {
 		return fmt.Errorf("unsupported output format %q", opts.OutputFormat)
 	}
 
-	nonce := forge.Nonce(tenantNamespace)
+	// Before printing the namespace we need to check whether a tenant namespace for the given clusterid
+	// already exists. In that case forge a new namespace with the name of the existing one and place the
+	// nonce secret in the existing namespace.
+	exitingTenantNs, err := o.namespaceManager.GetNamespace(ctx, o.clusterID.GetClusterID())
+	if err != nil && !kerrors.IsNotFound(err) {
+		opts.Printer.CheckErr(fmt.Errorf("unable to get tenant namespace: %v", output.PrettyErr(err)))
```
fra98, line 142:
```suggestion
	if client.IgnoreNotFound(err) != nil {
```

pkg/liqoctl/rest/nonce/create.go
```diff
@@ -132,10 +134,30 @@ func (o *Options) output(tenantNamespace string) error {
...
+		return err
+	}
+
+	var nsname *string = nil
+	if exitingTenantNs != nil {
```
fra98, line 147:
```suggestion
	var nsname *string
```

PR #2636, opened 2024-07-19
[Test] E2E network CI

cheina97:
# Description

This PR adds the network E2E tests inside the CI pipeline.

### DIFF
--- a/test/e2e/cruise/network/network_test.go
+++ b/test/e2e/cruise/network/network_test.go
@@ -1,0 +1,222 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//nolint:gosec // Need to run liqoctl command
+package network
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	liqov1 "github.com/liqotech/liqo/apis/core/v1alpha1"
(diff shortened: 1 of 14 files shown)


test/e2e/pipeline/infra/cluster-api/setup.sh
```diff
@@ -100,6 +100,11 @@ do
   # Install metrics-server
   "${KUBECTL}" apply -f https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.6.4/components.yaml --kubeconfig "${TMPDIR}/kubeconfigs/liqo_kubeconf_${i}"
   "${KUBECTL}" -n kube-system patch deployment metrics-server --type json --patch '[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--kubelet-insecure-tls"}]' --kubeconfig "${TMPDIR}/kubeconfigs/liqo_kubeconf_${i}"
+
+  # Install kyverno for network tests
+  "${HELM}" repo add kyverno https://kyverno.github.io/kyverno/
+  "${HELM}" repo update
+  "${HELM}" install kyverno kyverno/kyverno -n kyverno --create-namespace --kubeconfig "${TMPDIR}/kubeconfigs/liqo_kubeconf_${i}"
```
aleoli, line 107: Are kyverno images pulled from docker? If that, can we pull them from the docker proxy?

aleoli (commented):
Additional question: does liqoctl test use images from docker? If so, I would add a flag to specify which registry to pull images from

test/e2e/pipeline/infra/cluster-api/setup.sh
cheina97: Of course

test/e2e/cruise/network/network_test.go
```diff
@@ -0,0 +1,221 @@
+// Copyright 2019-2024 The Liqo Authors
```
cheina97, line 1: Add an afterSuite to delete namespace if test fail. Also check for namespace deletion

test/e2e/cruise/network/network_test.go
```diff
@@ -0,0 +1,225 @@
...
+	for i := range testContext.Clusters {
+		if testContext.Clusters[i].Role == liqov1.ProviderRole {
+			providers = append(providers, testContext.Clusters[i].KubeconfigPath)
+		}
+	}
+	for i := range testContext.Clusters {
+		if testContext.Clusters[i].Role == liqov1.ConsumerRole {
+			consumer = testContext.Clusters[i].KubeconfigPath
+		}
+	}
```
fra98, line 77: I would break out of the loop after the first consumer is found to be more coeherent with the other tests and for a more predictable behaviour

test/e2e/cruise/network/network_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+			for i := range deploymentList.Items {
+				deployment := &deploymentList.Items[i]
+				if deployment.Status.ReadyReplicas == *deployment.Spec.Replicas {
+					return true, nil
+				}
+			}
+			return false, nil
+		}),
+	).To(Succeed())
```
fra98, line 224: Why not using an `Eventually()` with default timeut and interval?

fra98 (approved):

PR #2667, opened 2024-08-01
[Network] skip second connect in bidirectional peering

cheina97:
# Description

This PR disable a second connection establishment in case of a bidirectional peering.

### DIFF
--- a/pkg/liqoctl/network/handler.go
+++ b/pkg/liqoctl/network/handler.go
@@ -179,6 +179,20 @@ func (o *Options) RunConnect(ctx context.Context) error {
 		return err
 	}
 
+	// Check if the reverse Networking is already established on cluster 1
+	if established, err := cluster1.CheckAlreadyEstablishedForGwServer(ctx); err != nil {
+		return err
+	} else if established {
+		return nil
+	}
+
+	// Check if the reverse Networking is already established on cluster 2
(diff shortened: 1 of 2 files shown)


pkg/liqoctl/network/handler.go
```diff
@@ -63,6 +64,39 @@ func NewOptions(localFactory *factory.Factory) *Options {
...
+	} else if client.IgnoreNotFound(err) != nil {
+		return false, err
+	}
+
+	_, err = cluster2.GetGatewayClient(ctx, forge.DefaultGatewayServerName(cluster1.localClusterID))
+	if err == nil {
+		return true, nil
+	} else if client.IgnoreNotFound(err) != nil {
+		return false, err
+	}
```
aleoli, line 95: I would check for both client and server on both clusters

aleoli (approved):

fra98 (approved):
I think it's better to handle the check inside the liqoctl network commands (i.e., `network connect` and if necessary `network init`), and not on the peering command, so that also `network init/connect` works too.

pkg/liqoctl/network/handler.go
cheina97: It would become impossible to change network parameters using liqoctl peer without unpeer

fra98 (approved):

PR #2713, opened 2024-09-02
[Fix] Node IP

aleoli:
# Description

This pr fixes an issue updating the node IP on virtual kubelet deployment rolling update where the IP of the node was not updated to the new pod ip

### DIFF
--- a/pkg/virtualKubelet/liqoNodeProvider/reconciler.go
+++ b/pkg/virtualKubelet/liqoNodeProvider/reconciler.go
@@ -34,6 +34,11 @@ import (
 	"github.com/liqotech/liqo/pkg/utils/slice"
 )
 
+func (p *LiqoNodeProvider) reconcileNodeFromNode(_ watch.Event) error {
+	// enforce the node to be the same as the one we are managing
+	return p.updateNode()
+}
+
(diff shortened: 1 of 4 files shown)


pkg/virtualKubelet/liqoNodeProvider/reconciler.go
```diff
@@ -34,6 +34,21 @@ import (
...
+func (p *LiqoNodeProvider) reconcileNodeFromNode(event watch.Event) error {
+	var node v1.Node
+	unstruct, ok := event.Object.(*unstructured.Unstructured)
+	if !ok {
+		return errors.New("error in casting Node")
+	}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstruct.Object, &node); err != nil {
+		klog.Error(err)
+		return err
+	}
```
fra98, line 46: What is the purpose of de-unstructuring the node resource in this case?

fra98 (approved):

pkg/virtualKubelet/liqoNodeProvider/reconciler.go
aleoli: None I think 😅 I'm removing it

PR #2718, opened 2024-09-05
New `liqoctl info` command

claudiolor:
# Description

With the new version of Liqo the liqoctl status command has been removed because of some issues with the mantainabilty of the command, indeed it used to get the info from multiple CR and as with the new version of Liqo, a major refactoring of the resources has been done and complete rework on it was necessary.

The status command was a good way to check the:

- Health of the installation
- Current configuration
- Status of the peerings

This PR introduces the new `liqoctl info` command as a replacement for the status command. It not only shows the current status of the local Liqo instance, but provides some brief info about the active peerings. 
By default the output is presented in a human-readable form. However, to simplify automate retrieval of the data, via the `-o` option it is possible to format the output in `JSON` or `YAML`  and, via the `--get field.subfield` argument, each field of the reports can be individually retrieved.

E.g.
Get the output in YAML format
```
liqoctl info -o json
```

Get the podCIDR of the local Liqo instance:
```
liqoctl info --get network.podcidr
```

This PR add the boilerplate for the `liqoctl info peer <CLUSTED_ID>` command, **which is still not implemented**.

# How Has This Been Tested?

Test suite provided.

### DIFF
--- a/cmd/liqoctl/cmd/info.go
+++ b/cmd/liqoctl/cmd/info.go
@@ -1,0 +1,129 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package cmd
+
+import (
+	"context"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/liqotech/liqo/pkg/liqoctl/completion"
+	"github.com/liqotech/liqo/pkg/liqoctl/factory"
+	"github.com/liqotech/liqo/pkg/liqoctl/info"
+	"github.com/liqotech/liqo/pkg/liqoctl/info/localstatus"
+	"github.com/liqotech/liqo/pkg/liqoctl/output"
+	"github.com/liqotech/liqo/pkg/utils/args"
+)
+
+var outputFormat = args.NewEnum([]string{"json", "yaml"}, "")
+
+const liqoctlInfoLongHelp = `Show info about the current Liqo instance.
+
+Liqoctl provides a set of commands to verify the status of the Liqo control
+plane, its configuration, as well as the characteristics of the currently
+active peerings, and reports the outcome in human-readable or
+machine-readable format (either JSON or YAML).
+Additionally, via '--get', it allows to retrieve each single field of the reports
+using a query in dot notation (e.g. '--get field.subfield')
+
+This command shows information about the local cluster and checks the presence
+and the sanity of the Liqo namespace and the Liqo pods and some brief info about
+the active peerings and their status.
+
+Examples:
+  $ {{ .Executable }} info
+  $ {{ .Executable }} info --namespace liqo-system
+show the output in YAML format
+  $ {{ .Executable }} info -o yaml
+get a specific field
+  $ {{ .Executable }} info --get clusterid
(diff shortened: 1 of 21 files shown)


pkg/liqoctl/info/utils.go
```diff
@@ -0,0 +1,127 @@
...
+		o.Printer.Error.Printf("Unable to check if Liqo is installed: %v", err)
+		return err
+	} else if err != nil {
+		o.Printer.Error.Println("Liqo is not installed in the current cluster! \n\n" +
+			"You can install liqo via the 'liqoctl install' command.\n" +
+			"Check 'liqoctl install --help' for further information.")
+		return err
+	}
+
+	return nil
```
fra98, line 53: I know it works like this, but in general we handle this with a switch to be more clear and comprensible what is the if-else block that handles a NotFound error.

pkg/liqoctl/info/utils.go
```diff
@@ -0,0 +1,127 @@
...
+	if shortcut, ok := queryShortcuts[strings.ToLower(query)]; ok {
+		query = shortcut
+	}
+
+	query = strings.TrimPrefix(query, ".")
+	fields := strings.Split(query, ".")
+
+	currData, ok := data[fields[0]]
+	if !ok {
+		return "", fmt.Errorf("invalid query '%s': '%s' not found", query, fields[0])
```
fra98, line 71:
```suggestion
		return "", fmt.Errorf("invalid query %q: %q not found", query, fields[0])
```

pkg/liqoctl/info/handler.go
```diff
@@ -0,0 +1,100 @@
...
+		output, err = o.sPrintMachineReadable(checkers)
+	}
+
+	if err != nil {
+		o.Printer.Error.Println(err)
+	} else {
+		fmt.Println(output)
+	}
+
+	return err
```
fra98, line 94: Probably a three case switch would be more readable

pkg/liqoctl/info/localstatus/health.go
```diff
@@ -0,0 +1,131 @@
...
+// isPodReady checks the conditions of the pod to determine if all the containers in the Pod are up and running.
+func (l *HealthChecker) isPodReady(conditions []corev1.PodCondition) bool {
+	// Check the condition of the pod
+	for _, c := range conditions {
+		if c.Type == corev1.PodReady {
+			return c.Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
```
fra98, line 131: We can re-use this util function https://github.com/liqotech/liqo/blob/4992cda88235fd3c9da558bbfc348368c0463465/pkg/utils/pod/pod.go#L26

pkg/liqoctl/info/localstatus/local_info.go
```diff
@@ -0,0 +1,149 @@
...
+func (l *InstallationChecker) GetTitle() string {
+	return "Local installation info"
+}
+
+func (l *InstallationChecker) collectClusterLabels(ctrlDeployment *appsv1.Deployment) error {
+	version, err := getters.GetContainerImageVersion(ctrlDeployment.Spec.Template.Spec.Containers, ctrlManagerContainerName)
+	if err != nil {
+		return err
+	}
+	l.data.Version = version
```
fra98, line 116: Probably leftover as already managed in the dedicated function

cmd/liqoctl/cmd/info.go
```diff
@@ -0,0 +1,128 @@
...
+	if options.GetQuery != "" {
+		options.Verbose = true
+	}
+}
+
+func newPeerInfoCommand(ctx context.Context, f *factory.Factory, options *info.Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "peer",
+		Short:             "Show additional info about one or more specific peerings",
+		Long:              WithTemplate(""),
```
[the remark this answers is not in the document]
claudiolor: I left it empty on purpose. I'm planning to add this description in the next PR, once we add the implementation of the `info peer` command

pkg/liqoctl/info/handler.go
claudiolor: good idea

fra98 (approved):

PR #2720, opened 2024-09-06
[Feat] New In-Band Peering

aleoli:
# Description

This pr fixes the support for API server proxies and the in-band peering

### DIFF
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -382,51 +382,15 @@ func main() {
 			os.Exit(1)
 		}
 
-		go func() {
-			createWithRetry := func(f func() error) error {
-				maxRetries := 5
-				for {
-					err := f()
-					if err == nil {
-						break
-					}
+		if err := ipamips.EnforceAPIServerIPRemapping(ctx, uncachedClient, *liqoNamespace); err != nil {
+			klog.Errorf("Unable to enforce the API server IP remapping: %v", err)
+			os.Exit(1)
+		}
 
-					klog.V(3).Infof("Unable to create the IP mapping: %v, retrying...", err)
-
-					maxRetries--
-					if maxRetries == 0 {
-						klog.Errorf("Unable to create the IP mapping: %v", err)
-						os.Exit(1)
-					}
-
-					select {
-					case <-ctx.Done():
-						return nil
-					case <-time.After(10 * time.Second):
-					}
-				}
-
-				return nil
-			}
-
-			// enforce IP remapping for the API Server
-			err := createWithRetry(func() error {
-				return ipamips.EnforceAPIServerIPRemapping(ctx, uncachedClient, *liqoNamespace)
-			})
-			if err != nil {
-				klog.Errorf("Unable to enforce the API Server IP remapping: %v", err)
-				os.Exit(1)
-			}
-
-			// enforce IP remapping for the API Server Proxy
-			err = createWithRetry(func() error {
-				return ipamips.EnforceAPIServerProxyIPRemapping(ctx, uncachedClient, *liqoNamespace)
-			})
-			if err != nil {
-				klog.Errorf("Unable to enforce the API Server Proxy IP remapping: %v", err)
-				os.Exit(1)
-			}
-		}()
+		if err := ipamips.EnforceAPIServerProxyIPRemapping(ctx, uncachedClient, *liqoNamespace); err != nil {
+			klog.Errorf("Unable to enforce the API server proxy IP remapping: %v", err)
+			os.Exit(1)
+		}
 	}
 
 	// Configure the foreigncluster controller.
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -383,27 +383,48 @@ func main() {
 		}
 
 		go func() {
-			maxRetries := 5
-			for {
-				// enforce IP remapping for the API Server
-				err := ipamips.EnforceAPIServerIPRemapping(ctx, uncachedClient, *liqoNamespace)
-				if err == nil {
-					break
+			createWithRetry := func(f func() error) error {
+				maxRetries := 5
+				for {
+					err := f()
+					if err == nil {
+						break
+					}
+
+					klog.V(3).Infof("Unable to create the IP mapping: %v, retrying...", err)
+
+					maxRetries--
+					if maxRetries == 0 {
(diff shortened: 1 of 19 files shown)


docs/advanced/api-server-proxy.md
```diff
@@ -0,0 +1,23 @@
+# API Server Proxy
```
frisso, line 1: @aleoli This documentation looks nice, but it's not clear where this has to be set up (in the cluster that export the resources? IN the other?) I feel this should be clarified in the documentation.
In addition, it's not clear to me which is the IP address I would use, in my cluster, to connect to the remote API server.
Finally, this feature has nothing to do with the flag that enables an application to contact either its local or its remote API server, isn't it?

docs/features/peering.md
```diff
@@ -42,9 +42,9 @@ If this is not possible, refer to the advanced guide to learn how to perform the
 ```
 
 ```{warning}
-Starting from Liqo 1.0 the concept of *out-of-band* and *in-band* peering is deprecated.
 The user can adopt different peering approaches depending if it has contemporary access to both clusters or not, as described in the [dedicated page](/advanced/peering-strategies.md). 
 
-If the user can not expose the API server to the public, there is the possibility to leverage the Liqo networking to let the API server traffic flow into the cross-cluster VPN tunnel and use the *liqo-gateway* as a proxy for the API server.
-The procedure to achieve this will be explained in detail in the next releases.  
+If the user can not expose the API server to the public, there is the possibility to leverage the Liqo networking to let the API server traffic flow into the cross-cluster VPN tunnel by leveraging the [IP remapping](../advanced/external-ip-remapping.md) (of the `liqo-proxy` service) and the [API Server proxy](../advanced/api-server-proxy.md) features.
```
frisso, line 47:
```suggestion
If the user cannot expose the API server to the public, you can leverage the Liqo networking to let the API server traffic flow into the cross-cluster VPN tunnel by leveraging the [IP remapping](../advanced/external-ip-remapping.md) (of the `liqo-proxy` service) and the [API Server proxy](../advanced/api-server-proxy.md) features.
```

frisso: I'm not sure I would understand how to configure this feature. Would it be possible to provide an example?

aleoli: In a default installation, the only required step is to add a `--in-band` flag to liqoctl peer or liqoctl authenticate commands, as explained in the following sentence.

docs/advanced/api-server-proxy.md
aleoli: Users do not need to set up anything; this is done by default by Liqo at installation time. This page describes the functionality.

No, there are different things. This allows network interconnection, while the flag specifies which variable to inject in the pod to contact the local or remote API server. Of course, if it needs to contact the one in the other cluster, it will use the mapping done here.

docs/features/peering.md
```diff
@@ -42,9 +42,9 @@ If this is not possible, refer to the advanced guide to learn how to perform the
 ```
 
 ```{warning}
-Starting from Liqo 1.0 the concept of *out-of-band* and *in-band* peering is deprecated.
 The user can adopt different peering approaches depending if it has contemporary access to both clusters or not, as described in the [dedicated page](/advanced/peering-strategies.md). 
 
-If the user can not expose the API server to the public, there is the possibility to leverage the Liqo networking to let the API server traffic flow into the cross-cluster VPN tunnel and use the *liqo-gateway* as a proxy for the API server.
-The procedure to achieve this will be explained in detail in the next releases.  
+If you cannot expose the Kubernetes API Server to the public, you can leverage the Liqo networking to let the API server traffic flow into the cross-cluster VPN tunnel by leveraging the [IP remapping](../advanced/external-ip-remapping.md) (of the `liqo-proxy` service) and the [Kubernetes API Server proxy](../advanced/k8s-api-server-proxy.md) features.
```
frisso, line 47:
```suggestion
If you cannot expose the Kubernetes API Server to the public, you can leverage the Liqo networking to connect to the remote API server, and the related traffic will flow into the cross-cluster VPN tunnel by leveraging the [IP remapping](../advanced/external-ip-remapping.md) (of the `liqo-proxy` service) and the [Kubernetes API Server proxy](../advanced/k8s-api-server-proxy.md) features.
```

docs/features/peering.md
```diff
@@ -42,9 +42,9 @@ If this is not possible, refer to the advanced guide to learn how to perform the
...
+You can achieve this by setting the `--in-band` flag in the `liqoctl peer` or `liqoctl authenticate` commands.
```
frisso, line 48:
```suggestion
You can achieve this by setting the `--in-band` flag in the `liqoctl peer` or `liqoctl authenticate` commands, which automatically configure all the features needed for this mechanism to work (i.e., the API server proxy and the IP remapping).
```

docs/features/peering.md
```diff
@@ -42,9 +42,9 @@ If this is not possible, refer to the advanced guide to learn how to perform the
...
+Note that the networking module must be enabled.
```
frisso, line 49:
```suggestion
Note that, for this feature to work, the Liqo networking module must be enabled.
```

frisso (approved):

aleoli:
Thanks @frisso! I fixed them

cheina97 (approved):

cmd/liqo-controller-manager/main.go
```diff
@@ -378,28 +378,39 @@ func main() {
...
-				case <-ctx.Done():
-					return
-				case <-time.After(10 * time.Second):
-				}
-			}
+			// enforce IP remapping for the API Server Proxy
+			createWithRetry(func() error {
+				return ipamips.EnforceAPIServerProxyIPRemapping(ctx, uncachedClient, *liqoNamespace)
+			})
 		}()
```
cheina97, line 414:
```suggestion
		if err := ipamips.EnforceAPIServerIPRemapping(ctx, uncachedClient, *liqoNamespace); err != nil {
			klog.Errorf("Unable to enforce the API server IP remapping: %v", err)
			os.Exit(1)
		}

		if err := ipamips.EnforceAPIServerProxyIPRemapping(ctx, uncachedClient, *liqoNamespace); err != nil {
			klog.Errorf("Unable to enforce the API server proxy IP remapping: %v", err)
			os.Exit(1)
		}
```

I tried and it works. I think the problem was the webhook, but now we don't have a webhook on IP resources anymore.

fra98 (approved):

PR #2741, opened 2024-09-24
[Feat] New `liqoctl info peer` command

claudiolor:
# Description

This completes the work done in #2718 implementing the `liqoctl info peer` command, which provides detailed info about the current active peerings:

- Peer identity and role
- Status and info about each of the modules
- Amount of resources offloaded and shared
- Details about the currently configured virtual nodes

By default the output is presented in a human-readable form. However, to simplify automate retrieval of the data, via the `-o` option, it is possible to format the output in JSON or YAML and, via the `--get clusterID.field.subfield` argument, each field of the reports can be individually retrieved (when data about a single peer is returned, the clusterID can be omitted).

E.g.
Get info about all the currently active peerings in a human readable form:

```
liqoctl info peer
```

dump the result in json form:

```
liqoctl info peer -o json
```

### DIFF
--- a/pkg/liqoctl/info/handler.go
+++ b/pkg/liqoctl/info/handler.go
@@ -18,7 +18,10 @@ package info
 import (
 	"context"
 	"fmt"
+	"maps"
+	"slices"
 
+	liqov1beta1 "github.com/liqotech/liqo/apis/core/v1beta1"
 	"github.com/liqotech/liqo/pkg/liqoctl/factory"
 )
 
@@ -41,9 +44,10 @@ var LocalInfoQueryShortcuts = map[string]string{
 type Options struct {
 	*factory.Factory
 
-	Verbose  bool
-	Format   OutputFormat
-	GetQuery string
+	Verbose      bool
+	Format       OutputFormat
+	GetQuery     string
+	ClustersInfo map[liqov1beta1.ClusterID]*liqov1beta1.ForeignCluster
 }
 
 // NewOptions returns a new Options struct.
@@ -80,9 +84,11 @@ func (o *Options) RunInfo(ctx context.Context, checkers []Checker) error {
 		return nil
 	// If query specified try to retrieve the field from the output
 	case o.GetQuery != "":
-		output, err = o.sPrintField(o.GetQuery, checkers, LocalInfoQueryShortcuts)
+		data := o.collectDataFromCheckers(checkers)
+		output, err = o.sPrintField(o.GetQuery, data, LocalInfoQueryShortcuts)
 	default:
-		output, err = o.sPrintMachineReadable(checkers)
+		data := o.collectDataFromCheckers(checkers)
+		output, err = o.sPrintMachineReadable(data)
 	}
 
 	if err != nil {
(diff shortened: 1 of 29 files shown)


pkg/liqoctl/info/localstatus/peerings.go
```diff
@@ -79,19 +66,7 @@ func (p *PeeringChecker) Collect(ctx context.Context, options info.Options) {
...
-					if condition.Status == liqov1beta1.ConditionStatusEstablished || condition.Status == liqov1beta1.ConditionStatusReady {
-						moduleStatus[moduleName] = Healthy
-					} else {
-						moduleStatus[moduleName] = Unhealthy
-					}
-				}
-			} else {
-				moduleStatus[moduleName] = Disabled
-			}
+			moduleStatus[moduleName] = common.CheckModuleStatus(moduleInfo)
```
cheina97, line 69: The strings used from line 75 to 77 can be converted in constants?

pkg/liqoctl/info/handler.go
```diff
@@ -95,6 +101,65 @@ func (o *Options) RunInfo(ctx context.Context, checkers []Checker) error {
...
+	var output string
+	switch {
+	// If no format is specified, format and print a user-friendly output
+	case o.Format == "" && o.GetQuery == "":
+		for clusterID := range o.ClustersInfo {
+			for i := range checkers {
+				o.Printer.BoxSetTitle(checkers[i].GetTitle())
+				o.Printer.BoxPrintln(checkers[i].FormatForClusterID(clusterID, *o))
+			}
+			fmt.Println("---")
```
cheina97, line 132: Can we find another solution? It gives me the feeling that the output is buggy

cheina97 (approved):

pkg/liqoctl/info/peer/network.go
```diff
@@ -0,0 +1,190 @@
...
+
+// GatewayType represents the type of network gateway (Client or Server).
+type GatewayType string
+
+const (
+	// GatewayClientType indicates that the gateway acts as a client.
+	GatewayClientType GatewayType = "Client"
+	// GatewayServerType indicates that the gateway acts as a server.
+	GatewayServerType GatewayType = "Server"
+)
```
fra98, line 40: We can recycle consts already defined at https://github.com/liqotech/liqo/blob/master/pkg/consts/externalnetwork.go#L30-L33

pkg/liqoctl/info/peer/auth.go
```diff
@@ -0,0 +1,217 @@
...
+		// Check the status of the ResourceSlice
+		accepted := false
+		alerts := []string{}
+		for _, condition := range resSlice.Status.Conditions {
+			if condition.Status == authv1beta1.ResourceSliceConditionAccepted {
+				accepted = true
+			} else {
+				alerts = append(alerts, condition.Message)
+			}
+		}
```
fra98, line 187: With this implementation, if the first conditions is not accepted and the second is, it is shown as accepted.
I would either require that all conditions must be accepted or differentiate Authentication condition from Resource condition.

pkg/liqoctl/info/peer/auth.go
```diff
@@ -0,0 +1,217 @@
...
+
+		// To check who is the provider of the ResourceSlice we can verify whether the CR has been replicated from the cluster consumer
+		action := ConsumingAction
+		if _, isProvider := resSlice.ObjectMeta.Labels[consts.ReplicationStatusLabel]; isProvider {
+			action = ProvidingAction
+		}
```
fra98, line 193: Can you also check if the value is set to "false" (case insensitive)?
At the moment it does not impact as we never set it to false, but I would change it to be more resilient in case of future changes.

pkg/liqoctl/info/peer/offloading.go
```diff
@@ -0,0 +1,163 @@
...
+		// Collect the status of the offloading module
+		oc.collectStatusInfo(clusterID, options.ClustersInfo, &offloadingState)
+
+		labelFilter, _ := labels.NewRequirement(consts.RemoteClusterID, selection.Equals, []string{string(clusterID)})
+
+		// Get the VirtualNode resources pointing to the given remote clusterID
+		var virtualNodeList offloadingv1beta1.VirtualNodeList
+		if err := options.CRClient.List(ctx, &virtualNodeList, &client.ListOptions{
+			LabelSelector: labels.NewSelector().Add(*labelFilter),
+		}); err != nil {
```
fra98, line 74: We can use `ListVirtualNodesByClusterID()` from getters pkg/utils/getters.

pkg/liqoctl/info/peer/auth.go
```diff
@@ -0,0 +1,217 @@
...
+				ac.AddCollectionError(fmt.Errorf("unable to get API server address of cluster %q: %w", clusterID, err))
+			}
+		}
+
+		labelFilter, _ := labels.NewRequirement(consts.RemoteClusterID, selection.Equals, []string{string(clusterID)})
+		// Get the ResourceSlices related to the given remote clusterID
+		var resourceSliceList authv1beta1.ResourceSliceList
+		if err := options.CRClient.List(ctx, &resourceSliceList, &client.ListOptions{
+			LabelSelector: labels.NewSelector().Add(*labelFilter),
+		}); err != nil {
```
fra98, line 89: We can use `ListResourceSlicesByLabel()` from getters pkg/utils/getters and pass the label selector, but at this point I would directly add a utility function `ListResourceSlicesByClusterID()` that use internally `ListResourceSlicesByLabel()`, like we did with `ListVirtualNodesByClusterID()`.

pkg/liqoctl/info/peer/offloading.go
```diff
@@ -0,0 +1,163 @@
...
+	for i := range virtualNodes {
+		vNode := &virtualNodes[i]
+
+		// Check the status of the virtual node
+		status := common.ModuleUnhealthy
+		for _, condition := range vNode.Status.Conditions {
+			if condition.Status == offloadingv1beta1.RunningConditionStatusType {
+				status = common.ModuleHealthy
+			}
+		}
```
fra98, line 150: Also here i would show Healthy only if all conditions are Healthy.
It could be necessary to consider also `None` status as healthy, as for example when you create a VirtualNode with the k8s Node disabled.

pkg/liqoctl/info/peer/offloading.go
```diff
@@ -0,0 +1,163 @@
...
+
+		// Get the ResourceSlice assigned to VirtualNode secret
+		resourceSlice := vNode.ObjectMeta.Labels[consts.ResourceSliceNameLabelKey]
+
+		offloadingState.VirtualNodes = append(offloadingState.VirtualNodes, VirtualNodeStatus{
+			Name:          vNode.Name,
+			Status:        status,
+			ResourceSlice: resourceSlice,
+			Secret:        vNode.Spec.KubeconfigSecretRef.Name,
```
fra98, line 159: Possible nil panic, I would add a check

pkg/liqoctl/info/peer/network.go
claudiolor: I created new constants for consistency with `Provider` and `Consumer`, which have the first letter capitalized

fra98 (approved):

fra98 (approved):

PR #2746, opened 2024-09-25
[Feat] support cluster ip in gateway services

aleoli:
# Description

This pr adds the cluster to the list of the gateways-supported services. It can be helpful for particular use cases where service addresses are discoverable from the other cluster.


fra98:
This is a niche feature that can be confusing for most people, plus using the ClusterIP can be tempting and lead to misunderstandings. 
Should we document this and advise the user not to use it unless you really know what you are doing?

cheina97 (approved):

Issue #2732, opened 2024-09-16
[Feature] Improve GatewayServer declerative setup in 1.0


EladDolev:
**Is your feature request related to a problem? Please describe.**

With current implementation, creating the `GatewayServer` resource would generate a `Secret` resource with a key pair **unless** a secret already exists.
That makes it very hard for declarative setup that uses `SealedSecret` for example for creating this `Secret` as the sealed-secrets controller would refuse to touch the `Secret` if it already exists, hence there's a race condition here

**Describe the solution you'd like**

A way of preventing automatically creating the `Secret` resource with a key pair when creating the `GatewayServer` resource

**Describe alternatives you've considered**

I have no alternatives other than manual intervention or manual setup which kinda beats the purpose of declarative setup

claudiolor:
Fixed with #2747

PR #2747, opened 2024-09-30
[Feat] Gateway external secrets

aleoli:
# Description

This pr makes it possible to specify an existing secret or a secret that will exist for wireguard keys


fra98:
Fixes #2732

fra98 (approved):

pkg/liqo-controller-manager/networking/external-network/server-operator/server_controller.go
```diff
@@ -168,6 +178,7 @@ func (r *ServerReconciler) EnsureGatewayServer(ctx context.Context, gwServer *ne
 			Namespace:  gwServer.Namespace,
 			GatewayUID: string(gwServer.UID),
 			ClusterID:  remoteClusterID,
+			SecretName: gwServer.Spec.SecretRef.Name,
```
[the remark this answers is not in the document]
aleoli: No, secret ref is not a pointer because of go-templating. It will be an empty string

pkg/liqo-controller-manager/networking/external-network/wireguard/wggatewayclient_controller.go
```diff
@@ -197,32 +230,47 @@ func (r *WgGatewayClientReconciler) mutateFnWgClientDeployment(deployment *appsv
 	// Forge spec
 	deployment.Spec = wgClient.Spec.Deployment.Spec
 
+	if wgClient.Status.SecretRef != nil {
```
claudiolor, line 233: With this logic, doesn't a secret with the proper labels take over the secret configured in the spec? Shoudn't the latter have the priority?

pkg/liqo-controller-manager/networking/external-network/wireguard/utils.go
```diff
@@ -143,6 +143,27 @@ func ensureKeysSecret(ctx context.Context, cl client.Client, wgObj metav1.Object
...
 
+func checkExistingKeysSecret(ctx context.Context, cl client.Client, secretName, namespace string) error {
+	var s corev1.Secret
+	if err := cl.Get(ctx, types.NamespacedName{Name: secretName, Namespace: namespace}, &s); err != nil {
+		return err
+	}
+
+	// check labels
+	if s.Labels == nil {
+		return fmt.Errorf("missing labels in secret %q", secretName)
```
claudiolor, line 154: I would make the error more talkative indicating which labels are expected in the secret, so something like the following:

```suggestion
		return fmt.Errorf("mandatory labels %q: "true" and %q are missing in secret %q", consts.GatewayResourceLabel, consts.RemoteClusterID, secretName)
```

PR #2759, opened 2024-10-04
[Feat] Webhook HA

aleoli:
# Description

This pr makes it possible to set multiple replicas for the liqo webhook deployment

### DIFF
--- a/pkg/leaderelection/labeler.go
+++ b/pkg/leaderelection/labeler.go
@@ -1,0 +1,59 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


fra98 (approved):

pkg/leaderelection/labeler.go
```diff
@@ -0,0 +1,57 @@
...
+			maxRetries--
+			return true
+		}, func() error {
+			if err := handleLeaderLabelWithClient(ctx, cl, info); err != nil {
+				klog.Error(err, "retrying...")
+				return err
+			}
+			return nil
+		}); err != nil {
+			klog.Error(err)
```
claudiolor, line 54: Wondering whether it is worth panicking when the leader is not able to add itself the label. As in that case we will be in a "gray" condition, where the leader is running but it actualy doesn't handle the incoming requests

PR #2767, opened 2024-10-07
[Feat] Gateway HA

cheina97:
# Description

This PR adds support for HA (active/passive) in liqo gateways. The "gateway" container is now the only responsible for the lease acquisition and can start the "geneve" and "wireguard" containers using IPC.

The main changes are:
- added a "label" to target the active gateway pod (useful for gateway server services and to create internal network tunnels)
- added a "runnable" to add and remove the "label" from the active pod when it becomes the leader (using k8s lease)
- added an IPC (inter process communication) based on UNIX sockets shared between gateway containers

### DIFF
--- a/cmd/gateway/geneve/main.go
+++ b/cmd/gateway/geneve/main.go
@@ -22,7 +22,6 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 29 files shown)


pkg/gateway/concurrent/const.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package concurrent
+
+// UnixSocketPath is the path of the Unix socket.
+const unixSocketPath string = "/ipc/leader.sock"
+
+// ContainerNames is the list of container names.
+var containerNames = []string{"wireguard", "geneve"}
```
fra98, line 21: Should we make these configurable and not hardcoded? So that we better support different technologies and to ease eventual future changes

cmd/gateway/wireguard/main.go
```diff
@@ -178,6 +167,15 @@ func run(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("unable to register prometheus collector: %w", err)
 	}
 
+	runnable, err := concurrent.NewRunnableGuest("wireguard")
```
fra98, line 170: Can we make it a costant for better reusability?

cmd/gateway/geneve/main.go
```diff
@@ -120,6 +109,15 @@ func run(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("unable to setup internalnode reconciler: %w", err)
 	}
 
+	runnable, err := concurrent.NewRunnableGuest("geneve")
```
fra98, line 112: same here

fra98 (approved):

PR #2770, opened 2024-10-08
[Feat] VirtualKubelet HA

aleoli:
# Description

This pr enables HA active/passive virtual kubelet deployment


aleoli:
Waiting for pr #2759 to be merged

cmd/virtual-kubelet/root/root.go
```diff
@@ -102,6 +102,28 @@ func runRootCommand(ctx context.Context, c *Opts) error {
...
+			Namespace: c.TenantNamespace,
+		},
+		LeaderElectorName: c.NodeName,
+		LeaseDuration:     c.VirtualKubeletLeaseLeaseDuration,
+		RenewDeadline:     c.VirtualKubeletLeaseRenewDeadline,
+		RetryPeriod:       c.VirtualKubeletLeaseRetryPeriod,
+	}); err != nil {
+		return err
+	} else if !leader {
+		klog.Info("This virtual-kubelet is not the leader")
```
claudiolor, line 123: Isn't here more correct logging an error? If we exit the blocking function and the virtual kubelet is not the leader maybe something went wrong, isn't it?

claudiolor (approved):

fra98 (approved):

PR #2783, opened 2024-10-22
fix: network status in ForeignCluster resource

claudiolor:
# Description

This patch fixes the network status in the ForeignCluster resource. Previously, when the connection or the configuration were missing no conditions were shown, so who looked at the ForeignCluster status could think that everything is correctly configured.
Moreover, this patch makes the status check more robust, collecting all the errors first and only once we are sure that the module is enabled (there is at least a network resource present), it writes the errors in the status.

### DIFF
--- a/pkg/liqo-controller-manager/core/foreigncluster-controller/status.go
+++ b/pkg/liqo-controller-manager/core/foreigncluster-controller/status.go
@@ -35,6 +35,12 @@ import (
 	"github.com/liqotech/liqo/pkg/utils/pod"
 )
 
+type statusException struct {
+	liqov1beta1.ConditionStatusType
+	Reason  string
+	Message string
+}
+
 func (r *ForeignClusterReconciler) clearStatusExceptConditions(foreignCluster *liqov1beta1.ForeignCluster) {
 	foreignCluster.Status = liqov1beta1.ForeignClusterStatus{
 		Role: liqov1beta1.UnknownRole,
@@ -56,13 +62,8 @@ func (r *ForeignClusterReconciler) clearStatusExceptConditions(foreignCluster *l
 	}
 }
 
-func (r *ForeignClusterReconciler) handleNetworkingModuleStatus(ctx context.Context,
-	fc *liqov1beta1.ForeignCluster, moduleEnabled bool) error {
-	if !moduleEnabled {
-		clearModule(&fc.Status.Modules.Networking)
-		return nil
-	}
-
+func (r *ForeignClusterReconciler) handleConnectionStatus(ctx context.Context,
+	fc *liqov1beta1.ForeignCluster, statusExceptions map[liqov1beta1.ConditionType]statusException) error {
 	clusterID := fc.Spec.ClusterID
 
 	connection, err := getters.GetConnectionByClusterID(ctx, r.Client, string(clusterID))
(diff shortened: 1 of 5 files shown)


pkg/liqo-controller-manager/core/foreigncluster-controller/status.go
```diff
@@ -63,38 +63,43 @@ func (r *ForeignClusterReconciler) handleNetworkingModuleStatus(ctx context.Cont
 	}
 
 	clusterID := fc.Spec.ClusterID
+	type StatusException struct {
+		liqov1beta1.ConditionStatusType
+		Reason  string
+		Message string
+	}
```
cheina97, line 70: Can you declare it outside from the function?

pkg/liqo-controller-manager/core/foreigncluster-controller/status.go
```diff
@@ -63,38 +63,43 @@ func (r *ForeignClusterReconciler) handleNetworkingModuleStatus(ctx context.Cont
 	}
 
 	clusterID := fc.Spec.ClusterID
+	type StatusException struct {
```
cheina97, line 66: Can you split this function in more functions?

pkg/liqo-controller-manager/core/foreigncluster-controller/status.go
```diff
@@ -126,15 +131,50 @@ func (r *ForeignClusterReconciler) handleNetworkingModuleStatus(ctx context.Cont
...
+		}
+	}
+
 	gwClient, err := getters.GetGatewayClientByClusterID(ctx, r.Client, clusterID)
 	switch {
 	case errors.IsNotFound(err):
 		klog.V(6).Infof("GatewayClient resource not found for ForeignCluster %q", clusterID)
 		fcutils.DeleteModuleCondition(&fc.Status.Modules.Networking, liqov1beta1.NetworkGatewayClientStatusCondition)
+		// Do not write any exception, as if both gateways are missing we write a single condition telling that a gateway is missing.
+		// If instead, only the gateway client is missing, then this cluster acts as server.
```
cheina97, line 171: I would suggest to simplify the workflow. You can get the gateway client and server and add the condition only if the 2 arrors are NotFound

cheina97 (approved):

pkg/liqo-controller-manager/core/foreigncluster-controller/status.go
cheina97: And maybe you don't need to export it

pkg/liqo-controller-manager/core/foreigncluster-controller/status.go
```diff
@@ -89,15 +95,32 @@ func (r *ForeignClusterReconciler) handleNetworkingModuleStatus(ctx context.Cont
...
+func (r *ForeignClusterReconciler) handleGatewaysStatus(ctx context.Context,
+	fc *liqov1beta1.ForeignCluster, statusExceptions map[liqov1beta1.ConditionType]statusException) error {
+	clusterID := fc.Spec.ClusterID
+
+	gwServer, errServer := getters.GetGatewayServerByClusterID(ctx, r.Client, clusterID)
+	gwClient, errClient := getters.GetGatewayClientByClusterID(ctx, r.Client, clusterID)
+
+	if errors.IsNotFound(errServer) && errors.IsNotFound(errClient) {
+		klog.V(6).Infof("Both GatewayServer and GatewayClient resources not found for ForeignCluster %q", clusterID)
+		statusExceptions[liqov1beta1.NetworkGatewayClientStatusCondition] = statusException{
```
fra98, line 110: Defaulting to GatewayClient can be misleading if I expect to have a GatewayServer. I would use a new generic status condition or a better solution if any

fra98 (approved):

PR #2787, opened 2024-10-24
[IPAM] API and first stub

aleoli:
# Description

This pr includes the new IPAM gRPC API


pkg/ipam/ipam.proto
```diff
@@ -1,76 +1,62 @@
...
 }
 
-message SetSubnetsPerClusterRequest {
-    string remappedPodCIDR = 1;
-    string remappedExternalCIDR = 2;
-    string clusterID = 3;
+message NetworkAvailableResponse {
+    ResponseResult result = 1;
+    bool available = 2;
 }
```
fra98, line 62: isn't the `available` boolean redundant if there is already one in the `ResponseResult`? I get that they can have different meanings, but in this case it seems redundant and not very clear

fra98 (approved):

pkg/ipam/ipam.proto
aleoli: This function checks if the network is available or not; in the resource result, we store the errors. We can have no errors and both available or not available networks, they are different things

fra98: My question is more about the first boolean in the ResponseResult. What does it tell? If I am understanding correctly, I would just check if there is an error first, and then check the `available` bool, thus always discarding the first `success` bool

aleoli: The first success is like to check `err != nil`, while this one tells you if the network is available or not

PR #2791, opened 2024-10-25
[Feat] liqoctl gateway template check

cheina97:
# Description

This PR adds some checks in liqoctl (at peering time) about the gateway template used.
This checks are:
- template existence
- in case `--server-service-nodeport` flag has been specified, it checks if the template selected support static nodeports
- in case `--server-service-loadbalancerip` flag has been specified, it checks if the template selected support static load balancer ips

### DIFF
--- a/cmd/liqoctl/cmd/network.go
+++ b/cmd/liqoctl/cmd/network.go
@@ -65,6 +65,7 @@ func newNetworkCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
 
 	cmd.PersistentFlags().DurationVar(&options.Timeout, "timeout", 120*time.Second, "Timeout for completion")
 	cmd.PersistentFlags().BoolVar(&options.Wait, "wait", false, "Wait for completion")
+	cmd.PersistentFlags().BoolVar(&options.SkipValidation, "skip-validation", false, "Skip the validation")
 
 	options.LocalFactory.AddFlags(cmd.PersistentFlags(), cmd.RegisterFlagCompletionFunc)
 	options.RemoteFactory.AddFlags(cmd.PersistentFlags(), cmd.RegisterFlagCompletionFunc)
@@ -139,12 +140,7 @@ func newNetworkConnectCommand(ctx context.Context, options *network.Options) *co
 		Args:  cobra.NoArgs,
 
 		Run: func(_ *cobra.Command, _ []string) {
-			err := options.RunConnect(ctx)
-			if err != nil {
-				options.LocalFactory.Printer.CheckErr(
-					fmt.Errorf("`network connect` failed (error: %w). Issue `network disconnect` to cleanup the environment", err))
-			}
-			output.ExitOnErr(err)
+			output.ExitOnErr(options.RunConnect(ctx))
 		},
 	}
 
@@ -156,15 +152,15 @@ func newNetworkConnectCommand(ctx context.Context, options *network.Options) *co
 	cmd.Flags().StringVar(&options.ServerTemplateNamespace, "server-template-namespace", "",
 		"Namespace of the Gateway Server template")
 	cmd.Flags().Var(options.ServerServiceType, "server-service-type",
-		fmt.Sprintf("Service type of the Gateway Server. Default: %s."+
+		fmt.Sprintf("Service type of the Gateway Server service. Default: %s."+
 			" Note: use ClusterIP only if you know what you are doing and you have a proper network configuration",
 			forge.DefaultGwServerServiceType))
-	cmd.Flags().Int32Var(&options.ServerPort, "server-port", forge.DefaultGwServerPort,
-		fmt.Sprintf("Port of the Gateway Server. Default: %d", forge.DefaultGwServerPort))
-	cmd.Flags().Int32Var(&options.ServerNodePort, "node-port", 0,
-		"Force the NodePort of the Gateway Server. Leave empty to let Kubernetes allocate a random NodePort")
-	cmd.Flags().StringVar(&options.ServerLoadBalancerIP, "load-balancer-ip", "",
-		"Force LoadBalancer IP of the Gateway Server. Leave empty to use the one provided by the LoadBalancer provider")
+	cmd.Flags().Int32Var(&options.ServerServicePort, "server-service-port", forge.DefaultGwServerPort,
+		fmt.Sprintf("Port of the Gateway Server service. Default: %d", forge.DefaultGwServerPort))
+	cmd.Flags().Int32Var(&options.ServerServiceNodePort, "server-service-nodeport", 0,
+		"Force the NodePort of the Gateway Server service. Leave empty to let Kubernetes allocate a random NodePort")
+	cmd.Flags().StringVar(&options.ServerServiceLoadBalancerIP, "server-service-loadbalancerip", "",
+		"Force LoadBalancer IP of the Gateway Server service. Leave empty to use the one provided by the LoadBalancer provider")
 
 	// Client flags
 	cmd.Flags().StringVar(&options.ClientGatewayType, "client-type", forge.DefaultGwClientType,
--- a/pkg/liqoctl/network/cluster.go
+++ b/pkg/liqoctl/network/cluster.go
@@ -21,6 +21,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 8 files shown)


deployments/liqo/templates/liqo-wireguard-gateway-server-template.yaml
```diff
@@ -36,6 +36,7 @@ spec:
           - port: "{{"{{ .Spec.Endpoint.Port }}"}}"
             protocol: UDP
             targetPort: "{{"{{ .Spec.Endpoint.Port }}"}}"
+
```
aleoli, line 39: We should align this file to the eks one also on what concerns empty lines

pkg/liqoctl/network/handler.go
```diff
@@ -165,6 +165,16 @@ func (o *Options) RunConnect(ctx context.Context) error {
...
 
+	// Check if the Templates exists and is valid on cluster 1
+	if err := cluster1.CheckTemplateGwClient(ctx, o); err != nil {
+		return err
+	}
+
+	// Check if the Templates exists and is valid on cluster 2
+	if err := cluster2.CheckTemplateGwServer(ctx, o); err != nil {
+		return err
+	}
```
aleoli, line 176: We should add a flag to skip these checks

pkg/utils/maps/maps.go
```diff
@@ -197,3 +197,24 @@ func SerializeMap(m map[string]string) string {
 func DeSerializeCache(s string) []string {
 	return strings.Split(s, ",")
 }
+
+// GetNestedField returns the nested field of a map.
+// Example: GetNestedField(map[string]interface{}{"a": map[string]interface{}{"b": "c"}}, "a.b") returns "c".
+func GetNestedField(m map[string]interface{}, path string) (interface{}, error) {
```
aleoli, line 203: In modern go, I think `any` type is identical to `interface{}` and more idiomatic. Not entirely sure about this

pkg/liqoctl/network/cluster.go
```diff
@@ -194,6 +197,123 @@ func (c *Cluster) SetupConfiguration(ctx context.Context, conf *networkingv1beta
...
+	}
+
+	port, ok := templateServicePortsSlice[0].(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("unable to cast %s to map[string]interface{}", path)
+	}
+
+	_, err = maps.GetNestedField(port, "nodePort")
+	if err != nil {
+		return fmt.Errorf("unable to get spec.template.spec.service.spec.ports[0].nodePort int the server template, " +
```
claudiolor, line 297: Once we add a flag to skip this check (as suggested by aleoli), we can add a suggestion of how to skip this check. Morevoer, we might also add a documentation page explaining the concept of template and how to add a field in the template, so that we can add a link to it in the error message

pkg/liqoctl/network/cluster.go
```diff
@@ -194,6 +197,123 @@ func (c *Cluster) SetupConfiguration(ctx context.Context, conf *networkingv1beta
...
+func (c *Cluster) CheckTemplateGwServer(ctx context.Context, opts *Options) error {
+	templateName := opts.ServerTemplateName
+	templateNamespace := opts.ServerTemplateNamespace
+	templateGVR := opts.ServerGatewayType
+
+	s := c.local.Printer.StartSpinner(fmt.Sprintf("Checking gateway client template \"%s/%s\"", templateName, templateNamespace))
+
+	template, err := c.checkTemplate(ctx, templateName, templateNamespace, templateGVR)
+	if err != nil {
+		s.Fail(fmt.Sprintf("An error occurred while checking gateway template \"%s/%s\": %v", templateName, templateNamespace, output.PrettyErr(err)))
```
fra98, line 228:
```suggestion
		s.Fail(fmt.Sprintf("An error occurred while checking gateway server template \"%s/%s\": %v", templateName, templateNamespace, output.PrettyErr(err)))
```

pkg/liqoctl/network/cluster.go
```diff
@@ -194,6 +197,123 @@ func (c *Cluster) SetupConfiguration(ctx context.Context, conf *networkingv1beta
...
+		return err
+	}
+
+	if err := c.checkTemplateServerService(template, opts); err != nil {
+		s.Fail(fmt.Sprintf("An error occurred while checking gateway template \"%s/%s\": %v", templateName, templateNamespace, output.PrettyErr(err)))
```
fra98, line 233:
```suggestion
		s.Fail(fmt.Sprintf("An error occurred while checking gateway server template \"%s/%s\": %v", templateName, templateNamespace, output.PrettyErr(err)))
```

pkg/liqoctl/network/cluster.go
```diff
@@ -194,6 +197,123 @@ func (c *Cluster) SetupConfiguration(ctx context.Context, conf *networkingv1beta
...
+		return err
+	}
+
+	s.Success(fmt.Sprintf("Gateway client template \"%s/%s\" correctly checked", templateName, templateNamespace))
```
fra98, line 237:
```suggestion
	s.Success(fmt.Sprintf("Gateway server template \"%s/%s\" correctly checked", templateName, templateNamespace))
```

pkg/liqoctl/network/cluster.go
```diff
@@ -194,6 +197,123 @@ func (c *Cluster) SetupConfiguration(ctx context.Context, conf *networkingv1beta
...
+
+	port, ok := templateServicePortsSlice[0].(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("unable to cast %s to map[string]interface{}", path)
+	}
+
+	_, err = maps.GetNestedField(port, "nodePort")
+	if err != nil {
+		return fmt.Errorf("unable to get spec.template.spec.service.spec.ports[0].nodePort int the server template, " +
+			"since you specified the flag \"--server-service-nodeport\" you need to add the \"nodePort\" field in the template")
```
fra98, line 298: Probably small detail but can you use the %q directives when possible? Just for coeherence and better autosuggestions

cmd/liqoctl/cmd/network.go
```diff
@@ -153,18 +149,18 @@ func newNetworkConnectCommand(ctx context.Context, options *network.Options) *co
 		"Type of Gateway Server. Leave empty to use default Liqo implementation of WireGuard")
 	cmd.Flags().StringVar(&options.ServerTemplateName, "server-template-name", forge.DefaultGwServerTemplateName,
 		"Name of the Gateway Server template")
-	cmd.Flags().StringVar(&options.ServerTemplateNamespace, "server-template-namespace", "",
+	cmd.Flags().StringVar(&options.ServerTemplateNamespace, "server-template-namespace", forge.DefaultGwServerTemplateNamespace,
```
fra98, line 152: This was originally left empty as it is defaulted after with provided liqo namespace (to support cases where the user install liqo in a custom namespace).
It is handled here: https://github.com/liqotech/liqo/blob/d0486815e9aa5a06f03f4dc7553089eadd294d8f/pkg/liqoctl/network/handler.go#L322

Same for the ClientTemplateNamespace

pkg/liqoctl/network/cluster.go
```diff
@@ -194,6 +197,123 @@ func (c *Cluster) SetupConfiguration(ctx context.Context, conf *networkingv1beta
 	return nil
 }
 
+// CheckTemplateGwClient checks if the GatewayClient template is correctly set up.
+func (c *Cluster) CheckTemplateGwClient(ctx context.Context, opts *Options) error {
+	templateName := opts.ServerTemplateName
+	templateNamespace := opts.ServerTemplateNamespace
+	templateGVR := opts.ServerGatewayType
+
```
fra98, line 205:
```suggestion
	templateName := opts.ClientTemplateName
	templateNamespace := opts.ClientTemplateNamespace
	templateGVR := opts.ClientGatewayType
```

pkg/liqoctl/network/cluster.go
cheina97: I know, but I think is better to not use %q since it is not a variable

fra98: Yes you are right. I referenced the wrong lines. I meant the ones with `\"%s/%s\"`

cheina97: s := c.local.Printer.StartSpinner(fmt.Sprintf("Checking gateway client template %q", fmt.Sprintf("%s/%s", templateNamespace, templateName)))

That should be a solution, but I think is too much

cmd/liqoctl/cmd/network.go
cheina97: Understood, so it must be moved in the outer function

pkg/liqoctl/network/cluster.go
fra98: Yes it's a minor thing, we can overlook it.

fra98 (approved):

cmd/liqoctl/cmd/network.go
fra98: Correct

PR #2794, opened 2024-10-29
[Fix] argocd install

aleoli:
# Description

This pr fixes the sync of liqo with argocd by removing the pre/post install/update helm hooks not correctly handled by argocd


pkg/consts/controllers.go
```diff
@@ -51,6 +51,7 @@ const (
 	CtrlPodInternalNet         = "pod_internalnet"
 	CtrlPublicKey              = "publickey"
 	CtrlRouteConfiguration     = "routeconfiguration"
+	CtrlSecretWebhook          = "secret_webhook"
```
fra98, line 54: Can you move this to the Core section?

aleoli: Yes, sorry, i focused on the alphabetical order and missed the group 😅

PR #2798, opened 2024-10-30
[IPAM] Sync routine

fra98:
# Description

This PR implements the sync routine for the internal IPAM state.

### DIFF
--- a/cmd/ipam/main.go
+++ b/cmd/ipam/main.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/health"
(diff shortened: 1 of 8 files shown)


cmd/ipam/main.go
```diff
@@ -102,14 +105,14 @@ func run(cmd *cobra.Command, _ []string) error {
 
 	// Get the rest config.
 	cfg := restcfg.SetRateLimiter(ctrl.GetConfigOrDie())
-	options.Config = cfg
```
aleoli, line 105: Is it ok here not to pass the config to options?

aleoli (approved):

cmd/ipam/main.go
fra98: Yes it wasn't used, I also removed it from the struct itself

PR #2806, opened 2024-11-06
[IPAM]: added unit tests for sync routine

fra98:
# Description

This PR adds unit tests for sync routine

### DIFF
--- a/pkg/ipam/sync_test.go
+++ b/pkg/ipam/sync_test.go
@@ -1,0 +1,175 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ipam
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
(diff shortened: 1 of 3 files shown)


pkg/ipam/sync_test.go
```diff
@@ -0,0 +1,175 @@
...
+			})
+
+			It("should remove networks from cache if they are not present in the cluster", func() {
+				// Run sync
+				Expect(fakeIpamServer.syncNetworks(ctx, expiredThreshold)).To(Succeed())
+
+				// Check the cache
+				Expect(fakeIpamServer.cacheNetworks).To(HaveKey("10.0.0.0/16"))    // network in cluster and cache
+				Expect(fakeIpamServer.cacheNetworks).To(HaveKey("10.1.0.0/16"))    // network in cluster and cache before expired threshold
+				Expect(fakeIpamServer.cacheNetworks).NotTo(HaveKey("10.2.0.0/16")) // network in cluster but not in cache
```
aleoli, line 131: Shouldn't it add the network when available in the cluster since the cluster is the source of truth?

pkg/ipam/sync_test.go
```diff
@@ -0,0 +1,175 @@
...
+				// Run sync
+				Expect(fakeIpamServer.syncIPs(ctx, expiredThreshold)).To(Succeed())
+
+				// Check the cache
+				Expect(fakeIpamServer.cacheIPs).To(HaveKey(
+					ipCidr{ip: "10.0.0.0", cidr: "10.0.0.0/24"}.String())) // IP in cluster and cache
+				Expect(fakeIpamServer.cacheIPs).To(HaveKey(
+					ipCidr{ip: "10.0.0.1", cidr: "10.0.0.0/24"}.String())) // IP in cluster and cache before expired threshold
+				Expect(fakeIpamServer.cacheIPs).NotTo(HaveKey(
+					ipCidr{ip: "10.0.0.2", cidr: "10.0.0.0/24"}.String())) // IP in cluster but not in cache
```
aleoli, line 167: here again

pkg/ipam/sync_test.go
```diff
@@ -0,0 +1,175 @@
...
+		syncFrequency = 3 * time.Second
+		testNamespace = "test"
+	)
+
+	var (
+		ctx               context.Context
+		fakeClientBuilder *fake.ClientBuilder
+		now               time.Time
+		expiredThreshold  time.Time
+		expired           time.Time
```
aleoli, line 43: Expired looks misleading to me... isn't it a threshold under which you are not considering changes in the local state since the k8s client can be unaligned with the actual k8s state?

And cache is misleading too; it is the internal (or local) state of the ipam, isn't it?

aleoli (approved):

PR #2817, opened 2024-11-18
Nftables monitor disable

cheina97:
# Description

This PR introduces a flag to disable the nftables monitoring routine in liqo-fabric.
This is useful because the monitoring routine uses many resources in some cases (like k3s).
It also disables by default the routine in liqoctl k3s provider.


frisso:
@cheina97 Thanks for this fix, I was encountering exactly this issue (i.e., the 'liqo-fabric' pod constantly consuming one full CPU core).
However, it is not clear to me what this fix leaves behind. Which feature do I loose with this flag turned on?

PR #2818, opened 2024-11-18
Metrics: changed port from 8080 to 8082

cheina97:
# Description

This PR changes the ports used for metrics from 8080 to 8082. This is useful to avoid using port 8080 with host network pods, as port 8080 should be used by another service on the host.


fra98 (commented):
Grepping for 8080, I see some deployments templates to still have 8080 as metric port

aleoli:
@cheina97 this is probably breaking something since both tests (overlapped and not) are broken while waiting for the connection to be set up in the peer command

fra98 (approved):

PR #2822, opened 2024-11-21
docs: removed community meeting from readme

fra98:
# Description

This PR removes from the readme file the community meeting paragraph.


frisso:
@fra98 I agree to modify this file, because a community meeting once a week is an overkill.
However, I would not remove it at all, as the community should be reinforced, not killed.
For example, this project schedules a "demo day" (which looks a sort of community meeting" once a month: https://www.edgee.cloud/demo-day
I think this is something we should consider, creating an event every month or so.

fra98:
> @fra98 I agree to modify this file, because a community meeting once a week is an overkill. However, I would not remove it at all, as the community should be reinforced, not killed. For example, this project schedules a "demo day" (which looks a sort of community meeting" once a month: [edgee.cloud/demo-day](https://www.edgee.cloud/demo-day) I think this is something we should consider, creating an event every month or so.

@frisso I think the once-a-month meeting is a good idea and we should evaluate it in case. Regarding this PR, we thought to remove it from the readme for now to not give false information to possible attendees. Once we have a more clear idea of what to do we can re-introduce it.

PR #2825, opened 2024-11-21
[Fix] improve liqoctl uninstall error message

### LINKED ISSUE #2821: Issue on liqo uninstallation
Hello everyone.

I have a central kubernetes cluster implemented on OpenStack with 3 machines (1 server and 2 agents) with k3s installed on it using:
`curl -sfL https://get.k3s.io | sh -s - server --disable traefik --node-external-ip <PUBLIC_IP_ADDRESS>`

and an edge node which is a minipc with k3s installed using:
`sudo curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--disable traefik" sh -s - --write-kubeconfig-mode 644-`

This is the installed version of liqo on both sides:
Screenshot from 2024-11-21 11-58-50

I was trying to establish liqo peering on the edge node (minipc) using this command:
`liqoctl peer --in-band --remote-kubeconfig <KUBECONFIG_FILE> --server-service-type NodePort`
 but it stuck at this stage:
Waiting for Connection status to be established

Then I tried to uninstall liqo using `liqoctl uninstall --purge` but it returned me this message:
 ERRO  Pre-uninstall checks failed: you should:                                                                                                           
       disable networking for clusters:
       - snowy-water

I need to remove all the clusterroles, clusterrolebindings, customresourcedefinition, services, deployments, daemonsets, pods, cronjob.batch, namespaces, mutatingwebhookconfigurations and validatingwebhookconfiguration related to liqo to be able to uninstall liqo itself.

claudiolor:
# Description

this patch improves the liqoctl uninstall error message when the pre-uninstall checks fail, by suggesting the user which command has to be used to unpeer a cluster.

Closes #2821

### DIFF
--- a/pkg/liqoctl/uninstall/handler.go
+++ b/pkg/liqoctl/uninstall/handler.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
@@ -67,6 +68,25 @@ func (o *Options) Run(ctx context.Context) error {
 	s := o.Printer.StartSpinner("Running pre-uninstall checks")
 	if err := utils.PreUninstall(ctx, o.CRClient); err != nil {
 		s.Fail("Pre-uninstall checks failed: ", output.PrettyErr(err))
+		var uninstallErr utils.UninstallError
+		if errors.As(err, &uninstallErr) {
+			errorTypes := uninstallErr.GetErrorTypes()
+			if slices.Contains(errorTypes, utils.PendingActivePeerings) {
+				o.Printer.Error.Printfln(
+					"You must remove all the active peerings with other clusters before uninstalling Liqo.\n" +
+						"You can disable the active peerings via the 'liqoctl unpeer' command.\n" +
+						"Check 'liqoctl unpeer --help' for further information.",
+				)
+			}
+
+			if slices.Contains(errorTypes, utils.PendingOffloadedNamespaces) {
+				o.Printer.Error.Printfln(
+					"You must remove all the offloaded namespaces before uninstalling Liqo.\n" +
+						"You can disable the namespace offloading 'liqoctl unoffload' command.\n" +
+						"Check 'liqoctl unoffload --help' for further information.\n",
+				)
+			}
+		}
 		return err
 	}
 	s.Success("Pre-uninstall checks passed")
--- a/pkg/utils/preuninstall.go
+++ b/pkg/utils/preuninstall.go
@@ -17,6 +17,7 @@ package utils
 import (
 	"context"
 	"fmt"
+	"maps"
 	"slices"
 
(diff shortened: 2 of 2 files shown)


pkg/liqoctl/uninstall/handler.go
```diff
@@ -66,7 +66,10 @@ func (o *Options) Run(ctx context.Context) error {
 
 	s := o.Printer.StartSpinner("Running pre-uninstall checks")
 	if err := utils.PreUninstall(ctx, o.CRClient); err != nil {
-		s.Fail("Pre-uninstall checks failed: ", output.PrettyErr(err))
+		errMsg := fmt.Sprintf("Pre-uninstall checks failed: %s\n"+
```
aleoli, line 69: Are we checking here also for offloaded namespaces? if so, I would add the command to unoffload namespaces too

aleoli (approved):

pkg/liqoctl/uninstall/handler.go
```diff
@@ -66,7 +66,10 @@ func (o *Options) Run(ctx context.Context) error {
...
+			"You can disable the active peerings with other clusters with the 'liqoctl unpeer' command.\n"+
```
frisso, line 70: I would say "You must tear down the active peerings..." instead of "You can disable the active peering...."
Saying "you can", it means (to me) that it's just a suggestion, not a mandatory requirement to proceed.

claudiolor: The issue here is that the errors might not be related to the active peerings, but as @aleoli said, there might be also the possibility that there still is an offloaded namespace or other resources to remove. So I wouldn't always print to "You must tear down the active peerings..." because this is not always the case.
We might add at the end of the error message something like:
- To disable active peering launch liqoctl X
- To unoffload a namespace launch liqoctl Y

Actually, the ideal solution would be changing the returned error message to suggest a solution for each of the returned error, but since this is a shared function, I wouldn't suggest to use `liqoctl` when the user is not actually using it. So we should add a logic to understand when we are using liqoctl and print the suggestion. However, I'm not sure if it is worth it.
What do you guys think? @aleoli @fra98 @cheina97 

Maybe in future, it would be nice having a flag like `--force` that automatically removes everything that prevents Liqo from being uninstalled.

frisso: Is there any command we can suggest to look for active resources? (e.g., some "status"?)
In this case, the message could be "cannot uninstall because there are still some active resources. Check the status of your setup with command XXX and clean them properly to uninstall, or use the "--force" flag.

claudiolor: We finally decided to proceed with the best solution for the user: for each of the problems that prevent Liqo from being uninstalled we suggest the resolution

pkg/liqoctl/uninstall/handler.go
claudiolor: Fixed

PR #2827, opened 2024-11-21
[Docs] improve NAT and manual peering documentation

claudiolor:
# Description

This PR improves documentation about manual peering and the way to configure the network when there is a NAT in the topology


frisso:
Two comments on the picture.
1) Typo: "gatway" instead of "gateway" :-)
2) To me, one of the biggest advantage of the manual Liqo peering is that for the network setup we don't care who is the provider, and who is the consumer. I think this should be stated clearly in the documentation, ans possibly reflected in the picture, where clusters could be named "cluster1" and "cluster2" instead of "consumer/provider".

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
-# NAT Firewall
+# Configure a Liqo gateway server behind a NAT
 
-In Case a **gateway server** is behind a NAT firewall, the following steps are required to establish the connection:
-You can configure the settings for the connection by setting the following *annotations* in the `values.yaml` file or by using the `liqoctl` command line `--set` option:
+There might be some cases, especially when working in lab environments, where **it is not possible to configure the gateway server service as `LoadBalancer`**, requiring it to be exposed as a `NodePort` service instead.
+However, the nodes of the cluster might not be directly reachable, as for example, they are behind a NAT or a phisycal load balancer.
```
frisso, line 4:
```suggestion
However, the nodes of the cluster might not be directly reachable, e.g., when they are behind a NAT or a physical load balancer.
```

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
...
 
-Under the `networking.gatewayTemplates.server.service.annotations` key, you can set the following annotations:
+In this documentation page, we are going to describe how to configure Liqo in these scenarios.
```
frisso, line 6:
```suggestion
This page describes how to configure Liqo in the above scenarios.
```

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
...
-* **liqo.io/override-address**: the public IP address of the NAT firewall.
-* **liqo.io/override-port**: the public port of the NAT firewall.
+## Option A: Reverse Liqo network
 
-```{admonition} Tip
-In case you need to have multiple gateways behind the same NAT firewall, you need to override the port for each peer using the `--server-port` flag at peering time.
-This flag works with `liqoctl peer` and `liqoctl network connect` commands.
+![The provider is behind a NAT](../_static/images/advanced/nat/provider-nat.svg)
+
+By default using the `peer` command, the gateway server is configured on the provider cluster.
```
frisso, line 12:
```suggestion
The `liqoctl peer` command assumes that the gateway server is on the provider cluster.
```

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
...
+However, there may be cases where the provider cluster's nodes are not directly reachable, such as when they are behind a NAT, while the consumer cluster is directly accessible.
+For instance, in the image above, cluster 2 is behind a NAT and is therefore not directly reachable.
+
+To fix this issue, we can swap the roles of the gateways, configuring the client on the cluster provider and the server on the consumer.
```
frisso, line 16:
```suggestion
This problem can be solved by swapping the roles of the gateways, hence configuring the client on the cluster provider and the server on the consumer.
```

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
...
+
+In the case you would like to keep the gateway server on the provider, you can follow the instructions on the next paragraph.
+
+## Option B: configure the gateway server to be reachable through the NAT (or physical load balancer)
+
+The solution presented in this section allows to reach the gateway server even when it is behind a NAT (or, for example, a physical load balancer) which prevents it from being directly reachable.
+
+![The provider is behind a NAT](../_static/images/advanced/nat/port-address-override.svg)
+
+In this case, **you must configure port-mapping on the NAT** so that the traffic directed on a specific port of the NAT can be forwarded to one of nodes of the cluster.
```
frisso, line 31:
```suggestion
In this case, **you must configure the appropriate ip/port-mapping on the NAT device** so that the traffic directed on a specific port of the NAT can be forwarded to one of nodes of the cluster.
```

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
...
+
+**When network is configured via `liqoctl` commands**, if the service is of type `NodePort`, Liqo automatically uses the external IPs of the nodes to configure the gateway client.
```
frisso, line 33: I do not understand the term "automatically". If it is automatic, it means that the user has to do nothing to use the "external IP". However, from the next sentence, it seems that the user has to do something to let liqoctl knowing that we're in that situation. This means that it not automatic.
Am I right?
If so, I suggest to rearrange a bit the sentence to make things clearer.

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
...
+
+You can force the Liqo gateway client to use the public IP or FQDN and the port of the NAT where we configured port mapping, by providing some options to `liqoctl peer` or `liqoctl network connect` commands when using `liqoctl`.
+[With declarative peering](./peering/peering-via-cr.md#configuring-the-client-gateway-consumer-cluster) you can directly provide these values in the `GatewayClient` manifest.
+
+For example, let's imagine we expose the gateway server with a `NodePort` on port `30742` and we configure port mapping on the NAT so that port 40582 of the NAT redirects to port `30742` of a node of the cluster.
```
frisso, line 38:
```suggestion
For example, let's imagine we expose the gateway server with a `NodePort` on port `30742` and we configure port mapping on the NAT (which has public IP address equal to `203.0.113.8`) so that port 40582 of the NAT redirects to port `30742` of a node of the cluster.
```

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
...
+
+Of course, the client, running on cluster 1 cannot directly reach the gateway, but it needs to pass through the NAT.
+To do so, you need to pass the IP address or FQDN of the NAT and port where configured port mapping (`203.0.113.8` and `40582` in this case):
+
+```bash
+liqoctl peer --remote-kubeconfig $PATH_TO_CLUSTER2_KUBECONFIG --client-address $NAT_PUBBLIC_ADDR --client-port $NAT_MAPPING_PORT --server-service-nodeport $GATEWAY_SERVER_NODEPORT
+```
```
frisso, line 45: I fell the explanation below may is rather verbose. I would suggest to write something like:

In our precise case, the above command line will become the following:
```bash
liqoctl peer --remote-kubeconfig $PATH_TO_CLUSTER2_KUBECONFIG --client-address 203.0.113.8 --client-port 40582 --server-service-nodeport 30742
```
From this syntax, I have some concerns about the terminology used, namely "client" and "server". Personally, I was more in favor of something like:

```bash
liqoctl peer --remote-kubeconfig $PATH_TO_CLUSTER2_KUBECONFIG --public-address 203.0.113.8 --public-port 40582 --server-service-nodeport 30742
```
(instead of "public" we can also use "external")

docs/advanced/nat.md
```diff
@@ -1,14 +1,59 @@
...
+Where:
+
+- `NAT_PUBBLIC_ADDR` is the IP address of FQDN where the NAT is reachable (in this case `203.0.113.8`)
+- `NAT_MAPPING_PORT` is the port of the NAT redirecting to the `NodePort` (in this case `40582`)
+- `GATEWAY_SERVER_NODEPORT` is the port of the cluster 2's nodes where we are exposing the gateway server (in this case `30742`)
+
+The command above sets up a complete peering between cluster 1 and cluster 2.
+**To configure only the network**, you can pass the same parameters to the `liqoctl network connect` command:
+
+```bash
```
frisso, line 56: I would add some comments in the command line below, just briefly staring what the 'init' command is supposed to do, and what the 'connect' command is supposed to do.

docs/advanced/peering/inter-cluster-network.md
```diff
@@ -26,9 +26,11 @@ The unpeer process will automatically remove the Liqo Gateway from the tenant na
 
 ## Manual on cluster couple
```
frisso, line 27: Is it something like "option C"?
(previous subsections of type '##' started with "Option A" and "Option B".)

In any case, I don't understand what the title means.

docs/advanced/peering/inter-cluster-network.md
```diff
@@ -26,9 +26,11 @@ The unpeer process will automatically remove the Liqo Gateway from the tenant na
...
 
-When you have access to both clusters, you can configure the network connectivity for all the successive peering creations.
+When you have access to both clusters, you can configure the inter-cluster network connectivity via the `liqoctl network` command.
 
-First, you need to initialize the network:
+Note that when you use the `liqoctl network` command, the remote kubeconfig/context provided as argument, is the one of the cluster where we want to configure the server gateway of Wireguard tunnel.
```
frisso, line 31:
```suggestion
Note that when you use the `liqoctl network` command, the argument specifying the remote kubeconfig/context corresponds to the cluster that acts as server gateway for the Wireguard tunnel.
```

docs/advanced/peering/inter-cluster-network.md
```diff
@@ -26,9 +26,11 @@ The unpeer process will automatically remove the Liqo Gateway from the tenant na
...
+
+The first step to configure networking, is initializing the network configuration:
```
frisso, line 33: Still missing what "initialize" means. What are we actually achieving with this command? Why is it needed?

docs/usage/namespace-offloading.md
```diff
@@ -37,10 +37,11 @@ Regardless of the approach adopted, namespace offloading can be further configur
 
 ### Namespace mapping strategy
 
+When a namespace is offloaded, Liqo needs to create a *twin* namespace on the clusters where the namespaces is extended.
 The *namespace mapping strategy* defines the naming strategy used to create the remote namespaces, and can be configured through the `--namespace-mapping-strategy` flag.
 The accepted values are:
 
-* **DefaultName** (default): to **prevent conflicts** on the target cluster, remote namespace names are generated as the concatenation of the local namespace name, the cluster name of the local cluster and a unique identifier (e.g., *foo* could be mapped to *foo-lively-voice-dd8531*).
+* **DefaultName** (default): to **prevent conflicts** on the target cluster, remote namespace names are generated as the concatenation of the local namespace name and the cluster name of the local cluster (e.g., *foo* could be mapped to *foo-lively-voice*).
```
frisso, line 44:
```suggestion
* **DefaultName** (default): to **prevent conflicts** on the target cluster, remote namespace names are generated as the concatenation of the local namespace name and the cluster name of the local cluster (e.g., *foo* could be mapped to *foo-name-of-the-local-cluster*).
```

docs/usage/namespace-offloading.md
```diff
@@ -37,10 +37,11 @@ Regardless of the approach adopted, namespace offloading can be further configur
...
 * **EnforceSameName**: remote namespaces are named after the local cluster's namespace.
 This approach ensures **naming transparency**, which is required by certain applications, as well as guarantees that **cross-namespace DNS queries** referring to reflected services work out of the box (i.e., without adapting the target namespace name).
```
frisso, line 46:
```suggestion
This approach ensures **naming transparency**, which is required by certain applications (e.g., Istio), as well as guarantees that **cross-namespace DNS queries** referring to reflected services work out of the box (i.e., without adapting the target namespace name).
```

docs/usage/namespace-offloading.md
```diff
@@ -37,10 +37,11 @@ Regardless of the approach adopted, namespace offloading can be further configur
...
 Yet, it can lead to **conflicts** in case a namespace with the same name already exists inside the selected remote clusters, ultimately causing the remote namespace creation request to be rejected.
```
frisso, line 47:
```suggestion
However, it can lead to **conflicts** in case a namespace with the same name already exists inside the selected remote clusters, ultimately causing the remote namespace creation request to be rejected.
For instance, this mapping strategy prevents two consumer clusters to offload a namespace with the same name (e.g., `istio-system`) on the same provider cluster. 
```

docs/usage/namespace-offloading.md
```diff
@@ -72,23 +74,20 @@ In other words, an empty *cluster selector* matches all virtual clusters.
...
-You have two different ways to schedule pods onto virtual nodes, and they are mutually exclusive per Liqo installation:
+You have two different ways to determine whether a pod can be scheduled on a virtual node (so on a remote cluster):
 
-* **Pod offloading strategy** (default): it is defined at the namespace level, and it is applied to all pods deployed in the namespace. It will enable all pods to be scheduled onto virtual nodes; you will need to set affinities or node selectors to specify which pods should be offloaded.
-* **RuntimeClass**: it is defined at the pod level, and it is applied to the single pod. In this case, the pod offloading strategy is ignored, and the pod will be scheduled to the virtual nodes.
-
-```{admonition} Note
-In both cases, the pod offloading is enabled only if the namespace is offloaded.
-```
+* Defining a **pod offloading strategy** for the offloaded namespaces (default), which tells where the pods created on that namespace should be scheduled (whether in the local cluster, the remote clusters, or both letting the vanilla K8s scheduler decide).
```
frisso, line 82: I would suggest to add an example of a command that implements this strategy. The same comment applies also to the 'runtimeclass' sentence below.

docs/usage/namespace-offloading.md
```diff
@@ -72,23 +74,20 @@ In other words, an empty *cluster selector* matches all virtual clusters.
...
+* **LocalAndRemote** (default): pods deployed in the local namespace can be scheduled **both onto local nodes and onto virtual nodes**, hence possibly offloaded to remote clusters. This will leave the Kubernetes scheduler to decide about the best placement, based on the available resources and the pod requirements. You can still influence the scheduler decision on which pods should be scheduled onto virtual nodes using the [standard Kubernetes mechanisms to assign Pods to Nodes](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/).
 * **Local**: pods deployed in the local namespace are enforced to be scheduled onto **local nodes only**, hence never offloaded to remote clusters.
 The extension of a namespace, forcing at the same time all pods to be scheduled locally, enables the consumption of local services from the remote cluster, as shown in the [*service offloading* example](/examples/service-offloading).
```
frisso, line 92: I would remove this sentence, moving into a more general sentence below (after "*Remote*).
E.g.,

It is worth mentioning that, independently from the selected pod offloading strategy, services associated to them are propagated to the entire namespace (both locally and in the remote cluster), hence enabling the above pods to be consumed from anywhere in the Liqo domain, as shown in the [*service offloading* example](/examples/service-offloading).

docs/advanced/nat.md
claudiolor: First part of the comment fixed.
About the syntax, the names of these flags have been a little bit controversed. The problem here is that, with the peer/connect command you set up both the client and the server of the tunnel. So, a user, reading "public" or "external", might think he is configuring how the server is exposed. I believe that the message that should pass here is that with those flags you are overriding address and port used by the client.

docs/advanced/peering/inter-cluster-network.md
claudiolor: This title belongs to another page :)

docs/advanced/peering/inter-cluster-network.md
claudiolor: Fixed!
Initialize retrieves the network configurations of the clusters that needs to be connected, and creates all the resources to enable the IP remapping. The reason why this is a separate command is that this is a one shot action, but we are internally discussing whether it makes sense removing this command and making the `network init` implicit in the `network connect` command

docs/advanced/peering/inter-cluster-network.md
frisso: Ops... got it.
In any case, the title is complex to understand to me. I would simply say "Manual".

docs/usage/namespace-offloading.md
claudiolor: We have some example in the following sections, so I would proceed with your second proposal.

docs/advanced/peering/inter-cluster-network.md
claudiolor: Agree that the title is not so clear. Trying to make them more descriptive

docs/usage/namespace-offloading.md
claudiolor: I agree that the statement you proposed is useful and it is worth to be mentioned.
The one you want to remove is about not about services but about pods scheduling, so I would keep it. The idea is saying: "even though with the `LocalAndRemote` policy, the decision about where to schedule pod is up to the K8s vanilla scheduler, you can still influence its decision, like you would do with any other nodes of the cluster."

fra98 (approved):

docs/advanced/nat.md
frisso: I understand your point, but the fact that you're "overriding" a value is just in the mind of who defined the command, not in the user's mind :-) 
The parameter says "--client-address" and "--client-port", which to me (as a user) I feel misleading. The client address is the internal one, not the external one.
This would be much more clear if you say "--override-client-address" and "--override-client-port", just in case you can accept a longer command.

docs/usage/namespace-offloading.md
frisso: I'm not sure I got your point.
The sentence I was referring to is the following:
`The extension of a namespace, forcing at the same time all pods to be scheduled locally, enables the consumption of local services from the remote cluster, as shown in the [*service offloading* example](/examples/service-offloading).`
This sentence is about service offloading, not about pod scheduling. I feel this should not stay here, as this should be already included in the new sentence we added at line L97.

docs/advanced/peering/inter-cluster-network.md
```diff
@@ -24,11 +24,13 @@ Automatic network configuration
 
 The unpeer process will automatically remove the Liqo Gateway from the tenant namespace.
 
-## Manual on cluster couple
+## Setup the inter-cluster network via liqoctl network
```
frisso, line 27:
```suggestion
## Setup the inter-cluster network via 'liqoctl network' command
```

docs/advanced/peering/inter-cluster-network.md
```diff
@@ -199,7 +201,7 @@ You can configure how to expose the Liqo Gateway Server service by using the fol
 
 Use the `liqoctl network connect --help` command to see all the available options.
 
-## Manual on single cluster
+## Manually setup the inter-cluster network on each cluster separately
 
 When you don't have access to both clusters, or you want to configure it in a declarative way, you can configure it by applying CRDs.
```
frisso, line 206:
```suggestion
When you do not have contemporary access to both clusters, or you would like to configure the inter-cluster network in a declarative way, you can configure this component by applying the proper CRDs on either clusters.
```

docs/advanced/peering/inter-cluster-network.md
```diff
@@ -209,6 +211,8 @@ The process consists of the following steps:
 3. **Cluster client**: the other cluster defines a `GatewayClient` resource, which will configure a client that will connect to the gateway server exposed on the other cluster.
 4. **Cluster client and server**: the cluster client and server need to exchange the public keys to allow secure communication.
 
+More info about the declarative setup of the network can he found [here](./peering-via-cr.md#declarative-network-configuration).
```
frisso, line 214:
```suggestion
More info about the declarative setup of the network can be found [here](./peering-via-cr.md#declarative-network-configuration).
```

PR #2828, opened 2024-11-21
[IPAM] Core implementation

cheina97:
# Description

This PR introduces the new liqo ipam core

It's main features are:

- Define custom network pools used by the ipam
- Allocate networks
- Remap networks and allocate them
- Allocate IPs from a network
- Visualize the IPAM using a graphviz file

The IPAM is managed internally using a binary tree, where each node is a network and contains a list of IPs
We used this algorithm https://en.wikipedia.org/wiki/Buddy_memory_allocation to optimize the network allocation

image

### DIFF
--- a/cmd/ipam/main.go
+++ b/cmd/ipam/main.go
@@ -70,8 +70,12 @@ func main() {
 
 	// Server options.
 	cmd.Flags().IntVar(&options.ServerOpts.Port, "port", consts.IpamPort, "The port on which to listen for incoming gRPC requests.")
-	cmd.Flags().DurationVar(&options.ServerOpts.SyncFrequency, "interval", consts.SyncFrequency,
+	cmd.Flags().DurationVar(&options.ServerOpts.SyncFrequency, "sync-interval", consts.SyncInterval,
 		"The interval at which the IPAM will synchronize the IPAM storage.")
+	cmd.Flags().BoolVar(&options.ServerOpts.GraphvizEnabled, "enable-graphviz", false, "Enable the graphviz output for the IPAM.")
+	cmd.Flags().StringSliceVar(&options.ServerOpts.Pools, "pools",
+		[]string{"10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12"}, "The pools used by the IPAM.",
+	)
 
 	// Leader election flags.
 	cmd.Flags().BoolVar(&options.EnableLeaderElection, "leader-election", false, "Enable leader election for IPAM. "+
@@ -132,7 +136,7 @@ func run(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
-	liqoIPAM, err := ipam.New(ctx, cl, &options.ServerOpts)
+	liqoIPAM, err := ipam.New(ctx, cl, options.ServerOpts.Pools, &options.ServerOpts)
 	if err != nil {
 		return err
 	}
--- a/pkg/ipam/core/ipam.go
+++ b/pkg/ipam/core/ipam.go
@@ -1,0 +1,218 @@
+// Copyright 2019-2024 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ipamcore
+
+import (
+	"fmt"
+	"net/netip"
+	"slices"
+)
+
+// Ipam represents the IPAM core structure.
+type Ipam struct {
+	roots []node
+}
+
+// NewIpam creates a new IPAM instance.
+func NewIpam(pools []string) (*Ipam, error) {
+	ipamRootsPrefixes := make([]netip.Prefix, len(pools))
+	for i, root := range pools {
+		ipamRootsPrefixes[i] = netip.MustParsePrefix(root)
+	}
+
+	if err := checkRoots(ipamRootsPrefixes); err != nil {
+		return nil, err
+	}
+
+	ipamRoots := make([]node, len(pools))
+	for i := range ipamRootsPrefixes {
+		ipamRoots[i] = newNode(ipamRootsPrefixes[i])
+	}
+
+	ipam := &Ipam{
+		roots: ipamRoots,
+	}
+
+	return ipam, nil
+}
+
+// NetworkAcquire allocates a network of the given size.
+// It returns the allocated network or nil if no network is available.
+func (ipam *Ipam) NetworkAcquire(size int) *netip.Prefix {
+	for i := range ipam.roots {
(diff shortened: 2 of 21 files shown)


pkg/ipam/sync.go
```diff
@@ -53,61 +59,110 @@ func (lipam *LiqoIPAM) sync(ctx context.Context, syncFrequency time.Duration) {
...
+}
 
-	// Remove networks that are present in the cache but not in the cluster, and were added before the threshold.
-	for key := range lipam.cacheNetworks {
-		if _, inCluster := setClusterNetworks[key]; !inCluster && lipam.cacheNetworks[key].creationTimestamp.Before(expiredThreshold) {
-			lipam.freeNetwork(lipam.cacheNetworks[key].network)
+func syncIPsFreeWithNetwork(lipam *LiqoIPAM, clusterIPs []ipCidr, currentIPs []netip.Addr, currentIPNetwork netip.Prefix) error {
+	for i := range currentIPs {
+		currentIP := currentIPs[i]
+		if !isIPInCluster(clusterIPs, ipCidr{ip: currentIP.String(), cidr: currentIPNetwork.String()}) {
```
fra98, line 131:
```suggestion
		currentIP := &currentIPs[i]
```

pkg/ipam/sync.go
```diff
@@ -53,61 +59,110 @@ func (lipam *LiqoIPAM) sync(ctx context.Context, syncFrequency time.Duration) {
...
+			if err := lipam.ipRelease(ipCidr{ip: currentIPs[i].String(), cidr: currentIPNetwork.String()}); err != nil {
+				return fmt.Errorf("failed to free IP %q in network %q: %w", currentIPs[i].String(), currentIPNetwork.String(), err)
+			}
 		}
 	}
+	return nil
+}
```
fra98, line 138: Here we do not have optimized the research which is still O(n) instead of O(1). We should cache in a set/map the clusterIPs for a faster lookup

pkg/ipam/sync.go
```diff
@@ -53,61 +59,110 @@ func (lipam *LiqoIPAM) sync(ctx context.Context, syncFrequency time.Duration) {
...
+	return false
+}
 
-	// Remove networks that are present in the cache but not in the cluster, and were added before the threshold.
-	for key := range lipam.cacheNetworks {
-		if _, inCluster := setClusterNetworks[key]; !inCluster && lipam.cacheNetworks[key].creationTimestamp.Before(expiredThreshold) {
-			lipam.freeNetwork(lipam.cacheNetworks[key].network)
+func syncIPsFreeWithNetwork(lipam *LiqoIPAM, clusterIPs []ipCidr, currentIPs []netip.Addr, currentIPNetwork netip.Prefix) error {
+	for i := range currentIPs {
+		currentIP := currentIPs[i]
```
fra98, line 130:
```suggestion
		currentIP := &currentIPs[i]
```

pkg/ipam/networks.go
```diff
@@ -16,95 +16,102 @@ package ipam
...
 	}
-	_, ok := lipam.cacheNetworks[nw.String()]
 
-	return ok
+	return !lipam.IpamCore.NetworkIsAvailable(prefix), nil
 }
 
-func listNetworksOnCluster(ctx context.Context, cl client.Client) ([]network, error) {
-	var nets []network
+func (lipam *LiqoIPAM) listNetworksOnCluster(ctx context.Context, cl client.Client) ([]netip.Prefix, error) {
```
fra98, line 113: Here the original function returned the network struct which also containes the preAllocated field. We can not remove it since we need this info the initialize and sync routines

pkg/ipam/core/ipam.go
```diff
@@ -0,0 +1,225 @@
...
+func NewIpam(roots, preallocated []string) (*Ipam, error) {
+	ipamRootsPrefixes := make([]netip.Prefix, len(roots))
+	for i, root := range roots {
+		ipamRootsPrefixes[i] = netip.MustParsePrefix(root)
+	}
+
+	ipamPreallocated := make([]netip.Prefix, len(preallocated))
+	for i, prefix := range preallocated {
+		ipamPreallocated[i] = netip.MustParsePrefix(prefix)
+	}
```
fra98, line 38: I think this feature it is not needed anymore with https://github.com/liqotech/liqo/pull/2829, as the reserved networks (podcidr, servicecidr) are allocated with grpc calls before starting controllers

pkg/ipam/ipam.go
```diff
@@ -83,31 +100,89 @@ func (lipam *LiqoIPAM) IPAcquire(_ context.Context, req *IPAcquireRequest) (*IPA
...
+			return &NetworkAcquireResponse{}, err
+		}
+	}
+
+	for i := 0; i < int(req.GetPreAllocated()); i++ {
+		_, err := lipam.ipAcquire(prefix)
+		if err != nil {
+			return &NetworkAcquireResponse{}, errors.Join(err, lipam.networkRelease(prefix))
+		}
 	}
```
fra98, line 153: Shouldn't we acquire IPs from the remapped CIDR and not from the original CIDR?

pkg/ipam/ipam.go
```diff
@@ -83,31 +100,89 @@ func (lipam *LiqoIPAM) IPAcquire(_ context.Context, req *IPAcquireRequest) (*IPA
...
+		remappedCidr, err = lipam.networkAcquire(prefix)
+		if err != nil {
+			return &NetworkAcquireResponse{}, err
+		}
+	}
+
+	for i := 0; i < int(req.GetPreAllocated()); i++ {
+		_, err := lipam.ipAcquire(prefix)
+		if err != nil {
+			return &NetworkAcquireResponse{}, errors.Join(err, lipam.networkRelease(prefix))
```
fra98, line 151: We should release the remappedCidr

pkg/ipam/sync.go
```diff
@@ -53,61 +60,126 @@ func (lipam *LiqoIPAM) sync(ctx context.Context, syncFrequency time.Duration) {
 	}
 }
 
-func (lipam *LiqoIPAM) syncNetworks(ctx context.Context, expiredThreshold time.Time) error {
+func syncNetworkAcquire(lipam *LiqoIPAM, clusterNetworks map[netip.Prefix]prefixDetails, currentNetworks map[netip.Prefix]any) error {
+	// Add networks that are present in the cluster but not in the cache.
+	for clusterNetwork, clusterNetworkDetails := range clusterNetworks {
+		if _, ok := currentNetworks[clusterNetwork]; !ok {
+			if _, err := lipam.networkAcquireSpecific(clusterNetwork); err != nil {
+				return fmt.Errorf("failed to acquire network %q: %w", clusterNetworks[clusterNetwork], err)
```
fra98, line 68:
```suggestion
				return fmt.Errorf("failed to acquire network %q: %w", clusterNetwork, err)
```

cmd/ipam/main.go
```diff
@@ -72,6 +72,7 @@ func main() {
 	cmd.Flags().IntVar(&options.ServerOpts.Port, "port", consts.IpamPort, "The port on which to listen for incoming gRPC requests.")
 	cmd.Flags().DurationVar(&options.ServerOpts.SyncFrequency, "interval", consts.SyncFrequency,
```
aleoli, line 73: I know it is not in this pr, but I noticed it in the manifest. Here, we have a name for the flag (interval) different from the const (sync frequency). I would use something like sync interval, also because the value is not a frequency but an interval

aleoli (commented):
In general, I would add some comments on how the algorithm works and the single functions work in the core package

PR #2831, opened 2024-11-25
[liqoctl] allow override gateway client address and port

claudiolor:
# Description

At the moment to override the address and port used by the gateway client when peering is done via liqoctl is handled via an annotation to be placed on the gateway server template resource. However, this has some limitations:

- It's not really used friendly
- It doesn't work when behind the same NAT there are multiple gateways (as the template is shared by all the instances of gateway, so override is performed on everybody)

To fix this issue this PR introduces the `--client-address` and `--client-port` to the `peer` and `connect` commands of `liqoctl` to override the value written in the endpoint field of the status of the GatewayServer resource, which is used to configure the GatewayClient resource. This is useful especially when the gateway server is not directly reachable by the client (e.g. it is behind a NAT).


cmd/liqoctl/cmd/network.go
```diff
@@ -165,6 +165,14 @@ func newNetworkConnectCommand(ctx context.Context, options *network.Options) *co
 	// Client flags
 	cmd.Flags().StringVar(&options.ClientGatewayType, "client-type", forge.DefaultGwClientType,
 		"Type of Gateway Client. Leave empty to use default Liqo implementation of WireGuard")
+	cmd.Flags().StringVar(&options.ClientConnectAddress, "client-address", "",
+		"Define the address used by the gateway client to connect to the gateway server."+
+			"This value overrides the one automatically retrieved by Liqo and it is useful when the server is "+
```
aleoli, line 170:
```suggestion
		"Define the address used by the gateway client to connect to the gateway server. "+
			"This value overrides the one automatically retrieved by Liqo and it is useful when the server is "+
```
I think we are missing a space here, the same for the line below and for the peer command

aleoli (approved):

cmd/liqoctl/cmd/network.go
claudiolor: Fixed! Thank you

examples/networking/wireguard-server-nat.yaml
fra98, line 1: Should we remove this file now? Since it's identical to the default

fra98 (approved):

examples/networking/wireguard-server-nat.yaml
claudiolor: Removed

PR #2832, opened 2024-11-25
Add global labels and annotations to liqo-created resources

aleoli:
# Description

This pr adds global labels and annotations to all liqo-created resources.

It can be useful to add the resources created by liqo controllers to an argocd application

### DIFF
--- a/pkg/liqo-controller-manager/authentication/utils/nonce.go
+++ b/pkg/liqo-controller-manager/authentication/utils/nonce.go
@@ -20,19 +20,19 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	liqov1beta1 "github.com/liqotech/liqo/apis/core/v1beta1"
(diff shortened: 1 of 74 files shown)


claudiolor (approved):

pkg/liqo-controller-manager/authentication/utils/nonce.go
```diff
@@ -55,7 +55,9 @@ func EnsureSignedNonceSecret(ctx context.Context, cl client.Client,
 			return fmt.Errorf("nonce not provided and no nonce secret found")
 		}
 		secret := forge.SignedNonce(remoteClusterID, tenantNamespace, *nonce)
-		if err := cl.Create(ctx, secret); err != nil {
+		if _, err := resource.CreateOrUpdate(ctx, cl, secret, func() error {
+			return nil
+		}); err != nil {
 			return fmt.Errorf("unable to create nonce secret: %w", err)
 		}
 		return nil
```
fra98, line 63: I find this potentially risky or misused. If the object already exists, the CreateOrUpdate does not apply the previously forged object (since it only gets the exisiting object and the passed mutateFn does not do anything). I wonder if it's safer to use a proper Create() wrapper (or just add the global labels before). 
Note: of course, for this particular case it works, as we check before that the secret does not exist

fra98 (approved):

fra98 (approved):

PR #2836, opened 2024-11-29
[IPAM] fix uninstall

fra98:
# Description

This PR fixes the uninstall with the new IPAM.
In particular, it sets a specific annotation to networks installed with Helm. Since these are required through the whole lifecycle of Liqo, we do not add a finalizer since we do not need to unmap them (as we are uninstalling anyway).


pkg/liqo-controller-manager/networking/ip-controller/ip_controller.go
```diff
@@ -114,7 +115,7 @@ func (r *IPReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 		}
 
 		// Remove finalizer from the IP resource if present.
-		if controllerutil.ContainsFinalizer(&ip, ipamIPFinalizer) {
+		if controllerutil.ContainsFinalizer(&ip, ipamIPFinalizer) && !utils.IsPreinstalledResource(&ip) {
```
aleoli, line 118:
```suggestion
		if controllerutil.ContainsFinalizer(&ip, ipamIPFinalizer) {
```

The preinstalled resources should not have the finalizer, but I would not check this if, for some strange reason,n they have

pkg/liqo-controller-manager/networking/network-controller/network_controller.go
```diff
@@ -137,7 +138,7 @@ func (r *NetworkReconciler) handleNetworkStatus(ctx context.Context, nw *ipamv1a
 				return err
 			}
 		}
-	} else if controllerutil.ContainsFinalizer(nw, ipamNetworkFinalizer) {
+	} else if controllerutil.ContainsFinalizer(nw, ipamNetworkFinalizer) && !utils.IsPreinstalledResource(nw) {
```
aleoli, line 141: the same

aleoli (approved):

PR #2841, opened 2024-12-04
[IPAM] function NetworkIsAvailable refactoring

cheina97:
# Description

This PR changes how the NetworkIsAvailable internally work

### DIFF
--- a/pkg/ipam/core/node.go
+++ b/pkg/ipam/core/node.go
@@ -121,6 +121,33 @@ func networkRelease(prefix netip.Prefix, node *node) *netip.Prefix {
 	return result
 }
 
+func networkIsAvailable(prefix netip.Prefix, node *node) bool {
+	if node.prefix.Addr().Compare(prefix.Addr()) == 0 && node.prefix.Bits() == prefix.Bits() {
+		if node.left != nil && node.left.left.isSplitted() {
+			return false
+		}
+		if node.right != nil && node.right.isSplitted() {
+			return false
+		}
+
+		// If node children are not splitted and node is not acquired, then network is available
+		return !node.acquired
+	}
+
+	if node.left == nil && node.right == nil {
+		return true
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/ipam/core/node.go
```diff
@@ -121,6 +121,31 @@ func networkRelease(prefix netip.Prefix, node *node) *netip.Prefix {
 	return result
 }
 
+func networkIsAvailable(prefix netip.Prefix, node *node) bool {
+	if node == nil {
+		return true
+	}
+
+	if node.prefix.Addr().Compare(prefix.Addr()) == 0 && node.prefix.Bits() == prefix.Bits() {
+		if node.left != nil && (node.left.left != nil || node.left.right != nil) {
```
claudiolor, line 130:
```suggestion
		if node.left != nil && node.left.isSplitted() {
```

pkg/ipam/core/node.go
```diff
@@ -121,6 +121,31 @@ func networkRelease(prefix netip.Prefix, node *node) *netip.Prefix {
...
+			return false
+		}
+		if node.right != nil && (node.right.left != nil || node.right.right != nil) {
```
claudiolor, line 133:
```suggestion
		if node.right != nil && node.right.isSplitted() {
```

pkg/ipam/core/node.go
```diff
@@ -121,6 +121,31 @@ func networkRelease(prefix netip.Prefix, node *node) *netip.Prefix {
...
+			return false
+		}
+		return !node.acquired
```
claudiolor, line 136:
```suggestion
                // If node is not splitted nor acquired, then network is available
		return !node.acquired
```

pkg/ipam/core/node.go
cheina97: Same

fra98 (approved):

PR #2847, opened 2024-12-05
[Deploy] Fixes for smooth ArgoCD uninstall

claudiolor:
# Description

Cleanly uninstall Liqo with ArgoCD was not possible due to the fact that sometimes the deleting process was stuck due to resources having finalizers that were removed after the removal of the controller-manager (in charge of removing those finalizers) and some other residual resources polluted the cluster even after Liqo removal.

This PR introduces some fixes allowing to smoothly uninstall ArgoCD provided the following conditions are verified:

- There are no pending peerings of offloaded namespace
- The following configuration is defined at installation time:
```
common:
          globalLabels:
            app.kubernetes.io/instance: <ARGOCD_APP_NAME>
          globalAnnotations:
            argocd.argoproj.io/ignore-resource-updates: "true"
            argocd.argoproj.io/sync-options: Delete=true
            argocd.argoproj.io/sync-wave: 10
            argocd.argoproj.io/compare-options: IgnoreExtraneous
 
```
Where:
  - `app.kubernetes.io/instance` in `globalLabels` tells Liqo to add that labels to all the resources created by the controllers, in this case this allows ArgoCD to track all the resources created afterwards by controllers. While:
  - ` argocd.argoproj.io/sync-options: Delete=true`: tells ArgoCD to remove the resources tracked with this annotation during deletion
  - `argocd.argoproj.io/sync-wave`: anticipate the deletion of the resources annotated with that label, so that controllers can remove the finalizers and perform the necessary cleanup

### DIFF
--- a/pkg/liqo-controller-manager/networking/internal-network/route/internalnode_controller.go
+++ b/pkg/liqo-controller-manager/networking/internal-network/route/internalnode_controller.go
@@ -164,10 +164,13 @@ func (r *InternalNodeReconciler) genericEnqueuerfunc(ctx context.Context, _ clie
 		klog.Error(err)
 		return nil
 	}
+
 	var requests []reconcile.Request
 	for i := range internalNodes.Items {
+		iNode := &internalNodes.Items[i]
+
 		requests = append(requests, reconcile.Request{
-			NamespacedName: client.ObjectKeyFromObject(&internalNodes.Items[i]),
+			NamespacedName: client.ObjectKeyFromObject(iNode),
 		})
 	}
 	return requests
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -274,7 +274,7 @@ func main() {
 			ipamClient = ipam.NewIpamClient(connection)
 		}
 
-		if err := modules.SetupNetworkingModule(ctx, mgr, &modules.NetworkingOption{
+		opts := &modules.NetworkingOption{
 			DynClient: dynClient,
 			Factory:   factory,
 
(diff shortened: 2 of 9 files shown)


pkg/liqo-controller-manager/networking/internal-network/route/internalnode_controller.go
```diff
@@ -155,20 +172,29 @@ func (r *InternalNodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+		// When the object that triggered the watch is a Node, enqueue the event only for the InternalNode related to that Node.
+		if isNode && iNode.Name == obj.GetName() {
+			requests = append(requests, reconcile.Request{
+				NamespacedName: client.ObjectKeyFromObject(iNode),
+			})
+		}
 	}
 	return requests
 }
```
fra98, line 200: I would do a custom enqueuer for the node Watch(), to avoid casting and reduce logic complexity

pkg/liqo-controller-manager/networking/internal-network/route/internalnode_controller.go
fra98, line 1: I don't get why this controller should call the deletion on his own reconciled resource, it seems a strange pattern. Wasn't the idea to watch for event deletion in the node_controller (the one that actually creates and manage the lifecycle of the inernalnodes), and add a checks at pod startup to cleanup eventual internalnodes (if missed deletion events on pod crashes)

claudiolor: I believe reconciling on InternalNodes is a more robust solution, as there is no need to perform initial sync up and if an InternalNode resource not matching a Node is created, this is automatically reconciled by the controller. In any case, I agree this is not actually straightforward: fixing it according to your solution

fra98 (approved):

PR #2867, opened 2024-12-12
[Liqoctl] control-planes in liqoctl network np-nodes flag

cheina97:
# Description

This pr adds the value `control-planes` inside the `np-nodes` flag in `liqoctl test network`


cheina97:
&rebase test)true

fra98 (approved):

Issue #2869, opened 2024-12-16
RuntimeClassName not transmitted to the provider cluster


remmen-io:
#### What happened:
Using namespace offloading, the runtimeClassName is not transferred from the consumer cluster to the provider cluster.

Deployed on the consumer cluster is a YAML with the runtimeClassName
``` 
kind: Deployment
apiVersion: apps/v1
metadata:
  name: gpu-test
  labels:
    app: gpu-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: gpu-test
  template:
    metadata:
      labels:
        app: gpu-test
    spec:
      runtimeClassName: nvidia
      containers:
        - resources:
            limits:
              cpu: "1"
              memory: "1Gi"
              nvidia.com/gpu: "1"
            requests:
              cpu: "1"
              nvidia.com/gpu: "1"
          name: gpu
          imagePullPolicy: IfNotPresent
          image: "nvidia/cuda:12.6.1-base-ubuntu24.04"
          command: ["/bin/bash"]
          args:
           - "-c"
           - "while true; do nvidia-smi;sleep 100000; done"
```

On the provider cluster, the runtimeclass is missing.

```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2024-12-13T13:56:15Z"
  labels:
    app: gpu-test
    liqo.io/creator-user: gpupool-9529260abb9b
    liqo.io/managed-by: shadowpod
    offloading.liqo.io/destination: e1-k8s-lab-t
    offloading.liqo.io/nodename: gpupool
    offloading.liqo.io/origin: e1-k8s-lab-b
    pod-template-hash: 79b6ccb455
  name: gpu-test-79b6ccb455-gtpz7
  namespace: liqo-demo-e1-k8s-lab-b
  ownerReferences:
  - apiVersion: offloading.liqo.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: ShadowPod
    name: gpu-test-79b6ccb455-gtpz7
    uid: 61675316-6f37-49d4-8f5b-197896752371
  resourceVersion: "9930727"
  uid: 5d56a142-afce-43e8-b25a-e9b255689a74
spec:
  automountServiceAccountToken: false
  containers:
  - args:
    - -c
    - while true; do nvidia-smi;sleep 100000; done
    command:
    - /bin/bash
    env:
    - name: KUBERNETES_SERVICE_HOST
      value: kubernetes.default
[...]

PR #2887, opened 2024-12-27
[virtual-kubelet] reflect pod RuntimeClass

fra98:
# Description

This PR introduces the support for pod RuntimeClass reflection.

RuntimeClass will be reflected from local to remote if set. 
If you want to use the liqo RuntimeClass, you can annotate the pod with `liqo.io/remote-runtime-class-name: <RUNTIMECLASS_NAME>`.
It is also possible to enforce a remote RuntimeClass to all pods scheduled on a virtual node, by specifying it in the OffladingPatch of the virtualnode (`spec.offloadingPatch.runtimeClassName`).

To determine the remote RuntimeClass, the following priority will be used (from higher to lower priority):
1. pod Annotation (`liqo.io/remote-runtime-class-name`)
2. pod RuntimeClass (`spec.runtimeClassName`). It is ignored if set to "liqo"
3. virtualNode OffloadingPatch (`spec.offloadingPatch.runtimaClassName`)


fra98:
fixes #2869

aleoli (approved):

PR #2889, opened 2025-01-02
Use liqo.selectorTemplate for label definitions

aleoli:
# Description

This pr fixes the gateway label selectors making it possible to update a running gateway

### DIFF
--- a/deployments/liqo/templates/_helpers.tpl
+++ b/deployments/liqo/templates/_helpers.tpl
@@ -92,9 +92,16 @@ labels:
 Common Labels for Gateway Templates
 */}}
 {{- define "liqo.labelsTemplate" -}}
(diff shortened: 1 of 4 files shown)


deployments/liqo/templates/_helpers.tpl
```diff
@@ -93,8 +93,16 @@ Common Labels for Gateway Templates
 */}}
 {{- define "liqo.labelsTemplate" -}}
 {{ include "liqo.selectorLabelsTemplate" . }}
+{{ include "liqo.selectorTemplate" . }}
```
fra98, line 96: In `liqo.labelsTemplate`, `liqo.selectorLabelsTemplate` seems redudant as already included in `liqo.selectorTemplate`

fra98 (approved):

PR #2913, opened 2025-01-31
[liqoctl] set gateway server Service location in `peer` command

fra98:
# Description

This PR introduces the support to specify the gateway client and server location in a peering.
The `--server-service-location` flag in `liqoctl peer` allows to place the gateway server service in the provider (default) or consumer.
This can be useful if you cannot expose Services (or don't have LB UDP support) in the provider cluster or if the provider is behind a NAT (without going through the manual peering procedure or setting up the networking module individually).

### DIFF
--- a/cmd/liqoctl/cmd/peer.go
+++ b/cmd/liqoctl/cmd/peer.go
@@ -22,6 +22,7 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/util/runtime"
 
+	liqov1beta1 "github.com/liqotech/liqo/apis/core/v1beta1"
 	nwforge "github.com/liqotech/liqo/pkg/liqo-controller-manager/networking/forge"
 	"github.com/liqotech/liqo/pkg/liqoctl/completion"
 	"github.com/liqotech/liqo/pkg/liqoctl/factory"
@@ -90,8 +91,11 @@ func newPeerCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
 
 	// Networking flags
 	cmd.Flags().BoolVar(&options.NetworkingDisabled, "networking-disabled", false, "Disable networking between the two clusters")
+	cmd.Flags().Var(options.ServerServiceLocation, "server-service-location",
+		fmt.Sprintf("Location of the service to expose the Gateway Server (%q or %q). Default: %q",
+			liqov1beta1.ConsumerRole, liqov1beta1.ProviderRole, nwforge.DefaultGwServerLocation))
 	cmd.Flags().Var(options.ServerServiceType, "server-service-type",
-		fmt.Sprintf("Service type of the Gateway Server service. Default: %s."+
+		fmt.Sprintf("Service type of the Gateway Server service. Default: %q."+
 			" Note: use ClusterIP only if you know what you are doing and you have a proper network configuration",
 			nwforge.DefaultGwServerServiceType))
 	cmd.Flags().Int32Var(&options.ServerServicePort, "server-service-port", nwforge.DefaultGwServerPort,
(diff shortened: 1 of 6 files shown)


claudiolor (approved):

aleoli (approved):

frisso:
Difficult for me to understand what "server" means in the command line, "--server-service-location".
I would suggest to propose a more intuitive name (e.g., --gateway-server-location", perhaps?)

docs/usage/peer.md
```diff
@@ -36,7 +36,8 @@ To perform a peering without having access to both clusters, you need to manuall
 The peering command enables all 3 liqo modules and performs the following steps:
 
 1. **enables networking**.
-Exchanges network configurations and creates the two **gateways** (server in the provider, client in the consumer) to let the two clusters communicate over a secure tunnel.
+Exchanges network configurations and creates the two **gateways** (one server in a cluster and one client in the other) to let the two clusters communicate over a secure tunnel.
+By default the gateway server in placed in the provider, while the gateway client in the consumer, but you can configure it with the `--server-service-location` flag.
```
frisso, line 40:
```suggestion
By default the gateway server is located in the provider, while the gateway client in the consumer; this can be customized with the `--server-service-location` flag in `liqoctl`.
```

docs/usage/peer.md
```diff
@@ -36,7 +36,8 @@ To perform a peering without having access to both clusters, you need to manuall
 The peering command enables all 3 liqo modules and performs the following steps:
 
 1. **enables networking**.
-Exchanges network configurations and creates the two **gateways** (server in the provider, client in the consumer) to let the two clusters communicate over a secure tunnel.
+Exchanges network configurations and creates the two **gateways** (one server in a cluster and one client in the other) to let the two clusters communicate over a secure tunnel.
```
frisso, line 39:
```suggestion
Exchanges network configurations and creates the two **gateways** (one acting as _server_ and located in the provider cluster, another acting as _client_ in the consumer cluster) to let the two clusters communicate over a secure tunnel.
```

cmd/liqoctl/cmd/peer.go
```diff
@@ -90,8 +91,11 @@ func newPeerCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
 
 	// Networking flags
 	cmd.Flags().BoolVar(&options.NetworkingDisabled, "networking-disabled", false, "Disable networking between the two clusters")
+	cmd.Flags().Var(options.ServerServiceLocation, "server-service-location",
+		fmt.Sprintf("Location of the service to expose the Gateway Server (%q or %q). Default: %q",
```
frisso, line 95: Not sure I would understand correctly.
I would say this way:
fmt.Sprintf("Cluster that acts as gateway server when setting up the network between two Liqo instances (%q or %q). Default: %q",

fra98:
> Difficult for me to understand what "server" means in the command line, "--server-service-location". I would suggest to propose a more intuitive name (e.g., --gateway-server-location", perhaps?)

Hi @frisso, the purpose of this flag is to allow the user to customize the location of the k8s *Service* that exposes the Gateway Server. 
We decided to keep the word *service* as this is the most important thing to know for the common `liqoctl peer` user that does not need to know the internal of Liqo, nor what type of Server this is or what it does. For this flag, the thing that matters the most is the ability to expose a LoadBalancer *Service* (if not using NodePort ofc), therefore the user should choose the cluster equipped with a working LoadBalancer provider.
Other reasons are:
- coherence with other flags that customize the k8s *Service* (e.g,, `--server-service-port,` `--server-service-loadbalancerip`)
- a more exhaustive name (e.g., `--gateway-server-service-location`) would be too verbose, and the flag description already indicates that the flag customizes the location of the Service exposing the Gateway Server.

frisso:
Hello @fra98 
As a user, I'm definitely confused by the string "--server-service". Service of what? You know that is the gateway service, but this is not really evident from the flag.

fra98:
@frisso 
A good compromise between verbosity and clarity would be to add a short prefix (e.g., `gw-`) to the flag name.
In this case, a small refactoring is needed to change the other flags related to the gateway server/client, but I would leave that to a separate PR.

PR #2920, opened 2025-02-06
fix: examples

kariya-mitsuru:
# Description

`serviceSubnet` "10.70.0.0/12" and "10.90.0.0/12" are invalid.

### DIFF
--- a/examples/stateful-applications/manifests/cluster1.yaml
+++ b/examples/stateful-applications/manifests/cluster1.yaml
@@ -1,7 +1,7 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 networking:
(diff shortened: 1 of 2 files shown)


examples/stateful-applications/manifests/cluster1.yaml
```diff
@@ -1,7 +1,7 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 networking:
-  serviceSubnet: "10.90.0.0/12"
```
cheina97, line 4: Can you keep the same CIDR and use `10.80.0.0/12` ? Thanks in advice

cheina97 (changes requested):

cheina97 (approved):

PR #2921, opened 2025-02-06
fix: PodMonitor

kariya-mitsuru:
# Description

Fix `PodMonitor`.

* crd-replicator

    Commit 10ef3087 changed the default port for metrics from `8080` to `8082`, but crd-replicator was not changed because there is no command line option. However, only the `Deployment` port was changed to 8082, so `PodMonitor` no longer works properly.

    This commit reverts the `Deployment` port back to `8080`.

* VirtualKubelet

    Commit ddafea075 changed the label of VirtualKubelet but did not change the `PodMonitor` selector so it no longer works correctly.

    This commit makes `PodMonitor`'s selector follow commit ddafea075.

### DIFF
--- a/cmd/crd-replicator/main.go
+++ b/cmd/crd-replicator/main.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	liqov1beta1 "github.com/liqotech/liqo/apis/core/v1beta1"
 	crdreplicator "github.com/liqotech/liqo/internal/crdReplicator"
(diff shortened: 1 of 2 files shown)


deployments/liqo/templates/liqo-crd-replicator-deployment.yaml
```diff
@@ -50,7 +50,7 @@ spec:
           resources: {{- toYaml .Values.crdReplicator.pod.resources | nindent 12 }}
           ports:
           - name: metrics
-            containerPort: 8082
```
cheina97, line 53: Hi @kariya-mitsuru, thanks for your PR. Can you keep the "8082" port and add a flag inside the CRD replicator to set the port please?

You need to edit /cmd/crd-replicator/main.go, check this https://github.com/liqotech/liqo/blob/c2f889a26d38a1785310b56290c28419b8892761/cmd/liqo-controller-manager/main.go#L106 to have a guideline about how to implement it.

cheina97, line 53: Hi @kariya-mitsuru, thanks for your PR. Can you keep the "8082" port and add a flag inside the CRD replicator to set the port, please?

Could you edit /cmd/crd-replicator/main.go.

Check this https://github.com/liqotech/liqo/blob/c2f889a26d38a1785310b56290c28419b8892761/cmd/liqo-controller-manager/main.go#L106 to have a guideline about how to implement it.

cheina97 (changes requested):

cheina97 (approved):

PR #2954, opened 2025-02-20
[liqoctl] reduce permissions required on provider cluster

claudiolor:
# Description

This PR reduces the permissions required on the remote cluster, to create a peering connection. It includes:
- **Reduction of permissions on namespaces**: liqoctl checks whether a Tenant namespace exists, to do it, we listed all the namespaces of the provider with some given labels. This PR adds a fallback so that when the user has no permissions to list the namespaces, it directly gets the tenant namespace with the default name. 
- **Removed access to remote ForeignCluster resource**: during unpeering, liqoctl needs to check whether peering is bidirection, as, in that case, it should not remove the network. To do so it checked the foreigncluster resource of both consumer and provider clusters. However, a check only on the consumer is enough. 
- **Limit the scope of the list made on the provider cluster**: made all the list of network resources (Configurations, Networks, PublicKeys) namespace-scoped.


aleoli (approved):

aleoli:
@claudiolor, please add a description to the pr

claudiolor:
> @claudiolor, please add a description to the pr

Sorry, I completely forgot it. Description updated!

aleoli (approved):

fra98 (approved):

pkg/utils/getters/k8sGetters.go
```diff
@@ -712,10 +712,12 @@ func ListConfigurationsByLabel(ctx context.Context, cl client.Client, lSelector
...
 func GetConfigurationByClusterID(ctx context.Context, cl client.Client,
-	clusterID liqov1beta1.ClusterID) (*networkingv1beta1.Configuration, error) {
+	clusterID liqov1beta1.ClusterID, tenantNamespace string) (*networkingv1beta1.Configuration, error) {
 	remoteClusterIDSelector := labels.Set{consts.RemoteClusterID: string(clusterID)}.AsSelector()
-	configurations, err := ListConfigurationsInNamespaceByLabel(ctx, cl, corev1.NamespaceAll, remoteClusterIDSelector)
+
+	configurations, err := ListConfigurationsInNamespaceByLabel(ctx, cl, tenantNamespace, remoteClusterIDSelector)
 	if err != nil {
 		return nil, err
 	}
```
fra98, line 723: Why tenantnamespace? Since it is a networking resource, it is not guaranteed

pkg/liqoctl/network/cluster.go
```diff
@@ -503,11 +513,12 @@ func (c *Cluster) EnsurePublicKey(ctx context.Context, remoteClusterID liqov1bet
 }
 
 // DeleteConfiguration deletes a Configuration.
-func (c *Cluster) DeleteConfiguration(ctx context.Context, remoteClusterID liqov1beta1.ClusterID) error {
+// If tenantNamespace is empty this function searches in all the namespaces in the cluster.
+func (c *Cluster) DeleteConfiguration(ctx context.Context, remoteClusterID liqov1beta1.ClusterID, tenantNs string) error {
```
fra98, line 517: Same here

PR #2955, opened 2025-02-20
Adding information about community (how-to, meetings, etc)

frisso:


cheina97:
Hi @frisso, i think we should merge this PR when the links will be available. For the moment I think we should keep this PR in draft.

PR #2956, opened 2025-02-20
[E2E] kubeadm provider refactoring

cheina97:
# Description

This PR replace the infrastructure name from "cluster-api" to "kubeconfig" and implement the download of clusterctl.


fra98 (approved):
We also need to update the e2e targets that use ../liqoctl/.. in the Makefile

PR #2971, opened 2025-03-03
fix(liqoctl): rename `producer` occurences to `provider`

fra98:
# Description

This PR renames all occurrences of `producer` to `provider` to be more coherent with the terminology adopted.


frisso:
Such a beautiful PR: I like so much these readability-oriented refactoring in the code!
(and usually I clashes with developers that hate doing that :-) )

frisso (approved):

PR #3039, opened 2025-05-07
[Telemetry] UUID

cheina97:
# Description

This PR allows telemetry job to generate a UUID to use as ClusterID
This increases the privacy and avoid cases in which 2 clusters have been set with the same ClusterID


pkg/telemetry/builder.go
```diff
@@ -34,9 +36,16 @@ import (
...
-		return nil, err
+		if apierrors.IsNotFound(err) {
+			klog.V(4).InfoS("ClusterID not found, creating a new one")
+			if err := createClusterIDTelemetryConfigMap(ctx, c.Client, c.Namespace); err != nil {
+				return nil, fmt.Errorf("failed to create configmap: %w", err)
+			}
+		} else {
+			return nil, err
+		}
 	}
```
aleoli, line 49: Can we use a switch here for error handling and better readability?

aleoli (approved):

PR #3054, opened 2025-05-27
feat: unpeering force

### LINKED ISSUE #3051: Force unpeering

When I try to unpeer a provider cluster and it fails because the remote no more available or reachable, I can’t  unpeer anymore it. This doesn’t allow me to remove outdated peering from the liqo cluster.

When I cannot contact the provider cluster, I would like to be able to use an option that allows me to forcibly close the peering connection with the provider cluster.

This feature can help the network team to be able to remove the connection between two clusters despite dynamic changes in the IP of the remote cluster or in case of cluster provider unrecoverable unavailability, ensuring the continued use of the liqo cluster.

### Describe your proposed solution

Two flags can be added to perform a forced unpeering:
 `--force`, a Boolean flag that force the unpeering even if the provider cluster is unreachable, and `--remote-cluster-id`, a flag that must be entered in conjunction with `--force` and must contain the cluster id of the provider cluster.

### Code of Conduct

giuliafalaschi:
# Description

Two flags have been added to perform a forced unpeering:
` --force`, a Boolean flag that force the unpeering even if the provider cluster is unreachable, and `--remote-cluster-id`, a flag that must be entered in conjunction with` --force` and must contain the cluster id of the provider cluster.

Fixes #3051  

# How Has This Been Tested?
I created two local clusters using kind, installed Liqo on them, and established a peering relationship between the two. To simulate a catastrophic scenario, I modified the IP address of the provider cluster to make it unreachable. Finally, I executed the unpeer command with the options I implemented, specifically:
`liqoctl unpeer --force --remote-cluster-id <cluster-id>`
In this way, the unpeering was successfully performed on the consumer cluster, even though the provider was no longer reachable.


claudiolor:
Hi @giuliafalaschi, thanks for this PR. Is this ready for review?

claudiolor:
Hi Giulia 😊

Thank you so much for your contribution and the effort you put into this PR, it's really appreciated! 🙏
Apologies for the delay in getting back to you, I got caught up with other tasks but I’ve now had the chance to review and test your changes.

That said, I noticed some issues related to how the forced unpeering is handled. Specifically, there are some challenges that need to be handled when the remote cluster becomes unreachable, which are not addressed in your implementation. The two main components that cause problems in such scenarios are:

- **CRDReplicator**: whose role is [reflecting Liqo CR](https://github.com/liqotech/liqo/blob/master/internal/crdReplicator/reflection/handler.go#L53)  (e.g., ResourceSlice) on the remote cluster, and It attempts to clean them up when they are deleted. If the remote cluster is no longer reachable, the finalizers won’t be removed, preventing the resources from being deleted.

- **VirtualNode**: When a ResourceSlice is deleted, the associated virtual node also gets removed, triggering the [VirtualNode controller](https://github.com/liqotech/liqo/blob/c39e9fce0a99cc247a3fae7b27852ad2ae683ca6/pkg/liqo-controller-manager/offloading/virtualnode-controller/virtualnode_controller.go#L116), which starts a process that [tries to evict pods scheduled on the virtual node](https://github.com/liqotech/liqo/blob/master/pkg/liqo-controller-manager/offloading/virtualnode-controller/deletion-routine.go#L185). The [Virtual Kubelet's reflector](https://github.com/liqotech/liqo/blob/c39e9fce0a99cc247a3fae7b27852ad2ae683ca6/pkg/virtualKubelet/reflection/workload/podns.go#L117) will take in charge the eviction of the pods, but if the remote cluster is unreachable, this eviction gets stuck.

Because of this, if you try to force unpeer two kind clusters and then delete one of them, with your version of the code you’ll notice that the forced unpeer hangs on the ResourceSlice deletion.

To properly handle these situations, we need to introduce a way to inform that the remote cluster is no longer expected to respond. That way, components like the CRDReplicator or Virtual Kubelet can stop trying to reach it endlessly. Here are a couple of ideas I could come up with:

1. Use the [controller manager’s API server checker](https://github.com/liqotech/liqo/blob/c39e9fce0a99cc247a3fae7b27852ad2ae683ca6/pkg/liqo-controller-manager/core/foreigncluster-controller/apiserverchecker.go#L39) to flag a ForeignCluster as "dead" after a number of failed health checks. However, this might introduce issues in case of temporary downtimes, if the cluster is considered "dead", the Virtual Kubelet will mark the pod scheduled on the virtual node as terminated. However, if the connection comes back up, then the pod running remotely will never be stopped.

2. Explicitly mark a cluster as unreachable in the ForeignCluster spec (for example, when forcing unpeer and the cluster doesn’t respond). Then the component listed above can skip trying to delete remote resources, so cleanly removing the peering connection on the consumer cluster.

Once we handle this properly in the Liqo core, we can refine the behavior of the liqoctl unpeer command. I saw you modified the flow a bit, but I would suggest keeping the original flow and falling back to forced unpeer only if the standard one fails. For example, if we can’t get the cluster ID of the provider during unpeering, we can attempt with a short timeout. If it fails and `force` is enabled, we fall back to the user-supplied cluster ID and performing the operation required to force the unpeer (for example, following the second option above, marking the ForeignCluster as "dead").
Pay attention to keep the utility functions as "agnostic" as possible, not introducing on them the concept of "force". For example, you can add a flag to skip remote operations or giving the possibility to manually provide the data it tries to get from the remote cluster via API server.

Lastly, regarding the CLI flags: instead of having both `--force` and `--remote-cluster-id` (where the latter only makes sense with `--force`), it might be cleaner to have a single flag like `--force-with-cluster-id` that provides the fallback cluster ID directly.

Thanks again for your work! Let me know your thoughts on this and feel free to ask if you any questions 😊

frisso:
@giuliafalaschi Is this PR ready for the second review?

giuliafalaschi:
@frisso Good morning, I made some modifications to make my change more robust. I have verified that it works and leaves no dangling objects. I am cleaning up the code from unnecessary logs and as soon as possible I will update my pull request.

claudiolor:
As agreed with Giulia offline: closing this PR in favor of two separate PRs, one implementing this feature in the Liqo core (#3122) another implementing the liqoctl command to force unpeer with a remote cluster, she is going to open.

Issue #3057, opened 2025-06-05
[Feature] Namespace offloading in batches


air-31:
### Describe the problem you are having (if any)

We are working in an environment with a considerably high number of namespaces (on the order of thousands). We would like to offload a larger part of them, however it is somewhat impractical to do so with the currently available offloading procedure.

### Describe the solution you would like

It would be nice to upgrade the `liqoctl offload` command to include an option to offload namespaces in batches via the usage of label selectors (e.g. `liqoctl offload namespaces --label 'k8s.io/type=tenant'`). Subsequently, the `liqoctl unoffload` command is to be upgraded accordingly as well.

### Describe the user value of this feature

This would provide users with a direct way to offload a number of namespaces at once.

### Do you volunteer to implement this feature?

- [ ] I want to implement this feature

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

claudiolor:
Hi @air-31, 
Thanks for opening this issue! 
I agree this would be a nice addition to the liqoctl CLI, it should be fairly straightforward to implement and could be a great candidate for a first contribution

That said, I think you can still offload/unoffload namespaces in batch by first selecting the namespaces and then piping their names into the `liqoctl offload/unoffload` command. 
For example, you can use the following command to offload a batch of namespaces based on the label:

```bash
k get ns -l k8s.io/type=tenant -o custom-columns=":metadata.name" --no-headers | xargs -n 1 liqoctl offload namespace
```

Similarly to unoffload the namespaces:

```bash
k get ns -l k8s.io/type=tenant -o custom-columns=":metadata.name" --no-headers | xargs -n 1 liqoctl unoffload namespace --skip-confirm
```

mayooot:
I can do this.

claudiolor:
Hi @mayooot!
thank you for volunteering! I will be more than happy to review your PR once ready :) 
If you need any help or you have any questions feel free ask! You can join our [slack workspace](https://liqo-io.slack.com/join/shared_invite/zt-h20212gg-g24YvN6MKiD9bacFeqZttQ) to directly reach us.

mayooot:
I am new to the liqo project and know very little about it.
I have added an offload subcommand: namespaces which can be used: `liqoctl offload namespaces --label-selector='app=namespaces'`.

@claudiolor Cloud you give me some suggest about this https://github.com/liqotech/liqo/pull/3060.
If possible, I will finish the remaining development work abount `unoffload`.

[...]

PR #3060, opened 2025-06-12
feat(offload): Add the liqoctl offload namespaces command.

mayooot:
Multiple namespaces can be offloaded.

# Description

Added the liqoctl offload namespaces command. Matching namespaces can be found using `--label-selector`.

Fixes #(issue)

#3057

# How Has This Been Tested?
Yes, I hava tested it locally.

~~~
 $ liqoctl offload namespaces --label-selector='app=namespaces'
 INFO  Offloading of namespace "need01" correctly enabled
 INFO  Offloading completed successfully
 INFO  Offloading of namespace "need03" correctly enabled
 INFO  Offloading completed successfully
~~~

### DIFF
--- a/cmd/liqoctl/cmd/offload.go
+++ b/cmd/liqoctl/cmd/offload.go
@@ -64,6 +64,20 @@ or (output the NamespaceOffloading resource as a yaml manifest, without applying
   $ {{ .Executable }} offload namespace foo --output yaml
 `
 
+const liqoctlOffloadNamespacesLongHelp = `Offload all namespaces with optional label filtering.
+
+This command offloads all namespaces in the cluster to remote clusters, optionally filtered by a label selector.
+It applies the same offloading configuration to each matching namespace, similar to 'liqoctl offload namespace <name>'.
+
+Examples:
+  $ {{ .Executable }} offload namespaces
+or
+  $ {{ .Executable }} offload namespaces --selector 'key=value'
+or
+  $ {{ .Executable }} offload namespaces --selector 'key=value,env!=prod' \
+      --pod-offloading-strategy Remote --namespace-mapping-strategy EnforceSameName
+`
+
 func newOffloadCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "offload",
@@ -73,6 +87,7 @@ func newOffloadCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
 	}
 
 	utils.AddCommand(cmd, newOffloadNamespaceCommand(ctx, f))
+	utils.AddCommand(cmd, newOffloadNamespacesCommand(ctx, f))
 	return cmd
 }
 
@@ -137,6 +152,76 @@ func newOffloadNamespaceCommand(ctx context.Context, f *factory.Factory) *cobra.
 	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("pod-offloading-strategy", completion.Enumeration(podOffloadingStrategy.Allowed)))
 	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("namespace-mapping-strategy", completion.Enumeration(namespaceMappingStrategy.Allowed)))
 	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("selector", completion.LabelsSelector(ctx, f, completion.NoLimit)))
+	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("output", completion.Enumeration(outputFormat.Allowed)))
+
+	return cmd
+}
+
+func newOffloadNamespacesCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
+	var selectors []string
+	var labelSelector string
+
+	podOffloadingStrategy := args.NewEnum([]string{
+		string(offloadingv1beta1.LocalAndRemotePodOffloadingStrategyType),
+		string(offloadingv1beta1.RemotePodOffloadingStrategyType),
+		string(offloadingv1beta1.LocalPodOffloadingStrategyType)},
+		string(offloadingv1beta1.LocalAndRemotePodOffloadingStrategyType))
+
+	namespaceMappingStrategy := args.NewEnum([]string{
+		string(offloadingv1beta1.EnforceSameNameMappingStrategyType),
+		string(offloadingv1beta1.DefaultNameMappingStrategyType),
+		string(offloadingv1beta1.SelectedNameMappingStrategyType)},
+		string(offloadingv1beta1.DefaultNameMappingStrategyType))
+
+	var remoteNamespaceName = ""
+
+	outputFormat := args.NewEnum([]string{"json", "yaml"}, "")
+
+	options := offload.Options{Factory: f}
+
+	cmd := &cobra.Command{
+		Use:     "namespaces",
+		Aliases: []string{"ns"},
+		Short:   "Offload all namespaces with optional label filtering",
+		Long:    liqoctlOffloadNamespacesLongHelp,
+
+		Args: cobra.NoArgs,
+
+		PreRun: func(_ *cobra.Command, _ []string) {
+			options.PodOffloadingStrategy = offloadingv1beta1.PodOffloadingStrategyType(podOffloadingStrategy.Value)
+			options.NamespaceMappingStrategy = offloadingv1beta1.NamespaceMappingStrategyType(namespaceMappingStrategy.Value)
+			options.RemoteNamespaceName = remoteNamespaceName
+			options.OutputFormat = outputFormat.Value
+			options.LabelSelector = labelSelector
+			options.Printer.CheckErr(options.ParseClusterSelectors(selectors))
+		},
+
+		Run: func(_ *cobra.Command, _ []string) {
+			output.ExitOnErr(options.OffloadNamespaces(ctx))
+		},
+	}
+
+	cmd.Flags().Var(podOffloadingStrategy, "pod-offloading-strategy",
+		"The constraints regarding pods scheduling in this namespace, among Local, Remote and LocalAndRemote")
+	cmd.Flags().Var(namespaceMappingStrategy, "namespace-mapping-strategy",
+		"The naming strategy adopted for the creation of remote namespaces, among DefaultName, EnforceSameName and SelectedName")
+	cmd.Flags().DurationVar(&options.Timeout, "timeout", 20*time.Second, "The timeout for the offloading process")
+	cmd.Flags().StringVar(&remoteNamespaceName, "remote-namespace-name", "",
+		"The name of the remote namespace, required when using the SelectedName NamespaceMappingStrategy. "+
+			"Otherwise, it is ignored")
+
+	cmd.Flags().StringArrayVarP(&selectors, "selector", "l", []string{},
+		"The selector to filter the target clusters. Can be specified multiple times, defining alternative requirements (i.e., in logical OR)")
+	cmd.Flags().StringVar(&labelSelector, "label-selector", "",
+		"Selector (label query) to filter namespaces, supports '=', '==', and '!=' (e.g., -l key1=value1,key2=value2).")
+
+	cmd.Flags().VarP(outputFormat, "output", "o",
+		"Output the resulting NamespaceOffloading resource, instead of applying it. Supported formats: json, yaml")
+
+	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("pod-offloading-strategy", completion.Enumeration(podOffloadingStrategy.Allowed)))
+	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("namespace-mapping-strategy", completion.Enumeration(namespaceMappingStrategy.Allowed)))
+	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("selector", completion.LabelsSelector(ctx, f, completion.NoLimit)))
+	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("label-selector", completion.NamespacesSelector(ctx, f, completion.NoLimit)))
 	f.Printer.CheckErr(cmd.RegisterFlagCompletionFunc("output", completion.Enumeration(outputFormat.Allowed)))
 
 	return cmd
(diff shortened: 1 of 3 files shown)


frisso:
This looks very good to me, thanks!
Final words to @claudiolor , who has a deeper understanding of the source code.

mayooot:
@claudiolor Please check it.

claudiolor:
Hi @mayooot thank you so much for your PR :D

I would not create a new command with a single letter of difference to another as it seems to me really prone to user errors, while instead I would reuse the same command with [some aliases](https://github.com/liqotech/liqo/blob/master/cmd/liqoctl/cmd/offload.go#L101)  so that `liqoctl offload namespace` can be `liqoctl offload namespaces` or `liqoctl offload ns` and add the `--selector` option to it.
So according to this, we can allow the user to provide more than one namespace name [removing this option from the command](https://github.com/liqotech/liqo/blob/master/cmd/liqoctl/cmd/offload.go#L105).
At this point we can move the logic of [offload/handler.go from line 75 below](https://github.com/liqotech/liqo/blob/master/pkg/liqoctl/offload/handler.go#L75) on a separate function, so that we iterate over the namespaces indicated by the user or the ones retrieved via the label selector and call this function for each of the namespaces.

Of course, we should return an error if the user specified both the name of the namespaces of the label selectors or none of them.

In this way I think we are reducing the possibility that a user makes a mistake, we are giving the possibility to offload more than one namespace by name at a time and there is no need to duplicate code in order to introduce a new command.
Let me know what you think about it :)

mayooot:
@claudiolor You're right, `namespace` and `namespaces` do make it easy to get confused.
I tried to minimize changes to the source code during development, so I added a new command that minimizes the impact on the original functionality.

It's right to make changes to the `namespace` command directly, so that the processing of a single namespace and multiple namespaces can be put into a `Run` function, and it's more optimized for the user.

I will update the code as we discussed.
Thanks a lot!

mayooot:
@claudiolor I've done that.
It can be used below:

~~~bash
$ liqoctl offload namespace lowkey01
 INFO  Offloading of namespace "lowkey01" correctly enabled
 INFO  Offloading completed successfully

$ liqoctl offload namespace lowkey02 lowkey03 lowkey04
 INFO  Offloading of namespace "lowkey02" correctly enabled
 INFO  Offloading completed successfully
 INFO  Offloading of namespace "lowkey03" correctly enabled
 INFO  Offloading completed successfully
 INFO  Offloading of namespace "lowkey04" correctly enabled
 INFO  Offloading completed successfully

$ liqoctl offload namespace --label-selector="bunch=yes"
 INFO  Offloading of namespace "highkey01" correctly enabled
 INFO  Offloading completed successfully
 INFO  Offloading of namespace "highkey02" correctly enabled
 INFO  Offloading completed successfully
 INFO  Offloading of namespace "highkey03" correctly enabled
 INFO  Offloading completed successfully
~~~

claudiolor:
Hi @mayooot thanks a lot! :) I will check it ASAP

cmd/liqoctl/cmd/offload.go
```diff
@@ -98,23 +104,29 @@ func newOffloadNamespaceCommand(ctx context.Context, f *factory.Factory) *cobra.
...
 		},
-
 		Run: func(_ *cobra.Command, args []string) {
-			options.Namespace = args[0]
+			if len(args) == 0 && labelSelector == "" {
+				output.ExitOnErr(fmt.Errorf("namespace name or label selector must be specified"))
+			}
+			if len(args) != 0 && labelSelector != "" {
+				output.ExitOnErr(fmt.Errorf("namespace name and label selector must not be specified together"))
+			}
```
claudiolor: Here you are not printing the error message, I'd suggest to use the `options.Printer.ExitWithMessage` function, as done elsewhere in the codebase:
```suggestion
			if len(args) == 0 && labelSelector == "" {
				options.Printer.ExitWithMessage("namespace name or label selector must be specified")
			}
			if len(args) != 0 && labelSelector != "" {
				options.Printer.ExitWithMessage("namespace name and label selector must not be specified together")
			}
```

pkg/liqoctl/offload/handler.go
```diff
@@ -37,6 +40,8 @@ type Options struct {
 	*factory.Factory
 
 	Namespace                string
```
claudiolor: I'd remove single namespace from options
```suggestion
```

pkg/liqoctl/offload/handler.go
```diff
@@ -72,6 +77,63 @@ func (o *Options) Run(ctx context.Context) error {
...
+			NamespaceMappingStrategy: o.NamespaceMappingStrategy,
+			RemoteNamespaceName:      o.RemoteNamespaceName,
+			ClusterSelector:          o.ClusterSelector,
+			OutputFormat:             o.OutputFormat,
+			Timeout:                  o.Timeout,
+		}
+
+		if err := nsOptions.runOffload(ctx); err != nil {
+			errors = append(errors, err)
+		}
```
claudiolor: I couldn't understand why you create a new option struct for each namespace you need to offload.

I would just call an internal function where we provide the namespace name, it seems more straightforward, and there is less risk to forget passing some of the properties of the option struct:
```suggestion
		if err := o.runOffload(ctx, ns); err != nil {
			errors = append(errors, err)
		}
```

pkg/liqoctl/offload/handler.go
```diff
@@ -72,6 +77,63 @@ func (o *Options) Run(ctx context.Context) error {
...
+	}
+
+	return nil
+}
+
+// runOffload encapsulates the core logic of the offload namespace command.
+func (o *Options) runOffload(ctx context.Context) error {
 	// Output the NamespaceOffloading resource, instead of applying it.
 	if o.OutputFormat != "" {
 		o.Printer.CheckErr(o.output())
```
claudiolor: When the output is `yaml`, we need to append the end directive (`---`) to each of the printed object to produce a valid yaml output

cmd/liqoctl/cmd/offload.go
```diff
@@ -98,23 +104,29 @@ func newOffloadNamespaceCommand(ctx context.Context, f *factory.Factory) *cobra.
...
 	cmd := &cobra.Command{
 		Use:     "namespace name",
-		Aliases: []string{"ns"},
-		Short:   "Offload a namespace to remote clusters",
+		Aliases: []string{"ns", "namespaces"},
+		Short:   "Offload namespaces to remote clusters",
 		Long:    liqoctlOffloadNamespaceLongHelp,
 
-		Args:              cobra.ExactArgs(1),
 		ValidArgsFunction: completion.Namespaces(ctx, f, 1),
```
claudiolor: To enable suggestions for more than one arg
```suggestion
		ValidArgsFunction: completion.Namespaces(ctx, f, completion.NoLimit),
```

cmd/liqoctl/cmd/offload.go
```diff
@@ -130,13 +142,16 @@ func newOffloadNamespaceCommand(ctx context.Context, f *factory.Factory) *cobra.
 
 	cmd.Flags().StringArrayVarP(&selectors, "selector", "l", []string{},
 		"The selector to filter the target clusters. Can be specified multiple times, defining alternative requirements (i.e., in logical OR)")
+	cmd.Flags().StringVar(&labelSelector, "label-selector", "",
```
claudiolor: Wondering whether make sense calling this option something like `--ns-selector` or `--ns-label-selector` to avoid confusing this flag with `--selector` option, which is about the target cluster.
Your and @frisso thoughts are appreciated on this

mayooot: I prefer `--ns-selector`.

frisso: Fine for me.

claudiolor: Nice, let's do `--ns-selector`

frisso:
@mayooot This looks very good to me (@claudiolor for more precise notes on the code).
 To be completed, we would need two more steps:
 * align also the 'unoffload' command with the new features
 * align the documentation (https://docs.liqo.io/en/latest/usage/namespace-offloading.html) with the new features

pkg/liqoctl/offload/handler.go
```diff
@@ -72,12 +77,59 @@ func (o *Options) Run(ctx context.Context) error {
...
+	}
+
+	return nil
+}
+
+// runOffload encapsulates the core logic of the offload namespace command.
+func (o *Options) runOffload(ctx context.Context, namespace string) error {
 	// Output the NamespaceOffloading resource, instead of applying it.
 	if o.OutputFormat != "" {
 		o.Printer.CheckErr(o.output())
```
claudiolor: I believe this doesn't work, you are not providing the name of the namespace to the output function:
```suggestion
		o.Printer.CheckErr(o.output(namespace))
```

pkg/liqoctl/offload/handler.go
```diff
@@ -119,6 +171,7 @@ func (o *Options) Run(ctx context.Context) error {
 // output implements the logic to output the generated NamespaceOffloading resource.
 func (o *Options) output() error {
```
claudiolor: Because of the command above:
```suggestion
func (o *Options) output(namespace string) error {
```

pkg/liqoctl/offload/handler.go
```diff
@@ -139,6 +192,18 @@ func (o *Options) output() error {
 		},
 	}
 
+	// Print the object to a buffer to capture YAML output
+	if o.OutputFormat == "yaml" {
```
claudiolor: If you move the definition of the `nsoff` variable just above the switch...case, I believe we can avoid this if condition, and place its content inside the "yaml" case of the switch...case above

cmd/liqoctl/cmd/offload.go
```diff
@@ -98,23 +103,28 @@ func newOffloadNamespaceCommand(ctx context.Context, f *factory.Factory) *cobra.
...
 		},
-
 		Run: func(_ *cobra.Command, args []string) {
-			options.Namespace = args[0]
+			if len(args) == 0 && labelSelector == "" {
+				options.Printer.ExitWithMessage("namespace name or label selector must be specified")
+			}
+			if len(args) != 0 && labelSelector != "" {
+				options.Printer.ExitWithMessage("namespace name and label selector must not be specified together")
+			}
```
claudiolor, line 126: We need to add an additional check to make sure that no custom remote namespace name is specified when multiple namespaces are targeted for offloading:
```suggestion
			}
			if (len(args) > 1 || labelSelector != "") && remoteNamespaceName != "" {
				options.Printer.ExitWithMessage("the remote namespace name cannot be specified when offloading multiple namespaces at once")
			}
```

mayooot:
@claudiolor All done. Cloud you check it when you have a moment?

cmd/liqoctl/cmd/offload.go
```diff
@@ -98,23 +103,31 @@ func newOffloadNamespaceCommand(ctx context.Context, f *factory.Factory) *cobra.
...
+			if (len(args) > 1 && labelSelector != "") && remoteNamespaceName != "" {
```
claudiolor: This should be an OR:
```suggestion
			if (len(args) > 1 || labelSelector != "") && remoteNamespaceName != "" {
```

mayooot:
> accordingly

Ok, I've fixed the bugs, and I'll implement `unoffload` command ASAP.

mayooot:
> Thank you for fixing the issues I reported! 💯 Once the small issue I underlined is fixed we can consider the offload command completed. To merge this PR we still need to implement a similar behavior with the `unoffload` command and the [documentation](https://github.com/liqotech/liqo/blob/master/docs/usage/namespace-offloading.md) needs to be updated accordingly.

Ok, I've fixed the bugs, and I'll implement unoffload command ASAP.

mayooot:
@claudiolor All is done. Please check it and let me know your suggestions about the code and docs :)

claudiolor:
Thanks @mayooot, I will check these changes ASAP

docs/usage/namespace-offloading.md
```diff
@@ -136,6 +145,10 @@ The offloading of a namespace can be disabled through the dedicated *liqoctl* co
 
 ```bash
 liqoctl unoffload namespace foo
+
+liqoctl unoffload namespaces foo bar 
+
+liqoctl unoffload namespaces --ns-selector='foo=bar'
 ```
```
claudiolor:
```suggestion
```bash
# unoffload single namespace by name
liqoctl unoffload namespace foo

# unoffload multiple namespaces by name
liqoctl unoffload namespaces foo bar 

# unoffload multiple namespaces by label selector
liqoctl unoffload namespaces --ns-selector='foo=bar'
```
```

File: pkg/liqoctl/unoffload/handler.go [hunk 13]
```diff
@@ -40,6 +46,51 @@ func (o *Options) Run(ctx context.Context) error {
...
+
+	if len(errors) > 0 {
+		return fmt.Errorf("encountered %d errors during unoffloading, error details: %v", len(errors), errors)
+	}
+
+	return nil
+}
+
+func (o *Options) runUnoffload(ctx context.Context, namespace string) error {
+	o.Namespace = namespace
```
claudiolor: Please do not use o.Namespace, I think you can directly use the `namespace` variable

claudiolor (commented):
Thank you for implementing also unoffload!
- I left just a couple of comment.
- Before merging we also need to update the automatically generated `liqoctl` documentation, you can update the documentation by launching `make ctldocs`, then you can commit the updated files.

Once those two points are addressed I believe we can merge

mayooot:
> 

Ok, I'me done :)

mayooot:
> Thank you for implementing also unoffload!
> 
> * I left just a couple of comment.
> * Before merging we also need to update the automatically generated `liqoctl` documentation, you can update the documentation by launching `make ctldocs`, then you can commit the updated files.
> 
> Once those two points are addressed I believe we can merge

@claudiolor I'm done here :)

claudiolor:
@mayooot thank you! @aleoli can you please check and merge this?

claudiolor (approved):

aleoli:
Hi @mayooot! It looks good to me.

Could you please squash all your commits into a single commit? Then we can proceed with the merge. Thanks!

mayooot:
> Hi @mayooot! It looks good to me.
> 
> Could you please squash all your commits into a single commit? Then we can proceed with the merge. Thanks!

Sure, I just squashed them.

PR #3077, opened 2025-07-15
[Network] refactor InternalFabric name source

cheina97:
# Description

Fixes #(issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B


.github/workflows/integration.yml
```diff
@@ -7,6 +7,7 @@ on:
       - master
       - release
       - ipam
+      - frc/internalfabric
```
aleoli:
```suggestion
```

pkg/utils/getters/dataGetters.go
```diff
@@ -180,3 +180,18 @@ func RetrieveClusterIDsFromObjectsLabels[T metav1.Object](objectList []T) []stri
 	}
 	return slices.Collect(maps.Keys(clusterIDs))
 }
+
+// RetrieveClusterIDFromObject retrieves the remote cluster ID from the labels of a generic kubernetes object.
+func RetrieveClusterIDFromObject[T metav1.Object](object T) (liqov1beta1.ClusterID, error) {
```
aleoli: I think you can use interfaces here instead of generics, since Object is an interface

pkg/liqo-controller-manager/networking/internal-network/client-controller/client_controller.go
```diff
@@ -106,9 +106,14 @@ func (r *ClientReconciler) ensureInternalFabric(ctx context.Context, gwClient *n
 		return fmt.Errorf("internal endpoint not found for the gateway client %q", gwClient.Name)
 	}
 
+	internalFabricName, err := getters.RetrieveClusterIDFromObject(gwClient)
+	if err != nil {
+		return fmt.Errorf("unable to retrieve the cluster ID from the gateway client %q", gwClient.Name)
+	}
```
aleoli: In that way, if we update the liqo controllers, we will have two internal fabrics created for each remote cluster. Can we check that no other internal fabric exists for the same cluster ID? In that way, we can support a non-breaking update for this version. The same for the other controller

frisso:
@cheina97 First, thanks for this nice contribution.
A suggestion: if possible, it would be nice to have the name as "gw-clusterid" or something similar, in order to easily understand that this is a gateway pod, and not just the "clusterid".
Finally, a note: in the future we may have multiple gateways connected to the same cluster, e.g., for load balancing, so even a name that includes the "remote cluster id" may not be enough to distinguish the pods.

cheina97:
> @cheina97 First, thanks for this nice contribution. A suggestion: if possible, it would be nice to have the name as "gw-clusterid" or something similar, in order to easily understand that this is a gateway pod, and not just the "clusterid". Finally, a note: in the future we may have multiple gateways connected to the same cluster, e.g., for load balancing, so even a name that includes the "remote cluster id" may not be enough to distinguish the pods.

The name for gateway is still fully customizable, the only resource that is changing name is the InternalFabric that will use the remote cluster ID

pkg/gateway/fabric/geneve/internalfabric.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	networkingv1beta1 "github.com/liqotech/liqo/apis/networking/v1beta1"
+)
+
+// getInternalFabric retrieves the InternalFabric resource associated with the given GatewayServer.
+// WARNING: this function contains 2 calls to the Kubernetes API using 2 different names.
+// This is intended to avoid breaking changes, since the InternalFabric name has changed from GatewayServer to the GatewayServer name.
```
aleoli:
```suggestion
// This is intended to avoid breaking changes, since the InternalFabric name has changed from GatewayServer name to the GatewayServer cluster ID.
```

pkg/liqo-controller-manager/networking/internal-network/client-controller/client_controller.go
```diff
@@ -106,9 +107,14 @@ func (r *ClientReconciler) ensureInternalFabric(ctx context.Context, gwClient *n
 		return fmt.Errorf("internal endpoint not found for the gateway client %q", gwClient.Name)
 	}
 
+	internalFabricName, err := netutils.ForgeInternalFabricName(ctx, r.Client, &gwClient.ObjectMeta)
+	if err != nil {
+		return fmt.Errorf("unable to retrieve the cluster ID from the gateway client %q", gwClient.Name)
+	}
```
fra98, line 113: Nit: why are we getting the remote cluster ID from the object if we have it already here? I think we can just return the string statically (e.g., `func ForgeInternalFabricName(remoteClusterID liqov1beta1.ClusterID) { return remoteClusterID }`).
It seems safer (no possible errors raised, no contacting the apiserver).

pkg/gateway/fabric/geneve/internalfabric.go
```diff
@@ -0,0 +1,53 @@
...
+// This is intended to avoid breaking changes, since the InternalFabric name has changed from GatewayServer name to the GatewayServer cluster ID.
+func getInternalFabric(ctx context.Context, cl client.Client, gatewayName, remoteID, ns string) (*networkingv1beta1.InternalFabric, error) {
+	internalFabric := &networkingv1beta1.InternalFabric{}
+	err := cl.Get(ctx, client.ObjectKey{
+		Name:      remoteID,
+		Namespace: ns,
+	}, internalFabric)
+	if client.IgnoreNotFound(err) != nil {
+		return nil, fmt.Errorf("unable to get the internal fabric %q: %w", remoteID, err)
+	}
```
fra98, line 38: The real reason for the above comment arises since here it is not clear why we are using the remoteID, and it is harder to maintain or lead to bugs if we change the logic of the ForgeInternalFabric() name again. Instead, we can call the forge method again, passing as argument only the remoteClusterID (also here we have it directly in the function)

PR #3083, opened 2025-07-18
[Networking] internalfabric getter name refactoring

cheina97:
This PR refactors the logic used by **getInternalFabric** to get the internalfabric name. We have a function called ForgeInternalFabricName which is already used in other code places.


fra98:
Closing PR as fixed by https://github.com/liqotech/liqo/pull/3271

PR #3098, opened 2025-08-10
fix(setup.sh): Added a sleep to ensure correct creation of both clusters

Deblazz:
# Description
Added a sleep between clusters creation in setup.sh, since otherwise the second one would fail due to a timeout.


claudiolor:
Hi @Deblazz, thanks for this contribution!
I tried to reproduce the issue you reported but the Kind clusters were correctly created.

I'm not sure about the reason why you are adding a sleep between the two cluster creations, as the `kind create` command returns once the cluster is created and each cluster has its own independent docker containers.
I would instead investigate the real reason why Kind is failing during the creation of the second cluster, and if there is something to wait, we need to understand what needs to be waited, as I believe 15 seconds is an arbitrary value, which might not fit all the cases.

I will close this PR, but feel free to open a new one once you find a solution :)

PR #3115, opened 2025-08-28
feat: Direct connection support between provider clusters

### LINKED ISSUE #3011: [Feature] Enhance inter-cluster networking: direct path support between 'leaf' clusters

In a [multi-cluster deployment](https://docs.liqo.io/en/v1.0.0/examples/replicated-deployments.html#) with a central cluster and multiple provider clusters (leaves), I’ve successfully established a direct connection between two leaf clusters using `liqoctl network connect`.
While the network connection itself works and pods in the leaf clusters can communicate directly, at the exposure of a deployment, the resulting Service makes the traffic go through the central cluster even if would be possible to use the direct connection.

[As discussed in the Slack channel of Liqo](https://liqo-io.slack.com/archives/C014Z7F25KP/p1743695454044279) these are the possible solutions:

1. Change how the Virtual Kubelet reflects the EndpointSlices to allow direct connectivity. This is a simple solution but can create conflicts in case the PodCIDRs are overlapped. 
2. Develop a network manager that stays on top of every cluster and watches them. This manager can write the correct EndpointSlice in every cluster since it has total visibility on the topology. This can be made by using the ServiceImport and ServiceExport kubernetes standard resources to expose services and make that reachable from other clusters.

This feature proposal together with the possiblity to create shortcuts among leaves (as mentioned in #3006) directly from the central cluster would enable the users to improve the performance of communications between provider clusters and also centralize the management of these connections. 

### Code of Conduct

Gabbe64:
# Description

Fixes #3011 

This PR introduces the possibility to make pods deployed on provider clusters communicate directly by using an established network connection (`liqoctl network connect`).

This feature is **opt-in**: users can enable it by adding the annotation key `use-direct-connections: true` to the `Service` exposing the pods that should communicate directly. This way, the default behavior remains unchanged, while also giving users the possibility to choose whether to route traffic through the consumer cluster or directly.

## Implementation

As discussed during the last community meetings, the modified components are the  **virtual-kubelet** and the **ShadowEndpointslice controller**.

### Transmission - virtual-kubelet
When reflecting `Endpointslices` and the corresponding `Service` has the right annotation, some data is collected from the informers (without direct the API server calls) then serialized as JSON and reflected in the annotation field of the `ShadowEndpointslice`.

This data is:

1. The `clusterID` where the endpoint is deployed,
2. The local IPs contained in the `Endpointslice ` of the consumer,
3. The same IPs, remapped by the consumer cluster.

### Reception
The ShadowEndpointslice controller (in the provider cluster) receives this data and uses it to remap the addresses so to use the direct connections towards other provider clusters (if available).

To achieve this:
- Endpoints leading to pods on a different provider are identified using point 3. 
- They are replaced with direct addresses, derived from the remote cluster’s `podCIDR` (point 1) and the host part extracted from the local IP (point 2); this required the implementation of a "forced mapping".


pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/mapping.go
```diff
@@ -18,25 +18,52 @@ import (
...
+	klog "k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	liqov1beta1 "github.com/liqotech/liqo/apis/core/v1beta1"
+	directconnectioninfo "github.com/liqotech/liqo/pkg/utils/directconnection"
 	ipamips "github.com/liqotech/liqo/pkg/utils/ipam/mapping"
 )
 
 // MapEndpointsWithConfiguration maps the endpoints of the shadowendpointslice.
 func MapEndpointsWithConfiguration(ctx context.Context, cl client.Client,
```
cheina97, line 33: This function is has a parametr called clusterID whic is overrided by the GetConnectionDataByIP function. The options are 2:
- the parameter is useless
- the GetConnectionDataByIP is shadowing the value

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/mapping.go
```diff
@@ -18,25 +18,52 @@ import (
...
+
+					if err == nil {
+						endpoints[i].Addresses[j] = rAddr.String()
+						addrHasBeenRemapped = true
+					} else {
+						klog.Errorf("error while mapping address %q using the cidr from clusterID %q: %v", ip, clusterID, err)
+					}
+				}
+				if addrHasBeenRemapped {
+					break
```
cheina97: The break statement exits the inner address loop (for j := range endpoints[i].Addresses) after the first address is remapped. If an endpoint has multiple addresses, only the first one gets processed

pkg/utils/directconnection/directconnection.go
```diff
@@ -0,0 +1,95 @@
...
+func (l *InfoList) GetConnectionDataByIP(ip string) (clusterID, originalIP string, found bool) {
+	for _, entry := range l.Items {
+		for i, remappedIP := range entry.RemappedIPs {
+			if ip == remappedIP {
+				return entry.ClusterID, entry.IPs[i], true
+			}
+		}
+	}
+	return "", "", false
+}
```
cheina97: Assumes len(entry.IPs) == len(entry.RemappedIPs) without validation. If these slices have different lengths (which is possible via the Add() method), accessing entry.IPs[i] will cause a panic.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -196,12 +201,91 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
...
+				continue
+			}
+
+			clusterID, err := getters.RetrieveRemoteClusterIDFromNode(node)
+			if err != nil {
+				klog.Errorf("Failed to retrieve remote cluster ID from node %q: %v", *endpoint.NodeName, err)
+				continue
+			}
+
+			objectName := endpoint.TargetRef.Name
```
cheina97: endpoint.TargetRef is an optional pointer (*v1.ObjectReference). According to Kubernetes API specifications, it can be nil. This code will panic if an endpoint has no TargetRef

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -97,12 +101,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// Check foreign API server status
 	apiServerReady := foreigncluster.IsAPIServerReadyOrDisabled(fc)
 
+	// Check if direct connections data is provided
+	var remoteConnectionsData directconnectioninfo.InfoList
+	if val, ok := shadowEps.Annotations[directConnectionAnnotationLabel]; ok {
+		err := remoteConnectionsData.FromJSON([]byte(val))
+
+		if err != nil {
+			klog.Errorf("failed to unmarshal direct connection data for shadowendpointslice %q: %v", nsName, err)
```
cheina97: Why not returning an error?

pkg/utils/directconnection/directconnection.go
```diff
@@ -0,0 +1,95 @@
...
+// for which there may be a direct connection that can be used.
+//
+// ClusterID is the ID of the remote cluster where the pods are located.
+// IPs are the Addresses in the Endpoint as seen by the consumer cluster BEFORE REMAPPING -> needed to extract the host part to remap correctly.
+// RemappedIPs are the same IPs AFTER the remapping of the consumer -> needed to identify which IPs to replace.
+type Info struct {
+	ClusterID   string   `json:"ID"`
+	IPs         []string `json:"IPs"`
+	RemappedIPs []string `json:"rIPs"`
+}
```
cheina97: It's not clear to me why you need all these info in the annotation. I think just an annotation with the clusterID should be enough.
In theory you should just reflect the endpointslie as it is on the remote cluster, then the remote cluster will reconcile the shadow_endpointslice and will use the "configuration" resource related to the clusterID contained inside the annotation, to remap the IP

cheina97: Am I missing something?

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -97,12 +101,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		}
+		// JSON is not propagated to the EndpointSlice
+		delete(shadowEps.Annotations, directConnectionAnnotationLabel)
```
cheina97: Can you remove it in a more explicit way? For example create a function that "forge" the endpointslice annotations starting from the shadowendpointslice

Gabbe64:
Given your feedback I decided to refactor this feature. The core logic remains the same, but the exchanged data has changed slightly.

Changes:

- **EndpointSlice Reflection**: When an endpoint belongs to a cluster other than the local or remote one, we no longer send the address on the consumer's external CIDR. Instead, the address is sent as is (as seen by the consumer cluster).

- **Annotations**: The data payload in the annotation is now smaller since it no longer includes the external CIDR address. This required an update to the data structure used for marshalling/unmarshalling.

I also added a few utilities, including a helper for annotation forging and an optimized lookup for the data structure embedded within the annotations.

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/mapping.go
Gabbe64: Yes, that was a mistake.

The name of the parameter was changed and it stores the clusterID of the local cluster.

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
Gabbe64: In the previous implementation the error was not returned because in case of problems during the unmarshalling, the controller could just skip the remapping of the direct address and use the standard external CIDR address (of the consumer) provided by the ShadowEndpointSlice. 
It thought it to be some sort of fallback mechanism.

With the new implementation, since the address in the ShadowEPS is not the one from the external CIDR, it's mandatory to return an error, because there is no "usable" endpoint in case the unmarshalling goes wrong.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
Gabbe64: https://github.com/liqotech/liqo/pull/3115/changes#diff-a3538c69a2c86d46cc2692a9b1e5f24a200f2c05cb99cce1cac12423393248bbR219-R222

Added a check at the start of the block.

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
Gabbe64: https://github.com/liqotech/liqo/pull/3115/changes#diff-f1a76a3c052d25f7d5002fed269741de3338ba5c47f717baed29695a1f9b1c02R158-R167

Done!

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -51,7 +53,10 @@ var _ manager.NamespacedReflector = (*NamespacedEndpointSliceReflector)(nil)
 
 const (
 	// EndpointSliceReflectorName -> The name associated with the EndpointSlice reflector.
-	EndpointSliceReflectorName = "EndpointSlice"
+	EndpointSliceReflectorName      = "EndpointSlice"
+	maxAnnotationSize               = 256 * 1024 // 256 KiB
+	directConnectionAnnotationLabel = "direct-connections-data"
+	directConnectionServiceLabel    = "use-direct-connections"
```
aleoli: I would use something like "liqo.io/use-direct-connections" (also for the other annotation)

And, the variable is named "label", but we are usgin it as an annotation

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -183,6 +187,8 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
 		return nil
 	}
 
+	shouldProvideDirectConnectionData := ner.ShouldProvideDirectConnectionData(local)
```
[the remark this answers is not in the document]
Gabbe64: That's how it's supposed to work.

- Endpoints belonging to the local cluster would not be remapped anyways.
- Endpoints belonging to other provider clusters must not be remapped: we don't want to use the external CIDR of the sender but the direct connection, so the address should be untouched.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -192,16 +198,82 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
...
 	target := forge.RemoteShadowEndpointSlice(local, remote, ner.localNodeClient, ner.RemoteNamespace(), translator, ner.ForgingOpts)
 	if terr != nil {
 		klog.Errorf("Reflection of local EndpointSlice %q to %q failed: %v", ner.LocalRef(name), ner.RemoteRef(name), terr)
 		ner.Event(local, corev1.EventTypeWarning, forge.EventFailedReflection, forge.EventFailedReflectionMsg(terr))
 		return terr
 	}
+	
+	if marshaledData != nil {
+		target.Annotations[consts.DirectConnectionDataAnnotationKey] = string(marshaledData)
+	}
```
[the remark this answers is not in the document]
Gabbe64: Annotations is never nil at that point, it's created by the function called a couple lines before:

`target := forge.RemoteShadowEndpointSlice(local, remote, ner.localNodeClient, ner.RemoteNamespace(), translator, ner.ForgingOpts)`

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -192,16 +198,82 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
 		}
 
 		var translations []string
-		translations, terr = ner.MapEndpointIPs(name, originals)
+		translations, terr = ner.MapEndpointIPs(name, originals, shouldProvideDirectConnectionData)
```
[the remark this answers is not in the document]
Gabbe64: Replied in the previous comment.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -277,7 +349,16 @@ func (ner *NamespacedEndpointSliceReflector) MapEndpointIPFromIPResource(origina
...
-func (ner *NamespacedEndpointSliceReflector) MapEndpointIPs(endpointslice string, originals []string) ([]string, error) {
+//
+// skipTranslation parameter is needed when direct connections are enabled: in that case 
+// we want to skip the mapping on ExternalCIDR for endpoints that are reachable through direct connections between providers.
+func (ner *NamespacedEndpointSliceReflector) MapEndpointIPs(endpointslice string, originals []string, skipTranslation bool) ([]string, error) {
+	
+	if skipTranslation {
+		klog.V(4).Infof("Skipping translation of endpoint IPs for EndpointSlice %q, beacuse direct connections are enabled.", ner.LocalRef(endpointslice))
+		return originals, nil
+	}
```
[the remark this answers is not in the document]
Gabbe64: Replied in the previous comment.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -192,16 +198,82 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
...
+				// Max size of data in annotations is 256KB
+				klog.Errorf("marshaled direct connection data exceeds maximum size of %d bytes: %d bytes", maxAnnotationSize, len(marshaledData))
+				marshaledData = nil
+			} else if marshaledData != nil {
+				klog.V(4).Infof("Direct connection data for endpointslice %q marshaled successfully", local.Name)
+			}
+		}
+	}
+
 	target := forge.RemoteShadowEndpointSlice(local, remote, ner.localNodeClient, ner.RemoteNamespace(), translator, ner.ForgingOpts)
```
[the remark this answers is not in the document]
Gabbe64: Replied in the previous comment.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -192,16 +198,82 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
...
 
 		var translations []string
-		translations, terr = ner.MapEndpointIPs(name, originals)
+		translations, terr = ner.MapEndpointIPs(name, originals, shouldProvideDirectConnectionData)
 		return translations
 	}
 
+	var marshaledData []byte
+	if shouldProvideDirectConnectionData {
+		// Gather the data needed to make the providers use the direct connections beteen them.
```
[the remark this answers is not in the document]
Gabbe64: Tests added.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -192,16 +198,82 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
...
+				// In this way, we avoid to break the reflection of endpointslices with direct connections enabled, but with no cross-provider endpoints (e.g., because all endpoints are local).
+				// If the data cannot be marshaled, then it is equivalent to not having any data at all, hence we can proceed with the reflection anyway.
+			} else {
+				klog.V(4).Infof("marshaled direct connection data: %s", string(marshaledData))
+			}
+
+			if len(marshaledData)+totalAnnotationsSize(local) >= maxAnnotationSize {
+				// Max size of data in annotations is 256KB
+				klog.Errorf("marshaled direct connection data exceeds maximum size of %d bytes: %d bytes", maxAnnotationSize, len(marshaledData))
+				marshaledData = nil
```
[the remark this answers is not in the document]
Gabbe64: Tests added.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -196,16 +202,81 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
...
+			}
+			if endpoint.TargetRef == nil {
+				klog.Errorf("Endpoint with IP %q has no targetRef, skipping direct connection data collection for it", endpoint.Addresses)
+				continue
+			}
+			node, err := ner.localNodeClient.Get(*endpoint.NodeName)
+			if err != nil {
+				klog.Errorf("Failed getting the node %q: %v", *endpoint.NodeName, err)
+				continue
+			}
```
fra98: We need to handle the error. In case of a transient error, we would skip this endpoint. It should `continue` with the loop only in case of NotFound error

pkg/utils/directconnection/utils.go
```diff
@@ -0,0 +1,48 @@
...
+		return false
+	}
+
+	if node.Labels["liqo.io/type"] != "virtual-node" {
+		return false
+	}
+
+	if node.Labels["liqo.io/remote-cluster-id"] == remoteClusterID {
+		return false
+	}
```
fra98: We should use already defined constants. For virtual node check, I think we also have the utility `utils.IsVirtualNode()`.

Also, for remote cluster id I think we retrieve it at the next step. Maybe you can anticipate the retrieval and only do the comparison here

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -365,6 +443,35 @@ func (ner *NamespacedEndpointSliceReflector) ServiceToEndpointSlicesKeyer(metada
...
+	// Check if a service is associated to the EndpointSlice
+	svcname, ok := obj.GetLabels()[discoveryv1.LabelServiceName]
+	if !ok {
+		return false
+	}
+	// Retrieve the service from the local cache, done again after ShouldSkipReflection
+	svc, err := ner.localServices.Get(svcname)
+	if err != nil {
+		return false
+	}
```
fra98, line 466: Also here we probably should handle generic error from NotFound and return error in such case to make feature more resilient

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -39,6 +39,9 @@ import (
 	offloadingv1beta1clients "github.com/liqotech/liqo/pkg/client/clientset/versioned/typed/offloading/v1beta1"
 	ipamv1alpha1listers "github.com/liqotech/liqo/pkg/client/listers/ipam/v1alpha1"
 	offloadingv1beta1listers "github.com/liqotech/liqo/pkg/client/listers/offloading/v1beta1"
+	consts "github.com/liqotech/liqo/pkg/consts"
+	directconnectioninfo "github.com/liqotech/liqo/pkg/utils/directconnection"
```
fra98: NIT
```suggestion
	"github.com/liqotech/liqo/pkg/utils/directconnection"
```

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -196,16 +202,81 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
...
+			klog.V(4).Infof("Service is set for direct connections but no data found for this endpointslice: %s", local.Name)
+		} else {
+			var err error
+
+			marshaledData, err = remoteConnectionsData.ToJSON()
+			if err != nil {
+				klog.Errorf("Failed to marshal direct connection data: %v", err)
+				marshaledData = nil
+				// We do not return the error, since this is not a critical failure, as the reflection can proceed without the direct connection data.
+				// If the data cannot be marshaled, then it's equivalent to not having any data at all, so we can proceed with the reflection anyway.
```
fra98: This is arguable, as if this happens the code will silently ignore the feature and traffic will flow to the main cluster. If we return an error it will be more debuggable. Let's discuss about it

pkg/utils/ipam/ips.go
```diff
@@ -111,3 +111,28 @@ func FirstNIPsFromPrefix(prefix netip.Prefix, num uint32) iter.Seq[netip.Addr] {
...
+	if obj == nil {
+		return "", fmt.Errorf("IP object is nil")
+	}
+
+	if obj.Status == (ipamv1alpha1.IPStatus{}) {
+		return "", fmt.Errorf("IP object has no remapped IP")
+	}
+
+	return obj.Status.IP.String(), nil
+}
```
fra98: Are these two functions outdated code? I don't see it used anywhere

pkg/utils/directconnection/utils.go
```diff
@@ -0,0 +1,48 @@
...
+
+	return true
+}
```
fra98: Is this needed on this package or can it stay under reflection/exposition pkg?

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -41,12 +41,15 @@ import (
 	"github.com/liqotech/liqo/pkg/consts"
 	"github.com/liqotech/liqo/pkg/utils"
 	clientutils "github.com/liqotech/liqo/pkg/utils/clients"
+	directconnectioninfo "github.com/liqotech/liqo/pkg/utils/directconnection"
```
fra98:
```suggestion
	"github.com/liqotech/liqo/pkg/utils/directconnection"
```

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -97,25 +100,46 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// Check foreign API server status
 	apiServerReady := foreigncluster.IsAPIServerReadyOrDisabled(fc)
 
+	// Check if direct connections data is provided
+	var remoteConnectionsData directconnectioninfo.Data
+	if val, ok := shadowEps.Annotations[consts.DirectConnectionDataAnnotationKey]; ok {
+		err := remoteConnectionsData.FromJSON([]byte(val))
+
+		if err != nil {
```
fra98: I know we don't do it in this reconciler, but we should always returned wrapped error and avoid logging with klog.Error() as this will results in a double error log

pkg/utils/ipam/mapping/ips.go
```diff
@@ -186,6 +187,28 @@ func remapWithin(ip net.IP, address string, spec, status []networkingv1beta1.CID
...
+	cfg, err := getters.GetConfigurationByClusterID(ctx, cl, clusterID, corev1.NamespaceAll)
+	if err != nil {
+		return addr.String(), err
+	}
+
+	podCidr := cidrutils.GetPrimary(cfg.Status.Remote.CIDR.Pod).String()
+	_, podnet, err := net.ParseCIDR(podCidr)
+	if err != nil {
+		return addr.String(), err
+	}
```
fra98: Not sure if this is compatible with multiple pod CIDR support. @claudiolor what do you think?

pkg/virtualKubelet/forge/endpointslices.go
```diff
@@ -154,3 +154,14 @@ func RemoteEndpointSlicePorts(locals []discoveryv1.EndpointPort) []discoveryv1.E
...
+// EndpointSliceAnnotations builds annotations for a reflected EndpointSlice,
+// filtering out keys that must not be propagated.
+func EndpointSliceAnnotations(shadow *offloadingv1beta1.ShadowEndpointSlice,
+	annotationsNotReflected []string) map[string]string {
+	if shadow == nil {
+		return nil
+	}
+
+	return FilterNotReflected(shadow.GetAnnotations(), annotationsNotReflected)
+}
```
fra98: NIT: function is defined in VirtualKubelet but it's used only by controller-manager. I know here we have `FilterNotReflected()` utility but it would be just easier and more readable to directly remove the key in shadowendpointslice controller

pkg/virtualKubelet/reflection/exposition/endpointslice.go
Gabbe64: The error is logged on the virtual kubelet

In case of error the default behaviour is restored, otherwise the reflection would not take place at all. I think this is preferred, as it prevents a total disruption instead of only disabling a feature.

pkg/utils/directconnection/utils.go
Gabbe64: I was thinking to have it there only for clarity as it is used only for the direct connections feature. It ended up being the only function in that package.

Is it better to move it to reflection/exposition?

pkg/utils/ipam/mapping/ips.go
claudiolor: Yes, we need to consider all the POD cidrs. @Gabbe64 you can use the same approach of the regular MapAddressWithConfiguration, or maybe just wrapping that function and return the original address based on the returned error. BTW Can you please clarify why we need the `ForceMapAddressWithConfiguration` function?

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -97,25 +100,43 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		if err := MapEndpointsWithConfiguration(ctx, r.Client, clusterID, remappedEndpoints, rcindex); err != nil {
 			klog.Errorf("an error occurred while remapping endpoints for shadowendpointslice %q: %v", nsName, err)
 			return ctrl.Result{}, err
 		}
+	} else if rcindex != nil {
+		// Networking between consumer and provider is disabled, but direct provider-to-provider connections
+		// may still be present. Remap only those addresses so they can be reached via the direct link.
+		if err := MapOnlyDirectConnectionEndpoints(ctx, r.Client, remappedEndpoints, rcindex); err != nil {
+			klog.Errorf("an error occurred while remapping direct-connection endpoints for shadowendpointslice %q: %v", nsName, err)
+			return ctrl.Result{}, err
```
fra98: Can you return wrapped error and avoid logging with klog?

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -256,3 +277,20 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, wor
...
+
+	filtered := make(map[string]string, len(annotations))
+	for k, v := range annotations {
+		if k == consts.DirectConnectionDataAnnotationKey {
+			continue
+		}
+		filtered[k] = v
+	}
+
+	return filtered
```
fra98, line 298: NIT: you can use `maps.Clone()` and then delete key

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/mapping.go
```diff
@@ -18,24 +18,75 @@ import (
...
+			var err error
+			// if address is found, then the remapping is forced using the clusterID instead of the local one.
+			if mappingClusterID != localClusterID {
+				rAddr, err = ipamips.ForceMapAddressWithConfiguration(ctx, cl, mappingClusterID, addr)
+			} else {
+				rAddr, err = ipamips.MapAddress(ctx, cl, localClusterID, addr)
+			}
+			klog.V(4).Infof("Mapped address %q using clusterID %q; result is: %q", addr, mappingClusterID, rAddr)
 			if err != nil {
 				return err
```
fra98, line 58: Can you wrap error?

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/mapping.go
```diff
@@ -18,24 +18,75 @@ import (
...
+			if !found {
+				// Address is not a direct-connection address; leave it unchanged.
+				continue
+			}
 
+			rAddr, err := ipamips.ForceMapAddressWithConfiguration(ctx, cl, liqov1beta1.ClusterID(directClusterID), addr)
+			klog.V(4).Infof("Mapped direct-connection address %q using clusterID %q; result is: %q", addr, directClusterID, rAddr)
+			if err != nil {
+				return err
+			}
```
fra98, line 119: This way you log even if error was returned.
I would anticipate the check and return wrapped error, it will be logged automatically. This way you can log later assuming you are in happy path scenario.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -187,6 +192,12 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
 		return nil
 	}
 
+	shouldProvideDirectConnectionData, err := ner.ShouldProvideDirectConnectionData(local)
+	if err != nil {
+		klog.Errorf("Failed to check direct connection data eligibility for local EndpointSlice %q: %v", ner.LocalRef(name), err)
+		return err
+	}
```
fra98, line 199: Can you wrap error and avoid logging? (same for the rest of file)

pkg/virtualKubelet/reflection/exposition/endpointslice.go
```diff
@@ -196,16 +207,85 @@ func (ner *NamespacedEndpointSliceReflector) Handle(ctx context.Context, name st
...
+		// 2) ClusterID of the cluster on which that endpoint is running.
+
+		var remoteConnectionsData directconnection.Data
+
+		for _, endpoint := range local.Endpoints {
+			if endpoint.NodeName == nil {
+				continue
+			}
+			if endpoint.TargetRef == nil {
+				klog.Errorf("Endpoint with IP %q has no targetRef, skipping direct connection data collection for it", endpoint.Addresses)
```
fra98: This should be a high verbosity log or warning? Doesn't seem like an error

pkg/virtualKubelet/reflection/exposition/endpointslice.go
fra98: Why would the marshal/unmarshal fail, though? If it fails, it means there is an error in the code itself (and not a transient runtime error), which would be silently ignored and can only be spotted if you read the VK logs. Using a fallback for code errors seems strange to me.

pkg/utils/ipam/mapping/ips.go
Gabbe64: Hi Claudio, the `ForceMapAddressWithConfiguration` is needed for this feature because with the standard remapping, an address is remapped using the receiving cluster context.

In order to use a direct connection between providers we need to remap the addresses belonging to another provider with the `Configuration`  of that specific provider; this is why for me that function is necessary.

pkg/virtualKubelet/reflection/exposition/endpointslice.go
Gabbe64: @cheina97 @aleoli @cheina97 @claudiolor 

What happens in this situation is that in case a marshaling error occurs, it is logged on the VK but the endpoint reflection continues, bypassing the direct connection feature although it's enabled, like a "silent" fallback.

As suggested by Francesco, it could be changed to stop entirely the reflection: this way errors that lead to an incorrect marshalling can be detected more easily.

Both solutions have pros and cons. What do you think?

pkg/utils/ipam/mapping/ips.go
claudiolor: Ok, it's clear now, in the case you are not able to do remapping you go on passing through the consumer cluster

pkg/virtualKubelet/reflection/exposition/endpointslice.go
claudiolor: Better to stop the reflection, black or white failures are easier to handle

pkg/utils/ipam/mapping/ips.go
Gabbe64: Solved.
Now, instead of getting the first CIDR, a lookup on the `Configuration` is performed: based on the input address, index of the sourcePodCIDR is retrieved and it's used to return the right remappedPodCIDR.

pkg/utils/ipam/mapping/ips.go
claudiolor: I believe here there's no need to define your own custom function. You can use the `remapWithin` function:
```suggestion
	remappedAddr, isRemapped, err := remapWithin(addr, conf.spec, conf.status)
	if err != nil || !isRemapped {
		return addr.String(), err
	}
	return remappedAddr
```

Gabbe64: You are totally right.

pkg/utils/directconnection/directconnection.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Data is a struct representing the addresses of the pods deployed on a remote cluster
+// for which there may be a direct connection that can be used.
+//
+// Used to collect data and marshal to an annotation on the ShadowEndpointSlice.
+//
+// Key: clusterID. Value: list of IPs associated to that cluster.
+type Data struct {
+	ByCluster map[string][]string `json:"clstrIDtoAddrs"`
+}
```
aleoli: Can you use more meaningful names for the struct and the json tag?

Gabbe64: Changed to more descriptive names.

Also, I left some comments on that module so that every structure is well documented.

deployments/liqo/values.yaml
```diff
@@ -425,6 +425,10 @@ controllerManager:
     # This feature can be useful in case of remote node failure to guarantee better service continuity and to have the expected pods workload on the remote cluster.
     # However, enabling this feature could produce zombies in the worker node, in case the node returns Ready again without a restart.
     enableNodeFailureController: false
+    # -- Prevent endpoints reachable only through a direct provider-to-provider connection from being
+    # added to the reflected EndpointSlice. When enabled, the shadowendpointslice controller drops the
+    # endpoints referring to pods hosted on other providers (listed in the direct-connection annotation data).
+    denyDirectConnections: false
```
fra98: NIT: this is too technical for helm value. You should only say what it does and its high-level purpose

pkg/liqo-controller-manager/offloading/shadowendpointslice-controller/shadowendpointslice_controller.go
```diff
@@ -256,3 +287,13 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, wor
 		WithOptions(controller.Options{MaxConcurrentReconciles: workers}).
 		Complete(r)
 }
+
+// RemoveDirectConnectionAnnotation returns a copy of annotations without direct-connection data.
+func RemoveDirectConnectionAnnotation(annotations map[string]string) map[string]string {
```
fra98: NIT: this should be private as its only intended to use in this pkg

pkg/utils/ipam/mapping/ips.go
```diff
@@ -186,6 +186,33 @@ func remapWithin(ip net.IP, address string, spec, status []networkingv1beta1.CID
...
+// ForceMapAddressWithConfiguration forces the mapping of the provided address using the podCIDR of the remote cluster, retrieved using its clusterID.
+//
+// In case of failure returns the address unchanged.
+func ForceMapAddressWithConfiguration(ctx context.Context, cl client.Client,
+	clusterID liqov1beta1.ClusterID, address string) (string, error) {
+	addr := net.ParseIP(address)
+
+	cfg, err := getters.GetConfigurationByClusterID(ctx, cl, clusterID, corev1.NamespaceAll)
+	if err != nil {
+		return addr.String(), err
```
fra98: NIT: we should wrap error. Also when returning an error you usually return 0-value for the other args. So here (and below ) returning `""` it's cleaner

fra98 (approved):

aleoli (approved):

aleoli:
@Gabbe64 you shoul update the commit message before merging

cheina97 (approved):

deployments/liqo/values.yaml
```diff
@@ -437,6 +437,9 @@ controllerManager:
     # This feature can be useful in case of remote node failure to guarantee better service continuity and to have the expected pods workload on the remote cluster.
     # However, enabling this feature could produce zombies in the worker node, in case the node returns Ready again without a restart.
     enableNodeFailureController: false
+    # -- Prevents the usage of direct connections by provider clusters. 
+    # When enabled, the provider cluster will not route traffic directed to another provider through their direct connection. 
+    denyDirectConnections: false
```
claudiolor: I'd change the location of this field. I think from an user perspective it shouldn't stay under controllerManager, but networking

Gabbe64: You are right, it's functionally correct but counter-intuitive, placed it under "networking".

claudiolor (approved):

PR #3122, opened 2025-09-10
[Core] feat: mark foreigncluster as permanently unreachable

claudiolor:
# Description

This PR fixes a known issue with Liqo: when a provider cluster becomes unreachable, it is not easy to disable a peering connection with that cluster, as the CRDreplicator and the ControllerManager attempt to remove resources from a cluster that will never come back up. A mechanism based on the
`liqo.io/foreign-cluster-permanently-unreachable: true` has been added: when a ForeignCluster is marked with that annotations, and the API server of the remote cluster is considered unreachable, then replicated resources are removed from the local cluster, ignoring the remote ones.

### DIFF
--- a/internal/crdReplicator/foreignclusterstate_controller.go
+++ b/internal/crdReplicator/foreignclusterstate_controller.go
@@ -1,0 +1,79 @@
+// Copyright 2019-2026 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crdreplicator
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	liqov1beta1 "github.com/liqotech/liqo/apis/core/v1beta1"
+	"github.com/liqotech/liqo/internal/crdReplicator/reflection"
+	"github.com/liqotech/liqo/pkg/consts"
+	"github.com/liqotech/liqo/pkg/utils/foreigncluster"
+)
+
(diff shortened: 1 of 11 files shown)


pkg/utils/foreigncluster/conditions.go
```diff
@@ -20,8 +20,44 @@ import (
...
+				default:
+					message = condition.Message
+				}
+				break
+			}
+		}
+	}
+
+	return isDead, message
+}
```
fra98, line 55:
```suggestion
	// In all the other cases we consider the connection dead, as there is an annotation marking the cluster unreachable
	// unless the API server is still responding.

	// Check whether the API server is still responding.
	if foreignCluster.Status.Conditions != nil {
		// Check the API server reachability condition.
		for _, condition := range foreignCluster.Status.Conditions {
			if condition.Type == liqov1beta1.APIServerStatusCondition {
				switch condition.Status {
				case liqov1beta1.ConditionStatusEstablished:
					return false, ""
				default:
					return true, condition.Message
				}
				break
			}
		}
	}

	return true, "The foreign cluster has been marked as permanently unreachable"
}
```

internal/crdReplicator/foreignclusterstate_controller.go
```diff
@@ -0,0 +1,83 @@
...
+		return ctrl.Result{}, fmt.Errorf("no reflector found for ForeignCluster %q", foreignCluster.Name)
+	}
+
+	if dead, message := foreigncluster.IsDead(foreignCluster); dead {
+		reflector.RemoteReachable = false
+		klog.Warningf("Remote cluster %q is dead: %s", foreignCluster.Name, message)
+	} else {
+		reflector.RemoteReachable = true
+		if message != "" {
+			klog.Warningf("Remote cluster %q is temporary unreachable: %s", foreignCluster.Name, message)
```
fra98: If not dead we have no message, so this if is probably useless / can never happen

internal/crdReplicator/reflection/handler.go
```diff
@@ -96,11 +96,16 @@ func (r *Reflector) handle(ctx context.Context, key item) error {
...
 	if !localUnstr.GetDeletionTimestamp().IsZero() {
 		klog.Infof("[%v] Deleting remote %v with name %v, since the local one is being deleted", r.remoteClusterID, key.gvr, key.name)
-		vanished, err := r.deleteRemoteObject(ctx, resource, key)
-		if err != nil {
-			return err
+
+		// If the remote cluster is not reachable, then do not try to delete the remote object.
+		vanished := true
+		if r.RemoteReachable {
+			vanished, err = r.deleteRemoteObject(ctx, resource, key)
```
fra98: deleteRemoteObject already check status of r.remoteReachable and return correct vanished. Seems like duplicated logic or is there a reason?

PR #3128, opened 2025-09-22
feat: avoid kyverno dependency in tests

cheina97:
# Description

This PR removes the kyverno dependency from network tests.
If kyverno is not installed in one of the tested clusters a warning will be triggered.
In case you are not using kyverno you cannot be sure that all the pods will be scheduled one per node

### DIFF
--- a/pkg/liqoctl/test/network/setup/kyverno.go
+++ b/pkg/liqoctl/test/network/setup/kyverno.go
@@ -22,8 +22,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 
 	"github.com/liqotech/liqo/pkg/liqoctl/test/network/client"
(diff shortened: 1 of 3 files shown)


fra98 (approved):

pkg/liqoctl/test/network/setup/kyverno.go
```diff
@@ -32,32 +34,54 @@ var KyvernoPolicyGroupVersionResource = schema.GroupVersionResource{Group: "kyve
...
+			policy = ForgeKyvernoPodAntiaffinityPolicy(k, true)
+			if _, err := cl.ProvidersDynamic[k].Resource(KyvernoPolicyGroupVersionResource).
+				Namespace(NamespaceName).Create(ctx, policy, metav1.CreateOptions{}); err != nil && !apierrors.IsAlreadyExists(err) {
+				return fmt.Errorf("provider %q failed to create policy: %w", k, err)
+			}
+		} else {
+			kyvernoNotInstalled = true
+			printer.Logger.Warn(fmt.Sprintf("Kyverno not available on provider %q, skipping policy creation.", k))
 		}
 	}
```
fra98: NIT: can we refactor it so that both consumer and providers call the same function, avoiding code duplicates?

PR #3151, opened 2025-11-24
docs(install): add Cilium device config note

aleoli:
# Description

This pr adds some notes on Cilium device configurations

### DIFF
--- a/docs/installation/install.md
+++ b/docs/installation/install.md
@@ -632,6 +632,25 @@ affinity:
             operator: DoesNotExist
 ```
 
+#### Device Configuration
+
+When using **advanced Cilium eBPF features** such as eBPF-based host routing, host firewall, or BPF masquerading, Cilium automatically attaches eBPF programs to all network interfaces it detects.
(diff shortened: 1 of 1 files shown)


docs/installation/install.md
```diff
@@ -632,6 +632,25 @@ affinity:
...
+
+When using **advanced Cilium eBPF features** such as eBPF-based host routing, host firewall, or BPF masquerading, Cilium automatically attaches eBPF programs to all network interfaces it detects.
+However, Liqo creates its own network interfaces (e.g., `liqo.*`) that should not be managed by Cilium's eBPF datapath.
+
+```{admonition} Note
+This configuration is **not required** if you are using Cilium with default settings.
+It is only necessary when enabling advanced eBPF features that attach programs directly to network interfaces.
+```
+
+To prevent conflicts and ensure Liqo traffic is handled correctly when using these advanced features, you should explicitly configure which network interfaces Cilium should manage using the `devices` parameter.
```
frisso: "... using the `devices` parameter"
Where? Within the Cilium configuration file?

PR #3157, opened 2025-11-28
fix: reuse existing client endpoint when configuring server interface

### LINKED ISSUE #3090: the connection beween k8s cluster reconnect every about 10 hours
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Version

1.0.1

### What happened?

1.liqo gateway wireguard pod Reconcile every about 10 hours even nothing change. https://github.com/liqotech/liqo/blob/525e6eff35d816210b524864ec452ac0f79e4b66/pkg/gateway/tunnel/wireguard/publickeys_controller.go#L66
2.In gateway client (wireguard client) reconfigure won't close udp tunnel. https://github.com/liqotech/liqo/blob/525e6eff35d816210b524864ec452ac0f79e4b66/pkg/gateway/tunnel/wireguard/publickeys_controller.go#L82
3.But in gateway server  (wireguard server ) reconfigure will close udp tunnel. Thus will influence cross k8s network.

### Relevant log output

```shell

```

### How can we reproduce the issue?

1. use flow test-short.sh script to test network between k8s cluster.
```
#!/bin/bash
#  test-short.sh
#  ./test-short.sh target

TARGET_URL=${1}
LOG_FILE="short_conn.csv"
LOG_FILE_ERROR="short_conn_error.csv"

echo "timestamp,http_code,response_time,error" > $LOG_FILE

while true; do
    TIMESTAMP=$(date +"%Y-%m-%d %H:%M:%S")
    
    OUTPUT=$(curl -o /dev/null -s -w "%{http_code},%{time_total}" \
              --connect-timeout 3 -m 3 "$TARGET_URL" 2>&1)
    
    if [ $? -eq 0 ]; then
        echo "$TIMESTAMP,$OUTPUT," >> $LOG_FILE
    else
        ERROR=$(echo $OUTPUT | cut -d ' ' -f 1-3 | tr ',' ' ')
        echo "$TIMESTAMP,,,$ERROR" >> $LOG_FILE_ERROR
    fi
    
    sleep 1
done
```

### Provider or distribution

Rancher

### CNI version

flannel v1.4.1-rancher1

### Kernel Version

6.1.0-32

### Kubernetes Version

1.30

### Code of Conduct

xDev789:
Reuse existing client endpoint when configuring server interface.

# Description

Wireguard container now preserves peer's endpoint when configuring the server interface, which allows for uninterrupted connectivity in an event of PublicKey resource reconciliation.

Fixes #3090

# How Has This Been Tested?

- [x] make unit
- [X] liqoctl install on Kind clusters


cheina97 (dismissed):

cheina97:
Hi @xDev789, I've checked your PR and something is not clear to me. You are forcing the wg interface to use always the same peer's endpoint, but what happens if it changes? I tested it with cilium and I've noticed that this field is populated with an IP which is the one assigned to the `cilium_host@cilium_net` interface on the node. What happens if the gateway is rescheduled on another node? How you tested this PR?

cheina97:
I tried to move pods from one node to another, keeping the wrong IP in the server, and it seems that everything is working. It seems that the peer's endpoint is set in server mode but it's balue is ignored. I need to take some additional tests

xDev789:
Hi @cheina97! Thank you for reviewing my PR. You are right to be sceptical about reusing the same peer endpoint but it is only being explicitly set when configureDevice function gets called (i.e. when PublicKey custom resource gets reconciled). In case the peer endpoint changes, wireguard will update the endpoint due to its automatic peer discovery. It is the same mechanism Liqo currently relies on, except without this PR, reconciling PublicKey CR results in connection interruption because discovered peer endpoint gets erased. As for the testing part, I've tested it on two Kind clusters with two nodes each and it worked as expected. We also use the stable version of Liqo with these patches applied to establish tunnels between the control and worker clusters.

cheina97:
> Hi @cheina97! Thank you for reviewing my PR. You are right to be sceptical about reusing the same peer endpoint but it is only being explicitly set when configureDevice function gets called (i.e. when PublicKey custom resource gets reconciled). In case the peer endpoint changes, wireguard will update the endpoint due to its automatic peer discovery. It is the same mechanism Liqo currently relies on, except without this PR, reconciling PublicKey CR results in connection interruption because discovered peer endpoint gets erased. As for the testing part, I've tested it on two Kind clusters with two nodes each and it worked as expected. We also use the stable version of Liqo with these patches applied to establish tunnels between the control and worker clusters.

I've tried and it seems that wireguard was not able to reconfigure automatically the IP since it is set forcefully by the controller. Instead, without explicit setup, that IP is updated correctly. I'm not against this change, but I would prefer to wait a little bit and test it properly.

Can you share all the scenarios you have tried? I would like to know what provider have you tried and which CNI were you using (also with Kind clusters)

xDev789:
> I've tried and it seems that wireguard was not able to reconfigure automatically the IP since it is set forcefully by the controller. Instead, without explicit setup, that IP is updated correctly. I'm not against this change, but I would prefer to wait a little bit and test it properly.
> 
> Can you share all the scenarios you have tried? I would like to know what provider have you tried and which CNI were you using (also with Kind clusters)

I totally agree with you, I strongly believe in shipping quality product and I'm not trying to rush this change either. We use RKE2 with Cilium in Kube-Proxy replacement mode for both server and client clusters. With Kind clusters, I also used Cilium. I've tried rescheduling the gateway client pod on another node. I've also tried the HA scenario when another pod obtains the lease and becomes the active gateway. Both scenarios resulted in connection being reestablished as expected. Could you share more details on the problematic case? You said it worked initially but some other tests helped you spot an issue.

cheina97:
> > I've tried and it seems that wireguard was not able to reconfigure automatically the IP since it is set forcefully by the controller. Instead, without explicit setup, that IP is updated correctly. I'm not against this change, but I would prefer to wait a little bit and test it properly.
> > Can you share all the scenarios you have tried? I would like to know what provider have you tried and which CNI were you using (also with Kind clusters)
> 
> I totally agree with you, I strongly believe in shipping quality product and I'm not trying to rush this change either. We use RKE2 with Cilium in Kube-Proxy replacement mode for both server and client clusters. With Kind clusters, I also used Cilium. I've tried rescheduling the gateway client pod on another node. I've also tried the HA scenario when another pod obtains the lease and becomes the active gateway. Both scenarios resulted in connection being reestablished as expected. Could you share more details on the problematic case? You said it worked initially but some other tests helped you spot an issue.

Thanks for your help, we just need to test it properly with other CNIs. I just need some time to test it and to process

cheina97:
Hi @xDev789 , sorry for the delay! We chatted with the other maintainers about this feature, and we’re leaning towards keeping it under a feature flag for now. Would you mind adding an if statement around your feature and including a flag in the helm chart to enable or disable it?

xDev789:
Hi, @cheina97, apologies for the delay. I've wrapped the feature in a feature flag as per your request.

fra98 (approved):

fra98:
@xDev789 thanks for your contribution. We are running extensive tests on most CNIs and if all pass we were thinking to **always** enable the feature on Helm value. We have a dedicated [PR](https://github.com/liqotech/liqo/pull/3302) for this, which also includes other smaller fixes:
- distinguish between device not found error and generic error
- avoid mutating status to `Connecting` at every reconcile

PR #3166, opened 2025-12-15
[Docs] Advanced networking documentation

riccardotornesello:
# Description

This PR aims to add documentation concerning advanced concepts of the liqo's network fabric.

## Contributors

- @cheina97 
- @MircoBarone 
- @riccardotornesello 
- @Gabbe64

### Documentation

The main contribution concerns the addition of a section on Liqo's architecture, specifically its network architecture.

The following pages have been added to Architecture / Liqo's network:

- Basic Concepts
- Pod-to-Pod Example
- Leaf-to-Leaf Communication
- NodePort
- Firewall Configuration
- Route Configuration
- IP
- InternalNode

### Sphinx

The Sphinx installation system has been changed: currently, the packages needed to compile the documentation are installed without a Python virtual environment, but this is no longer supported by default in new versions of PIP, so it has been necessary to use virtual environments in the Makefile.

Furthermore, to make local execution lighter and simpler, Nginx (via Docker) has been replaced with the native Python “http” module, which allows a static http server to be exposed.

Finally, the README has been modified to suggest that users use the autoserve command, which is more useful during development than serve thanks to its autoreload feature.


docs/README.md
```diff
@@ -33,5 +33,5 @@ Images should be put in the [docs/_static/images folder](docs/_static/images).
 In case you want to browse the documentation website on your local computer, the Makefile provides a simple way to do so and make it available at [http://localhost:8080](http://localhost:8080).
```
frisso:
```suggestion
In case you would like to browse the documentation website on your local computer, the `Makefile` provides a simple way to do so and make it available at [http://localhost:8080](http://localhost:8080).
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
+# Basic Concepts
+
+Liqo network is a cloud-native solution that enables seamless communication between Kubernetes clusters. In particular, it allows pod-to-pod communication across clusters. This means that the **services** and **endpointslices** reflection is not managed by the networking module (it is managed by the virtual-kubelet).
```
frisso:
```suggestion
The **Liqo network** is a cloud-native solution that enables seamless communication between Kubernetes clusters.
In particular, it allows pod-to-pod communication across clusters, i.e., enabling the IP traffic exiting from a pod on a first cluster to be delivered to another pod running on a second cluster.
This means that the **services** and **endpointslices** reflection is not managed by the networking module (it is managed by the virtual-kubelet).
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
...
+
+The two main areas of the Liqo network are:
```
frisso:
```suggestion
The Liqo network is split in to main parts:
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
...
+
+- **Internal network**: it manages the traffic inside the cluster and forwards it between nodes and gateways. It uses the Geneve protocol to encapsulate the traffic (the green tunnels in the image below).
+- **External network**: it manages the traffic between different clusters. It uses WireGuard to create a secure tunnel between gateways (the orange tunnels in the image below).
+
+![basic](../../_static/images/architecture/network/basev2.excalidraw.svg)
+
+## Internal network
+
+In the internal network, every node is connected to each gateway (one for each peering) using a dedicated Geneve tunnel. This architecture ensures that:
```
frisso:
```suggestion
In the internal network, each node is connected to the pod acting as **cluster gateway** (one for each peering) using a dedicated Geneve tunnel.
According to this architecture, each node has one Geneve interface that takes all the traffic generated by local pods and directed to a remote pod, and sends it to the gateway pod.
On the other side, the cluster gateway has a number of Geneve interfaces (hence, tunnels) equal to the number of nodes in the Kubernetes cluster.
As a consequence, the number of Geneve tunnels in Liqo is equal to the number of _nodes_, and it does not depend on the number of _pods_.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
...
+
```
frisso, line 20:
```suggestion
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
...
+- For each node, there is a dedicated Geneve interface on every gateway.
```
frisso:
```suggestion
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
...
+- For each gateway, there is a dedicated Geneve interface on every node.
```
frisso:
```suggestion
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
...
+
+Each Geneve interface is assigned an IP address from the **internal CIDR**. This allows the interfaces to be reachable and enables you to ping the Geneve interfaces both from the node and from the gateway, facilitating connectivity checks and troubleshooting.
```
frisso:
```suggestion
Each Geneve interface is assigned an IP address from the **internal CIDR**.
This allows the interfaces to be reachable and enables you to ping the Geneve interfaces both from the node and from the gateway, facilitating connectivity checks and troubleshooting.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Details on GENEVE
+
+[Geneve](https://github.com/torvalds/linux/blob/master/drivers/net/geneve.c) is used to traverse the CNI and reach the Gateway efficiently, without altering the original traffic generated by the pod.
+The GENEVE interface encapsulates Ethernet traffic into a UDP datagram without encryption (Ethernet over IP).
+![basic](../../_static/images/architecture/network/doc-Geneve.drawio.svg)
+
+The external IP packet has the node's address as its source address and the gateway as its destination address. The destination port is, by default for Liqo, 6091.
+Once encapsulated, the packet can be processed by the CNI without issues, as its destination is clearly defined.
+The encapsulated traffic is reintroduced into the CNI and routed to the Gateway. If the Gateway resides on a different node, the traffic may undergo additional encapsulation, resulting in nested encapsulation. This behavior depends on the CNI's handling of inter-node communication and is beyond the scope of Liqo.
```
frisso:
```suggestion
The encapsulated traffic is taken by the CNI and routed to the Gateway, as formally this traffic looks like a packet generated by the note itself.
If the Gateway resides on a different node, the traffic may undergo additional encapsulation depending on the CNI technology in use, possibly resulting in nested encapsulation.
This behavior depends on the CNI's handling of inter-node communication and is beyond the scope of Liqo.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### Known Limitations: Geneve and Kubernetes Services
+
+The use of Kubernetes Services (ClusterIP, NodePort, or LoadBalancer) as endpoints for Geneve tunnels is **not supported**. Kubernetes Services do not represent physical network interfaces but are implemented via DNAT rules and Connection Tracking.
```
frisso:
```suggestion
Technologies that leverage UDP tunnels, such as Geneve, do not support the traversal of intermediate NATs.
The reason can be found on the [RFC 7348](https://www.rfc-editor.org/rfc/rfc7348.html):

> It is recommended that the UDP source port number
> be calculated using a hash of fields from the inner packet --
> one example being a hash of the inner Ethernet frame's headers.
> This is to enable a level of entropy for the ECMP/load-
> balancing of the VM-to-VM traffic across the VXLAN overlay.

In other words, Linux performs _source port hashing_ when encapsulating the pod traffic in the Geneve tunnel, meaning the source port in the UDP header is ephemeral and is calculated per-packet based on 5-tuple of the internal packet.
Therefore, traffic tunneled in UDP cannot traverse the NAT because the UDP 5-tuple in the incoming/outgoing directions are not equal (due to the UDP source port hashing), hence the tuple cannot be found in the connection tracking table of the NAT.

Since Kubernetes Services are implemented with DNAT rules and Connection Tracking (hence, are not associated to any network interface), Geneve tunnels cannot support any kind of Kubernetes Service as endpoints (ClusterIP, NodePort, or LoadBalancer), but must be directed to a _physical_ IP of the pod/node, which is guaranteed to handled by the CNI without any NAT mechanism.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+> It is recommended that the UDP source port number
+> be calculated using a hash of fields from the inner packet --
+> one example being a hash of the inner Ethernet frame's headers.
+> This is to enable a level of entropy for the ECMP/load-
+> balancing of the VM-to-VM traffic across the VXLAN overlay.
+
+In other words, Linux performs _source port hashing_ when encapsulating the pod traffic in the Geneve tunnel, meaning the source port in the UDP header is ephemeral and is calculated per-packet based on 5-tuple of the internal packet.
+Therefore, traffic tunneled in UDP cannot traverse the NAT because the UDP 5-tuple in the incoming/outgoing directions are not equal (due to the UDP source port hashing), hence the tuple cannot be found in the connection tracking table of the NAT.
+
+Since Kubernetes Services are implemented with DNAT rules and Connection Tracking (hence, are not associated to any network interface), Geneve tunnels cannot support any kind of Kubernetes Service as endpoints (ClusterIP, NodePort, or LoadBalancer), but must be directed to a _physical_ IP of the pod/node, which is guaranteed to handled by the CNI without any NAT mechanism.
```
frisso:
```suggestion
Since Kubernetes Services are implemented with DNAT rules and Connection Tracking (hence, are not associated to any network interface), Geneve tunnels cannot support any kind of Kubernetes Service as endpoint (ClusterIP, NodePort, or LoadBalancer), but must be directed to the _real_ IP of the pod/node, which is guaranteed to handled by the CNI without any NAT mechanism.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+This architecture is incompatible with the Geneve protocol, as the modification of IP headers and the lack of a direct physical binding prevent proper tunnel establishment and packet decapsulation. **Geneve Tunnels must be established directly on Pod or Node IPs.**
```
frisso:
```suggestion
This is the reason why **Liqo creates Geneve tunnels that are established directly on Pod or Node IPs**.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
```
frisso:
```suggestion
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+This is the reason why internal networking does not use ClusterIPs, but instead relies on dedicated controllers that dynamically track the actual Gateway IP, with specific implementations for the [client](https://github.com/liqotech/liqo/blob/8543b7ff7c34e601b60cde3f1666f3146ed29b71/pkg/liqo-controller-manager/networking/internal-network/client-controller/client_controller.go#L130) and [server](https://github.com/liqotech/liqo/blob/8543b7ff7c34e601b60cde3f1666f3146ed29b71/pkg/liqo-controller-manager/networking/internal-network/server-controller/server_controller.go#L130) roles.
```
frisso:
```suggestion
As a consequence, the Liqo internal network relies on dedicated controllers that dynamically track the actual Gateway IP, with specific implementations for the [client](https://github.com/liqotech/liqo/blob/8543b7ff7c34e601b60cde3f1666f3146ed29b71/pkg/liqo-controller-manager/networking/internal-network/client-controller/client_controller.go#L130) and [server](https://github.com/liqotech/liqo/blob/8543b7ff7c34e601b60cde3f1666f3146ed29b71/pkg/liqo-controller-manager/networking/internal-network/server-controller/server_controller.go#L130) roles.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+## External network
+
+In the external network, traffic between different clusters is managed.
```
frisso:
```suggestion
The external network handles the traffic between two different clusters peered with Liqo.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+Each WireGuard tunnel is created inside a dedicated gateway pod. For every peering, a dedicated gateway pod is deployed in both clusters involved in the connection. Each gateway pod establishes a secure WireGuard tunnel that connects only to a single remote cluster. This means that every gateway can point to just one remote cluster, ensuring isolation and security for each inter-cluster connection.
```
frisso:
```suggestion
Although Liqo provides extensibility mechanisms that facilitate the adoption of different inter-cluster technologies, at the current stage only Wireguard is implemented.

Each cluster has a dedicated gateway pod per each peering; a point-to-point Wireguard tunnel with the remote cluster is terminated inside this gateway pod.
Hence, each gateway pod handles only a single, secure WireGuard tunnel, which is connected to the remote cluster, ensuring isolation and security for each inter-cluster connection.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+There are two types of gateway pods: **gateway-client** and **gateway-server**.
+
+- The **gateway-client** acts as a WireGuard client, initiating the connection to the remote cluster.
+- The **gateway-server** acts as a WireGuard server, accepting incoming connections from remote clients. The gateway-server must expose itself using a Kubernetes Service, which can be of type LoadBalancer, NodePort, or ClusterIP, depending on the cluster's networking setup and requirements.
+
```
frisso, line 68:
```suggestion

The role of the gateway, either _client_ or _server_, can be dynamically configured before starting the peering.
In other words, this role does not depend whether the cluster is a Liqo resource _provider_ or _consumer_, although the default choice is to have the cluster provider acting as a gateway server, and the consumer acting as gateway client.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+### Gateway Details
+
+Each gateway in Liqo is equipped with at least three distinct interfaces:
```
frisso:
```suggestion
Each gateway in Liqo handles at least three distinct interfaces:
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+- The default interface, which serves as the direct link to external networks.
+- A GENEVE interface for each node within the cluster (at least one in the case of a single-node cluster).
```
frisso:
```suggestion
- A GENEVE interface to each node within the cluster (at least one in the case of a single-node cluster).
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+- A tunnel interface for communication with the remote Gateway, WireGuard by default.
```
frisso:
```suggestion
- A tunnel interface to the remote Gateway, using WireGuard by default.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+The Gateway is also responsible for performing DNAT and SNAT to support the remapping of the remote cluster.
```
frisso:
```suggestion
The gateway pod is also responsible for performing DNAT and SNAT to support the remapping of the pod addresses used in the remote cluster, which is enabled automatically by Liqo in case both clusters use the same Pod CIDR.
More details on the [CIDR Remapping](#cidr-remapping) section below.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Further details about GENEVE
```
frisso:
```suggestion
#### Routing traffic inside the gateway
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+GENEVE-encapsulated traffic arrives at the gateway via the default interface. A GENEVE socket listens on port 6091. Here, the traffic is processed, decapsulated, and reintroduced into the gateway through the GENEVE interface itself. Once the packets are ready for transmission to the remote cluster, they are directed through the WireGuard interface using policy routing.
```
frisso:
```suggestion
The traffic generated by user pods, and directed to the remote cluster, arrives at the gateway pod via the default interface, encapsulated in a GENEVE tunnel.
This traffic is transported by the underlying CNI, as it appears a normal traffic from the IP address of a Node, and the IP address of the gateway pod; hence, the CNI does not have any glue that this traffic is actually a GENEVE tunnel.
This traffic arrives at the gateway via the default interface, directed to a GENEVE socket listening on port 6091.
Here, the traffic is processed, decapsulated, and reintroduced into the gateway through the GENEVE interface itself.
Once the packets are ready for transmission to the remote cluster, they are redirected to the WireGuard interface using policy routing.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+_Note_: Within the Gateway, a route is also configured to direct all traffic arriving from the tunnel and destined for a local pod back to the appropriate GENEVE interface.
```
frisso: "arriving from the tunnel": it means "arriving from the Wireguard tunnel", isn't?

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Details on Wireguard
+
+[WireGuard](https://www.wireguard.com/) is a secure protocol that encapsulates IP packets within UDP, ensuring encryption of the inner content. Unlike GENEVE, WireGuard operates at Layer 3 (L3), meaning its interface does not possess a MAC address.
```
frisso:
```suggestion
[WireGuard](https://www.wireguard.com/) is a secure protocol that encapsulates IP packets within UDP, ensuring encryption of the inner content.
Unlike GENEVE, WireGuard operates at Layer 3 (L3), meaning its interface does not have a MAC address.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+In Liqo, during configuration, the local WireGuard interface is set to contact the appropriate endpoint, which in this case is the address of the node hosting the remote Gateway.
```
frisso: "during configuration" means "when the tunnel is established in the Liqo peering phase", isn't it?

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,124 @@
...
+Once encapsulated, the traffic is reintroduced into the Gateway and exits via the default interface. It then traverses the node and exits through its primary interface to reach the remote endpoint.
```
frisso:
```suggestion
Once encapsulated, the traffic is re-injected into the Gateway and exits via the default interface of the pod.
It then traverses the node and exits through its primary interface to reach the remote endpoint.
Hence, also this traffic, tunneled with Wireguard, is handled as normal traffic by the underlying CNI, and it appears as generated by a local pod and directed to an IP address outside the cluster.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+The gateway pod is also responsible for performing DNAT and SNAT to support the remapping of the pod addresses used in the remote cluster, which is enabled automatically by Liqo in case both clusters use the same Pod CIDR.
+More details on the [CIDR Remapping](#cidr-remapping) section below.
+
+#### Routing traffic inside the gateway
+
+The traffic generated by user pods, and directed to the remote cluster, arrives at the gateway pod via the default interface, encapsulated in a GENEVE tunnel.
+This traffic is transported by the underlying CNI, as it appears a normal traffic from the IP address of a Node, and the IP address of the gateway pod; hence, the CNI does not have any glue that this traffic is actually a GENEVE tunnel.
+This traffic arrives at the gateway via the default interface, directed to a GENEVE socket listening on port 6091.
+Here, the traffic is processed, decapsulated, and reintroduced into the gateway through the GENEVE interface itself.
```
frisso: I'm not sure I understood the sentence.
I thought this traffic would be encapsulated in Wireguard, not in the Geneve tunnel.
Am I wrong? Can you please add some more details?

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+Once encapsulated, the traffic is re-injected into the Gateway and exits via the default interface of the pod.
+It then traverses the node and exits through its primary interface to reach the remote endpoint.
+Hence, also this traffic, tunneled with Wireguard, is handled as normal traffic by the underlying CNI, and it appears as generated by a local pod and directed to an IP address outside the cluster.
+![basic](../../_static/images/architecture/network/doc-Wireguard.drawio.svg)
+
+### Recap of Packets flow
+
+The flow of a packet from a pod in one cluster to a pod in another cluster involves several steps:
+
+1. The packet originates from a pod and reaches its node through a virtual Ethernet interface. This is managed by the CNI and is standard Kubernetes networking behavior (Liqo is not responsible for this part).
```
frisso:
```suggestion
1. The packet originates from a pod and reaches the TCP/IP stack of the node through a virtual Ethernet interface. This step is managed by the CNI and follows the standard behavior of Kubernetes networking (Liqo is not involved in this step).
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+2. From the node, the packet enters a Geneve tunnel using a specific [route](routeconfiguration.md#remote-cluster-id-node-gw-node).The packet arrives at the other end of the Geneve tunnel, inside the gateway pod.
```
frisso:
```suggestion
2. From the node, the packet enters into a Geneve tunnel using a specific [route](routeconfiguration.md#remote-cluster-id-node-gw-node), added by Liqo. The Liqo route captures all the traffic that goes towards the Pod CIDR of the remote Liqo cluster. The packet arrives at the other end of the Geneve tunnel, inside the gateway pod. 
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+3. The gateway does DNAT and routes the packet into the WireGuard tunnel([check [this route](routeconfiguration.md#remote-cluster-id-gw-ext-gateway)]).
```
frisso:
```suggestion
3. In case the local and remote clusters have overlapped Pod CIDRs, the gateway applies proper DNAT rules; then, it routes the packet into the WireGuard tunnel ([check [this route](routeconfiguration.md#remote-cluster-id-gw-ext-gateway)]), where is properly encrypted, then passed to the node TCP/IP stack for further processing.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+4. The traffic reaches the gateway pod in the remote cluster. It is decrypted by the WireGuard driver and reinserted into the stack. It is then routed to the correct Geneve interface to reach the node hosting the target pod.This is achieved using the following [routes](routeconfiguration.md#local-cluster-id-node-name-gw-node-gateway). After routing, SNAT is applied.
```
frisso:
```suggestion
4. From the node, the packet follows the normal TCP/IP stack processing and exits from the local cluster. This step is managed by the CNI and follows the standard behavior of Kubernetes networking (Liqo is not involved in this step).
5. The tunneled Wireguard traffic reaches the gateway pod in the remote cluster. It is decrypted by the WireGuard driver and re-injected into the stack, within the pod itself. Then, it is routed to the Geneve interface of the tunnel that terminates on the node hosting the target pod. This is achieved using the following [routes](routeconfiguration.md#local-cluster-id-node-name-gw-node-gateway). After routing, and in case the local and remote clusters have overlapped Pod CIDRs, SNAT is applied.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+5. The packet is then forwarded from the node to the target pod, leveraging the CNI.
```
frisso:
```suggestion
6. Once reached the target node, the packet exits from the Geneve tunnel and is then forwarded to the target pod. This step is managed by the CNI and follows the standard behavior of Kubernetes networking (Liqo is not involved in this step).
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+The return traffic follows the same path in reverse, ensuring symmetric routing between the two pods across clusters.
```
frisso:
```suggestion
The return traffic follows the same path in opposite order, ensuring symmetric routing between the two pods across clusters.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+![Packets flow](../../_static/images/architecture/network/baseflow-v2.excalidraw.svg)
+
+## CIDR Remapping
+
+Liqo enables two clusters to peer even if they have the same pod CIDR. Each cluster can independently remap the remote cluster’s pod CIDR to a different, locally chosen CIDR. This remapped CIDR is decided by each cluster on its own and is not shared with other clusters.
```
frisso:
```suggestion
Liqo enables two clusters to peer even if they have the same pod CIDR.
Each cluster can independently remap the remote cluster’s pod CIDR to a different, locally chosen CIDR.
This remapped CIDR is decided locally by each cluster and is not shared, nor need to be negotiated, with other clusters.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+For example, if both Cluster A and Cluster B use `10.0.0.0/16` as their pod CIDR, Cluster A can remap Cluster B’s pod CIDR (`10.0.0.0/16`) to a new CIDR, such as `11.0.0.0/16`. As a result, Cluster A will be able to reach a pod in Cluster B with IP `10.0.0.6` using the remapped IP `11.0.0.6`. The same applies in the opposite direction if Cluster B chooses to remap Cluster A’s CIDR.
```
frisso:
```suggestion
For example, if both Cluster A and Cluster B use `10.0.0.0/16` as their pod CIDR, Cluster A can remap Cluster B’s pod CIDR (`10.0.0.0/16`) to a new CIDR, such as `11.0.0.0/16`.
As a result, Cluster A will be able to reach a pod in Cluster B with IP `10.0.0.6` using the remapped IP `11.0.0.6`.
The same applies in the opposite direction if Cluster B chooses to remap Cluster A’s CIDR.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+When a packet directed to a remote pod enters the gateway, its destination IP is the remapped one. The gateway performs a DNAT operation to replace this remapped address (used only within the local cluster) with the corresponding real IP address used by the remote cluster.
```
frisso:
```suggestion
When a packet directed to a remote pod enters the gateway, its destination IP address is the remapped one.
The gateway performs a DNAT operation to replace this remapped address (used only within the local cluster) with the corresponding real IP address used by the remote cluster.
Hence, when the packet enters in the Wireguard tunnel, the destination IP address is the actual one of the remote destination pod.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+![Packets flow](../../_static/images/architecture/network/SNAT-DNAT.svg)
+
+In the example above a packet originating from a local pod destined for a remote pod enters the Gateway with the IP address `11.0.0.6` and will be translated to `10.0.0.6` by the Gateway before being sent to the remote cluster.
```
frisso:
```suggestion
In the above example, a packet originating from a local pod and directed to a remote pod enters the Gateway with the IP address `11.0.0.6`, which will be translated to `10.0.0.6` by the Gateway before sending it to the remote cluster.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+In a similar way, when the gateway receives incoming traffic, its source IP is the real one used by the remote cluster, so it is SNATed to the remapped one before being sent to the appropriate pod.
```
frisso:
```suggestion
However, the DNAT on the outgoing traffic is not enough to guarantee a successful communication to the remote pod.
In fact, when the gateway on the other side of the Wireguard tunnel receives the above packet as an incoming traffic, it applies a SNAT rule to the packet, hence remapping the source address (which still belongs to the Pod CIDR of the first cluster) into the proper remapped Pod CIDR of the first cluster.
Then, the packet continues its journey to the final pod.

Liqo clearly splits the responsibilities of the DNAT and SNAT steps (which are both required when two networks with overlapped IP addresses needs to be connected seamlessly without IP renumbering) into the two involved gateway pods: the _exiting_ gateway performs DNAT, the _entering_ gateway performs SNAT.
This splitting of responsibilities allow both gateways to run independently, without having to coordinate themselves about the addresses used in the IP remapping process, hence simplifying the overall management of this process.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+All the remapping logic is managed by the IPAM component of liqo, which tracks the used IPs and CIDRs inside the cluster.
```
frisso:
```suggestion
In fact, all the remapping logic is managed locally by the Liqo IPAM component, which tracks the used IPs and CIDRs inside the cluster, and therefore can simply decide which IP address ranges could be used in the remapping process.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+For example, if both Cluster A and Cluster B use `10.0.0.0/16` as their pod CIDR, Cluster A can remap Cluster B’s pod CIDR (`10.0.0.0/16`) to a new CIDR, such as `11.0.0.0/16`. As a result, Cluster A will be able to reach a pod in Cluster B with IP `10.0.0.6` using the remapped IP `11.0.0.6`. The same applies in the opposite direction if Cluster B chooses to remap Cluster A’s CIDR.
+
+When a packet directed to a remote pod enters the gateway, its destination IP is the remapped one. The gateway performs a DNAT operation to replace this remapped address (used only within the local cluster) with the corresponding real IP address used by the remote cluster.
+![Packets flow](../../_static/images/architecture/network/SNAT-DNAT.svg)
+
+In the example above a packet originating from a local pod destined for a remote pod enters the Gateway with the IP address `11.0.0.6` and will be translated to `10.0.0.6` by the Gateway before being sent to the remote cluster.
+
+In a similar way, when the gateway receives incoming traffic, its source IP is the real one used by the remote cluster, so it is SNATed to the remapped one before being sent to the appropriate pod.
+
```
frisso, line 146:
```suggestion
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+All the remapping logic is managed by the IPAM component of liqo, which tracks the used IPs and CIDRs inside the cluster.
+
+All the firewall rules for managing this remapping is handled inside the gateway pods, ensuring seamless communication even in the presence of overlapping CIDRs.
```
frisso: I cannot understand this sentence.
Why are we talking about "firewall rules" here? We never mentioned firewalling so far.

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+The same logic is applied also to the **external CIDR**. This means that every cluster will be able to remap its neighboring cluster's external CIDR to a different, locally chosen CIDR. Pay attention that if you don't change the default external CIDR used by Liqo (check the helm values), every cluster will use the same external CIDR and neighbour's external CIDRs will appear always remapped.
```
frisso:
```suggestion
The same logic is applied also to the **External CIDR**.
This means that each cluster will be able to remap its neighboring cluster's External CIDR to a different, locally chosen CIDR.
Pay attention that if you do not change the default External CIDR used by Liqo (check the Helm values), each cluster will use the same External CIDR, hence neighbour's external CIDRs will appear always remapped.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Example
+
+A full example of the entire path can be found [here](./basic_example.md)
```
frisso:
```suggestion
A complete example of the entire journey of a packet can be found [here](./basic_example.md)
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,147 @@
...
+Then, the packet continues its journey to the final pod.
+
+Liqo clearly splits the responsibilities of the DNAT and SNAT steps (which are both required when two networks with overlapped IP addresses needs to be connected seamlessly without IP renumbering) into the two involved gateway pods: the _exiting_ gateway performs DNAT, the _entering_ gateway performs SNAT.
+This splitting of responsibilities allow both gateways to run independently, without having to coordinate themselves about the addresses used in the IP remapping process, hence simplifying the overall management of this process.
+In fact, all the remapping logic is managed locally by the Liqo IPAM component, which tracks the used IPs and CIDRs inside the cluster, and therefore can simply decide which IP address ranges could be used in the remapping process.
+
+All the firewall rules for managing this remapping is handled inside the gateway pods, ensuring seamless communication even in the presence of overlapping CIDRs.
+
+The same logic is applied also to the **External CIDR**.
```
frisso: Here we start talking about External CIDR, but:
1) we didn't mention what this is
2) the explanation is very basic, and to me very difficult to understand. I would suggest to expand a little more, and explain better why we do have to change the default setting of the External CIDR in the Helm values.

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
+# Pod-to-Pod Example
```
frisso:
```suggestion
# Pod-to-Pod Packet Journey Example
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+
+Let's analyze and debug the complete path of an ICMP packet sent from a pod in the consumer cluster to a pod in the provider cluster.
```
frisso:
```suggestion
This example presents the complete journey of an ICMP packet sent from a pod in a consumer cluster to a pod in a provider cluster.
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Setup
+
+For this example, we will use two Kubernetes-in-Docker (KinD) clusters with only one worker each and Calico as the CNI.
+
+The entire flow of the packet is depicted in the image below
```
frisso:
```suggestion
The entire journey of the packet is shown in the image below.
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+  - **Pod:** `nginx-local` (`10.200.1.2`)
+  - **Gateway:** `gw-local` (`10.200.1.3`)
+  - **Node:** `rome-worker` (`172.18.0.2`)
+- **Cluster Milan (Remote):**
+  - **Pod:** `nginx-remote` (`10.200.1.4`)
+  - **Gateway:** `gw-remote` (`10.200.1.5`)
+  - **Node:** `milan-worker` (`172.18.0.3`)
+
+Since both clusters use `10.200.0.0/16` as the pod CIDR, Liqo remaps the pod CIDR of the remote cluster to `10.201.0.0/16` on the local cluster.
+
```
frisso:
```suggestion
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+This means that, from the perspective of the local cluster, the remote pod `nginx-remote` will appear to have the IP address `10.201.1.4`.
+
+This examples follows the path of an ICMP Echo Request packet sent from `nginx-local` to `nginx-remote` using the command: `ping 10.201.1.4`.
+
+## The packet flow
+
+### Local Node
+
+The ICMP packet exits the pod's default interface (e.g. `eth0`) and immediately enters the worker node's root network namespace via the other end of the veth pair (e.g., `caliaaaaa`). The node must now decide where to send this packet.
```
frisso:
```suggestion
The ICMP packet exits the pod's default interface (e.g. `eth0`) and immediately enters the worker node's root network namespace via the other end of the `veth` pair (e.g., `caliaaaaa`).
The node must now decide where to send this packet.
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+
+- **Policy Routing:** A custom policy [routing rule](routeconfiguration.md#remote-cluster-id-node-gw-node) matches the destination IP `10.201.0.0/16`. The matching rule is as follows: `10.201.0.0/16 via 10.71.0.3 dev liqo.00000`. This route sends the packet to a special interface named `liqo.00000`, which is a **GENEVE tunnel**.
```
frisso:
```suggestion
- **Policy Routing:** A custom [policy routing rule](routeconfiguration.md#remote-cluster-id-node-gw-node) matches the destination IP `10.201.0.0/16`. The matching rule is as follows: `10.201.0.0/16 via 10.71.0.3 dev liqo.00000`. This route sends the packet to a special interface named `liqo.00000`, which is a **GENEVE tunnel**.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,147 @@
...
+4. From the node, the packet follows the normal TCP/IP stack processing and exits from the local cluster. This step is managed by the CNI and follows the standard behavior of Kubernetes networking (Liqo is not involved in this step).
+5. The tunneled Wireguard traffic reaches the gateway pod in the remote cluster. It is decrypted by the WireGuard driver and re-injected into the stack, within the pod itself. Then, it is routed to the Geneve interface of the tunnel that terminates on the node hosting the target pod. This is achieved using the following [routes](routeconfiguration.md#local-cluster-id-node-name-gw-node-gateway). After routing, and in case the local and remote clusters have overlapped Pod CIDRs, SNAT is applied.
+6. Once reached the target node, the packet exits from the Geneve tunnel and is then forwarded to the target pod. This step is managed by the CNI and follows the standard behavior of Kubernetes networking (Liqo is not involved in this step).
+
+The return traffic follows the same path in opposite order, ensuring symmetric routing between the two pods across clusters.
+
+![Packets flow](../../_static/images/architecture/network/baseflow-v2.excalidraw.svg)
+
```
frisso, line 113:
```suggestion

### TCP Clamping
When the IP traffic traverses a tunnel, there is always the risk that large IP packets need to be fragmented before encapsulating them into the tunnel.
Unfortunately, not all the packets can be fragmented (e.g., the ones that have the IP _Don't Fragment_ flag turned on), and in this case the packet is dropped.

Although Liqo does it best to limit this problem, e.g., by reducing the MTU at the traffic source, this is not always effective.
Therefore, starting from version 1.1, Liqo implements the TCP Clamping mechanism on all the traffic traversing the Wireguard tunnel.
This behavior is controlled by a new flag that is active by default, although it can be changed in the [Helm values file](https://github.com/liqotech/liqo/blob/6b287acbfe0c2bb2ea065aa6f4ae7928b00a4d95/deployments/liqo/values.yaml#L60).

In a nutshell, when the TCP handshake mechanism establishes a new TCP session, the two TCP endpoints agree on the maximum chunk of data that can be exchanged on that connection, the so called Maximum Segment Size (MSS).
The TCP clamping mechanism acts as a _bump in the wire_ and it intercepts this negotiation, adapting the MSS value to the MTU of the Wireguard tunnel, so that the TCP endpoints will never generate a chunk of data that requires the IP packet to be fragmented.

For more information, check the _TCP Clamping_ section on this [documentation](https://www.cloudflare.com/learning/network-layer/what-is-mss/) provided by Cloudflare.

```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+- **Encapsulation:** This interface encapsulates the _entire L2 Ethernet frame_ (not just the IP packet) inside a new UDP packet.
+  - **Inner Packet (Original):** `10.200.1.2` > `10.201.1.4`
+  - **Outer Packet (New):**
+    - **Source IP:** `172.18.0.2` (the `rome-worker` node)
+    - **Destination IP:** `10.200.1.3` (the `gw-local` pod)
+    - **Destination Port:** `6091` (the GENEVE port)
+
+This new, larger UDP packet is then sent via the standard CNI network to the `gw-local` pod.
+
+In this case, with just one node, the packet is sent directly to the `gw-local` pod, however in a multi-node cluster, the CNI may route it to another node first.
```
frisso:
```suggestion
In this case, with just one node, the packet is sent directly to the `gw-local` pod, however in a multi-node cluster, the CNI may route it to another node first, actually the node that hosts the gateway pod.
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+
+```{admonition} TIP
+To inspect the policy routing, use the command `ip rule show all` to find the correct table and `ip route show table <table_id>`.
```
frisso:
```suggestion
To inspect the routing rules set by policy routing, use the command `ip rule show all` to find the correct table and then `ip route show table <table_id>`.
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+### Local Gateway
+
+The encapsulated packet arrives at the gateway's default interface (e.g. `eth0`).
```
frisso:
```suggestion
The encapsulated packet arrives at the default interface (e.g. `eth0`) of the gateway pod.
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+  > `SRC: 10.200.1.2` > `DST: 10.201.1.4`
+- **DNAT (Destination NAT):** This packet's destination (`10.201.1.4`) is meaningless to the remote cluster. The gateway must translate it back to the pod's _real_ IP. This is done using **`nftables`**. A rule in the `remap-podcidr` table performs a Destination NAT:
+  > `ip daddr 10.201.0.0/16 ... dnat prefix to 10.200.0.0/16`
+- **Packet Transformation:**
+  - **Before DNAT:** `10.200.1.2` > `10.201.1.4`
+  - **After DNAT:** `10.200.1.2` > `10.200.1.4`
+- **Routing to WireGuard:** Now that the packet has its _real_ destination, it's routed to the inter-cluster tunnel. The [route](routeconfiguration.md#remote-cluster-id-gw-ext-gateway) is: `10.200.0.0/16 via 169.254.18.1 dev liqo-tunnel`
+
+```{admonition} TIP
+To know which is the other end of a GENEVE interface, use the command `ip -d link`, both ends must have the same ID.
```
frisso:
```suggestion
To know which is the other end of a GENEVE interface, use the command `ip -d link`; both ends must have the same ID.
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+### Local Gateway
+
+The encapsulated packet arrives at the gateway's default interface (e.g. `eth0`).
+
+- **Decapsulation:** The gateway's networking stack recognizes the `6091` destination port and directs the packet to its own GENEVE interface (e.g., `liqo.11111`), which is paired with the worker's interface. The gateway strips the outer UDP/IP headers, extracting the original inner packet:
+  > `SRC: 10.200.1.2` > `DST: 10.201.1.4`
+- **DNAT (Destination NAT):** This packet's destination (`10.201.1.4`) is meaningless to the remote cluster. The gateway must translate it back to the pod's _real_ IP. This is done using **`nftables`**. A rule in the `remap-podcidr` table performs a Destination NAT:
+  > `ip daddr 10.201.0.0/16 ... dnat prefix to 10.200.0.0/16`
```
frisso, line 61: Can we add the nftables command that shows this rule?

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+
+- **Encapsulation:** WireGuard encrypts the IP packet and wraps it in a new UDP packet for transport.
+  - **Inner Packet (NATTed):** `10.200.1.2` > `10.200.1.4`
+  - **Outer Packet:**
+    - **Source IP:** `10.200.1.3` (the `gw-local` pod)
+    - **Destination IP:** `172.18.0.3` (the `milan-worker` node, where `gw-remote` runs)
+- **Node-Level SNAT:** This encrypted packet exits the `gw-local` pod and goes back to the `rome-worker` node. The node performs _another_ SNAT (Source NAT) to make the packet routable on the external network.
+  - **Before SNAT:** `10.200.1.3:36252` > `172.18.0.3:31864`
+  - **After SNAT:** `172.18.0.2:17806` > `172.18.0.3:31864`
+
```
frisso, line 85: I would suggest to add that this second SNAT is done by the CNI, and has nothing to do with Liqo. It is actually done on all the packets that, from this cluster, go to the public Internet.

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+This final, encrypted, and twice-NATted packet now travels "across the internet" (or in this case, the Docker network) from `rome-worker` to `milan-worker`.
+
+### Remote Node
+
+The encrypted packet arrives at the `milan-worker`'s default interface (e.g. `eth0`).
+
+- **Node-Level DNAT:** Since the packet is addressed to the node itself, the node's kube-proxy intercepts it and performs a Destination NAT to send it to the `gw-remote` pod.
```
frisso: Idem.
Also here I would specify that this DNAT is performed by default by the CNI on all incoming packets, and has nothing to do with Liqo.

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Final Delivery
+
+1. The GENEVE packet exits the `gw-remote` pod and enters the `milan-worker` node, which in our case is also the node hosting the target pod.
+2. The node's GENEVE interface (`liqo.33333`) decapsulates the packet, retrieving:
+    > `SRC: 10.201.1.2` > `DST: 10.200.1.4`
+3. The node's routing table (created by the CNI plugin) directs this packet to the final pod via the appropriate end of the `veth` pair.
+4. The packet travels across the `veth` and arrives at the default interface of the `nginx-remote` pod.
+
+The `nginx-remote` pod successfully receives the ICMP Echo Request from the (remapped) source `10.201.1.2` and sends an Echo Reply, which follows the entire path in reverse.
```
frisso:
```suggestion
The `nginx-remote` pod successfully receives the ICMP Echo Request from the (remapped) source `10.201.1.2` and sends an Echo Reply, which follows the entire path in the opposite direction.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,151 @@
...
+### TCP Clamping
+TODO
```
frisso: Complete this part.

docs/conf.py
```diff
@@ -54,11 +54,13 @@
...
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ['_build', '_legacy', 'Thumbs.db', '.DS_Store', 'README.md', 'requirements.txt']
+exclude_patterns = ['_build', '_legacy', 'Thumbs.db', '.DS_Store', 'README.md', 'requirements.txt', '.venv']
 
 linkcheck_ignore = [
     r'http://localhost:\d+/?',  # Localhost links
     'https://github.com/virtual-kubelet/virtual-kubelet#liqo-provider',  # This anchor is not found by the linter, but it works
+    'https://github.com/liqotech/liqo/blob/8543b7ff7c34e601b60cde3f1666f3146ed29b71/pkg/liqo-controller-manager/networking/internal-network/client-controller/client_controller.go#L130', # This anchor is not found by the linter, but it works
+    'https://github.com/liqotech/liqo/blob/8543b7ff7c34e601b60cde3f1666f3146ed29b71/pkg/liqo-controller-manager/networking/internal-network/server-controller/server_controller.go#L130', # This anchor is not found by the linter, but it works
     'https://www.wireguard.com/', # knwon recurrent error 500
```
frisso:
```suggestion
    'https://www.wireguard.com/', # known recurrent error 500
```

docs/architecture/network/basic.md
frisso:
```suggestion

When the IP traffic traverses a tunnel, there is always the risk that large IP packets need to be fragmented before encapsulating them into the tunnel.
Unfortunately, not all the packets can be fragmented (e.g., the ones that have the IP _Don't Fragment_ flag turned on), and in this case the packet is dropped.
	
Although Liqo does it best to limit this problem, e.g., by reducing the MTU at the traffic source, this is not always effective.
Therefore, starting from version 1.1, Liqo implements the TCP Clamping mechanism on all the traffic traversing the Wireguard tunnel.
This behavior is controlled by a new flag that is active by default, although it can be changed in the [Helm values file](https://github.com/liqotech/liqo/blob/6b287acbfe0c2bb2ea065aa6f4ae7928b00a4d95/deployments/liqo/values.yaml#L60).
	
In a nutshell, when the TCP handshake mechanism establishes a new TCP session, the two TCP endpoints agree on the maximum chunk of data that can be exchanged on that connection, the so called Maximum Segment Size (MSS).
The TCP clamping mechanism acts as a _bump in the wire_ and it intercepts this negotiation, adapting the MSS value to the MTU of the Wireguard tunnel, so that the TCP endpoints will never generate a chunk of data that requires the IP packet to be fragmented.

```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+5. The tunneled Wireguard traffic reaches the gateway pod in the remote cluster. It is decrypted by the WireGuard driver and re-injected into the stack, within the pod itself. Then, it is routed to the Geneve interface of the tunnel that terminates on the node hosting the target pod. This is achieved using the following [routes](routeconfiguration.md#local-cluster-id-node-name-gw-node-gateway). After routing, and in case the local and remote clusters have overlapped Pod CIDRs, SNAT is applied.
+6. Once reached the target node, the packet exits from the Geneve tunnel and is then forwarded to the target pod. This step is managed by the CNI and follows the standard behavior of Kubernetes networking (Liqo is not involved in this step).
+
+The return traffic follows the same path in opposite order, ensuring symmetric routing between the two pods across clusters.
+
+![Packets flow](../../_static/images/architecture/network/baseflow-v2.excalidraw.svg)
+
+### TCP Clamping
```
frisso, line 114: @cheina97 Could you please check this section?

docs/architecture/network/main.md
```diff
@@ -0,0 +1,24 @@
+# Liqo's Network
```
cheina97: I would call it only network

docs/architecture/network/main.md
```diff
@@ -0,0 +1,24 @@
...
+
+Below you will find links to the available topics in this section:
+
+- [Basic Concepts](basic.md): This page presents the foundational concepts and general information about the Liqo networking model.
+  - [Pod-to-Pod Example](basic_example.md) An example illustrating the basic concepts of Liqo.
```
cheina97: This indentation looks ugly to me when rendered. What about avoiding it there and includes this link in the basic.md page?

docs/architecture/network/main.md
```diff
@@ -0,0 +1,24 @@
...
+- [InternalNode](internalnode.md): This page provides detailed information about the InternalNode resource and its role in the Liqo network.
+
+## Useful debug commands for Liqo networking
+
+- `tcpdump -tnl -i any \<protocol\>`: capture all traffic for a specific protocol (e.g., `tcp`, `udp`, `icmp`, etc.)
+- `tcpdump -tnl -i any tcp port 8080`: capture all TCP traffic on port 8080
+- `tcpdump -tnl -i any tcp dst port 8080`: capture traffic destined to port 8080
+- `conntrack -D`: delete all the conntrack entries
+- `conntrack -L`: list all the conntrack entries
+- `nft list ruleset`: list all the rules in the nftables
```
cheina97: I think they are pretty useless here without a context, We already have a debug section under FAQ, maybe we can increase it or move it in another place

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+The Liqo network is split in to main parts:
+
+- **Internal network**: it manages the traffic inside the cluster and forwards it between nodes and gateways. It uses the Geneve protocol to encapsulate the traffic (the green tunnels in the image below).
+- **External network**: it manages the traffic between different clusters. It uses WireGuard to create a secure tunnel between gateways (the orange tunnels in the image below).
+
+![basic](../../_static/images/architecture/network/basev2.excalidraw.svg)
+
+## Internal network
+
+In the internal network, each node is connected to the pod acting as **cluster gateway** (one for each peering) using a dedicated Geneve tunnel.
```
cheina97: We call it just gateway

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Details on GENEVE
+
+[Geneve](https://github.com/torvalds/linux/blob/master/drivers/net/geneve.c) is used to traverse the CNI and reach the Gateway efficiently, without altering the original traffic generated by the pod.
+The GENEVE interface encapsulates Ethernet traffic into a UDP datagram without encryption (Ethernet over IP).
+![basic](../../_static/images/architecture/network/doc-Geneve.drawio.svg)
+
+The external IP packet has the node's address as its source address and the gateway as its destination address. The destination port is, by default for Liqo, 6091.
+Once encapsulated, the packet can be processed by the CNI without issues, as its destination is clearly defined.
+The encapsulated traffic is taken by the CNI and routed to the Gateway, as formally this traffic looks like a packet generated by the note itself.
```
cheina97:
```suggestion
The encapsulated traffic is taken by the CNI and routed to the Gateway, as formally this traffic looks like a packet generated by the node itself.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+
+In the internal network, each node is connected to the pod acting as **cluster gateway** (one for each peering) using a dedicated Geneve tunnel.
+According to this architecture, each node has one Geneve interface that takes all the traffic generated by local pods and directed to a remote pod, and sends it to the gateway pod.
+On the other side, the cluster gateway has a number of Geneve interfaces (hence, tunnels) equal to the number of nodes in the Kubernetes cluster.
+As a consequence, the number of Geneve tunnels in Liqo is equal to the number of _nodes_, and it does not depend on the number of _pods_.
+
+Each Geneve interface is assigned an IP address from the **internal CIDR**.
+This allows the interfaces to be reachable and enables you to ping the Geneve interfaces both from the node and from the gateway, facilitating connectivity checks and troubleshooting.
+
+### Details on GENEVE
```
cheina97, line 24: It would be cool to specify the size and the overhead of the GENEVE encapsulation

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+5. The tunneled Wireguard traffic reaches the gateway pod in the remote cluster. It is decrypted by the WireGuard driver and re-injected into the stack, within the pod itself. Then, it is routed to the Geneve interface of the tunnel that terminates on the node hosting the target pod. This is achieved using the following [routes](routeconfiguration.md#local-cluster-id-node-name-gw-node-gateway). After routing, and in case the local and remote clusters have overlapped Pod CIDRs, SNAT is applied.
+6. Once reached the target node, the packet exits from the Geneve tunnel and is then forwarded to the target pod. This step is managed by the CNI and follows the standard behavior of Kubernetes networking (Liqo is not involved in this step).
+
+The return traffic follows the same path in opposite order, ensuring symmetric routing between the two pods across clusters.
+
+![Packets flow](../../_static/images/architecture/network/baseflow-v2.excalidraw.svg)
+
+### TCP Clamping
+
+When the IP traffic traverses a tunnel, there is always the risk that large IP packets need to be fragmented before encapsulating them into the tunnel.
```
cheina97, line 116: I would add some details.
The packet should become too big because the CNI MTU (which sets the maximum size of the packets generated by the pods) should be equal to the VPC or precisely configured to prevent fragmentation during CNI’s encapsulation. So in a common environment (without Liqo) the MTU of the CNI should be X-Y where X is the VPC (or LAN) MTU and Y the overhead caused by the CNI encapsulation protocol. Liqo adds the GENEVE and the WireGuard overhead to this equation, so the equation should become CNI_MTU=VPC_MTU-CNI_OVERHEAD-WIREGUARD_OVERHEAD (I'm not considering Geneve overhead because it is lower to the wireguard's one). The problem is that we cannot change the CNI MTU, but the MSS clamping allows us to override the MTU for the TCP connections directed to the remote cluster

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+Unfortunately, not all the packets can be fragmented (e.g., the ones that have the IP _Don't Fragment_ flag turned on), and in this case the packet is dropped.
+	
+Although Liqo does it best to limit this problem, e.g., by reducing the MTU at the traffic source, this is not always effective.
```
cheina97, line 118: We never did it, and I don't even think it's possible

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+Therefore, starting from version 1.1, Liqo implements the TCP Clamping mechanism on all the traffic traversing the Wireguard tunnel.
+This behavior is controlled by a new flag that is active by default, although it can be changed in the [Helm values file](https://github.com/liqotech/liqo/blob/6b287acbfe0c2bb2ea065aa6f4ae7928b00a4d95/deployments/liqo/values.yaml#L60).
+	
+In a nutshell, when the TCP handshake mechanism establishes a new TCP session, the two TCP endpoints agree on the maximum chunk of data that can be exchanged on that connection, the so called Maximum Segment Size (MSS).
+The TCP clamping mechanism acts as a _bump in the wire_ and it intercepts this negotiation, adapting the MSS value to the MTU of the Wireguard tunnel, so that the TCP endpoints will never generate a chunk of data that requires the IP packet to be fragmented.
```
cheina97, line 122: I would add that in the 99% of cases, it is not an isue with UDP, since the Don't Fragment is usually disabled for UDP traffic (while it is always enabled for TCP)

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+
+## CIDR Remapping
+
+Liqo enables two clusters to peer even if they have the same pod CIDR.
+Each cluster can independently remap the remote cluster’s pod CIDR to a different, locally chosen CIDR.
+This remapped CIDR is decided locally by each cluster and is not shared, nor need to be negotiated, with other clusters.
+
+For example, if both Cluster A and Cluster B use `10.0.0.0/16` as their pod CIDR, Cluster A can remap Cluster B’s pod CIDR (`10.0.0.0/16`) to a new CIDR, such as `11.0.0.0/16`.
+As a result, Cluster A will be able to reach a pod in Cluster B with IP `10.0.0.6` using the remapped IP `11.0.0.6`.
+The same applies in the opposite direction if Cluster B chooses to remap Cluster A’s CIDR.
```
cheina97, line 134: I would add that the "brain" for the CIDRS and IPs management in Liqo is our IPAM service, which is a little bit different from common IPAMs. They usually only store the used cidr or ips used. In our case i is able to always return a free CIDR/IP also if the requested one is not free

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,161 @@
...
+![Packets flow](../../_static/images/architecture/network/SNAT-DNAT.svg)
+
+In the above example, a packet originating from a local pod and directed to a remote pod enters the Gateway with the IP address `11.0.0.6`, which will be translated to `10.0.0.6` by the Gateway before sending it to the remote cluster.
+
+However, the DNAT on the outgoing traffic is not enough to guarantee a successful communication to the remote pod.
+In fact, when the gateway on the other side of the Wireguard tunnel receives the above packet as an incoming traffic, it applies a SNAT rule to the packet, hence remapping the source address (which still belongs to the Pod CIDR of the first cluster) into the proper remapped Pod CIDR of the first cluster.
+Then, the packet continues its journey to the final pod.
+
+Liqo clearly splits the responsibilities of the DNAT and SNAT steps (which are both required when two networks with overlapped IP addresses needs to be connected seamlessly without IP renumbering) into the two involved gateway pods: the _exiting_ gateway performs DNAT, the _entering_ gateway performs SNAT.
```
cheina97, line 147: Can you explain which is the entering and the exiting gateway with a dedicated sketch or referring to another one?

docs/architecture/network/basic.md
cheina97: Me too.

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,115 @@
+# Pod-to-Pod Packet Journey Example
```
cheina97, line 1: I would start with a reminder to read the basic.md documents before starting this journey

docs/architecture/network/main.md
cheina97: I would keep the link about the "packet journey" on the same level of the other links

cheina97 (changes requested):
A general comments. I would group all the crds in thi dropdown, in order to have 2 main sections called concepts and CRDs with all the pages inside. Is it possible or sphinx doesn't allow it?

Image

cheina97 (changes requested):
I've noticed that some images are in excalidraw format. Can you redraw it in draw.io? I know it's a lot of work, so for me it's ok if you focus on the docs "content" and redraw them in a next PR

cheina97 (changes requested):
For some reason, Images like this one uses a "dark" palette, even if the original draw.io file is not using it. In general, we never uses dark palette, even in "dark mode". Can you check what is happening? Some draw.io behave like that, while some others do not

Image

This is the only draw.io added by you which is using the "light palette"

Image

cheina97 (changes requested):
https://github.com/liqotech/liqo/pull/3198
In the next few days we will merge this PR, which updates all the Python packages used for the documentation. I suggest you to rebase and check if everything is working good

MircoBarone:
> For some reason, Images like this one uses a "dark" palette, even if the original draw.io file is not using it. In general, we never uses dark palette, even in "dark mode". Can you check what is happening? Some draw.io behave like that, while some others do not
> 
@cheina97 I've re-exported all the draw.io images with a transparent background. I believe this should fix the dark palette issue. Could you please check if it looks correct now?

docs/architecture/network/basic.md
MircoBarone: Fixed. Replaced with "gateway".

docs/architecture/network/basic.md
MircoBarone: Applied

docs/architecture/network/basic.md
MircoBarone: Applied the suggestion: included the overhead of GENEVE encapsulation (14 bytes Ethernet + 20 bytes IPv4 + 8 bytes UDP + 8 bytes GENEVE).

docs/architecture/network/basic.md
MircoBarone: Applied: expanded the IPAM explanation to clarify its role in CIDR and IP management. please let me know if further clarification is needed.

docs/architecture/network/basic.md
MircoBarone: This is a bit tricky because in the current sketch both GW-Local and GW-Remote act as entering and exiting gateways depending on the traffic direction. Do you suggest adding another, very similar sketch showing only the transmission direction, or would that be too redundant?

docs/architecture/network/basic_example.md
MircoBarone: Added!

docs/architecture/network/main.md
MircoBarone: Put at the same level. There is already a link to the example at the end of basic.md. Do you think that is sufficient?

docs/architecture/network/basic_example.md
MircoBarone: Clarified

docs/architecture/network/basic_example.md
MircoBarone: Clarified: this DNAT is performed by the node/CNI because the Remote Gateway is behind a NodePort Service

docs/architecture/network/basic_example.md
MircoBarone: Added an {admonition} TIP  section with the command

cheina97:
> > For some reason, Images like this one uses a "dark" palette, even if the original draw.io file is not using it. In general, we never uses dark palette, even in "dark mode". Can you check what is happening? Some draw.io behave like that, while some others do not
> 
> @cheina97 I've re-exported all the draw.io images with a transparent background. I believe this should fix the dark palette issue. Could you please check if it looks correct now?

It worked, thx :)

docs/architecture/network/basic.md
MircoBarone: Yes. Changed to "arriving from the Wireguard interface" that is more clear

docs/architecture/network/basic.md
MircoBarone: Changed to "when the Gateway is created". I think "during the peering phase" is too generic, since gateways can also be created without peering (e.g., using `liqo network connect`). Am I right?

docs/architecture/network/basic.md
MircoBarone: That part was unclear. The traffic arrives encapsulated on the Geneve interface, and the Geneve implementation decapsulates it and reintroduces it into the gateway networking stack. After that, it is routed to the WireGuard interface. I tried to clarify this, please let me know if it is still unclear.

docs/architecture/network/basic.md
MircoBarone: Changed to "NAT rules". The confusion might have been with the FirewallConfiguration CRD that contains t
he high level spec for the rules

docs/architecture/network/basic.md
MircoBarone: Tried to clarify. Not sure if the explanation of the External CIDR is now complete.

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,120 @@
...
```
frisso, line 1:
```suggestion
# Pod-to-Pod Packet Journey Example

```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,120 @@
...
+This example presents the complete journey of an ICMP packet sent from a pod in a consumer cluster to a pod in a provider cluster.
+
+Before reading this example, see the [Basic Concepts](basic.md) section
+## Setup
```
frisso, line 7:
```suggestion

## Setup
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Local Gateway
+
+The encapsulated packet arrives at the default interface (e.g. `eth0`) of the gateway pod.
+
+- **Decapsulation:** The gateway's networking stack recognizes the `6091` destination port and directs the packet to its own GENEVE interface (e.g., `liqo.11111`), which is paired with the worker's interface. The gateway strips the outer UDP/IP headers, extracting the original inner packet:
+  > `SRC: 10.200.1.2` > `DST: 10.201.1.4`
+- **DNAT (Destination NAT):** This packet's destination (`10.201.1.4`) is meaningless to the remote cluster. The gateway must translate it back to the pod's _real_ IP. This is done using **`nftables`**. A rule in the `remap-podcidr` table performs a Destination NAT:
+  > `ip daddr 10.201.0.0/16 ... dnat prefix to 10.200.0.0/16`
+```{admonition} TIP
```
frisso:
```suggestion

```{admonition} TIP
```

docs/architecture/network/basic_example.md
```diff
@@ -0,0 +1,120 @@
...
+It is possible to see this rule, together with the SNAT rule for incoming traffic, using the command:
+nft list table ip remap-podcidr
+```
+- **Packet Transformation:**
```
frisso, line 63:
```suggestion

- **Packet Transformation:**
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,163 @@
...
+It then traverses the node and exits through its primary interface to reach the remote endpoint.
+Hence, also this traffic, tunneled with Wireguard, is handled as normal traffic by the underlying CNI, and it appears as generated by a local pod and directed to an IP address outside the cluster.
+![basic](../../_static/images/architecture/network/doc-Wireguard.drawio.svg)
+
+### Recap of Packets flow
+
+The flow of a packet from a pod in one cluster to a pod in another cluster involves several steps:
+
+1. The packet originates from a pod and reaches the TCP/IP stack of the node through a virtual Ethernet interface. This step is managed by the CNI and follows the standard behavior of Kubernetes networking (Liqo is not involved in this step).
+2. From the node, the packet enters into a Geneve tunnel using a specific [route](routeconfiguration.md#remote-cluster-id-node-gw-node), added by Liqo. The Liqo route captures all the traffic that goes towards the Pod CIDR of the remote Liqo cluster. The packet arrives at the other end of the Geneve tunnel, inside the gateway pod.
```
frisso:
```suggestion
2. From the node, the packet enters into a Geneve tunnel using a specific [route](routeconfiguration.md#remote-cluster-id-node-gw-node), added by Liqo. The Liqo route captures all the traffic that goes towards the Pod CIDR of the remote Liqo cluster. The packet arrives at the other end of the Geneve tunnel, inside the gateway pod.
```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,163 @@
...
+
+The return traffic follows the same path in opposite order, ensuring symmetric routing between the two pods across clusters.
+
+![Packets flow](../../_static/images/architecture/network/baseflow-v2.excalidraw.svg)
+
+### TCP Clamping
+
+When the IP traffic traverses a tunnel, there is always the risk that large IP packets need to be fragmented before encapsulating them into the tunnel.
+Unfortunately, not all the packets can be fragmented (e.g., the ones that have the IP _Don't Fragment_ flag turned on), and in this case the packet is dropped.
+
```
frisso:
```suggestion

```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,163 @@
...
+### TCP Clamping
+
+When the IP traffic traverses a tunnel, there is always the risk that large IP packets need to be fragmented before encapsulating them into the tunnel.
+Unfortunately, not all the packets can be fragmented (e.g., the ones that have the IP _Don't Fragment_ flag turned on), and in this case the packet is dropped.
+	
+Although Liqo does it best to limit this problem, e.g., by reducing the MTU at the traffic source, this is not always effective.
+Therefore, starting from version 1.1, Liqo implements the TCP Clamping mechanism on all the traffic traversing the Wireguard tunnel.
+This behavior is controlled by a new flag that is active by default, although it can be changed in the [Helm values file](https://github.com/liqotech/liqo/blob/6b287acbfe0c2bb2ea065aa6f4ae7928b00a4d95/deployments/liqo/values.yaml#L60).
+
```
frisso:
```suggestion

```

docs/architecture/network/basic.md
```diff
@@ -0,0 +1,163 @@
...
+All the NAT rules for managing this remapping are handled inside the gateway pods, ensuring seamless communication even in the presence of overlapping CIDRs.
+
+The same remapping logic also applies to the **External CIDR**.
+External CIDR is a dedicated IP range, its addressed are used as proxy destinations when the cluster has no direct route to a remote Pod. Traffic sent to an external-CIDR address is recognized by the network fabric and routed across the inter-cluster tunnels to the provider cluster that actually hosts the Pod.
+Since the External CIDR may overlap across clusters, Liqo applies the same remapping mechanism used for Pod CIDRs, allowing each cluster to locally remap a neighbor’s External CIDR to a non-conflicting range.
+By default, Liqo assigns the same External CIDR to every cluster. If this default value is not changed (see the Helm values), all neighboring clusters will appear to use overlapping External CIDRs, and remapping will always be required. To avoid remapping, you must change the default value.
+
+
```
frisso:
```suggestion
```

PR #3226, opened 2026-03-14
feat: handle creation and configuration of multiple WireGuard tunnels

MircoBarone:
# Description

Part of the multi-tunnel WireGuard implementation. This is the first PR related to issue #3225.

Add `--listen-ports` and `--endpoint-ports` flags to the WireGuard container.

These flags allow specifying multiple ports for the Gateway server and client, respectively.  
The original idea was to also include the `--num-interfaces` flag to indicate the number of WireGuard interfaces to create, but it is unnecessary and redundant: the same information can be inferred from the number of ports, maintaining a single source of truth.

The existing `--listen-port` and `--endpoint-port` flags are preserved for backward compatibility and default values. They **can be ignored** if values are provided in the new multi-port flags, ensuring that the existing infrastructure continues to work without changes.

### DIFF
--- a/apis/networking/v1beta1/gatewayserver_types.go
+++ b/apis/networking/v1beta1/gatewayserver_types.go
@@ -36,16 +36,32 @@ var GatewayServerGroupResource = schema.GroupResource{Group: GroupVersion.Group,
 var GatewayServerGroupVersionResource = GroupVersion.WithResource(GatewayServerResource)
 
 // Endpoint defines the endpoint of the gatewayserver.
+// +kubebuilder:validation:XValidation:rule="!(has(self.port) && has(self.ports))",message="port and ports are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!(has(self.nodePort) && has(self.nodePorts))",message="nodePort and nodePorts are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="has(self.port) || (has(self.ports) && size(self.ports) > 0)",message="at least one of port or ports must be set"
+// +kubebuilder:validation:XValidation:rule="!(has(self.ports)) || self.ports.all(p, self.ports.filter(q, q == p).size() == 1)",message="ports must not contain duplicates"
+// +kubebuilder:validation:XValidation:rule="!(has(self.nodePorts)) || self.nodePorts.all(p, self.nodePorts.filter(q, q == p).size() == 1)",message="nodePorts must not contain duplicates"
+//
+//nolint:lll // ignore long lines given by Kubebuilder marker annotations
 type Endpoint struct {
+	// Deprecated: use Ports instead
 	// Port specifies the port of the endpoint.
 	Port int32 `json:"port,omitempty"`
+	// Ports specifies the ports of the endpoint.
+	// +kubebuilder:validation:MaxItems=64
+	Ports []int32 `json:"ports,omitempty"`
 	// ServiceType specifies the type of the service.
 	// +kubebuilder:default=ClusterIP
 	// +kubebuilder:validation:Enum=ClusterIP;NodePort;LoadBalancer;ExternalName
 	ServiceType corev1.ServiceType `json:"serviceType,omitempty"`
+	// Deprecated: use NodePorts instead
 	// NodePort allocates a static port for the NodePort service.
 	// +optional
 	NodePort *int32 `json:"nodePort,omitempty"`
+	// NodePorts allocates a list of static ports for the NodePort service.
+	// +optional
+	// +kubebuilder:validation:MaxItems=64
+	NodePorts []int32 `json:"nodePorts,omitempty"`
 	// LoadBalancerIP override the LoadBalancer IP to use a specific IP address (e.g., static LB). It is used only if service type is LoadBalancer.
 	// LoadBalancer provider must support this feature.
 	// +optional
@@ -74,14 +90,23 @@ type GatewayServerSpec struct {
 }
 
 // EndpointStatus defines the observed state of the endpoint.
+// +kubebuilder:validation:XValidation:rule="!(has(self.port) && has(self.ports))",message="port and ports are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="has(self.port) || (has(self.ports) && size(self.ports) > 0)",message="at least one of port or ports must be set"
+// +kubebuilder:validation:XValidation:rule="!(has(self.ports)) || self.ports.all(p, self.ports.filter(q, q == p).size() == 1)",message="ports must not contain duplicates"
+//
+//nolint:lll // ignore long lines given by Kubebuilder marker annotations
 type EndpointStatus struct {
 	// Addresses specifies the addresses of the endpoint.
 	Addresses []string `json:"addresses,omitempty"`
+	// Deprecated: use Ports instead
 	// Port specifies the port of the endpoint.
 	Port int32 `json:"port,omitempty"`
 	// Protocol specifies the protocol of the endpoint.
 	// +kubebuilder:validation:Enum=TCP;UDP
 	Protocol *corev1.Protocol `json:"protocol,omitempty"`
+	// Ports specifies the ports of the endpoint.
+	// +kubebuilder:validation:MaxItems=64
+	Ports []int32 `json:"ports,omitempty"`
 }
 
 // InternalGatewayEndpoint defines the endpoint for the internal network.
--- a/cmd/gateway/wireguard/main.go
+++ b/cmd/gateway/wireguard/main.go
@@ -148,10 +148,18 @@ func run(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("unable to load keys: %w", err)
 	}
 
-	// Create the wg-liqo interface and init the wireguard configuration depending on the mode (client/server).
-	if err := wireguard.InitWireguardLink(cmd.Context(), options); err != nil {
-		return fmt.Errorf("unable to init wireguard link: %w", err)
+	// Get interface list
+	ports, err := wireguard.GetWireguardPorts(options)
+	if err != nil {
+		return fmt.Errorf("failed to parse wireguard ports: %w", err)
 	}
+	// Create the wg-liqo interface and init the wireguard configuration depending on the mode (client/server).
+	for i := range ports {
+		if err := wireguard.InitWireguardLink(cmd.Context(), options, i); err != nil {
+			return fmt.Errorf("failed to create WireGuard interface %d/%d: %w", i+1, len(ports), err)
+		}
+	}
+	klog.Infof("Successfully setup %d WireGuard interfaces", len(ports))
 
 	// Create the Prometheus collector and register it inside the controller-runtime metrics server.
 	promcollect, err := wireguard.NewPrometheusCollector(mgr.GetClient(), &wireguard.MetricsOptions{
--- a/deployments/liqo/templates/liqo-wireguard-gateway-client-template.yaml
+++ b/deployments/liqo/templates/liqo-wireguard-gateway-client-template.yaml
@@ -94,6 +94,7 @@ spec:
                 {{- end }}
                 - --enable-nft-monitor={{ .Values.networking.gatewayTemplates.nftablesMonitor }}
                 - --enable-route-monitor={{ .Values.networking.gatewayTemplates.routeMonitor }}
+                - --enable-multipath-hash-policy={{ printf "{{ if gt (len .Spec.Endpoint.Ports) 1 }}true{{ else }}false{{ end }}" }}
                 volumeMounts:
                 - name: ipc
                   mountPath: /ipc
@@ -141,7 +142,7 @@ spec:
                 - --container-name=wireguard
                 - --mtu={{"{{ .Spec.MTU }}"}}
                 - --endpoint-address={{"{{ index .Spec.Endpoint.Addresses 0 }}"}}
-                - --endpoint-port={{"{{ .Spec.Endpoint.Port }}"}}
+                - --endpoint-ports={{ printf "{{ if .Spec.Endpoint.Ports }}{{ range $i, $p := .Spec.Endpoint.Ports }}{{ if $i }},{{ end }}{{ $p }}{{ end }}{{ else }}{{ .Spec.Endpoint.Port }}{{ end }}" }}
                 {{- if .Values.metrics.enabled }}
                 - --metrics-address=:8084
                 {{- end }}
@@ -170,6 +171,7 @@ spec:
                     add:
                     - NET_ADMIN
                     - NET_RAW
+                    - "{{ "{{ if gt (len .Spec.Endpoint.Ports) 1 }}SYS_ADMIN{{ else }}NET_RAW{{ end }}" }}"
                   {{ if .Values.networking.gatewayTemplates.wireguard.implementation | eq "userspace" }}
                   privileged: true
                   {{ end }}
--- a/deployments/liqo/templates/liqo-wireguard-gateway-server-template.yaml
+++ b/deployments/liqo/templates/liqo-wireguard-gateway-server-template.yaml
@@ -114,6 +114,7 @@ spec:
                 {{- end }}
                 - --enable-nft-monitor={{ .Values.networking.gatewayTemplates.nftablesMonitor }}
                 - --enable-route-monitor={{ .Values.networking.gatewayTemplates.routeMonitor }}
+                - --enable-multipath-hash-policy={{ printf "{{ if gt (len .Spec.Endpoint.Ports) 1 }}true{{ else }}false{{ end }}" }}
                 volumeMounts:
                 - name: ipc
                   mountPath: /ipc
@@ -160,7 +161,7 @@ spec:
                 - --mode=server
                 - --container-name=wireguard
                 - --mtu={{"{{ .Spec.MTU }}"}}
-                - --listen-port={{"{{ .Spec.Endpoint.Port }}"}}
+                - --listen-ports={{ printf "{{ if .Spec.Endpoint.Ports }}{{ range $i, $p := .Spec.Endpoint.Ports }}{{ if $i }},{{ end }}{{ $p }}{{ end }}{{ else }}{{ .Spec.Endpoint.Port }}{{ end }}" }}
                 {{- if .Values.metrics.enabled }}
                 - --metrics-address=:8084
                 {{- end }}
@@ -190,6 +191,7 @@ spec:
                     add:
                     - NET_ADMIN
                     - NET_RAW
+                    - "{{ "{{ if gt (len .Spec.Endpoint.Ports) 1 }}SYS_ADMIN{{ else }}NET_RAW{{ end }}" }}"
                   {{ if .Values.networking.gatewayTemplates.wireguard.implementation | eq "userspace" }}
                   privileged: true
                   {{ end }}
--- a/pkg/gateway/tunnel/wireguard/device.go
+++ b/pkg/gateway/tunnel/wireguard/device.go
@@ -30,7 +30,7 @@ import (
 
 var errWgEndpointPeerNotFound = errors.New("wg endpoint peer not found")
 
-func configureDevice(wgcl *wgctrl.Client, options *Options, peerPubKey wgtypes.Key) error {
+func configureDevice(wgcl *wgctrl.Client, options *Options, peerPubKey wgtypes.Key, idx int) error {
 	confdev := wgtypes.Config{
 		PrivateKey: &options.PrivateKey,
 		ListenPort: nil,
@@ -42,12 +42,13 @@ func configureDevice(wgcl *wgctrl.Client, options *Options, peerPubKey wgtypes.K
 		},
 		ReplacePeers: true,
 	}
+	name := tunnel.GetTunnelName(idx)
 
 	switch options.GwOptions.Mode {
 	case gateway.ModeServer:
-		confdev.ListenPort = &options.ListenPort
+		confdev.ListenPort = &options.ListenPorts[idx]
 		if options.PreserveClientEndpoint {
-			endpoint, err := getExistingPeerEndpoint(wgcl, peerPubKey)
+			endpoint, err := getExistingPeerEndpoint(wgcl, peerPubKey, name)
 			switch {
 			case err == nil:
 				klog.Infof("Found existing endpoint %s for current peer. Re-using it.", endpoint.String())
@@ -61,20 +62,20 @@ func configureDevice(wgcl *wgctrl.Client, options *Options, peerPubKey wgtypes.K
 	case gateway.ModeClient:
 		confdev.Peers[0].Endpoint = &net.UDPAddr{
 			IP:   options.EndpointIP,
-			Port: options.EndpointPort,
+			Port: options.EndpointPorts[idx],
 		}
 	}
 
-	if err := wgcl.ConfigureDevice(tunnel.TunnelInterfaceName, confdev); err != nil {
-		return fmt.Errorf("an error occurred while configuring the device: %w", err)
+	if err := wgcl.ConfigureDevice(name, confdev); err != nil {
+		return fmt.Errorf("configuring the device %q: %w", name, err)
 	}
-	klog.Infof("Device %s configured", tunnel.TunnelInterfaceName)
+	klog.Infof("Device %s configured", name)
 
 	return nil
 }
 
-func getExistingPeerEndpoint(wgcl *wgctrl.Client, peerPubKey wgtypes.Key) (*net.UDPAddr, error) {
-	peer, err := getExistingPeer(wgcl, peerPubKey)
+func getExistingPeerEndpoint(wgcl *wgctrl.Client, peerPubKey wgtypes.Key, name string) (*net.UDPAddr, error) {
+	peer, err := getExistingPeer(wgcl, peerPubKey, name)
 	if err != nil {
 		return nil, fmt.Errorf("getting existing peer: %w", err)
 	}
@@ -84,8 +85,8 @@ func getExistingPeerEndpoint(wgcl *wgctrl.Client, peerPubKey wgtypes.Key) (*net.
 	return peer.Endpoint, nil
 }
 
-func getExistingPeer(wgcl *wgctrl.Client, peerPubKey wgtypes.Key) (*wgtypes.Peer, error) {
-	dev, err := getExistingDevice(wgcl)
+func getExistingPeer(wgcl *wgctrl.Client, peerPubKey wgtypes.Key, name string) (*wgtypes.Peer, error) {
+	dev, err := getExistingDevice(wgcl, name)
 	if err != nil {
 		return nil, fmt.Errorf("getting existing device: %w", err)
 	}
@@ -97,11 +98,11 @@ func getExistingPeer(wgcl *wgctrl.Client, peerPubKey wgtypes.Key) (*wgtypes.Peer
 	return nil, fmt.Errorf("no matching peer: %w", errWgEndpointPeerNotFound)
 }
 
-func getExistingDevice(wgcl *wgctrl.Client) (*wgtypes.Device, error) {
-	dev, err := wgcl.Device(tunnel.TunnelInterfaceName)
+func getExistingDevice(wgcl *wgctrl.Client, name string) (*wgtypes.Device, error) {
+	dev, err := wgcl.Device(name)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return nil, fmt.Errorf("device %s not found: %w", tunnel.TunnelInterfaceName, errWgEndpointPeerNotFound)
+			return nil, fmt.Errorf("device %s not found: %w", name, errWgEndpointPeerNotFound)
 		}
 		return nil, fmt.Errorf("fetching device: %w", err)
 	}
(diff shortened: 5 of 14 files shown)


pkg/gateway/tunnel/wireguard/options.go
```diff
@@ -75,6 +75,9 @@ type Options struct {
 	DNSCheckInterval time.Duration
 
 	Implementation WgImplementation
+
+	ListenPorts   []int
+	EndpointPorts []int
```
fra98: Can you move ListenPorts close to ListenPort? Same for EndpointPorts.
I would also add an in-line comment to clarify what you said in the PR description

pkg/gateway/tunnel/wireguard/flags.go
```diff
@@ -68,6 +72,8 @@ func InitFlags(flagset *pflag.FlagSet, opts *Options) {
 	flagset.DurationVar(&opts.DNSCheckInterval, FlagNameDNSCheckInterval.String(), 5*time.Minute, "Interval between two DNS checks")
 
 	flagset.Var(&opts.Implementation, "implementation", "Implementation of the wireguard interface (kernel or userspace)")
+	flagset.IntSliceVar(&opts.EndpointPorts, FlagNameEndpointPorts.String(), nil, "List of endpoint ports (client only)")
+	flagset.IntSliceVar(&opts.ListenPorts, FlagNameListenPorts.String(), nil, "List of listen ports (server only)")
```
fra98: Same here, I would move them closer and expand description that these takes precedence over single-port flag

pkg/gateway/tunnel/wireguard/options.go
MircoBarone: Done

pkg/gateway/tunnel/wireguard/flags.go
MircoBarone: Done

MircoBarone:
I have updated the logic to support the creation of multiple interfaces.
This commit is an incremental step, as the interfaces are created but not yet fully configured.

I have added the `GetWireguardPorts` function to extract the list of ports from the flags (`--endpoint-port(s)`, `--listen-port(s)`) , which is also used to determine the number of interfaces. I tried to correctly handle priority between the new plural flags and the legacy single-port flags.

I have also extended `GetInterfaceIP` to assign a unique IP to each interface within the `169.254.0.0/16` range. Each interface now gets its own `/30` subnetwork.
I considered assigning the same IP to all interfaces, but in my opinion, it is problematic because it would generate overlapping routes where the kernel consistently picks only the first available route.
For example:
```yaml
169.254.18.0/30 dev liqo-tunnel proto kernel scope link src 169.254.18.2
169.254.18.0/30 dev liqo-tunnel1 proto kernel scope link src 169.254.18.2
169.254.18.0/30 dev liqo-tunnel2 proto kernel scope link src 169.254.18.2
169.254.18.0/30 dev liqo-tunnel3 proto kernel scope link src 169.254.18.2
```
This would cause, in particular, issues with keepalives.

The wireguard container entrypoint has been extended to initialize N interfaces using a fail-fast approach: if the creation of even a single interface fails, the entire container is considered failed and will exit. 

Additionally, all underlying interface creation functions have been refactored to support this index-based multi-tunnel logic.

MircoBarone:
In the commit ad92c6b5f2cca069d9e4eaf850774b1de6c213b2 I have removed the single-port flags of the wireguard container (`--endpoint-port` and `--listen-port`) and replaced them with `--endpoint-ports` and `--listen-ports`.

This change is meant to be integrated with PR #3233 and PR #3236. The idea is to update the template later on, introducing something like this:

```yaml
--endpoint-ports={{ if .Spec.Endpoint.Ports }}{{ range $i, $p := .Spec.Endpoint.Ports }}{{ if $i }},{{ end }}{{ $p }}{{ end }}{{ else }}{{ .Spec.Endpoint.Port }}{{ end }}
```

I will submit the template modifications once the two PRs are merged.

Let me know if you agree with this approach.

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
 )
 
 // InitWireguardLink inits the Wireguard interface.
-func InitWireguardLink(ctx context.Context, options *Options) error {
-	exists, err := existsLink()
+func InitWireguardLink(ctx context.Context, options *Options, idx, port int) error {
```
fra98: NIT: passing port seems redundant as we already have idx and options to derive the right port

MircoBarone: Solved

apis/networking/v1beta1/gatewayserver_types.go
```diff
@@ -36,16 +36,25 @@ var GatewayServerGroupResource = schema.GroupResource{Group: GroupVersion.Group,
...
 	// Port specifies the port of the endpoint.
 	Port int32 `json:"port,omitempty"`
+	// Ports specifies the ports of the endpoint.
+	// This field is preferred over the legacy Port field, which is kept for backward compatibility.
+	Ports []int32 `json:"ports,omitempty"`
 	// ServiceType specifies the type of the service.
 	// +kubebuilder:default=ClusterIP
 	// +kubebuilder:validation:Enum=ClusterIP;NodePort;LoadBalancer;ExternalName
 	ServiceType corev1.ServiceType `json:"serviceType,omitempty"`
 	// NodePort allocates a static port for the NodePort service.
```
claudiolor, line 58:
```suggestion
     // Deprecated: use NodePorts instead
	// NodePort allocates a static port for the NodePort service.
```

apis/networking/v1beta1/gatewayserver_types.go
```diff
@@ -36,16 +36,25 @@ var GatewayServerGroupResource = schema.GroupResource{Group: GroupVersion.Group,
...
 	// +optional
 	NodePort *int32 `json:"nodePort,omitempty"`
+	// NodePorts allocates a list of static ports for the NodePort service.
+	// This field is preferred over the legacy NodePort field, which is kept for backward compatibility.
+	// +optional
```
claudiolor, line 62:
```suggestion
	// NodePorts allocates a list of static ports for the NodePort service.
	// +optional
```

apis/networking/v1beta1/gatewayserver_types.go
```diff
@@ -36,16 +36,25 @@ var GatewayServerGroupResource = schema.GroupResource{Group: GroupVersion.Group,
 var GatewayServerGroupVersionResource = GroupVersion.WithResource(GatewayServerResource)
 
 // Endpoint defines the endpoint of the gatewayserver.
+// +kubebuilder:validation:XValidation:rule="!(has(self.port)&&has(self.ports)&&self.port!=self.ports[0])",message="port must match ports[0]"
+// +kubebuilder:validation:XValidation:rule="!(has(self.nodePort)&&has(self.nodePorts)&&self.nodePort!=self.nodePorts[0])",message="nodePort must match nodePorts[0]"
 type Endpoint struct {
 	// Port specifies the port of the endpoint.
 	Port int32 `json:"port,omitempty"`
```
claudiolor, line 49:
```suggestion
// Deprecated: use Ports instead
// Port specifies the port of the endpoint.
Port int32 `json:"port,omitempty"`
```

apis/networking/v1beta1/gatewayserver_types.go
```diff
@@ -36,16 +36,25 @@ var GatewayServerGroupResource = schema.GroupResource{Group: GroupVersion.Group,
...
+	// Ports specifies the ports of the endpoint.
+	// This field is preferred over the legacy Port field, which is kept for backward compatibility.
```
claudiolor:
```suggestion
```

apis/networking/v1beta1/gatewayserver_types.go
```diff
@@ -36,16 +36,25 @@ var GatewayServerGroupResource = schema.GroupResource{Group: GroupVersion.Group,
 var GatewayServerGroupVersionResource = GroupVersion.WithResource(GatewayServerResource)
 
 // Endpoint defines the endpoint of the gatewayserver.
+// +kubebuilder:validation:XValidation:rule="!(has(self.port)&&has(self.ports)&&self.port!=self.ports[0])",message="port must match ports[0]"
+// +kubebuilder:validation:XValidation:rule="!(has(self.nodePort)&&has(self.nodePorts)&&self.nodePort!=self.nodePorts[0])",message="nodePort must match nodePorts[0]"
```
claudiolor: Not sure about this validation, we shouldn't allow to specify both singular and plural form of the same field together

apis/networking/v1beta1/gatewayserver_types.go
```diff
@@ -82,6 +92,9 @@ type EndpointStatus struct {
 	// Protocol specifies the protocol of the endpoint.
 	// +kubebuilder:validation:Enum=TCP;UDP
 	Protocol *corev1.Protocol `json:"protocol,omitempty"`
+	// Ports specifies the ports of the endpoint.
+	// This field is preferred over the legacy Port field, which is kept for backward compatibility.
```
claudiolor: Same, here. Remove this message and add a deprecation warning in the Port field

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -184,3 +185,25 @@ func existsLink() (bool, error) {
...
 }
+
+// GetWireguardPorts returns the list of ports to be used for WireGuard interfaces.
+func GetWireguardPorts(opts *Options) ([]int, error) {
+	var ports []int
+
+	switch opts.GwOptions.Mode {
+	case gateway.ModeClient:
+		ports = opts.EndpointPorts
+
```
claudiolor: I feel these spaces between the cases don't improve readability

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -184,3 +185,25 @@ func existsLink() (bool, error) {
...
+	case gateway.ModeServer:
+		ports = opts.ListenPorts
+
+	default:
+		return nil, fmt.Errorf("invalid mode %v", opts.GwOptions.Mode)
+	}
+	if len(ports) > tunnel.MaxWireguardInterfaces {
+		klog.Warningf("Requested %d WireGuard interfaces, capping to maximum of %d", len(ports), tunnel.MaxWireguardInterfaces)
+		ports = ports[:tunnel.MaxWireguardInterfaces]
+	}
```
claudiolor, line 207: I don't like this kind of gray failures, let's document that it is not possible to configure more than MaxWireguardInterfaces and return an error if the user configured more.

cmd/gateway/wireguard/main.go
```diff
@@ -148,10 +148,18 @@ func run(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("unable to load keys: %w", err)
 	}
 
+	// Get interface list
+	ports, err := wireguard.GetWireguardPorts(options)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve ports: %w", err)
```
claudiolor:
```suggestion
		return fmt.Errorf("failed to parse wireguard ports: %w", err)
```

cmd/gateway/wireguard/main.go
```diff
@@ -148,10 +148,18 @@ func run(cmd *cobra.Command, _ []string) error {
...
+	}
 	// Create the wg-liqo interface and init the wireguard configuration depending on the mode (client/server).
-	if err := wireguard.InitWireguardLink(cmd.Context(), options); err != nil {
-		return fmt.Errorf("unable to init wireguard link: %w", err)
+	for i := range ports {
+		if err := wireguard.InitWireguardLink(cmd.Context(), options, i); err != nil {
+			return fmt.Errorf("failed to create WireGuard interface %d/%d: %w", i+1, len(ports), err)
+		}
 	}
+	klog.Infof("Successfully created all %d interfaces", len(ports))
```
claudiolor:
```suggestion
	klog.Infof("Successfully setup %d WireGuard interfaces", len(ports))
```

pkg/gateway/tunnel/wireguard/device.go
```diff
@@ -41,18 +41,18 @@ func configureDevice(wgcl *wgctrl.Client, options *Options, peerPubKey wgtypes.K
...
 	}
+	name := tunnel.GetTunnelName(idx)
+	klog.Infof("Configuring device %s", name)
 
-	klog.Infof("Configuring device %s", tunnel.TunnelInterfaceName)
-
-	if err := wgcl.ConfigureDevice(tunnel.TunnelInterfaceName, confdev); err != nil {
-		return fmt.Errorf("an error occurred while configuring the device: %w", err)
+	if err := wgcl.ConfigureDevice(name, confdev); err != nil {
+		return fmt.Errorf("an error occurred while configuring the device %q: %w", name, err)
```
claudiolor:
```suggestion
		return fmt.Errorf("configuring the device %q: %w", name, err)
```

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
...
-		return fmt.Errorf("cannot create Wireguard interface: %w", err)
+	if err := createLink(ctx, options, idx); err != nil {
+		return fmt.Errorf("cannot create Wireguard interface %q: %w", name, err)
 	}
 
-	link, err := tunnel.GetLink(tunnel.TunnelInterfaceName)
+	link, err := tunnel.GetLink(name)
 	if err != nil {
-		return fmt.Errorf("cannot get Wireguard interface: %w", err)
+		return fmt.Errorf("cannot get Wireguard interface %q: %w", name, err)
```
claudiolor:
```suggestion
		return fmt.Errorf("getting Wireguard interface %q: %w", name, err)
```

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
...
 	if exists {
-		klog.Infof("Wireguard interface %q already exists", tunnel.TunnelInterfaceName)
+		klog.Infof("Wireguard interface %q already exists", name)
 		return nil
 	}
 
-	if err := createLink(ctx, options); err != nil {
-		return fmt.Errorf("cannot create Wireguard interface: %w", err)
+	if err := createLink(ctx, options, idx); err != nil {
+		return fmt.Errorf("cannot create Wireguard interface %q: %w", name, err)
```
claudiolor:
```suggestion
		return fmt.Errorf("creating Wireguard interface %q: %w", name, err)
```

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
...
 
 // InitWireguardLink inits the Wireguard interface.
-func InitWireguardLink(ctx context.Context, options *Options) error {
-	exists, err := existsLink()
+func InitWireguardLink(ctx context.Context, options *Options, idx int) error {
+	name := tunnel.GetTunnelName(idx)
+	exists, err := existsLink(idx)
 	if err != nil {
-		return fmt.Errorf("cannot check if Wireguard interface exists: %w", err)
+		return fmt.Errorf("cannot check if Wireguard interface %q exists: %w", name, err)
```
claudiolor:
```suggestion
		return fmt.Errorf("checking if Wireguard interface %q exists: %w", name, err)
```

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
...
 		}
 	}
 
 	return nil
 }
 
 // createLinkKernel creates a new Wireguard interface using the kernel module.
-func createLinkKernel(options *Options) error {
+func createLinkKernel(options *Options, idx int) error {
 	link := netlink.Wireguard{
```
claudiolor: Since we are getting the name multiple times, let's define here a variable with the interface name, and we use it below:
```
intName := GetTunnelName(idx)
```

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
...
 		LinkAttrs: netlink.LinkAttrs{
 			MTU:  options.MTU,
-			Name: tunnel.TunnelInterfaceName,
+			Name: tunnel.GetTunnelName(idx),
```
claudiolor:
```suggestion
			Name: intName,
```

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
...
 		},
 	}
 
 	err := netlink.LinkAdd(&link)
 	if err != nil {
-		return fmt.Errorf("cannot add Wireguard interface: %w", err)
+		return fmt.Errorf("cannot add Wireguard interface %q: %w", tunnel.GetTunnelName(idx), err)
```
claudiolor:
```suggestion
		return fmt.Errorf("getting Wireguard interface %q: %w", intName), err)
```

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
...
 	mtu := options.MTU
 	if mtu <= 0 {
 		mtu = device.DefaultMTU
 	}
 
-	tunDev, err := tun.CreateTUN(tunnel.TunnelInterfaceName, mtu)
+	tunDev, err := tun.CreateTUN(tunnel.GetTunnelName(idx), mtu)
 	if err != nil {
-		return fmt.Errorf("failed to create wireguard TUN device %q: %w", tunnel.TunnelInterfaceName, err)
+		return fmt.Errorf("failed to create wireguard TUN device %q: %w", tunnel.GetTunnelName(idx), err)
```
claudiolor:
```suggestion
		return fmt.Errorf("creating wireguard TUN device %q: %w", tunnel.GetTunnelName(idx), err)
```

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -33,95 +33,96 @@ import (
...
 // embedded as a library. The device is kept running in-process for the lifetime of the gateway.
-func createLinkUserspace(ctx context.Context, options *Options) error {
+func createLinkUserspace(ctx context.Context, options *Options, idx int) error {
 	mtu := options.MTU
 	if mtu <= 0 {
 		mtu = device.DefaultMTU
 	}
 
-	tunDev, err := tun.CreateTUN(tunnel.TunnelInterfaceName, mtu)
+	tunDev, err := tun.CreateTUN(tunnel.GetTunnelName(idx), mtu)
```
claudiolor: Same, here: let's create a variable with the interface name

pkg/gateway/tunnel/wireguard/publickeys_controller.go
```diff
@@ -78,9 +78,15 @@ func (r *PublicKeysReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	if err := configureDevice(r.Wgcl, r.Options, wgtypes.Key(publicKey.Spec.PublicKey)); err != nil {
-		return ctrl.Result{}, err
+	// Get interface list
+	ports, err := GetWireguardPorts(r.Options)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to retrieve ports: %w", err)
+	}
+	for i := range ports {
+		if err := configureDevice(r.Wgcl, r.Options, wgtypes.Key(publicKey.Spec.PublicKey), i); err != nil {
+			return ctrl.Result{}, err
```
claudiolor: Let's wrap this error
```suggestion
			return ctrl.Result{}, fmt.Errorf("configuring interface %d": %w, i, err)
```

pkg/gateway/tunnel/wireguard/publickeys_controller.go
```diff
@@ -78,9 +78,15 @@ func (r *PublicKeysReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		klog.Warning("EndpointIP is not set yet. Maybe the DNS resolution is still in progress")
 		return ctrl.Result{}, nil
 	}
-
-	if err := configureDevice(r.Wgcl, r.Options, wgtypes.Key(publicKey.Spec.PublicKey)); err != nil {
-		return ctrl.Result{}, err
+	// Get interface list
+	ports, err := GetWireguardPorts(r.Options)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to retrieve ports: %w", err)
```
claudiolor:
```suggestion
		return ctrl.Result{}, fmt.Errorf("parsing wireguard ports: %w", err)
```

apis/networking/v1beta1/gatewayserver_types.go
fra98: @claudiolor I think this was inspired by similar cases in upstream Kubernetes like Node resource:

```yaml
spec:
  podCIDR: 10.10.1.0/24
  podCIDRs:
  - 10.10.1.0/24
```

We can relax this and allow to set both, the only important thing is that, if both are defined, `ports[0]` must match `port`

claudiolor: I believe this makes sense in Kubernetes nodes for backward compatibility, as some third party might expect that field defined. But here I can't see any point in having both fields defined together

deployments/liqo/templates/liqo-wireguard-gateway-client-template.yaml
```diff
@@ -154,6 +155,7 @@ spec:
                     add:
                     - NET_ADMIN
                     - NET_RAW
+                    - SYS_ADMIN
```
fra98: open question: is this the least privileged capability we can give for NAPI?

pkg/gateway/tunnel/wireguard/device.go
```diff
@@ -61,20 +61,22 @@ func configureDevice(wgcl *wgctrl.Client, options *Options, peerPubKey wgtypes.K
 	case gateway.ModeClient:
 		confdev.Peers[0].Endpoint = &net.UDPAddr{
 			IP:   options.EndpointIP,
-			Port: options.EndpointPort,
+			Port: options.EndpointPorts[idx],
 		}
 	}
+	name := tunnel.GetTunnelName(idx)
+	klog.Infof("Configuring device %s", name)
```
fra98:
```suggestion
```
NIT: can be removed as we already log error or info (happy path)

pkg/gateway/tunnel/wireguard/device.go
```diff
@@ -45,9 +45,9 @@ func configureDevice(wgcl *wgctrl.Client, options *Options, peerPubKey wgtypes.K
 
 	switch options.GwOptions.Mode {
 	case gateway.ModeServer:
-		confdev.ListenPort = &options.ListenPort
+		confdev.ListenPort = &options.ListenPorts[idx]
 		if options.PreserveClientEndpoint {
-			endpoint, err := getExistingPeerEndpoint(wgcl, peerPubKey)
+			endpoint, err := getExistingPeerEndpoint(wgcl, peerPubKey, idx)
```
fra98: NIT: for better re-usability, instead of passing index to the whole chain, you can pass interface name directly

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -30,98 +31,102 @@ import (
 
 	"github.com/liqotech/liqo/pkg/gateway"
 	"github.com/liqotech/liqo/pkg/gateway/tunnel"
+	"github.com/liqotech/liqo/pkg/utils/kernel"
 )
 
 // InitWireguardLink inits the Wireguard interface.
-func InitWireguardLink(ctx context.Context, options *Options) error {
-	exists, err := existsLink()
+func InitWireguardLink(ctx context.Context, options *Options, idx int) error {
```
fra98, line 38: NIT: probably also here, can we pass interface name directly?

pkg/gateway/tunnel/wireguard/netlink.go
```diff
@@ -184,3 +189,64 @@ func existsLink() (bool, error) {
...
+	if len(ports) > tunnel.MaxWireguardInterfaces {
+		return nil, fmt.Errorf("requested %d WireGuard interfaces, maximum allowed is %d", len(ports), tunnel.MaxWireguardInterfaces)
+	}
+
+	return ports, nil
+}
+
+// EnsureThreadedNAPI enables threaded NAPI for all WireGuard interfaces.
+// Retry up to maxNAPIAttempts times per interface to handle transient failures.
+func EnsureThreadedNAPI(interfaces int) {
```
fra98: Why do we not handle errors?

pkg/utils/kernel/threaded.go
```diff
@@ -0,0 +1,98 @@
...
+		return false, err
+	}
+
+	if len(data) > 0 && data[0] == '1' {
+		return false, nil
+	}
+
+	if err := os.WriteFile(path, []byte("1\n"), 0o600); err != nil {
+		return false, err
+	}
```
fra98, line 51: Let's wrap errors

pkg/utils/kernel/threaded.go
```diff
@@ -0,0 +1,98 @@
...
+// IsThreadedNAPISupported checks if the kernel supports threaded NAPI by probing the sysfs entry for the given interface.
+func IsThreadedNAPISupported(iface string) bool {
+	path := fmt.Sprintf("/sys/class/net/%s/threaded", iface)
+	_, err := os.Stat(path)
+	if errors.Is(err, os.ErrNotExist) {
+		return false
+	}
+	if err != nil {
+		klog.Warningf("Unexpected error checking threaded NAPI support for %s: %v", iface, err)
+		return false
```
fra98: Also here it's strange if an error when reading a file silently disable the feature

deployments/liqo/templates/liqo-wireguard-gateway-client-template.yaml
MircoBarone: As far as I know,  `CAP_SYS_ADMIN ` is not required by threaded NAPI itself. It is currently required only because we need to remount sysfs as read-write before writing to `/sys/class/net/<iface>/threaded`. 

Do you think there's an alternative way to avoid the remounting process and avoid introducing this capability?

pkg/gateway/tunnel/wireguard/netlink.go
MircoBarone: Tried to simplify the chain passing only the name where possible (specifically in `createLinkKernel`, `createLinkUserspace` and `existsLink`).

`InitWireguardLink` still needs the `idx` to calculate the interface IP, and `createLink` needs it to resolve the port from `options.ListenPorts[idx]`.

pkg/gateway/tunnel/wireguard/netlink.go
MircoBarone: Tried to better wrap the errors in all the threaded NAPI chain.

pkg/utils/kernel/threaded.go
MircoBarone: Same as above

pkg/utils/kernel/threaded.go
MircoBarone: Same as above

deployments/liqo/templates/liqo-wireguard-gateway-client-template.yaml
MircoBarone: Following up on this, I've just updated the PR to make the `SYS_ADMIN` capability optional. It is now dynamically injected only when multiple ports are configured in the GatewayClient/GatewayServer spec:

```yaml
- "{{ "{{ if gt (len .Spec.Endpoint.Ports) 1 }}SYS_ADMIN{{ else }}NET_RAW{{ end }}" }}"
```

The only minor side effect is that `NET_RAW` ends up duplicated in the manifest when a single interface is deployed. This is purely a cosmetic layout detail, as duplicate capabilities are safely ignored by the Linux kernel at runtime.

Let me know if this compromise works for you

apis/networking/v1beta1/gatewayserver_types.go
MircoBarone: I have updated the CEL validation following your feedback. The new rules now enforce the following constraints:

1. At least one of `port` or `ports` must be set, while `nodePort`/`nodePorts` remain fully optional.
2. The resource is rejected if both the singular and plural forms are specified at the same time.
3. Both `ports` and `nodePorts` arrays must not contain duplicate values.

deployments/liqo/templates/liqo-wireguard-gateway-server-template.yaml
```diff
@@ -190,6 +191,7 @@ spec:
                     add:
                     - NET_ADMIN
                     - NET_RAW
+                    - "{{ "{{ if gt (len .Spec.Endpoint.Ports) 1 }}SYS_ADMIN{{ else }}NET_RAW{{ end }}" }}"
```
fra98, line 194: NIT: I'd remove the above NET_RAW, to not duplicate NET_RAW

deployments/liqo/templates/liqo-wireguard-gateway-client-template.yaml
```diff
@@ -170,6 +171,7 @@ spec:
                     add:
                     - NET_ADMIN
                     - NET_RAW
+                    - "{{ "{{ if gt (len .Spec.Endpoint.Ports) 1 }}SYS_ADMIN{{ else }}NET_RAW{{ end }}" }}"
```
fra98, line 174: Same here

fra98 (approved):

claudiolor (approved):

PR #3229, opened 2026-03-22
feat(gateway): add fib_multipath_hash_policy support for L4-aware ECMP flow distribution

MircoBarone:
# Description

Part of the multi-tunnel WireGuard implementation. This is the second PR related to issue https://github.com/liqotech/liqo/issues/3225.

This PR adds the `enable-multipath-hash-policy` flag to the gateway. When set to `true`, the gateway writes `1` to `/proc/sys/net/ipv4/fib_multipath_hash_policy`, enabling 5-tuple (src/dst IP, src/dst port, protocol) hashing for multipath routing instead of L3-only (src/dst IP) hashing.

This is necessary to properly exploit ECMP across multiple WireGuard tunnels: without this setting, traffic between the same src/dst IP pair will always hash to the same tunnel regardless of the port, limiting the effectiveness of ECMP for those flows.

The flag defaults to `false`. The intended usage is to set it to `true` via a proper template when the number of interfaces specified in the `GatewayClient` or `GatewayServer` CRD is greater than one.

Since `/proc/sys/net/ipv4/fib_multipath_hash_policy` is namespaced, this setting only affects the network namespace of the gateway pod and has no impact on the host node.

If the write fails, the gateway returns an error and stops the execution to ensure the policy is strictly applied

A new file `pkg/utils/kernel/multipathpolicy.go` was added to implement the low-level write.

### DIFF
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -118,6 +118,12 @@ func run(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
+	// Enable Multipath Hash Policy if required.
+	if connoptions.GwOptions.EnableMultipathHashPolicy {
+		if err = kernel.EnableMultipathHashPolicy(); err != nil {
+			return fmt.Errorf("failed to enable multipath hash policy: %w", err)
+		}
+	}
 	// Set controller-runtime logger.
 	log.SetLogger(klog.NewKlogr())
 
--- a/pkg/gateway/flags.go
+++ b/pkg/gateway/flags.go
@@ -77,6 +77,8 @@ const (
 	FlagNameDisableKernelVersionCheck FlagName = "disable-kernel-version-check"
 	// FlagNameMinimumKernelVersion is the minimum kernel version required by Liqo.
 	FlagNameMinimumKernelVersion FlagName = "minimum-kernel-version"
+	// FlagNameEnableMultipathHashPolicy sets fib_multipath_hash_policy=1 to enable 5-tuple hashing for multipath routing.
+	FlagNameEnableMultipathHashPolicy FlagName = "enable-multipath-hash-policy"
 )
 
 // RequiredFlags contains the list of the mandatory flags.
@@ -125,6 +127,8 @@ func InitFlags(flagset *pflag.FlagSet, opts *Options) {
 
 	flagset.BoolVar(&opts.DisableKernelVersionCheck, FlagNameDisableKernelVersionCheck.String(), false, "Disable the kernel version check")
 	flagset.Var(&opts.MinimumKernelVersion, FlagNameMinimumKernelVersion.String(), "Minimum kernel version required by Liqo")
+	flagset.BoolVar(&opts.EnableMultipathHashPolicy, FlagNameEnableMultipathHashPolicy.String(), false,
+		"Set fib_multipath_hash_policy=1 to use 5-tuple hashing for multipath routing")
 }
 
 // MarkFlagsRequired marks the flags as required.
--- a/pkg/utils/kernel/multipathpolicy.go
+++ b/pkg/utils/kernel/multipathpolicy.go
@@ -1,0 +1,36 @@
+// Copyright 2019-2026 The Liqo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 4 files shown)


cheina97 (approved):

cheina97:
Can you temporarily enable the flag? Then I will run the e2e tests

MircoBarone:
> Can you temporarily enable the flag? Then I will run the e2e tests

@cheina97 Do you mean I should add a commit temporarily setting the default value of enable-multipath-hash-policy to true
?

cheina97:
> > Can you temporarily enable the flag? Then I will run the e2e tests
> 
> @cheina97 Do you mean I should add a commit temporarily setting the default value of enable-multipath-hash-policy to true
> ?

Yep

MircoBarone:
> > > Can you temporarily enable the flag? Then I will run the e2e tests
> > 
> > 
> > @cheina97 Do you mean I should add a commit temporarily setting the default value of enable-multipath-hash-policy to true
> > ?
> 
> Yep

@cheina97  done

pkg/gateway/flags.go
```diff
@@ -125,6 +127,8 @@ func InitFlags(flagset *pflag.FlagSet, opts *Options) {
 
 	flagset.BoolVar(&opts.DisableKernelVersionCheck, FlagNameDisableKernelVersionCheck.String(), false, "Disable the kernel version check")
 	flagset.Var(&opts.MinimumKernelVersion, FlagNameMinimumKernelVersion.String(), "Minimum kernel version required by Liqo")
+	flagset.BoolVar(&opts.EnableMultipathHashPolicy, FlagNameEnableMultipathHashPolicy.String(), true,
+		"Set fib_multipath_hash_policy=1 to use 5-tuple hashing for multipath routing")
```
fra98, line 131: Leaving comment just as a reminder to switch the default flag to `false` before merging

cmd/gateway/main.go
```diff
@@ -118,6 +118,13 @@ func run(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
+	// Enable Multipath Hash Policy if required.
+	if connoptions.GwOptions.EnableMultipathHashPolicy {
+		if err = kernel.EnableMultipathHashPolicy(); err != nil {
+			klog.Warningf("Failed to enable multipath hash policy, ECMP will fall back to L3-only hashing (src/dst IP): "+
+				"flow distribution may be suboptimal: %v", err)
+		}
+	}
```
fra98, line 126: Why not returning error and exit? If I enabled the multipath flag and this is not working, I would expect the pod to fail instead of silently falling back

pkg/utils/kernel/multipathpolicy.go
```diff
@@ -0,0 +1,33 @@
...
+// EnableMultipathHashPolicy enables 5-tuple hashing for multipath routing by writing 1 to /proc/sys/net/ipv4/fib_multipath_hash_policy.
+func EnableMultipathHashPolicy() error {
+	file, err := os.OpenFile(multipathpolicyFile, os.O_WRONLY, 0o600)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	if _, err := file.WriteString("1\n"); err != nil {
+		return err
```
fra98: nit: can you wrap errors?

cmd/gateway/main.go
MircoBarone: Without this policy, ECMP only distinguishes flows based on L3 data (IPs). This feature is an optimization (useful primarily for sessions between the same IP pair using different ports), but it's not strictly mandatory for ECMP to work.

My idea for the fallback was to ensure the gateway remains operational even if the kernel write fails for some reason. However, I see your point about avoiding suboptimal performance. If you prefer to ensure the policy is strictly applied, I can update the code to return the error and stop the gateway creation

pkg/utils/kernel/multipathpolicy.go
MircoBarone: Done

cmd/gateway/main.go
fra98: I thought this was a requirement, but since this is only an optimization I'd leave you to decide.
My 2 cents, since this is just a write on a file, I don't expect it to fail unless there is something seriously wrong. In such cases it's better to fail probably so user is notified. Also in this way it is guaranteed that if the flag is enabled, this optimization is actually enabled

fra98: For similar things (e.g.,`EnableIPForwarding` and `EnableRpFilter`) we return error too

MircoBarone: I've updated the code to return the error instead of a warning, ensuring consistency with `EnableIPForwarding` and `EnableRpFilter`.

You are right that failures should be treated as fatal. Since Liqo requires kernel 5.10+ and this flag was introduced in 4.12, the file is guaranteed to exist. Any error would simply indicate that something went wrong during the setup

fra98:
@MircoBarone test passed!

Can you switch flag to `false` and squash all commits? After that I will merge it

fra98 (approved):

PR #3255, opened 2026-04-14
fix(vk): synchronize local pods when vk miss remote pods deletion

claudiolor:
# Description

When the remote real pod is deleted while the local pod appears Running, the virtual kubelet's HandleStatus returns nil without updating the local pod status. The local pod stays Running indefinitely because the VK has no real pod to reflect from. This prevents Deployment controllers from creating replacements and makes pod logs inaccessible.

## Fix

If the shadow pod is in a terminal phase (Succeeded/Failed): propagates that phase to the local pod with the new RemotePodCompletedReason reason, allowing the workload controller to react correctly.

### DIFF
--- a/pkg/virtualKubelet/forge/pods.go
+++ b/pkg/virtualKubelet/forge/pods.go
@@ -38,6 +38,9 @@ const (
 	PodOffloadingBackOffReason = "OffloadingBackOff"
 	// PodOffloadingAbortedReason -> the reason assigned to pods rejected by the virtual kubelet after offloading has started.
 	PodOffloadingAbortedReason = "OffloadingAborted"
+	// RemotePodTerminatedReason -> the reason assigned to pods in the local cluster whose remote counterpart has completed,
+	// but it wasn't possible to reflect the real status (e.g. missed completion event), and the shadow pod reports a terminal phase.
+	RemotePodTerminatedReason = "RemotePodTerminated"
 
 	// ServiceAccountVolumeName is the prefix name that will be added to volumes that mount ServiceAccount secrets.
 	// This constant is taken from kubernetes/kubernetes (plugin/pkg/admission/serviceaccount/admission.go).
@@ -142,7 +145,15 @@ func LocalRejectedPodStatus(local *corev1.PodStatus, phase corev1.PodPhase, reas
 	}
 
 	for i := range local.ContainerStatuses {
-		local.ContainerStatuses[i].Ready = false
+		TerminateContainerState(&local.ContainerStatuses[i], phase, reason)
+	}
+
+	for i := range local.InitContainerStatuses {
+		TerminateContainerState(&local.InitContainerStatuses[i], phase, reason)
+	}
+
+	for i := range local.EphemeralContainerStatuses {
+		TerminateContainerState(&local.EphemeralContainerStatuses[i], phase, reason)
 	}
 
 	return *local
@@ -240,6 +251,25 @@ func RemotePodSpec(creation bool, local, remote *corev1.PodSpec, mutators ...Rem
 	}
 
 	return *remote
+}
+
+// TerminateContainerState modifies the container status to set it in a terminated state, if not already in a terminal state.
+func TerminateContainerState(cs *corev1.ContainerStatus, phase corev1.PodPhase, reason string) {
+	cs.Ready = false
+	if cs.State.Terminated == nil && (phase == corev1.PodFailed || phase == corev1.PodSucceeded) {
+		exitCode := int32(0)
+		if phase == corev1.PodFailed {
+			exitCode = 1
+		}
+		cs.State = corev1.ContainerState{
+			Terminated: &corev1.ContainerStateTerminated{
+				ExitCode: exitCode,
+				Reason:   reason,
+			},
+			Waiting: nil,
+			Running: nil,
+		}
+	}
 }
 
 // APIServerSupportMutator is a mutator which implements the support to enable offloaded pods to interact back with the local Kubernetes API server.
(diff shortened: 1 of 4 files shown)


pkg/virtualKubelet/forge/pods_test.go
```diff
@@ -135,6 +135,22 @@ var _ = Describe("Pod forging", func() {
...
+		Context("when the container is already terminated", func() {
+			BeforeEach(func() {
+				local.Status.ContainerStatuses[0].State = corev1.ContainerState{
+					Terminated: &corev1.ContainerStateTerminated{ExitCode: 137, Reason: "OOMKilled"},
+				}
+			})
+			It("should preserve the existing terminated state", func() {
+				Expect(output.Status.ContainerStatuses[0].State.Terminated).NotTo(BeNil())
+				Expect(output.Status.ContainerStatuses[0].State.Terminated.ExitCode).To(Equal(int32(137)))
+				Expect(output.Status.ContainerStatuses[0].State.Terminated.Reason).To(Equal("OOMKilled"))
```
fra98, line 175: Let's add tests to assert nil for State.Waiting and State.Running

pkg/virtualKubelet/forge/pods.go
```diff
@@ -242,6 +253,25 @@ func RemotePodSpec(creation bool, local, remote *corev1.PodSpec, mutators ...Rem
 	return *remote
 }
 
+// TerminateContainerState modifies the container status to set it in a terminated state, if not already in a terminal state.
+func TerminateContainerState(cs *corev1.ContainerStatus, phase corev1.PodPhase, reason string) {
+	cs.Ready = false
+	if cs.State.Terminated == nil && (phase == corev1.PodFailed || phase == corev1.PodSucceeded) {
+		cs.State.Waiting = nil
+		cs.State.Running = nil
```
fra98: This is probably no-op as we override the whole ContainerState below. 
Optionally, you can assign nil to Waiting and Running later just to signal intent in code

pkg/virtualKubelet/reflection/workload/podns.go
```diff
@@ -376,9 +376,31 @@ func (npr *NamespacedPodReflector) ShouldUpdateShadowPod(ctx context.Context, sh
...
+			po := forge.LocalRejectedPod(local, phase, forge.RemotePodTerminatedReason)
+			if reflect.DeepEqual(local.Status, po.Status) {
+				klog.V(4).Infof("Skipping local pod %q status update, as already synced", npr.LocalRef(local.GetName()))
+				return nil
+			}
+
+			klog.Infof("Updating local pod %q status to %q since the remote pod is no longer available", npr.LocalRef(local.GetName()), phase)
+			if _, err := npr.localPodsClient.UpdateStatus(ctx, po, metav1.UpdateOptions{FieldManager: forge.ReflectionFieldManager}); err != nil {
+				klog.Errorf("Failed to update local pod %q status: %v", npr.LocalRef(local.GetName()), err)
+				return err
```
fra98: let's wrap error

pkg/virtualKubelet/reflection/workload/podns.go
```diff
@@ -376,9 +376,31 @@ func (npr *NamespacedPodReflector) ShouldUpdateShadowPod(ctx context.Context, sh
...
-func (npr *NamespacedPodReflector) HandleStatus(ctx context.Context, local, remote *corev1.Pod, info *PodInfo) error {
-	// Do not handle the status in case the remote pod has not yet been created, or already terminated.
+func (npr *NamespacedPodReflector) HandleStatus(ctx context.Context, local, remote *corev1.Pod,
+	shadow *offloadingv1beta1.ShadowPod, info *PodInfo) error {
+	// The remote pod does not exist yet, or has been deleted.
 	if remote == nil {
+		switch {
+		case shadow == nil:
+			// Shadow pod still does not exist, wait for resources propagation.
+			klog.V(4).Infof("Remote pod %q and shadowpod do not exist, skipping status update", npr.RemoteRef(local.GetName()))
```
fra98: Is this only a redundant check for sanity reasons?

Because the code above if shadow does not exists is always terminal if I am not wrong: 
```go
	// If the remote shadowpod does not exist, then create it.
	if !shadowExists {
		defer tracer.Step("Ensured the presence of the remote object")
		if _, err := npr.remoteShadowPodsClient.Create(ctx, target, metav1.CreateOptions{FieldManager: forge.ReflectionFieldManager}); err != nil {
			if kerrors.IsAlreadyExists(err) {
				klog.Infof("Remote shadowpod %q already exists (local pod: %q)", npr.RemoteRef(name), npr.LocalRef(name))
				return nil
			}
			klog.Errorf("Failed to create remote shadowpod %q (local pod: %q): %v", npr.RemoteRef(name), npr.LocalRef(name), err)
			if !kerrors.IsConflict(err) {
				npr.Event(local, corev1.EventTypeWarning, forge.EventFailedReflection, forge.EventFailedReflectionMsg(err))
			}
			return err
		}

		// Do not attempt to create this shadowpod again until the informer has seen it.
		info.PreventCreationUntilSeen = true
		klog.Infof("Remote shadowpod %q successfully created (local: %q)", npr.RemoteRef(name), npr.LocalRef(name))
		npr.Event(local, corev1.EventTypeNormal, forge.EventSuccessfulReflection, forge.EventSuccessfulReflectionMsg())
		tracer.Step("Created the remote shadowpod")
		// Whatever the outcome, here we return without updating the status, as it will be triggered by future events.
		return nil
	}
```

fra98 (approved):

PR #3265, opened 2026-04-21
fix: move NsMap lifecycle ownership to controlplane secret

fra98:
# Description

This PR moves the ownership of NamespaceMaps lifecycle from the VirtualNode controller to a new controller that reconciles `ControlPlane` secrets and creates a child NamespaceMap for each.

## Why?

Since the multiple virtual nodes feature was introduced, virtualnodes can come and go rapidly. In the previous design, when no virtualnodes were left on the consumer, the NamespaceMap was deleted, causing a lot of unwanted side effects on offloaded namespaces on the provider cluster, like namespace termination and pod draining. This might be unwanted if a new virtual node spans a few seconds later. Offloaded namespaces should not be tied to virtualnode existence.

Since NamespaceMaps work only with a functional CRD replicator, which in turn can work only if a secret of type `ControlPlane` is set up, the most natural thing is to tie the NsMap lifecycle to the control plane secret. When performing a full unpeering, this secret will be deleted, and so the child NsMap.

The new controller also lets you fully manage the namespacemap lifecycle by just annotating the secret with `liqo.io/skip-namespacemap-creation: "true"` annotation


frisso:
@fra98 This looks so nice actually.
Is there any way to add also a proper page to document all those flags/labels we're currently introducing in Liqo?

fra98:
> @fra98 This looks so nice actually.
Is there any way to add also a proper page to document all those flags/labels we're currently introducing in Liqo?

Good question, we have a lot of labels/annotations that are not documented in the public docs. Some are internals, others are user-facing. We do have a lot of consts scattered among the codebase, and the distinction is not clear either. We should find a way to aggregate user-facing ones and keep them in sync with public docs 🤔

pkg/liqo-controller-manager/cross-module/controlplanesecret-controller/controlplanesecret_controller.go
```diff
@@ -0,0 +1,183 @@
...
+			return ctrl.Result{}, fmt.Errorf("getting secret %q: %w", req.NamespacedName, err)
+		}
+		return ctrl.Result{}, nil
+	}
+
+	remoteClusterID := liqocorev1beta1.ClusterID(secret.Labels[consts.RemoteClusterID])
+	if remoteClusterID == "" {
+		// This should not happen, as the controller is watching only secrets with the RemoteClusterID label,
+		// but we check it anyway for extra safety in case of empty value. We can return without doing anything.
+		logger.Info("Controlplane secret does not have the RemoteClusterID label, skipping")
```
claudiolor:
```suggestion
		logger.Warning("Controlplane secret does not have the RemoteClusterID label, skipping")
```

pkg/liqo-controller-manager/cross-module/controlplanesecret-controller/controlplanesecret_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	}}
+	op, err := resource.CreateOrUpdate(ctx, r.Client, &nm, func() error {
+		if nm.Labels == nil {
+			nm.Labels = map[string]string{}
+		}
+		nm.Labels[consts.RemoteClusterID] = string(remoteClusterID)
+		nm.Labels[consts.ReplicationRequestedLabel] = consts.ReplicationRequestedLabelValue
+		nm.Labels[consts.ReplicationDestinationLabel] = string(remoteClusterID)
+
+		// Set owner reference to the controlplane secret, to automatically delete the NamespaceMap when the secret is deleted.
```
claudiolor: We explicitly delete the namespacemap
```suggestion
```

claudiolor (commented):
Looks good, just some minor comments

claudiolor (approved):

PR #3271, opened 2026-04-28
fix(network): avoid geneve interface finalizer on internalfabric

fra98:
# Description
## Summary

This PR:
- fabric pod: removes the InternalFabric finalizer and replaces it with a periodic geneve interface cleanup, mirroring the approach already used on the gateway side. It replaces the `InternalFabricReconciler` with a `GeneveTunnelReconciler` that manages geneve interfaces without finalizers, reconciling on GeneveTunnel resources directly instead of Fabrics.
- gateway pod: replace the `InternalNodeReconciler` with a `GeneveTunnelReconciler`, reconciling on GeneveTunnel resources directly instead of Fabrics. This is to mirror what we did for the twin controller on the fabric pod. It also helps to avoid merging PR https://github.com/liqotech/liqo/pull/3083: the problem with the old approach is that we had to get the InternalFabric starting from the InternalNode (reconciled resource). This is difficult as we need to know how the Fabric's name was forged in the past (which already changed in the past or will potentially change again in the future, causing a breaking change). Instead, by reconciling on GeneveTunnel we can access real fabric name without inferring by just looking at GeneveTunnel spec (`.spec.internalFabricRef.name`), making it more resilient to future changes.

## Motivation

The InternalFabric finalizer was shared across all liqo-fabric DaemonSet pods. Only the first pod to reconcile the deletion would actually delete the geneve interface and remove the finalizer — the remaining pods might find the finalizer already gone and skip the interface cleanup entirely, leaving stale geneve interfaces on those nodes.

This makes the finalizer unreliable for interface cleanup. Instead, a periodic cleanup runnable (already used on the gateway side) garbage-collects orphan geneve interfaces by comparing the interfaces on the node against existing InternalFabric resources.


cmd/gateway/geneve/main.go
```diff
@@ -31,10 +31,8 @@ import (
 
 	networkingv1beta1 "github.com/liqotech/liqo/apis/networking/v1beta1"
 	"github.com/liqotech/liqo/pkg/gateway"
-	"github.com/liqotech/liqo/pkg/gateway/cleanup"
 	"github.com/liqotech/liqo/pkg/gateway/concurrent"
 	"github.com/liqotech/liqo/pkg/gateway/fabric"
```
claudiolor: Let's give an alias to this package, as it seems we are using the same controller for both the fabric and the gateway. It would be nice changing this package name, maybe in a separate PR

pkg/fabric/genevetunnel_controller.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func (r *GeneveTunnelReconciler) handleDeletion(ctx context.Context, gt *networkingv1beta1.GeneveTunnel) (ctrl.Result, error) {
+	var internalfabric networkingv1beta1.InternalFabric
+	err := r.Get(ctx, types.NamespacedName{
+		Name:      gt.Spec.InternalFabricRef.Name,
+		Namespace: gt.Spec.InternalFabricRef.Namespace,
+	}, &internalfabric)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
claudiolor: Let's add a comment to say this kind of deletion is best effort, and that leftover interfaces, will be garbage collected by the cleanup routine.

claudiolor (approved):

PR #3272, opened 2026-04-28
fix(networking): use returned route when deleting routes

fra98:
# Description

This PR fixes `EnsureRoutesAbsence` failing with `ESRCH: no such process` when deleting routes during RouteConfiguration finalization

## Cause

`EnsureRoutesAbsence` was constructing a minimal `netlink.Route` with only `Dst` set for the `RouteDel` call, while ignoring the full route object returned by `ExistsRoute`. The kernel's route deletion performs a strict match that includes attributes like `scope`, `dev`, and `gw`. When the actual route had `scope link` (253) but the deletion request defaulted to `scope universe` (0), the kernel couldn't find a matching entry and returned `ESRCH`.

## Fix

Use the actual `*netlink.Route` returned by `ExistsRoute` (which comes from `RouteListFiltered` and contains all kernel-resolved attributes) as the argument to `RouteDel`, ensuring an exact match.

### DIFF
--- a/pkg/route/route.go
+++ b/pkg/route/route.go
@@ -56,21 +56,13 @@ func EnsureRoutesAbsence(routes []networkingv1beta1.Route, tableID uint32) error
 		if routes[i].Dst == nil {
 			continue
 		}
-		_, dst, err := net.ParseCIDR(routes[i].Dst.String())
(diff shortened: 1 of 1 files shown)


fra98:
Closing it as already contained in https://github.com/liqotech/liqo/pull/3271

PR #3277, opened 2026-05-04
[Networking] Geneve tunnel conncheck

cheina97:
### Summary

This PR adds an active connectivity health-check (ping) mechanism for Geneve tunnels, tracking connection status and latency directly on the `GeneveTunnel` CRD.

**What changed:**

- **`GeneveTunnel` CRD extended** — added a `Status` subresource with `value` (Connected/Error) and `latency` fields, plus `kubectl get` print columns for quick status inspection.
- **Gateway `InternalNodeReconciler`** — on reconcile, lazily initialises a `ConnChecker` bound to the gateway's inner Geneve IP; registers a UDP ping sender per tunnel and wires a callback that writes connectivity results back to `GeneveTunnel` status. Senders are stopped when the corresponding `InternalNode` is deleted.
- **Fabric `InternalFabricReconciler`** — lazily starts a `ConnChecker` receiver bound to the node's inner Geneve IP, so the fabric side can respond to pings from the gateway.
- **`ConnChecker` improvements**:
  - Supports binding to a specific IP (`BindIP` option), enabling per-interface isolation between Geneve and WireGuard ping traffic.
  - Latency is now smoothed with an EWMA (`PingLatencyAlpha` option, default 0.1) instead of using raw per-packet measurements.
- **New flags** (gateway `geneve-fabric` component):
  - `--geneve-ping-enabled` (default `true`)
  - `--geneve-ping-port` (default `12346`)
  - `--geneve-ping-interval` (default `2s`)
  - `--geneve-ping-loss-threshold` (default `5`)
  - `--geneve-ping-update-status-interval` (default `10s`)
  - `--geneve-ping-latency-alpha` (default `0.1`)
- **New flag** (fabric component): `--geneve-ping-port` must match the gateway value.
- **New flag** (connection component): `--ping-latency-alpha` applies the same EWMA smoothing to WireGuard connection checks.

### How it works

```
Gateway InternalNodeReconciler
  └─ ConnChecker (sender, bound to gateway inner IP)
       └─ UDP ping → Fabric node inner IP : geneve-ping-port
            └─ ConnChecker (receiver, bound to node inner IP)
       └─ pong → callback → GeneveTunnel.Status.{Value, Latency}
```

When `--geneve-ping-enabled=false` the status is set to `Connected` immediately without running the ping loop.

### Test plan

- [x] Deploy two clusters peered with Geneve mode.
- [x] Verify `kubectl get genevetunnels` shows `Connected` and a latency value after a few seconds.
- [x] Delete an `InternalNode`; verify no goroutine leak (sender is stopped).
- [x] Test with `--geneve-ping-enabled=false`; verify status is immediately `Connected`.


deployments/liqo/charts/liqo-crds/crds/networking.liqo.io_genevetunnels.yaml
```diff
@@ -140,6 +151,26 @@ spec:
             - internalFabricRef
             - internalNodeRef
             type: object
+          status:
+            description: GeneveTunnelStatus defines the observed state of GeneveTunnel.
+            properties:
+              latency:
+                description: Latency of the tunnel.
```
frisso, line 158:
```suggestion
                description: Round-trip latency of the tunnel.
```

deployments/liqo/charts/liqo-crds/crds/networking.liqo.io_genevetunnels.yaml
```diff
@@ -140,6 +151,26 @@ spec:
...
+                properties:
+                  timestamp:
+                    description: Timestamp of the latency.
```
frisso, line 161:
```suggestion
                    description: Timestamp of the last measured round-trip latency.
```

deployments/liqo/charts/liqo-crds/crds/networking.liqo.io_genevetunnels.yaml
```diff
@@ -140,6 +151,26 @@ spec:
...
+                    format: date-time
+                    type: string
+                  value:
+                    description: Value of the latency.
+                    type: string
+                type: object
+              value:
+                description: Value of the tunnel connectivity.
```
frisso, line 169: I do not understand what this means.

pkg/gateway/tunnel/metrics.go
```diff
@@ -36,6 +36,20 @@ var (
 	MetricsPeerIsConnected *prometheus.Desc
 	// MetricsLabels is the labels that are used for the metrics.
 	MetricsLabels []string
+
+	// GeneveMetricsLabels is the labels used for geneve tunnel metrics.
```
frisso:
```suggestion
	// GeneveMetricsLabels are the labels used for geneve tunnel metrics.
```

pkg/gateway/tunnel/metrics.go
```diff
@@ -36,6 +36,20 @@ var (
...
+	// The labels are: internal_fabric, internal_node, namespace, remote_cluster_id.
+	GeneveMetricsLabels []string
+	// MetricsGeneveLatency is the metric that exposes the latency of a geneve tunnel.
+	MetricsGeneveLatency *prometheus.Desc
+	// MetricsGeneveLatencyHistogram is the metric that exposes the latency distribution of a geneve tunnel.
```
frisso: Is it for each tunnel? (btw, does it scale?) If so, I would say:
```suggestion
	// MetricsGeneveLatencyHistogram is the metric that exposes the latency distribution of each individual geneve tunnel.
```
If metrics are *per each tunnel*, I would also modify the following strings (e.g.,  "the geneve tunnel connection status" ==> "the status of each geneve tunnel", etc).

deployments/liqo/charts/liqo-crds/crds/networking.liqo.io_genevetunnels.yaml
cheina97: Me too, I need to check what field is it

cheina97:
```
const (
	// Connected used when the connection is up and running.
	Connected ConnectionStatusValue = "Connected"
	// Connecting used as temporary status while waiting for the vpn tunnel to come up.
	Connecting ConnectionStatusValue = "Connecting"
	// ConnectionError used to se the status in case of errors.
	ConnectionError ConnectionStatusValue = "Error"
)
```

It is an enum and it will be one of these values

pkg/gateway/fabric/flag.go
```diff
@@ -42,4 +55,15 @@ func InitFlags(flagset *pflag.FlagSet, opts *Options) {
...
+	flagset.BoolVar(&opts.PingEnabled, FlagNameGenevePingEnabled.String(), true,
+		"Enable geneve tunnel ping health check")
+	flagset.IntVar(&opts.ConnCheckOptions.PingPort, FlagNameGenevePingPort.String(), 12346,
+		"UDP port used for geneve tunnel ping health check")
+	flagset.DurationVar(&opts.ConnCheckOptions.PingInterval, FlagNameGenevePingInterval.String(), 2*time.Second,
+		"Interval between geneve tunnel pings")
+	flagset.UintVar(&opts.ConnCheckOptions.PingLossThreshold, FlagNameGenevePingLossThreshold.String(), 5,
+		"Number of consecutive lost pings before a geneve tunnel is declared down")
+	flagset.DurationVar(&opts.PingUpdateStatusInterval, FlagNameGenevePingUpdateStatusInterval.String(), 10*time.Second,
+		"Minimum interval between GeneveTunnel status updates")
```
fra98: Would be nice to have some of these configurable from Helm like we have for wireguard ping

fra98 (approved):

docs/_downloads/grafana/liqonetwork.json
claudiolor, line 1: Maybe this file shouldn't be here

pkg/conncheck/conncheck.go
```diff
@@ -113,7 +117,7 @@ func (c *ConnChecker) RunSender(clusterID string) {
 			klog.Warningf("failed to send ping: %s", err)
 		}
 		return false, nil
-	}); err != nil {
+	}); err != nil && sender.Ctx.Err() == nil {
```
claudiolor, line 121: Is this if correct? From the `PollUntilContextCancel` method documentation:
```
The returned error will be from ctx.Err(), the condition's err return value, or nil
```
So if I understand correctly this if would never be matched as err == sender.Ctx.Err().

pkg/conncheck/receiver.go
```diff
@@ -71,15 +74,14 @@ func (r *Receiver) SendPong(raddr *net.UDPAddr, msg *Msg) error {
 
 // ReceivePong receives a PONG message.
 func (r *Receiver) ReceivePong(msg *Msg) error {
-	r.m.Lock()
```
claudiolor, line 74: Wondering whether removing the lock here is fine. Since the main loop calls a go routine to receive the pong, if we receive two subsequent pongs we might end up writing on the same peer.

pkg/conncheck/receiver.go
```diff
@@ -139,24 +139,41 @@ func (r *Receiver) Run(ctx context.Context) {
...
+
 // RunDisconnectObserver starts the disconnect observer.
 func (r *Receiver) RunDisconnectObserver(ctx context.Context) {
 	klog.Infof("conncheck receiver disconnect checker: started")
 	// Ignore errors because only caused by context cancellation.
-	err := wait.PollUntilContextCancel(ctx, time.Duration(r.opts.PingLossThreshold)*r.opts.PingInterval/10, true,
+	threshold := r.opts.PingLossThreshold
+	if threshold > uint(math.MaxInt64) {
+		threshold = uint(math.MaxInt64)
+	}
```
claudiolor, line 173:
```suggestion
	threshold := min(r.opts.PingLossThreshold, uint(math.MaxInt64))
```

pkg/fabric/genevetunnel_controller.go
```diff
@@ -136,6 +167,32 @@ func (r *GeneveTunnelReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	klog.Infof("Enforced interface %s for genevetunnel %s", internalfabric.Spec.Interface.Node.Name, req.String())
 
+	if r.Options.ConnCheckOptions.PingEnabled {
+		bindIP := r.Options.ConnCheckOptions.PingBindIP
+		if bindIP == "" {
+			bindIP = internalnode.Spec.Interface.Node.IP.String()
+		}
+
+		if err := r.initConnCheckerReceiver(ctx, bindIP); err != nil {
```
claudiolor: I'm not sure whether we are handling the context correctly here. Since we are passing the context from the reconcile, we are scoping the connChecker to the reconcile request, so I think this might lead to issues if we set a timeout in the reconcile

pkg/conncheck/receiver.go
cheina97: Unfortunatelly the linter is not able to understand that syntax and will trigger an error

pkg/conncheck/receiver.go
cheina97: There are actually two distinct races if we remove the lock entirely:
1. Reading the map at the same time it's being written by InitPeer, or read by RunDisconnectObserver. This one is a real problem. Go maps are not safe for concurrent read+write, no matter which keys are involved. It can crash the whole program with fatal error: concurrent map read and map write. We need to keep this protected.
2. Two PONGs (or a PONG and RunDisconnectObserver) writing to the same peer at the same time. I don't think this needs protection:
	- If two PONGs arrive close together, whichever one "wins" is still a valid recent update. The next PONG will fix it a moment later.
	- The clash with RunDisconnectObserver can only happen when a PONG arrives right at the disconnect threshold, meaning the peer is already borderline anyway. Either outcome (connected or disconnected) gets corrected on the next check.
	
So my plan: keep the lock only around the map lookup, since that's the part that can crash the program. The field updates on the peer after that can stay unlocked, since getting overwritten there is fine.

fra98: I think even if we are protecting the map, we are still writing to the same `peer`, which is a struct pointer. This is not ideal: even if the Go runtime doesn't detect a panic, it is still considered undefined behavior for the Go memory model. 
Since the slow part is the `updateCallback()` rather than updating the peer fields, why not use an RW mutex and unlock it just before the updateCallback is called?

e.g.:
```go
    r.m.Unlock()

	if err := peer.updateCallback(true, peer.latency, now); err != nil {
		return fmt.Errorf("failed to update peer %s: %w", msg.ClusterID, err)
	}
```

Note: one could argue that the update callback function is also under `peer` struct, hence might be protected as well. Although I don't think this is a problem, as it is unlikely to change, maybe we could just move it out of peer struct

pkg/conncheck/receiver.go
```diff
@@ -164,17 +167,23 @@ func (r *Receiver) RunDisconnectObserver(ctx context.Context) {
...
+			for id, peer := range peers {
 				if time.Since(peer.lastPongTimestamp) <= r.opts.PingInterval*thresholdDuration {
 					continue
 				}
 				klog.V(8).Infof("conncheck receiver: %s unreachable", id)
 				peer.connected = false
 				peer.latency = 0
-				err := peer.updateCallback(false, 0, time.Time{})
-				if err != nil {
+				if err := peer.updateCallback(false, 0, time.Time{}); err != nil {
```
fra98: Here we are not protecting writes on `peer`. What about unlocking just before the updateCallback like we do in the other function?

pkg/conncheck/receiver.go
```diff
@@ -74,26 +74,29 @@ func (r *Receiver) SendPong(raddr *net.UDPAddr, msg *Msg) error {
...
-		if msg.TimeStamp.Before(peer.lastPingTimestamp) {
-			klog.V(8).Infof("dropped a PONG message from %s because out-of-order", msg.ClusterID)
-			return nil
-		}
-		now := time.Now()
-		peer.lastPingTimestamp = msg.TimeStamp
-		peer.lastPongTimestamp = now
-		peer.latency = now.Sub(msg.TimeStamp)
-		peer.connected = true
+	r.m.RLock()
```
fra98: I think we have to use `Lock()` as all `RLock()` will allow concurrent write to shared peer struct

pkg/conncheck/receiver.go
```diff
@@ -30,9 +31,11 @@ import (
...
 	connected bool
 	latency   time.Duration
-	// lastReceivedTimestamp is the timestamp when the last received PING has been sent.
-	lastReceivedTimestamp time.Time
-	updateCallback        UpdateFunc
+	// lastPingTimestamp is the timestamp of the last received PING (used for out-of-order detection).
+	lastPingTimestamp time.Time
+	// lastPongTimestamp is the time when the last PONG was received.
+	lastPongTimestamp time.Time
+	updateCallback    UpdateFunc
```
varnastadeus, line 38: can we remove this altogether and instead of `updateCallback`, do the update in reconcile loop fetching value directly from connChecker?

cheina97: But the reconcile loop will reconcile only when the Connection will be modified. A solution should be to reconcile every x seconds, but we will merge 2 different logics in the same reconcile. Right now that controller is used to manage the connchecks only. Honestly I would prefer to keep the updateFunction call in the conncheck logic

PR #3278, opened 2026-05-04
feat: allow multiple overlapping reserved subnet with refcount

fra98:
# IPAM: ref-counted reservations, exclusive ownership, and pool whitelisting

## What this PR does

Reworks the IPAM to support two acquisition modes — **shared** (ref-counted, overlapping) and **exclusive** (sole ownership) — and relaxes pool enforcement so that `ipam.pools` acts as a whitelist for Liqo's internal allocations.

### Shared (ref-counted) reservations

Networks like podCIDR, serviceCIDR, and reserved subnets can now overlap. Each acquisition increments a reference count; each release decrements it. The network is freed only when `refCount` reaches 0. This eliminates conflicts when these ranges overlap (e.g., podCIDR `10.4.0.0/16` inside a reserved `10.0.0.0/8`).

### Exclusive ownership

CIDRs actively used by Liqo (externalCIDR, internalCIDR, peering CIDRs) are acquired exclusively (`refCount = -1`). An exclusive node blocks all child acquisitions — both shared and exclusive. A shared parent also blocks exclusive children, so reserving `10.0.0.0/8` prevents Liqo from allocating `10.80.0.0/16` exclusively within it.

### Pool whitelisting

`ipam.pools` now acts as a whitelist. Shared reservations of CIDRs outside the pools succeed as no-ops (those ranges are implicitly excluded). Mutable requests for out-of-pool CIDRs get remapped into the pools. Only immutable+exclusive requests for out-of-pool CIDRs fail.

This allows configurations like `ipam.pools: [172.20.0.0/12]` — Liqo only uses that range, without needing to explicitly blacklist `10.0.0.0/8` and `192.168.0.0/16`.

### Dynamic external/internal CIDR allocation

`externalCIDR`/`internalCIDR` Helm values default to `""`. When empty, the IPAM allocates a `/16` automatically (mutable + exclusive). When set explicitly, the CIDR is immutable + exclusive. Both are now validated at controller-manager startup, failing early on conflicts.

## API: two orthogonal flags on `NetworkAcquireRequest`

- **`immutable`**: exact prefix only vs. remap if taken
- **`exclusive`**: sole ownership vs. shared (ref-counted). Defaults to exclusive when `immutable=false`.

| immutable | exclusive     | Behavior                                                                              |
| --------- | ------------- | ------------------------------------------------------------------------------------- |
| true      | false         | Shared reservation (blacklisting). podCIDR, serviceCIDR, reserved subnets.            |
| true      | true          | Exclusive, exact prefix. User-defined externalCIDR/internalCIDR.                      |
| false     | (always true) | Exclusive, remap if needed. Auto-allocated externalCIDR/internalCIDR, peering CIDRs.  |

Out-of-pool behavior:

| immutable | exclusive | In pool? | Result                      |
| --------- | --------- | -------- | --------------------------- |
| true      | false     | no       | No-op (implicitly reserved) |
| true      | true      | no       | Error                       |
| false     | *         | no       | Remap from pools            |

### IPAM core (`pkg/ipam/core/`)

`node.refCount` changed from `bool` to `int`: `-1` = exclusive, `0` = free, `>=1` = shared. Traversal in `allocateNetworkWithPrefix` blocks children when the parent is exclusive OR when the parent is shared and the child wants exclusive. `listNetworks` prunes subtrees under exclusive nodes (no descendants can be acquired there). Double-release of shared nodes is guarded against (refCount can't go below 0).

### IPAM server (`pkg/ipam/`)

gRPC handler dispatches on `immutable`/`exclusive` with early returns for out-of-pool cases. `initialize.go` and `sync.go` skip out-of-pool shared networks. Persistence via `network-shared` label on Network CRs.

### Controller manager (`cmd/liqo-controller-manager/modules/networking.go`)

New `initializeRequiredNetworks` runs after `initializeReservedNetworks` to acquire externalCIDR/internalCIDR at startup, catching conflicts early instead of deferring to the reconciler.

### Helm

`externalCIDR`/`internalCIDR` default to `""` with `10.70.0.0/16`/`10.80.0.0/16` as template fallbacks. `network-not-remapped` label is conditional on whether the value is set. `network-shared` label added to podCIDR, serviceCIDR, and reserved subnets.

### Proto

Added `exclusive` field (field 4) to `NetworkAcquireRequest`.


pkg/consts/ipam.go
```diff
@@ -30,6 +30,11 @@ const (
 	NetworkNotRemappedLabelKey = "ipam.liqo.io/network-not-remapped"
 	// NetworkNotRemappedLabelValue is the label value used to mark a Network that does not need CIDR remapping.
 	NetworkNotRemappedLabelValue = "true"
+	// NetworkSharedLabelKey is the label key used to mark a Network as shared (ref-counted, non-exclusive).
+	// Networks without this label are exclusive by default.
+	NetworkSharedLabelKey = "ipam.liqo.io/network-shared"
```
claudiolor, line 35: I'd use the "reserved" word, as you can allocate a CIDR on the IPAM, in that case you are taking ownership of the CIDR to use it. While, if you are reserving, you don't want people to use a CIDR, so it's fine if you reserve something multiple times
```suggestion
	NetworkReservedLabelKey = "ipam.liqo.io/network-reserved"
```

pkg/ipam/ipam.go
```diff
@@ -148,20 +149,30 @@ func (lipam *LiqoIPAM) NetworkAcquire(_ context.Context, req *NetworkAcquireRequ
...
-	if !lipam.isInPool(prefix) {
+	// Out-of-pool shared reservations are implicitly satisfied: Liqo cannot allocate from
+	// ranges outside its pools, so the CIDR is already excluded without tracking it.
+	// This only applies for:
+	// - CIDRs which don't allow remapping
+	// - CIDRs which are not intended to be used, hence we exclude "exclusive" reservations
+	if !lipam.isInPool(prefix) && req.GetImmutable() && !req.GetExclusive() {
+		klog.Infof("Network %q is outside pools, implicitly reserved", prefix.String())
+		return &NetworkAcquireResponse{Cidr: prefix.String()}, nil
+	}
```
claudiolor, line 160: nit: Maybe here I'd make the IPAM return an known error (like `OutOfRangeCIDRError`, which the controller manager can handle by just keeping empty the status of the resource

pkg/ipam/ipam.go
```diff
@@ -148,20 +149,30 @@ func (lipam *LiqoIPAM) NetworkAcquire(_ context.Context, req *NetworkAcquireRequ
...
+	case req.GetImmutable():
+		remappedCidr, err = lipam.networkAcquireSpecific(prefix, req.GetExclusive())
+	default:
 		remappedCidr, err = lipam.networkAcquire(prefix)
-		if err != nil {
-			return &NetworkAcquireResponse{}, err
-		}
+	}
+	if err != nil {
+		return &NetworkAcquireResponse{}, err
```
claudiolor:
```suggestion
		return &NetworkAcquireResponse{}, fmt.Errof("acquiring the network: %w", err)
```

PR_OVERVIEW.md
```diff
@@ -0,0 +1,73 @@
+# IPAM: ref-counted overlapping reservations and exclusive ownership
```
claudiolor: Committed by mistake?

pkg/consts/ipam.go
claudiolor, line 35: Wondering: what if we add a filed in the network resource specs?

fra98: My only fear is that it can be confused with `ipam.liqo.io/network-type=reserved` (for context, it's the label to indicate extra reserved cidr, but doesn't include podCIDR and serviceCIDR which they have their own network-type).
So network-shared would include podCIDR, serviceCIDR and all extra reserved

fra98: Yes I wanted to do that but there are two problems:
- the other grpc field (`immutable`) also is set through label and not CRD spec. I'd keep both as label for consistency at this point
- since this field is now key part of ipam, putting it under CRD requires crds upgrade so it's more breaking change. Label approach works with simple helm upgrade

pkg/ipam/ipam.go
fra98: I am not sure about UX in case of empty status. From the CR I cannot distinguish whatever the network failed to acquire because of real errors or because it was out-of-pool. Technically with whitelist approach this CIDR is implicitely reserved, so having status set seems right and would validate that IPAM is correctly configured. 
Moreover for external integration (e.g., controller) is simpler to always watch status. If empty, you can infer the IPAM didn't handle the CIDR correctly

pkg/consts/ipam.go
claudiolor: Ok

pkg/ipam/ipam.go
claudiolor: Ok

claudiolor (approved):

PR #3282, opened 2026-05-10
fix: serve stub OpenAPI documents from metric-agent

kariya-mitsuru:
# Description
kube-apiserver's OpenAPI aggregation controller queries /openapi/v2 and /openapi/v3 on every registered APIService. metric-agent did not handle these paths, causing the aggregator to log persistent errors.


fra98 (approved):

fra98:
Thanks for your contribution! @kariya-mitsuru

PR #3286, opened 2026-05-12
[Networking] avoid conntrack for geneve traffic

cheina97:
# Description

This PR introduces a `notrack` action for `FilterRule` resources, deploys a default `FirewallConfiguration` that exempts Geneve tunnel traffic from Linux connection tracking, and ensures that any pre-existing conntrack entries matching those `notrack` rules are evicted after the nftables rules are applied.

## Motivation

Geneve is the encapsulation protocol used by the Liqo fabric to carry cross-cluster traffic. Each Geneve packet was being tracked by the kernel's `conntrack` subsystem, which:

- introduces unnecessary CPU and memory overhead for tunnel traffic that does not need stateful inspection;
- risks flooding the conntrack table under high-throughput scenarios.

By marking UDP traffic on the configured Geneve port as `NOTRACK` in the `prerouting` and `output` hooks at priority `-300`, conntrack is bypassed entirely for that traffic.

However, simply adding `notrack` nftables rules does not affect connections that are already tracked: existing conntrack entries remain alive until they expire. To make the change effective immediately, the `firewallconfiguration` controller now removes existing conntrack flows that match the IPs, ports, and protocols declared in `notrack` rules.

### API (`apis/networking/v1beta1/firewall`)
- Added `ActionNotrack FilterAction = "notrack"` constant to `FilterRule` types.
- Extended the `+kubebuilder:validation:Enum` marker to include the new `notrack` value.

### Firewall utils (`pkg/firewall/utils`)
- Added `applyNotrackAction` which appends `&expr.Notrack{}` to the nftables rule expressions.
- Wired `ActionNotrack` into the `forgeFilterRule` switch statement.

### Firewall configuration controller (`pkg/firewall`)
- Added `pkg/firewall/conntrack.go`:
  - Abstracts a subset of `github.com/ti-mo/conntrack` behind a small `conntrackClient` interface.
  - `flushConntrackForNotrackRules` runs after nftables rules are applied.
  - It dumps the conntrack table and deletes only flows that match at least one `notrack` rule in either the original or reply direction.
  - Matching supports source/destination IP (single IP or CIDR), source/destination port (single port or range), protocol (`tcp`/`udp`), and `eq`/`neq` operators.
- Modified `pkg/firewall/firewallconfiguration_controller.go`:
  - Added `ConntrackClient` field to `FirewallConfigurationReconciler`.
  - Constructor opens a conntrack netlink connection and logs a warning if unavailable.
  - After the final `r.NftConnection.Flush()` and the `Applied firewallconfiguration` log, the controller calls `flushConntrackForNotrackRules` before returning.
- Added `pkg/firewall/conntrack_test.go` with unit tests for rule extraction, flow matching, port range parsing, and error propagation.

### Dependencies
- Added `github.com/ti-mo/conntrack v0.6.0` to `go.mod` and `go.sum`.

### Helm chart (`deployments/liqo`)
- Added `deployments/liqo/templates/liqo-fabric-notrack-firewallconfiguration.yaml`: a new `FirewallConfiguration` named `fabric-notrack` that creates an `liqo-notrack` nftables table with `prerouting` and `output` chains. Each chain contains two filter rules (matching UDP destination and source port equal to the configured Geneve port) with the `notrack` action and per-rule counters.
- Added `networking.fabric.config.notrack.enabled` (default: `true`) to `values.yaml` to allow disabling the feature at install time.

### Tests (`pkg/firewall/utils/filterule_test.go`)
- `Equal should return true for ActionNotrack` — basic notrack rule with no match.
- `Equal should return true for ActionNotrack with UDP port match` — notrack rule matching UDP + destination port, with counter enabled (mirrors the actual deployed rules).
- `Equal should return true for ActionNotrack with Counter` — notrack rule with counter only.

## How Has This Been Tested?

- Unit tests added in `pkg/firewall/utils/filterule_test.go` covering the new `ActionNotrack` path; all pass (`go test ./pkg/firewall/utils/...`).
- Unit tests added in `pkg/firewall/conntrack_test.go` covering the targeted conntrack cleanup logic; all pass (`go test ./pkg/firewall/...`).
- Manual verification on a local cluster: after deploying the chart, `nft list table ip liqo-notrack` shows the expected `prerouting` and `output` chains with `notrack` statements for Geneve UDP traffic.


deployments/liqo/templates/liqo-notrack-firewallconfiguration.yaml
```diff
@@ -0,0 +1,80 @@
...
+metadata:
+  name: {{ .name }}
+  labels:
+    networking.liqo.io/firewall-category: {{ .category }}
+    networking.liqo.io/firewall-subcategory: {{ .subcategory }}
+    liqo.io/webhook-skip: "true"
+spec:
+  table:
+    family: IPV4
+    name: liqo-notrack
```
fra98:
```suggestion
    name: liqo-notrack-geneve
```

deployments/liqo/values.yaml
```diff
@@ -196,6 +196,11 @@ networking:
       # -- Enable/Disable the route monitor for the fabric pod.
       # It means that the fabric pod will monitor the routing rules and will restore them in case of changes.
       routeMonitor: true
+      # -- Enable/Disable the creation of the NOTRACK firewallconfiguration for the fabric pod.
+      # When enabled, UDP traffic on the geneve port is exempted from connection tracking,
+      # which reduces overhead and prevents the conntrack table from being flooded by tunnel traffic.
+      notrack:
+        enabled: true
```
fra98:
```suggestion
        enabled: false
```
I would keep it disabled until we have https://github.com/liqotech/liqo/pull/3290 merged

fra98 (dismissed):

pkg/firewall/firewallconfiguration_controller.go
```diff
@@ -161,6 +170,12 @@ func (r *FirewallConfigurationReconciler) Reconcile(ctx context.Context, req ctr
 
 	klog.Infof("Applied firewallconfiguration %s", req.String())
 
+	// Existing conntrack flows are not affected by newly added notrack expressions,
+	// so they need to be evicted explicitly after the nftables rules have been applied.
+	if err = r.flushConntrackForNotrackRules(ctx, fwcfg); err != nil {
+		return ctrl.Result{}, fmt.Errorf("flushing conntrack entries: %w", err)
+	}
```
fra98, line 182: We should not run this if helm value is `notrack.enabled=false`

pkg/firewall/firewallconfiguration_controller.go
```diff
@@ -64,6 +66,12 @@ func newFirewallConfigurationReconciler(cl client.Client, s *runtime.Scheme, pod
 	if err != nil {
 		return nil, fmt.Errorf("unable to create nftables connection: %w", err)
 	}
+
+	conntrackClient, err := newConntrackConn()
```
fra98, line 70: Also this should be initialized only if helm value is enabled

pkg/firewall/firewallconfiguration_controller.go
```diff
@@ -64,6 +66,12 @@ func newFirewallConfigurationReconciler(cl client.Client, s *runtime.Scheme, pod
...
+	if err != nil {
+		klog.Warningf("unable to create conntrack connection: %v", err)
+	}
```
fra98, line 73: Why skip and not returning error? This seems applicative error we should block (of course if user enabled notrack)

pkg/firewall/conntrack.go
```diff
@@ -0,0 +1,244 @@
...
+			continue
+		}
+		if err := r.ConntrackClient.Delete(&flows[i]); err != nil {
+			return fmt.Errorf("deleting conntrack flow %v: %w", flows[i], err)
+		}
+		deleted++
+	}
+
+	klog.Infof("Deleted %d conntrack flows matching notrack rules in firewallconfiguration %s/%s",
+		deleted, fwcfg.Namespace, fwcfg.Name)
```
fra98: I would log only if deleted > 0 to avoid spamming

pkg/firewall/conntrack.go
```diff
@@ -0,0 +1,244 @@
...
+	flows, err := r.ConntrackClient.Dump(nil)
+	if err != nil {
+		return fmt.Errorf("dumping conntrack flows: %w", err)
+	}
+
+	deleted := 0
+	for i := range flows {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
```
fra98, line 85: Why this check here?

pkg/firewall/conntrack.go
```diff
@@ -0,0 +1,244 @@
...
+	networkingv1beta1 "github.com/liqotech/liqo/apis/networking/v1beta1"
+	firewallapi "github.com/liqotech/liqo/apis/networking/v1beta1/firewall"
+)
+
+// conntrackClient abstracts the subset of conntrack.Conn used by the reconciler.
+type conntrackClient interface {
+	Dump(*conntrack.DumpOptions) ([]conntrack.Flow, error)
+	Delete(*conntrack.Flow) error
+	Close() error
+}
```
fra98, line 36: What's the point of this interface?

pkg/firewall/firewallconfiguration_controller.go
```diff
@@ -161,6 +170,12 @@ func (r *FirewallConfigurationReconciler) Reconcile(ctx context.Context, req ctr
 
 	klog.Infof("Applied firewallconfiguration %s", req.String())
 
+	// Existing conntrack flows are not affected by newly added notrack expressions,
+	// so they need to be evicted explicitly after the nftables rules have been applied.
+	if err = r.flushConntrackForNotrackRules(ctx, fwcfg); err != nil {
```
fra98: Not sure if this is the right approach, this is gonna be a heavy operation as at every reconcile on any fwcfg it will list all conntrack which are potentially a lot. Not sure if we have other options as well 🤔

fra98: Maybe we can do it only when controller-runtime operation is `created`?

pkg/firewall/conntrack.go
```diff
@@ -0,0 +1,244 @@
...
+	case match.IP != nil:
+		matched = matchIP(tuple, match.IP)
+	case match.Port != nil:
+		matched = matchPort(tuple, match.Port)
+	case match.Proto != nil:
+		matched = matchProto(tuple, match.Proto)
+	case match.Dev != nil:
+		// Conntrack flows do not carry ingress/egress device information.
+		return true
+	}
```
fra98: What if a match entry have more than one between IP/Port/Proto/Dev under the same? With this code it will skip some if the first match?

deployments/liqo/templates/liqo-notrack-firewallconfiguration.yaml
```diff
@@ -0,0 +1,80 @@
...
+              counter: true
+              action: notrack
+              match:
+                - op: eq
+                  proto:
+                    value: udp
+                - op: eq
+                  port:
+                    position: dst
+                    value: {{ $genevePort | quote }}
```
fra98, line 37: NIT: this is a bit related to previous question, can this be written with a single match matching both proto and port?

pkg/firewall/firewallconfiguration_controller.go
cheina97: Why not? What about if a user creates a notrack FirewallConfiguration?

pkg/firewall/firewallconfiguration_controller.go
cheina97: Same answer as before

pkg/firewall/conntrack.go
cheina97: Makes sense. But I've testeda and it is not spamming a lot

pkg/firewall/firewallconfiguration_controller.go
cheina97: No, it list all the conntracks only when the firewallconfiguration contains a notrack rule

pkg/firewall/conntrack.go
cheina97: It's not possible, it is enforced by the webhook if I remember well

pkg/firewall/conntrack.go
cheina97: It is for safety, in case the deletion is tking too much time the context will be closed by controller-runtime. This will stop the for loop and triggers an error

cheina97: I've improved the error message

pkg/firewall/conntrack.go
cheina97: It's required for testing

pkg/firewall/conntrack.go
cheina97: My bad, I remembered differently. Nice catch

deployments/liqo/values.yaml
```diff
@@ -61,6 +61,11 @@ networking:
   denyDirectConnections: false
   # -- The port used by the geneve tunnels.
   genevePort: 6091
+  # -- Enable/Disable the creation of the NOTRACK firewallconfiguration for the geneve tunnel traffic.
+  # When enabled, UDP traffic on the geneve port is exempted from connection tracking,
+  # which reduces overhead and prevents the conntrack table from being flooded by tunnel traffic.
+  notrack:
+    enabled: true
```
cheina97: Need to revert it after tests pass

pkg/firewall/firewallconfiguration_controller.go
fra98: I mean if it happens it will dump all conntrack on the node/pod, potentially thousand or even tens of thousands

fra98 (approved):

PR #3290, opened 2026-05-15
[Networking] FirewallConfigurationBinding

cheina97:
## Background

The previous architecture let every host (fabric node or gateway pod) apply any
`FirewallConfiguration` whose labels matched its own identity. A single
controller in `liqo-controller-manager` managed a finalizer directly on the
`FirewallConfiguration` and wrote one status entry per host into the resource.
This model had two fundamental problems.

**Status does not scale.** For a `FirewallConfiguration` that targets all fabric
nodes, the status slice would accumulate one entry per node in the cluster. On
large deployments this makes the resource excessively large and causes
watch/update churn proportional to the number of nodes.

**Cleanup is broken.** There was a single finalizer on the
`FirewallConfiguration`. When the resource was deleted, every host raced to
remove it. Only the *first* host to win the race would execute the nftables
teardown; all subsequent hosts would find the finalizer already gone and silently
skip cleanup, leaving dangling nftables rules on every other node.

This PR replaces that model with one `FirewallConfigurationBinding` object per
(FirewallConfiguration, host) pair. Each binding is owned exclusively by its
host, carries its own finalizer, and holds its own status — eliminating both the
scalability and the correctness problems.

---

## New CRD: `FirewallConfigurationBinding`

This PR introduces `FirewallConfigurationBinding`
(`firewallconfigurationbindings.networking.liqo.io`), a new namespaced CRD in
the `networking.liqo.io` group.

A `FirewallConfigurationBinding` links a single **host entity** (a fabric node
or a gateway pod) to a **`FirewallConfiguration`**. The binding carries:

| Field | Purpose |
|---|---|
| `spec.firewallConfigurationRef` | Name of the `FirewallConfiguration` to apply. |
| `spec.targetID` | Identifier of the entity that must apply this binding (fabric node name or gateway pod name). |
| `status.conditions[Applied]` | Standard `metav1.Condition` reporting whether the nftables configuration has been successfully enforced. |
| `status.tableName` | The nftables table name, cached so cleanup can proceed even after the `FirewallConfiguration` is deleted. |

---

## How a binding attaches a host to a FirewallConfiguration

Each `FirewallConfiguration` carries a set of *selector labels*
(`liqo.io/firewall-category`, `liqo.io/firewall-subcategory`,
`liqo.io/firewall-unique`) that describe **which kind of hosts** should apply
it and **which scope** is involved (all gateways, a single gateway identified
by remote cluster ID, all fabric nodes, a single node, IP-mapping rules, …).

Two new **binding-creator controllers** watch `FirewallConfiguration` resources
and materialise the binding objects:

- **`GatewayBindingCreatorReconciler`** — runs inside each **gateway pod**
  (`cmd/gateway`). It inspects the category/subcategory labels of each
  `FirewallConfiguration` and creates one `FirewallConfigurationBinding` per
  matching `GatewayServer`/`GatewayClient`. The `spec.targetID` is set to the
  gateway pod name so that only the pod for that specific gateway will pick it
  up.

- **`FabricBindingCreatorReconciler`** — runs inside each **fabric DaemonSet
  pod** (`cmd/fabric`). It enumerates `InternalNode` objects and creates one
  `FirewallConfigurationBinding` per matching node. The `spec.targetID` is set
  to the node name.

Both reconcilers share a common `BindingCreatorBase` (in `pkg/firewall`) that
handles the create-or-update idempotency, stale-binding deletion, and legacy
finalizer migration.

The actual nftables enforcement is performed by a third controller,
**`FirewallConfigurationBindingReconciler`**, also running inside each gateway
and fabric pod. It watches `FirewallConfigurationBinding` resources filtered by
`spec.targetID == <own identity>` so that each pod processes only its own
bindings.

---

## Cleanup

Cleanup is split across two layers:

### Per-binding cleanup (nftables)

Each `FirewallConfigurationBinding` has a finalizer
(`firewallconfigurationbinding-controller.liqo.io/finalizer`) added by the
`FirewallConfigurationBindingReconciler` on the first successful reconcile. When
the binding is deleted:

1. The controller reads `status.tableName` (cached at apply time) to locate the
   nftables table — this avoids having to fetch the `FirewallConfiguration`,
   which may already be gone.
2. It deletes the table from the kernel via `nftables`.
3. It removes the finalizer, allowing the resource to be garbage-collected.

### Stale-binding cleanup (Kubernetes objects)

The `BindingCreatorBase.deleteStaleBindings` method is called on every
reconcile of the binding-creator controllers. It lists all
`FirewallConfigurationBinding` resources owned by the current
`FirewallConfiguration` and deletes any whose `spec.targetID` no longer
corresponds to an existing target (e.g. a gateway that has been removed or a
node that has been drained). Before deleting a stale binding the controller
explicitly strips the pod-side finalizer to avoid the binding being stuck when
the target pod is already gone.

### Legacy finalizer migration

The old architecture used a finalizer
(`firewallconfigurations-controller.liqo.io/finalizer`) directly on the
`FirewallConfiguration`. The binding-creator controllers transparently strip
this finalizer when they encounter it, so clusters upgraded from the previous
architecture are unblocked without manual intervention.

---

### The problem

The fabric DaemonSet pods hold a finalizer on each
`FirewallConfigurationBinding`. The finalizer is removed only by the pod
itself, after it has cleaned up the corresponding nftables rules. This creates
a dependency between the Kubernetes RBAC that authorises the pod and the pod's
ability to complete its cleanup: if the RBAC (ClusterRole / ClusterRoleBinding)
is deleted *before* the pod finishes, the pod can no longer patch the binding
to remove the finalizer, leaving it stuck forever.

Standard `helm uninstall` deletes all chart resources in a single pass, which
means the RBAC and the DaemonSet are torn down simultaneously with no ordering
guarantee.

### The solution: `helm.sh/resource-policy: keep` + post-delete job

The fabric RBAC resources (ClusterRole, ClusterRoleBinding, ServiceAccount) are
annotated with:

```yaml
helm.sh/resource-policy: keep
```

This instructs Helm to **skip** those resources during `helm uninstall`, leaving
them in place so the still-running fabric pods can complete their finalizer
cleanup.

A new **post-delete Helm hook job** (`post-delete-job.yaml`) runs after Helm
has deleted everything else. It invokes the `liqo-uninstaller` binary with
`--helm-delete-mode=post`, which explicitly deletes the kept RBAC objects once
cleanup is guaranteed to be complete.

The uninstaller now distinguishes two phases:

| Phase | Flag | What it does |
|---|---|---|
| Pre-delete | `--helm-delete-mode=pre` | Existing teardown logic: removes `ForeignCluster`, `InternalNode`, `Network`, `IP`, `Configuration` objects and waits for deletion. |
| Post-delete | `--helm-delete-mode=post` | Deletes all Liqo-owned resources annotated with `helm.sh/resource-policy: keep` (the fabric RBAC). |

### ArgoCD sync-waves

When Liqo is deployed via ArgoCD, resources are reconciled in *sync-wave*
order. Without explicit ordering, ArgoCD might delete the fabric RBAC (wave 0)
at the same time as the DaemonSet pods, again causing stuck finalizers.

To enforce the correct deletion order:

- The fabric RBAC resources are annotated with
  `argocd.argoproj.io/sync-wave: "-1"` — they are created *first* on install
  and deleted *last* on uninstall (higher absolute wave value = later deletion
  in ArgoCD's reverse-sync-wave ordering for deletions).
- The fabric DaemonSet pod template is annotated with
  `argocd.argoproj.io/sync-wave: "0"` — the DaemonSet is created after the
  RBAC is ready and deleted before (earlier in deletion order relative to the
  RBAC at wave -1).

Together these two annotations guarantee that on ArgoCD-managed uninstall the
fabric pods are given the opportunity to drain their finalizers while their RBAC
is still present.


apis/networking/v1beta1/firewallconfigurationbinding_types.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// FirewallConfigurationBindingResource is the name of the FirewallConfigurationBinding resources.
+// firewallconfigurationbindings is used for resource name pluralization because the k8s API does not manage false friends.
+// Waiting for this fix https://github.com/kubernetes-sigs/kubebuilder/pull/3408
+var FirewallConfigurationBindingResource = "firewallconfigurationbindings"
```
fra98: Is comment outdated? This doesn't seem like a false friend plural

pkg/consts/controllers.go
```diff
@@ -30,31 +30,34 @@ const (
...
-	CtrlWGGatewayServer        = "wggatewayserver"
+	CtrlConfigurationExternal                      = "configuration_external"
+	CtrlConfigurationInternal                      = "configuration_internal"
+	CtrlConfigurationRemapping                     = "configuration_remapping"
+	CtrlConfigurationRoute                         = "configuration_route"
+	CtrlConnection                                 = "connection"
+	CtrlFirewallConfigurationBinding               = "firewallconfigurationbinding"
+	CtrlFirewallConfigurationBindingCreator        = "firewallconfigurationbinding-creator"
+	CtrlFabricFirewallConfigurationBindingCreator  = "fabric-fwcfgbinding-creator"
+	CtrlGatewayFirewallConfigurationBindingCreator = "gateway-fwcfgbinding-creator"
```
fra98: Can you use the convention here? First is name of reconciled resource, plus eventual suffix separated by underscore

pkg/liqo-controller-manager/networking/firewall-binding-gc/doc.go
fra98, line 1: Can we use a more targeted name for this pkg instead of `firewall`?

.devcontainer/devcontainer-lock.json
```diff
@@ -0,0 +1,34 @@
+{
```
claudiolor, line 1: Committed by mistake?

.kimchi/plans/plan-1783507840329.md
claudiolor, line 1: Same here

apis/networking/v1beta1/firewallconfigurationbinding_types.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// FirewallConfigurationBindingSpec defines the desired state of FirewallConfigurationBinding.
+type FirewallConfigurationBindingSpec struct {
+	// FirewallConfigurationRef is the reference to the FirewallConfiguration to apply.
+	FirewallConfigurationRef corev1.LocalObjectReference `json:"firewallConfigurationRef"`
+	// TargetRef identifies the entity (e.g. a gateway pod or an InternalNode) that should apply this binding.
+	// The FirewallConfigurationBinding controller running on the matching entity filters resources by this field.
+	// It must be unique across all FirewallConfigurationBinding controllers instances in the cluster,
+	// otherwise multiple entities will apply the same FirewallConfiguration, which may cause unexpected behavior.
```
claudiolor:
```suggestion
    // TargetRef identifies the entity (for example, a gateway pod or Liqo node fabric)
    // to which this FirewallConfigurationBinding binds a FirewallConfiguration.
    //
    // A FirewallConfigurationBinding controller only reconciles bindings whose
    // TargetRef matches the target it manages. For a given target, there must be
    // exactly one controller instance in the cluster. Otherwise, multiple
    // controllers will apply the referenced FirewallConfiguration, resulting in
    // unexpected behavior.
```

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Normal path: fetch the FirewallConfiguration to get the full table spec.
+	fwcfg := &networkingv1beta1.FirewallConfiguration{}
+	if err = r.Get(ctx, types.NamespacedName{
+		Name:      fwbinding.Spec.FirewallConfigurationRef.Name,
+		Namespace: fwbinding.Namespace,
+	}, fwcfg); err != nil {
```
claudiolor: Since this is a multi-line function call maybe it's more clear if we split the if from the function call:
```suggestion
	err = r.Get(ctx, types.NamespacedName{
		Name:      fwbinding.Spec.FirewallConfigurationRef.Name,
		Namespace: fwbinding.Namespace,
	}, fwcfg) 
    
    if err != nil {
```

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("Referenced firewallconfiguration %q not found for binding %s; requeuing",
+				fwbinding.Spec.FirewallConfigurationRef.Name, req.String())
+			return ctrl.Result{RequeueAfter: 5 * time.Second}, fmt.Errorf("referenced firewallconfiguration %q not found",
```
claudiolor: I think this won't work, as since you returning an error the RequeueAfter is ignored: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller.go#L494
Why do we need to requeue after 5 seconds?

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	}()
+
+	// Remove outdated chains/rules that are no longer in the spec, and flush deletions before adding new ones.
+	if err = cleanTable(r.NftConnection, &fwcfg.Spec.Table); err != nil {
+		return ctrl.Result{}, fmt.Errorf("cleaning table %s: %w", ptr.Deref(fwcfg.Spec.Table.Name, ""), err)
+	}
+
+	if err = r.NftConnection.Flush(); err != nil {
+		return ctrl.Result{}, fmt.Errorf("flushing nftables connection: %w", err)
```
claudiolor:
```suggestion
		return ctrl.Result{}, fmt.Errorf("flushing nftables connection after table cleanup: %w", err)
```

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+
+	klog.V(4).Infof("Applying firewallconfigurationbinding %s (fwcfg %s)", req.String(), fwcfg.Name)
+
+	table := addTable(r.NftConnection, &fwcfg.Spec.Table)
+	if err = addChains(r.NftConnection, fwcfg.Spec.Table.Chains, table); err != nil {
+		return ctrl.Result{}, fmt.Errorf("adding chains to table %s: %w", ptr.Deref(fwcfg.Spec.Table.Name, ""), err)
+	}
+
+	if err = r.NftConnection.Flush(); err != nil {
+		return ctrl.Result{}, fmt.Errorf("flushing nftables connection: %w", err)
```
claudiolor:
```suggestion
		return ctrl.Result{}, fmt.Errorf("flushing nftables connection after table creation: %w", err)
```

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+
+// NewFirewallConfigurationBindingReconciler returns a new FirewallConfigurationBindingReconciler.
+func NewFirewallConfigurationBindingReconciler(cl client.Client, s *runtime.Scheme, nodename string,
+	er events.EventRecorder) (*FirewallConfigurationBindingReconciler, error) {
+	nftConnection, err := nftables.New()
+	if err != nil {
+		return nil, fmt.Errorf("unable to create nftables connection: %w", err)
+	}
+	return &FirewallConfigurationBindingReconciler{
+		NodeName:       nodename,
```
claudiolor: Is NodeName actually used? It seems it isn't. Can we drop this field?

pkg/firewall/binding_creator_base.go
```diff
@@ -0,0 +1,158 @@
...
+type BindingCreatorBase struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+// ReconcileBindings is the core orchestration method shared by both the fabric and
+// gateway binding creator reconcilers. It strips the legacy finalizer and ensures a
+// binding exists for each target.
+func (b *BindingCreatorBase) ReconcileBindings(ctx context.Context,
+	fwcfg *networkingv1beta1.FirewallConfiguration, targets []BindingTarget) error {
```
claudiolor: nit: since we always pass a slice with a single element, we might evaluate to just pass the element
```suggestion
	fwcfg *networkingv1beta1.FirewallConfiguration, target BindingTarget) error {
```

pkg/firewall/binding_creator_base.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	if op != controllerutil.OperationResultNone {
+		klog.V(4).Infof("FirewallConfigurationBinding %s/%s %s", fwcfg.Namespace, bindingName, op)
+	}
+	return nil
+}
+
+// EnqueueFirewallConfigurationsByCategory lists all FirewallConfiguration resources with the
+// given category label and returns reconcile requests for each.
+func (b *BindingCreatorBase) EnqueueFirewallConfigurationsByCategory(
```
claudiolor: I seems this function is not used.

pkg/firewall/watch.go
```diff
@@ -54,3 +55,65 @@ func NewFirewallWatchEventHandler(cl client.Client, labelsSets []labels.Set) han
 			return requests
 		})
 }
```
claudiolor: The function above, `NewFirewallWatchEventHandler` is unused and can be removed.

pkg/liqo-controller-manager/networking/firewall/bindinggc_controller.go
```diff
@@ -0,0 +1,234 @@
...
+// SetupWithManager registers the BindingGCReconciler with the manager. In addition to the
+// periodic re-check driven by the reconciler RequeueAfter, it reacts to target deletions
+// (e.g. gateway pods and Nodes) so that orphaned bindings are collected promptly.
+func (r *BindingGCReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).Named(consts.CtrlFirewallConfigurationBindingGC).
+		For(&networkingv1beta1.FirewallConfigurationBinding{}).
+		Watches(&corev1.Pod{},
+			handler.EnqueueRequestsFromMapFunc(r.enqueueBindingsByTarget)).
+		Watches(&corev1.Node{},
+			handler.EnqueueRequestsFromMapFunc(r.enqueueBindingsByTarget)).
```
claudiolor: Are these watches actually needed?
I mean, the garbage collector is the last resort if the fabric or the gateway couldn't clean up before termination. So I think we can just rely on the 10 minutes period and avoid such expesive watches.

claudiolor (approved):

pkg/firewall/firewallconfigurationbinding_finalizer.go
```diff
@@ -0,0 +1,42 @@
...
+	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	networkingv1beta1 "github.com/liqotech/liqo/apis/networking/v1beta1"
+)
+
+const (
+	firewallConfigurationBindingControllerFinalizer = "firewallconfigurationbinding-controller.liqo.io/finalizer"
+	// FirewallConfigurationBindingControllerFinalizer is the exported alias used by external
+	// packages (e.g. liqo-controller-manager) to reference the same finalizer string.
+	FirewallConfigurationBindingControllerFinalizer = firewallConfigurationBindingControllerFinalizer
```
claudiolor, line 29: Can't we just use the exported variable?

test/e2e/pipeline/utils.sh
```diff
@@ -197,9 +197,15 @@ function login_az() {
...
+  # Pin to a chart version compatible with Kubernetes < 1.31.
+  # Newer Kyverno charts use selectableFields in CRDs which requires K8s >= 1.31
+  # and causes Helm v4 to fail with a typed patch error on older clusters.
+  local KYVERNO_CHART_VERSION="3.2.7"
+
   "${HELM}" repo add kyverno https://kyverno.github.io/kyverno/
   "${HELM}" repo update
-  "${HELM}" install kyverno kyverno/kyverno -n kyverno --create-namespace --kubeconfig "${kubeconfig}"
+  "${HELM}" install kyverno kyverno/kyverno -n kyverno --create-namespace \
+    --version "${KYVERNO_CHART_VERSION}" --kubeconfig "${kubeconfig}"
```
fra98: Is this still needed? Kubernetes <1.31 is out of support window

test/e2e/pipeline/utils.sh
```diff
@@ -116,7 +116,7 @@ function install_helm() {
   local supported="darwin-amd64\ndarwin-arm64\nlinux-386\nlinux-amd64\nlinux-arm\nlinux-arm64\nlinux-ppc64le\nlinux-s390x\nwindows-amd64"
   check_supported_arch_and_os "${supported}" "${os}" "${arch}" helm
 
-  HELM_VERSION="v3.15.3"
+  HELM_VERSION="v4.2.0"
```
fra98: Is this bump in this PR needed also?

cmd/fabric/main.go
```diff
@@ -166,25 +180,21 @@ func run(cmd *cobra.Command, _ []string) error {
...
+		options.NodeName,
+		mgr.GetEventRecorder("firewall-binding-controller"),
 	)
 	if err != nil {
-		return fmt.Errorf("unable to create firewall configuration reconciler: %w", err)
+		return fmt.Errorf("unable to create firewall configuration binding reconciler: %w", err)
 	}
 
 	if err := fwcr.SetupWithManager(cmd.Context(), mgr,
+		"v1", "Node", options.NodeName, "",
```
fra98: NIT: what about passing `networkingv1beta1.TargetReference` struct instead of the 4 individual values for clearer intent?

pkg/consts/labels.go
```diff
@@ -31,6 +31,8 @@ const (
 	K8sAppComponentKey = "app.kubernetes.io/component"
 	// K8sAppPartOfKey = key of the label used to denote the application a resource is part of.
 	K8sAppPartOfKey = "app.kubernetes.io/part-of"
+	// K8sAppPartOfLiqoValue = value of the label used to denote a resource is part of Liqo.
+	K8sAppPartOfLiqoValue = "liqo"
```
fra98: Leftover I think

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+//nolint:revive // We usually use the name of the reconciled resource in the controller name.
+type FirewallConfigurationBindingReconciler struct {
+	NodeName      string
+	NftConnection *nftables.Conn
+	client.Client
+	Scheme         *runtime.Scheme
+	EventsRecorder events.EventRecorder
+}
+
+// NewFirewallConfigurationBindingReconciler returns a new FirewallConfigurationBindingReconciler.
```
fra98, line 58: Important: this PR is rebased before some changes we did to old firewallconfiguration_controller.go related to conntrack deletion. Hence we need to restore this capability in this PR as well otherwise we will just delete the file losing the changes

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+	if enableNftMonitor {
+		go func() {
+			utilruntime.Must(netmonitor.InterfacesMonitoring(ctx, src, &netmonitor.Options{Nftables: &netmonitor.OptionsNftables{Delete: true}}))
+		}()
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).Named(consts.CtrlFirewallConfigurationBinding).
+		For(&networkingv1beta1.FirewallConfigurationBinding{}, builder.WithPredicates(ForgeTargetRefPredicate(apiVersion, kind, name, namespace))).
+		Watches(&networkingv1beta1.FirewallConfiguration{},
+			NewFirewallConfigurationBindingByConfigurationHandler(r.Client, apiVersion, kind, name, namespace)).
```
fra98: Also here we can pass TargetReference

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+	if existing != nil && existing.Status == newStatus && existing.ObservedGeneration == fwbinding.Generation {
+		return nil
+	}
+
+	apimeta.SetStatusCondition(&fwbinding.Status.Conditions, metav1.Condition{
+		Type:               condType,
+		Status:             newStatus,
+		ObservedGeneration: fwbinding.Generation,
+		Reason:             string(networkingv1beta1.FirewallConfigurationBindingConditionTypeApplied),
+		Message:            string(newStatus),
```
fra98: reason and message should be set according to metav1 convention, here we are assigning respectively Type and Status. We can use reconcileErr as Message and ad-hoc consts for Reason based on success of failure

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+	})
+
+	r.EventsRecorder.Eventf(fwbinding, nil, "Normal", "FirewallConfigurationBindingUpdate", "Updated",
+		"FirewallConfigurationBinding %s/%s: %s", fwbinding.Namespace, fwbinding.Name, newStatus)
+	if clerr := r.Client.Status().Update(ctx, fwbinding); clerr != nil {
+		return errors.Join(reconcileErr, clerr)
```
fra98: here we are joining reconcileErr with the update err, but reconcileErr is already joined in the deferred function, so here we are duplicating it 🤔

pkg/firewall/firewallconfigurationbinding_controller.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// MatchesTargetRef returns true if the given TargetReference matches the provided
+// apiVersion, kind, name and namespace.
+func MatchesTargetRef(ref *networkingv1beta1.TargetReference, apiVersion, kind, name, namespace string) bool {
+	return ref.APIVersion == apiVersion &&
+		ref.Kind == kind &&
+		ref.Name == name &&
+		ref.Namespace == namespace
+}
```
fra98: This could be a method of networkingv1beta1.TargetReference, so it can re re-used across all packages instead of importing this one. It can also be re-used for future RouteConfigurationBinding refactoring

pkg/liqo-controller-manager/networking/external-network/remapping/label_test.go
fra98, line 1: These tests seem overkill in my opinion

pkg/fabric/label_test.go
fra98, line 1: Same here

pkg/gateway/label_test.go
fra98, line 1: and here

pkg/firewall/watch.go
fra98: Also `NewFirewallWatchSource()` seems unused

pkg/firewall/watch.go
```diff
@@ -54,3 +55,65 @@ func NewFirewallWatchEventHandler(cl client.Client, labelsSets []labels.Set) han
...
+}
+
+// NewFirewallBindingWatchSource creates a new Source for the FirewallConfigurationBinding watcher.
+func NewFirewallBindingWatchSource(src <-chan event.GenericEvent, eh handler.EventHandler) source.Source {
+	return source.Channel(src, eh)
+}
+
+// NewFirewallBindingWatchEventHandler creates a new EventHandler for FirewallConfigurationBinding resources.
+// It enqueues all bindings whose spec.targetRef matches the given apiVersion, kind, name and namespace.
+func NewFirewallBindingWatchEventHandler(cl client.Client, apiVersion, kind, name, namespace string) handler.EventHandler {
```
fra98: nit:

```suggestion
func NewFwCfgBindingEnqueuer(cl client.Client, apiVersion, kind, name, namespace string) handler.EventHandler {
```

pkg/firewall/watch.go
```diff
@@ -54,3 +55,65 @@ func NewFirewallWatchEventHandler(cl client.Client, labelsSets []labels.Set) han
 			return requests
 		})
 }
+
+// NewFirewallConfigurationBindingByConfigurationHandler returns an EventHandler that,
+// given a FirewallConfiguration object, enqueues all FirewallConfigurationBindings in
+// the same namespace that reference it through spec.firewallConfigurationRef.name and
+// match the given target (apiVersion, kind, name, namespace).
+func NewFirewallConfigurationBindingByConfigurationHandler(cl client.Client, apiVersion, kind, name, namespace string) handler.EventHandler {
```
fra98: nit:

```suggestion
func NewFwCfgToFwCfgBindingEnqueuer(cl client.Client, apiVersion, kind, name, namespace string) handler.EventHandler {
```

pkg/firewall/binding_creator_base.go
```diff
@@ -0,0 +1,158 @@
...
+type BindingTarget struct {
+	// APIVersion of the target (e.g. "v1" for a Pod or "networking.liqo.io/v1beta1" for an InternalNode).
+	APIVersion string
+	// Kind of the target (e.g. "Pod" or "InternalNode").
+	Kind string
+	// Name is the node name (for fabric) or gateway pod name (for gateway).
+	Name string
+	// Namespace of the target. Empty for cluster-scoped targets (e.g. an InternalNode).
+	Namespace string
+}
```
fra98: This seems a clone of networkingv1beta1.TargetReference, can we re-use that?

pkg/fabric/binding_creator_controller.go
```diff
@@ -0,0 +1,114 @@
...
+	labelPredicates := make([]predicate.Predicate, len(labelsSets))
+	for i := range labelsSets {
+		p, err := predicate.LabelSelectorPredicate(metav1.LabelSelector{MatchLabels: labelsSets[i]})
+		if err != nil {
+			return nil, err
+		}
+		labelPredicates[i] = p
+	}
+	return predicate.Or(labelPredicates...), nil
+}
```
fra98: This was also deprecated in a recent commit after the one this PR is rebased on. Now it uses a shared utility called `NewAnyLabelsSetPredicate()`. I'd suggest to look at how legacy controller called it

pkg/gateway/binding_creator_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	labelPredicates := make([]predicate.Predicate, len(labelsSets))
+	for i := range labelsSets {
+		p, err := predicate.LabelSelectorPredicate(metav1.LabelSelector{MatchLabels: labelsSets[i]})
+		if err != nil {
+			return nil, err
+		}
+		labelPredicates[i] = p
+	}
+	return predicate.Or(labelPredicates...), nil
+}
```
fra98: Same here

pkg/firewall/binding_creator_base.go
```diff
@@ -0,0 +1,158 @@
...
+			Name:       target.Name,
+			Namespace:  target.Namespace,
+		}
+		return controllerutil.SetControllerReference(fwcfg, binding, b.Scheme)
+	})
+	if err != nil {
+		return fmt.Errorf("creating/updating binding %s/%s: %w", fwcfg.Namespace, bindingName, err)
+	}
+	if op != controllerutil.OperationResultNone {
+		klog.V(4).Infof("FirewallConfigurationBinding %s/%s %s", fwcfg.Namespace, bindingName, op)
```
fra98: We can remove V(4) since we are using `if op != controllerutil.OperationResultNone`

pkg/liqo-controller-manager/networking/firewall/bindinggc_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	if linked {
+		// The target entity is still alive: nothing to do now, re-check later.
+		return ctrl.Result{RequeueAfter: r.GCPeriod}, nil
+	}
+
+	klog.Infof("FirewallConfigurationBinding %s is not linked to target %s/%s %s/%s: garbage collecting",
+		req.String(), binding.Spec.TargetRef.APIVersion, binding.Spec.TargetRef.Kind,
+		binding.Spec.TargetRef.Namespace, binding.Spec.TargetRef.Name)
```
fra98: NIT: maybe let's create a String method on TargetRef for easier printing

cmd/gateway/main.go
fra98, line 1: Question: is there a reason why on the gateway we don't call `CleanupFirewallConfigurationBindings()` like we do in the fabric and we rely only on the controller-manager garbage collection?

.devcontainer/devcontainer-lock.json
cheina97: nope

cmd/gateway/main.go
cheina97: Because if you reach that point it means that the gateway will be destroyed and cleaning is useless

pkg/firewall/firewallconfigurationbinding_finalizer.go
cheina97: It's a common pattern we already use in other code places. It's just for common style, but I agree is a little bit too much

pkg/liqo-controller-manager/networking/firewall-binding-gc/bindinggc_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	bindingList := &networkingv1beta1.FirewallConfigurationBindingList{}
+	if err := r.List(ctx, bindingList); err != nil {
+		klog.Errorf("Failed to list FirewallConfigurationBindings for target %s/%s: %v",
+			obj.GetNamespace(), obj.GetName(), err)
+		return nil
+	}
+
+	var requests []ctrl.Request
+	for i := range bindingList.Items {
+		ref := bindingList.Items[i].Spec.TargetRef
```
fra98: NIT: ranging over index doesn't increase performance if we copy the object in the loop. I'd either use the pointer (`&bindingList.Items[i].Spec.TargetRef` or loop directly with the copy (`for _, ref := range ...`)

pkg/liqo-controller-manager/networking/firewall/bindinggc_controller.go
fra98: @cheina97 still not convinced on the filtering by deleting predicate as the controller still have to watch all pods and nodes in the cluster. Does removing watch introduce unreliability?

fra98 (approved):

PR #3291, opened 2026-05-15
add ECMP support for <remote-cluster-id>-gw-ext RouteConfiguration

MircoBarone:
# Description
Part of the multi-tunnel WireGuard implementation. This is the 5th PR related to issue #3225.

This PR extends the semantics of the `RouteConfiguration` CR by introducing a new `NextHop` struct used to specify the hops of a multipath route. The struct contains a gateway IP and an associated weight. The weight indicates how much traffic should pass through that hop.
A new field, `NextHops []NextHop`, has been added to the `Route` struct as an alternative to the existing `Gw` field. This allows configuring multiple paths for the same destination.

The `pkg/route/route.go` file has been updated, specifically:
* `forgeNetlinkRoute` now translates routes including next-hops using the `netlink.NexthopInfo` struct.
* `IsEqualRoute` has been enhanced to handle deep comparison for multipath routes.

This is the first step. The next step involves deciding how and when to update the CR.

### DIFF
--- a/apis/networking/v1beta1/routeconfiguration_types.go
+++ b/apis/networking/v1beta1/routeconfiguration_types.go
@@ -56,6 +56,9 @@ type NextHop struct {
 	// Gw is the gateway (next hop) IP address.
 	// +kubebuilder:validation:Required
 	Gw IP `json:"gw"`
+	// Dev is the local network interface through which this next-hop is reachable.
+	// +kubebuilder:validation:Required
+	Dev string `json:"dev"`
 	// Weight is the weight of this next-hop for load balancing.
 	// Lower weight means more traffic.
 	// Default is 0, which is translated to a kernel weight of 1.
--- a/apis/networking/v1beta1/routeconfiguration_types.go
+++ b/apis/networking/v1beta1/routeconfiguration_types.go
@@ -51,7 +51,22 @@ const (
 	NowhereScope Scope = "nowhere"
 )
 
+// NextHop is a single next-hop in a multipath route (ECMP).
+type NextHop struct {
+	// Gw is the gateway (next hop) IP address.
+	// +kubebuilder:validation:Required
+	Gw IP `json:"gw"`
+	// Weight is the weight of this next-hop for load balancing.
+	// Lower weight means more traffic.
+	// Default is 0, which is translated to a kernel weight of 1.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=255
+	// +kubebuilder:default=0
+	Weight *int `json:"weight,omitempty"`
+}
+
 // Route is the route of the RouteConfiguration.
+// +kubebuilder:validation:XValidation:rule="!(has(self.gw) && has(self.nextHops))",message="cannot specify both gw and nextHops"
 type Route struct {
 	// Dst is the destination of the RouteConfiguration.
 	Dst *CIDR `json:"dst"`
@@ -59,9 +74,13 @@ type Route struct {
 	Src *IP `json:"src,omitempty"`
 	// Gw is the gateway of the RouteConfiguration.
 	Gw *IP `json:"gw,omitempty"`
+	// NextHops is the list of next-hops for ECMP (Equal-Cost Multi-Path) routing.
+	// Mutually exclusive with Gw.
+	// +optional
+	NextHops []NextHop `json:"nextHops,omitempty"`
 	// Dev is the device of the RouteConfiguration.
 	Dev *string `json:"dev,omitempty"`
-	// Onlink enables the onlink falg inside the route.
+	// Onlink enables the onlink flag inside the route.
 	Onlink *bool `json:"onlink,omitempty"`
 	// Scope is the scope of the RouteConfiguration.
 	// +kubebuilder:validation:Enum=global;link;host;site;nowhere
--- a/apis/networking/v1beta1/routeconfiguration_types.go
+++ b/apis/networking/v1beta1/routeconfiguration_types.go
@@ -51,7 +51,25 @@ const (
 	NowhereScope Scope = "nowhere"
 )
 
+// NextHop is a single next-hop in a multipath route (ECMP).
+type NextHop struct {
+	// Gw is the gateway (next hop) IP address.
+	// +kubebuilder:validation:Required
+	Gw IP `json:"gw"`
+	// Dev is the local network interface through which this next-hop is reachable.
+	// +kubebuilder:validation:Required
+	Dev string `json:"dev"`
+	// Weight is the weight of this next-hop for load balancing.
+	// Higher weight means more traffic.
+	// Default is 0, which is translated to a kernel weight of 1.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=255
+	// +kubebuilder:default=0
+	Weight *int `json:"weight,omitempty"`
+}
+
 // Route is the route of the RouteConfiguration.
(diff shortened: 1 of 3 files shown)


pkg/route/route.go
```diff
@@ -97,6 +97,7 @@ func ExistsRoute(route *networkingv1beta1.Route, tableID uint32) (*netlink.Route
 
 // IsEqualRoute checks if the two routes are equal.
 func IsEqualRoute(route1, route2 *netlink.Route) bool {
+
```
fra98: NIT: left new line

MircoBarone: Solved

pkg/route/route.go
```diff
@@ -123,6 +123,33 @@ func IsEqualRoute(route1, route2 *netlink.Route) bool {
...
+			if !found {
+				return false
+			}
+		}
+		return true
+	}
+
+	if (multipath1 > 0) != (multipath2 > 0) {
+		return false
+	}
```
claudiolor: No need to have two ifs:
```suggestion
	multipathLen1 := len(route1.MultiPath)
	multipathLen2 := len(route2.MultiPath)
	if multipathLen1 > 0 || multipathLen2 > 0 {
		if multipathLen1 != multipathLen2 {
			return false
		}

		for _, nh1 := range route1.MultiPath {
			found := false
			for _, nh2 := range route2.MultiPath {
				if nh1.Gw.String() == nh2.Gw.String() && nh1.Hops == nh2.Hops && nh1.LinkIndex == nh2.LinkIndex {
					found = true
					break
				}
			}
			if !found {
				return false
			}
		}
		return true
	}
```

pkg/route/route.go
```diff
@@ -219,10 +246,40 @@ func forgeNetlinkRoute(route *networkingv1beta1.Route, tableID uint32) (*netlink
...
+		// All next-hop specific information is contained within the MultiPath slice.
+		gw = nil
+		linkIndex = 0
+		multiPath = make([]*netlink.NexthopInfo, len(route.NextHops))
+		for i, nh := range route.NextHops {
+			nextHopGw := net.ParseIP(nh.Gw.String())
+			weight := 0
+			if nh.Weight != nil {
+				weight = *nh.Weight
+			}
```
claudiolor:
```suggestion
			weight := ptr.Defer(nh.Weight, 0)
```

pkg/route/route.go
```diff
@@ -219,10 +246,40 @@ func forgeNetlinkRoute(route *networkingv1beta1.Route, tableID uint32) (*netlink
...
+
+			link, err := netlink.LinkByName(nh.Dev)
+			if err != nil {
+				if errors.As(err, &netlink.LinkNotFoundError{}) {
+					return nil, fmt.Errorf("link %s not found: %w", nh.Dev, err)
+				}
+				return nil, fmt.Errorf("getting link %s: %w", nh.Dev, err)
+			}
```
claudiolor: Since we are doing this thing multiple times just to get the index, we can define a function to do it:

```
func getLinkIDByName(name str)(int, error) {
    link, err := netlink.LinkByName(name)
    if err != nil {
        if errors.As(err, &netlink.LinkNotFoundError{}) {
            return nil, fmt.Errorf("link %s not found: %w", name, err)
        }
        return nil, fmt.Errorf("getting link with name %q: %w", name, err)
    }
    return link.Attrs().Index
}
```

Then we can do
```suggestion
			linkID, err := getLinkIDByName(nh.Dev)
            if err != nil {
                return fmt.Errorf("getting link for nethop %d: %w", i, err)
            }
```

pkg/route/route.go
```diff
@@ -123,6 +123,33 @@ func IsEqualRoute(route1, route2 *netlink.Route) bool {
...
+	multipath2 := len(route2.MultiPath)
+	if multipath1 > 0 && multipath2 > 0 {
+		if multipath1 != multipath2 {
+			return false
+		}
+
+		for _, nh1 := range route1.MultiPath {
+			found := false
+			for _, nh2 := range route2.MultiPath {
+				if nh1.Gw.String() == nh2.Gw.String() && nh1.Hops == nh2.Hops && nh1.LinkIndex == nh2.LinkIndex {
```
claudiolor: I think here we can use the library function:
```suggestion
				if nh1.Equal(nh2) {
```

pkg/route/route.go
```diff
@@ -123,6 +123,33 @@ func IsEqualRoute(route1, route2 *netlink.Route) bool {
 	if route1.Flags != route2.Flags {
 		return false
 	}
+
+	multipath1 := len(route1.MultiPath)
```
claudiolor: Let's recall in the name it is a length
```suggestion
	multipathLen1 := len(route1.MultiPath)
```

apis/networking/v1beta1/routeconfiguration_types.go
```diff
@@ -51,17 +51,39 @@ const (
...
+// NextHop is a single next-hop in a multipath route (ECMP).
+type NextHop struct {
+	// Gw is the gateway (next hop) IP address.
+	// +kubebuilder:validation:Required
+	Gw IP `json:"gw"`
+	// Dev is the local network interface through which this next-hop is reachable.
+	// +kubebuilder:validation:Required
+	Dev string `json:"dev"`
+	// Weight is the weight of this next-hop for load balancing.
+	// Lower weight means more traffic.
```
fra98: Doesn't higher weight means more traffic in ECMP?

```suggestion
	// Higher weight means more traffic.
```

apis/networking/v1beta1/routeconfiguration_types.go
```diff
@@ -51,17 +51,39 @@ const (
...
+	// Default is 0, which is translated to a kernel weight of 1.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=255
+	// +kubebuilder:default=0
+	Weight *int `json:"weight,omitempty"`
+}
+
 // Route is the route of the RouteConfiguration.
+// +kubebuilder:validation:XValidation:rule="!(has(self.gw) && has(self.nextHops))",message="cannot specify both gw and nextHops"
```
fra98, line 72: This validation is fine. What if user doesn't set neither gw or nextHops? Are there cases where this makes sense? If not, maybe we can add XValidation for it as well

pkg/route/route.go
```diff
@@ -123,6 +124,25 @@ func IsEqualRoute(route1, route2 *netlink.Route) bool {
...
+				if nh1.Equal(*nh2) {
+					found = true
+					break
+				}
+			}
+			if !found {
+				return false
+			}
+		}
+	}
```
fra98, line 153: IsEqualRoute compares multipath routes with nested loops and nh1.Equal(*nh2). Because it does not track already-matched next-hops, duplicate next-hops in one route can match the same next-hop in the other route, producing false positives when the two routes actually differ.

What about using a deterministic comparison? such as sorting copies of both MultiPath slices (with a stable key of gateway, link index, and weight) before comparing element-by-element, or count next-hop occurrences in a map.

fra98 (approved):

apis/networking/v1beta1/routeconfiguration_types.go
MircoBarone: I think it makes sense to allow neither gw nor nextHops to be set, for example when building link-scope routes where the destination is directly reachable on the interface and no explicit gateway is needed. In fact, gw was not required in the original code either.

pkg/route/route.go
MircoBarone: Fixed by sorting copies of both slices by (Gw, LinkIndex, Hops) before a positional comparison.

fra98 (approved):

PR #3292, opened 2026-05-15
fix(vk): create node with VirtualNode and ForeignCluster info

claudiolor:
# Description 

This PR fixes an issue with the Virtual Kubelet. When the node is created, only the VirtualNode labels are applied but the status is set to Unknown. This causes a temporary transition of the status to unknown when the VK restarts. We now make sure that when the status is created both the VirtualNode and the ForeignCluster are taken into account. Additionally it refactors the code to simplify it and  avoid code duplications.

### DIFF
--- a/cmd/virtual-kubelet/root/root.go
+++ b/cmd/virtual-kubelet/root/root.go
@@ -258,32 +258,32 @@ func runRootCommand(ctx context.Context, c *Opts) error {
 		metrics.SetupMetricHandler(c.MetricsAddress)
 	}
 
-	// Initialize the node provider
-	nodecfg := nodeprovider.InitConfig{
-		HomeConfig:      localConfig,
-		RemoteConfig:    remoteConfig,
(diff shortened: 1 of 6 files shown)


cmd/virtual-kubelet/root/root.go
```diff
@@ -258,32 +258,33 @@ func runRootCommand(ctx context.Context, c *Opts) error {
...
-
-		InformerResyncPeriod: c.InformerResyncPeriod,
-		PingDisabled:         c.NodePingInterval == 0,
-		CheckNetworkStatus:   c.NodeCheckNetwork,
-
-		VirtualNode: vn,
-	}
-
 	var nodeReady chan struct{}
 	var nodeRunner *node.NodeController
```
fra98: NIT: let's move these also under `if c.createNode`

fra98 (approved):

PR #3294, opened 2026-05-19
[Core] Add support for multiple Pod CIDRs

claudiolor:
# Description

This PR adds full support for multiple Pod CIDRs in the Liqo codebase. While the Configuration CRD has previously allowed the definition of multiple Pod CIDRs, only the first CIDR was actually handled, and installations could only specify a single Pod CIDR. 

**Key changes:**

- **Installation interface update:**  
  - The single `podCIDR` field is now deprecated.  
  - Introduced a new `podCIDRs` field (list) in its place.
- **Core Liqo enhancements:**  
  - The codebase now fully supports multiple Pod CIDRs.
- **Configuration CR changes:**  
  - The Configuration resource now allows dynamic changes to both Pod and External CIDRs.
  - Introduced a new `ObservedGeneration` field in `ConfigurationStatus` to enhance status reconciliation tracking, replacing the previous mechanism that relied on the configuration.liqo.io/configured label. With this improvement, we can now reliably determine whether a Configuration resource has been reconciled: after a change, if metadata.generation does not match status.observedGeneration, it indicates that the new configuration has not yet been fully applied. This addresses the limitation of the previous approach, where the presence of the label did not guarantee that the latest configuration changes had been processed.


deployments/liqo/README.md
```diff
@@ -66,7 +66,8 @@
 | ipam.internal.syncGracePeriod | string | `"30s"` |  |
 | ipam.internal.syncInterval | string | `"2m"` | Set the interval at which the IPAM pod will synchronize it's in-memory status with the local cluster. If you want to disable the synchronization, set the interval to 0. |
 | ipam.internalCIDR | string | `""` | The subnet used for the internal CIDR. These IPs are assigned to the Liqo internal-network interfaces. If empty, a free network will be automatically allocated by the IPAM. If set, the IPAM will try to allocate the exact network, failing in case of conflicts. Set it only if you know what you are doing.  |
-| ipam.podCIDR | string | `""` | The subnet used by the pods in your cluster, in CIDR notation (e.g., 10.0.0.0/16). |
+| ipam.podCIDR | string | `""` | The subnet used by the pods in your cluster, in CIDR notation (e.g., 10.0.0.0/16). Deprecated: used as fallback if podCIDRs is empty. |
+| ipam.podCIDRs | list | `[]` | The subnets used by the pods in your cluster, in CIDR notation. |
```
frisso:
```suggestion
| ipam.podCIDRs | list | `[]` | The IP subnets used by the pods in your cluster, in CIDR notation (e.g., "10.0.0.0/16", "10.1.0.0/16"). |
```

deployments/liqo/templates/_helpers.tpl
fra98, line 1: Extra new line?

deployments/liqo/templates/liqo-ipam-networks.yaml
```diff
@@ -1,6 +1,23 @@
...
+{{- range $cidr := .Values.ipam.podCIDRs }}
+---
+apiVersion: ipam.liqo.io/v1alpha1
+kind: Network
+metadata:
+  name: pod-cidr-{{ $cidr | replace "/" "-" | replace "." "-" }}
+  labels:
+    {{- include "liqo.labels" $ipamConfig | nindent 4 }}
+    ipam.liqo.io/network-type: pod-cidr
+    ipam.liqo.io/network-not-remapped: "true"
```
fra98, line 14: network-shared label is missing

pkg/liqo-controller-manager/ipmapping/configuration_controller.go
```diff
@@ -51,10 +49,10 @@ type ConfigurationReconciler struct {
 }
 
 func forgeUnknownSourceIPName(cfg *networkingv1beta1.Configuration) string {
-	return cfg.Name + "-unknown-source"
+	return fmt.Sprintf("%s-unknown-source", cfg.Name)
```
fra98: NIT: why? Isn't string concatenation more recommended by linters?

pkg/liqo-controller-manager/ipmapping/configuration_controller.go
```diff
@@ -75,11 +73,15 @@ func (r *ConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 	klog.V(4).Infof("Reconciling configuration %q", req.NamespacedName)
 
-	extCIDR := cidr.GetPrimary(cfg.Status.Remote.CIDR.External)
-	remoteUnknownSourceIP, err := ipamutils.GetUnknownSourceIP(extCIDR.String())
+	remoteUnknownSourceIP, err := ipamutils.GetUnknownSourceIP(cfg.Status.Remote.CIDR.External[0].String())
```
fra98: This can panic if external is nil. I know that shouldn't happen becasuse we have the predicate filter, but seems unsafe and risky in case of future refatoring. Also previous utility (cidr.GetPrimary() didn't protect from this).
I would add explicit check for this

pkg/liqo-controller-manager/ipmapping/configuration_controller.go
```diff
@@ -123,3 +117,28 @@ func (r *ConfigurationReconciler) createOrUpdateUnknownSourceIPResource(ctx cont
...
+	}
+	if err == nil && unknownSourceIP.Spec.IP == networkingv1beta1.IP(remoteUnknownSourceIP) {
+		// The current IP is correct, no need to delete any orphan.
+		return nil
+	}
+
+	err = r.Client.Delete(ctx, unknownSourceIP)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("unable to delete orphan IP %q: %w", unknownSourceIP.Name, err)
+	}
```
fra98, line 148: NIT: you can compact this with something like:
if err := ... ; client.IgnoreNotFound(err) != nil {}

pkg/liqo-controller-manager/ipmapping/configuration_controller.go
```diff
@@ -123,3 +117,28 @@ func (r *ConfigurationReconciler) createOrUpdateUnknownSourceIPResource(ctx cont
...
+
+	unknownSourceIP := &ipamv1alpha1.IP{}
+	err := r.Client.Get(ctx, client.ObjectKey{Namespace: cfg.Namespace, Name: unknownSourceIPName}, unknownSourceIP)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("unable to get the IP %q: %w", unknownSourceIPName, err)
+	}
+	if err == nil && unknownSourceIP.Spec.IP == networkingv1beta1.IP(remoteUnknownSourceIP) {
+		// The current IP is correct, no need to delete any orphan.
+		return nil
+	}
```
fra98: you should also handle notfound here so we avoid making Delete call API later

pkg/liqo-controller-manager/networking/external-network/configuration/configuration_controller.go
```diff
@@ -84,25 +87,47 @@ func (r *ConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		configuration.Status.ObservedGeneration = configuration.Generation
 	}
 
-	if !isConfigurationConfigured(configuration) {
-		events.Event(r.EventsRecorder, configuration, "Waiting for all networks to be ready")
-	} else {
-		events.Event(r.EventsRecorder, configuration, "Configuration remapped")
-		if err := SetConfigurationConfigured(ctx, r.Client, configuration); err != nil {
-			return ctrl.Result{}, fmt.Errorf("unable to set configuration %q as configured: %w", req.NamespacedName, err)
+	if !equality.Semantic.DeepEqual(originalCfg, configuration) {
```
fra98: Shouldn't you compare only status?

pkg/liqo-controller-manager/networking/external-network/configuration/configuration_controller.go
```diff
@@ -84,25 +87,47 @@ func (r *ConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	if cfg.Status.Remote == nil {
+		return false
+	}
+	if len(cfg.Status.Remote.CIDR.Pod) != len(cfg.Spec.Remote.CIDR.Pod) ||
+		len(cfg.Status.Remote.CIDR.External) != len(cfg.Spec.Remote.CIDR.External) {
+		return false
+	}
+	return cidrutils.AllNonVoid(cfg.Status.Remote.CIDR.Pod) &&
+		cidrutils.AllNonVoid(cfg.Status.Remote.CIDR.External)
+}
```
fra98: Seems duplicate of what we do at some point in IsConfigurationObserved(). I think IsConfigurationObserved should call this function internally

pkg/liqo-controller-manager/networking/external-network/configuration/configuration_controller.go
```diff
@@ -84,25 +87,47 @@ func (r *ConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, err
 	}
 
-	if err := r.UpdateConfigurationStatus(ctx, configuration); err != nil {
-		return ctrl.Result{}, err
+	// Stamp ObservedGeneration only when the remap is fully complete; otherwise leave it at the
+	// previous value so downstream consumers see "still stale" until reconcile catches up.
+	if isRemapComplete(configuration) {
+		configuration.Status.ObservedGeneration = configuration.Generation
 	}
```
fra98: Are you sure this is the right way of using ObservedGeneration pattern? technically it should tell me whenever the controller has actually reconciled the given generation or not. Here we have successfully reconciled but other external conditions are still not met. If there is an error in IPAM and networks are not filled for some reasons, here I would see that the generation and observerdGeneration are not in sync, that would tell me this controller hasn;t reconciled succesfully yet. But that's far from truth, it has reconciled spec fully but remap is not complete yet

pkg/liqo-controller-manager/networking/external-network/configuration/network.go
```diff
@@ -31,75 +34,92 @@ import (
...
+		network.Labels = netLabels
+		network.Spec.CIDR = cidr
+		return ctrlutil.SetControllerReference(cfg, network, scheme)
+	})
 	if err != nil {
-		return err
+		return nil, err
 	}
-	return nil
+	if op != "" {
```
fra98:
```suggestion
	if op != controllerutil.OperationNone {
```

pkg/liqo-controller-manager/networking/external-network/remapping/configuration_controller.go
```diff
@@ -76,14 +74,14 @@ func (r *RemappingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 	klog.V(4).Infof("Reconciling configuration %q", req.NamespacedName)
 
-	if cidrutils.GetPrimary(conf.Spec.Remote.CIDR.Pod) != cidrutils.GetPrimary(conf.Status.Remote.CIDR.Pod) {
+	if !cidrutils.EqualOrdered(conf.Spec.Remote.CIDR.Pod, conf.Status.Remote.CIDR.Pod) {
```
fra98: Unrelated to this PR and maybe we don't have to fix it now, but why do we check for configuration changes instead of always creating or updating the FirewallCfg? Seems anti-pattern

pkg/liqoctl/install/validation.go
```diff
@@ -149,26 +149,56 @@ func (o *Options) validateAPIServerConsistency() error {
...
+		}
+
+		otherNets, err := parseCIDRs(cidrList)
+		if err != nil {
+			return fmt.Errorf("parsing %s CIDRs: %w", name, err)
+		}
+
+		// Validate no overlaps between the Pod CIDRs and the other CIDR list.
+		if err := validateNoPairwiseOverlap(podNets, otherNets); err != nil {
+			return fmt.Errorf("validating overlaps between Pod CIDRs and %s CIDRs: %w", name, err)
```
fra98, line 199: With new IPAM reserved cidrs can overlaps with podCIDR. E.g.: I ban all 10.0.0.0/8 and my pod CIDR is 10.0.0.0/16. We should allow this

pkg/liqoctl/install/handler.go
```diff
@@ -65,9 +65,10 @@ type Provider interface {
 type Options struct {
 	*CommonOptions
 
-	APIServer   string
-	PodCIDR     string
-	ServiceCIDR string
+	APIServer    string
+	PodCIDRs     []string
+	ServiceCIDR  string
+	ExternalCIDR string
```
fra98, line 71: Not sure about this too after IPAM refactoring. ExternalCIDR now arrives at "", it is defaulted by helm when applying the Network. If it arrives != "" it means user explicitely wants this externalCIDR to be set and not remapped by liqo. I think we can skip this check and leave responsibility to user if it explicitely sets externalCIDR

pkg/liqoctl/install/validation.go
```diff
@@ -233,3 +263,56 @@ func defaultPort(port string) string {
...
+	for i := range nets {
+		for j := i + 1; j < len(nets); j++ {
+			if cidrsOverlap(nets[i], nets[j]) {
+				return fmt.Errorf("CIDR %q overlaps with CIDR %q", nets[i].String(), nets[j].String())
+			}
+		}
+	}
+
+	return nil
+}
```
fra98, line 291: seems overengineerd, in ipam pools check we just do something like this:

```go
	for i := range roots {
		for j := i + 1; j < len(roots); j++ {
			if roots[i].Overlaps(roots[j]) {
				return fmt.Errorf("pools %s and %s overlap", roots[i], roots[j])
			}
		}
	}
```

pkg/utils/cidr/cidr.go
```diff
@@ -29,10 +35,73 @@ func SetPrimary(cidr networkingv1beta1.CIDR) []networkingv1beta1.CIDR {
...
+	if len(a) != len(b) {
+		return false
+	}
+	as := Strings(a)
+	bs := Strings(b)
+	sort.Strings(as)
+	sort.Strings(bs)
+
+	return slices.Equal(as, bs)
 }
```
fra98: Unused?

pkg/liqoctl/install/validation.go
claudiolor: It should be equivalent

pkg/liqoctl/install/handler.go
claudiolor: Which check are we talking about?

frisso (approved):

pkg/liqo-controller-manager/networking/external-network/remapping/configuration_controller.go
```diff
@@ -73,19 +72,13 @@ func (r *RemappingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		return ctrl.Result{}, err
 	}
 
-	if !cidrutils.EqualOrdered(conf.Spec.Remote.CIDR.External, conf.Status.Remote.CIDR.External) {
-		if err := CreateOrUpdateNatMappingCIDR(ctx, r.Client, r.Options, conf,
-			r.Scheme, ExternalCIDR); err != nil {
-			return ctrl.Result{}, err
-		}
+	if err := CreateOrUpdateNatMappingCIDR(ctx, r.Client, r.Options, conf, r.Scheme, ExternalCIDR); err != nil {
+		return ctrl.Result{}, err
```
fra98, line 81: Can you wrap errors?

fra98 (approved):

cheina97 (approved):

PR #3311, opened 2026-06-16
fix(networking): list LoadBalancer IPs before hostnames in GatewayServer status

### LINKED ISSUE #3310: Gateway client uses non-resolvable LoadBalancer hostname when cloud CCM sets both hostname and IP (prefer IP in forgeEndpointStatusLoadBalancer)
### Summary
When a Liqo Gateway Server is exposed as a LoadBalancer, some cloud providers (e.g. Vultr) populate `service.status.loadBalancer.ingress` with both a hostname and an IP. Liqo currently puts the hostname first in `GatewayServer.status.endpoint.addresses`, and the WireGuard client uses only `Addresses[0]`. If that hostname is not publicly resolvable, peering fails with repeated DNS lookup errors and the Connection resource never reaches Established.

This breaks `liqoctl peer --gw-server-service-type LoadBalancer` on affected providers, even though the LoadBalancer external IP is valid and UDP forwarding works.

### Environment

- Liqo version: v1.1.2 (also reproducible on current master)
- Cloud provider: Vultr VKE (also potentially other CCMs that set internal/non-public hostnames)
- Peering command:

```
liqoctl peer \
  --remote-kubeconfig <provider-kubeconfig> \
  --gw-server-service-location Provider \
  --gw-server-service-type LoadBalancer
```

### Steps to reproduce

1. Install Liqo on a Vultr VKE cluster (provider) and a consumer cluster.
2. Run liqoctl peer with --gw-server-service-type LoadBalancer and gateway server on the provider.
3. Wait until the gateway server Service gets an external IP (LB provisioning succeeds).
4. Observe WireGuard client logs on the consumer cluster.

### Actual behavior
- Gateway server Service is created and gets an external IP, e.g. 136.244.102.213.
- Vultr CCM also sets an ingress hostname, e.g. a9e6aca6ba67b4c9c8b770fb78451ea3 (internal, not resolvable from the public internet).
- GatewayServer.status.endpoint.addresses becomes something like:
```
addresses:
  - a9e6aca6ba67b4c9c8b770fb78451ea3   # hostname first
  - 136.244.102.213                     # IP second
```

- WireGuard client is configured with Addresses[0] (the hostname).
- Client logs show repeated errors like:
`DNS "a9e6aca6ba67b4c9c8b770fb78451ea3" not found
`
- `Connection` stays in `Error` / `Connecting`; peering times out.

### Expected behavior

When both hostname and IP are present in LoadBalancer ingress, Liqo should prefer a directly usable endpoint for WireGuard,  ideally the IP first, or the client should select the first parseable IP from `addresses`.

Peering with LoadBalancer should succeed without requiring manual `GatewayClient` patches or `--gw-client-address`.

sinadarbouy:
Fixes #3310

## Problem

Some cloud CCMs populate `service.status.loadBalancer.ingress` with both a hostname and an IP.
Liqo previously listed hostnames before IPs in `GatewayServer.status.endpoint.addresses`.
The WireGuard gateway client uses `addresses[0]`, so peering can fail when that hostname is not publicly resolvable, even though the LoadBalancer external IP is valid.

## Solution

- Add `getters.CollectLoadBalancerAddresses()` to collect ingress addresses with IPs listed before hostnames for each entry.
- Use it in `forgeEndpointStatusLoadBalancer` when building the gateway server endpoint status.
- Reuse the same helper in `retrieveIPFromService` for LoadBalancer services.

## Testing

- [x] Unit tests for `CollectLoadBalancerAddresses` (IP-only, hostname-only, both fields, multiple ingress entries)
- [x] `go test ./pkg/utils/getters/...`
- [x] Manual verification: `liqoctl peer --gw-server-service-type LoadBalancer` with a CCM that sets both hostname and IP; Connection reaches `Connected` and WireGuard handshake succeeds using the external IP.

### DIFF
--- a/pkg/liqo-controller-manager/networking/external-network/wireguard/wggatewayserver_controller.go
+++ b/pkg/liqo-controller-manager/networking/external-network/wireguard/wggatewayserver_controller.go
@@ -41,6 +41,7 @@ import (
 	"github.com/liqotech/liqo/pkg/gateway/forge"
 	enutils "github.com/liqotech/liqo/pkg/liqo-controller-manager/networking/external-network/utils"
 	"github.com/liqotech/liqo/pkg/utils"
+	"github.com/liqotech/liqo/pkg/utils/getters"
 	mapsutil "github.com/liqotech/liqo/pkg/utils/maps"
 	"github.com/liqotech/liqo/pkg/utils/resource"
 )
@@ -420,15 +421,7 @@ func (r *WgGatewayServerReconciler) forgeEndpointStatusLoadBalancer(service *cor
 	port := service.Spec.Ports[0].Port
 	protocol := &service.Spec.Ports[0].Protocol
 
-	var addresses []string
-	for i := range service.Status.LoadBalancer.Ingress {
-		if hostName := service.Status.LoadBalancer.Ingress[i].Hostname; hostName != "" {
-			addresses = append(addresses, hostName)
-		}
-		if ip := service.Status.LoadBalancer.Ingress[i].IP; ip != "" {
-			addresses = append(addresses, ip)
-		}
-	}
+	addresses := getters.CollectLoadBalancerAddresses(service.Status.LoadBalancer.Ingress)
 
 	return &networkingv1beta1.EndpointStatus{
 		Protocol:  protocol,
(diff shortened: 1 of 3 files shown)


sinadarbouy:
Unit tests and component builds pass. Both kubeadm cilium E2E variants failed in liqops run [27782970784](https://github.com/liqotech/liqops/actions/runs/27782970784). E2E peering uses NodePort, so this change shouldn't affect it. Could someone check the liqops logs or re-run tests?

fra98 (approved):

fra98:
@sinadarbouy thanks for your contribution

PR #3318, opened 2026-07-07
feat: access API server from remote through EndpointSlices

fra98:
# Description

## Problem                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                         
Offloaded pods on the consumer cluster reach the provider cluster's API server via the `kubernetes` Service ClusterIP, which Liqo remaps to a local IP through a static IP resource.

This approach does not work on some Kubernetes providers (known ones are GKE dataplane v2 and EKS), due to the CNI or VPC security groups layer.                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                         
## Approach                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                         
Instead of remapping the Service ClusterIP, watch the Kubernetes Service's EndpointSlices in the consumer's default namespace, and create one IP resource per backing address. Each of those IPs is marked with Masquerade: true so that, when the packet leaves the consumer node, it gets SNATed to the node IP — bypassing the CNI's          
ClusterIP-based filtering that causes the drop on EKS-like providers.                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                         
To preserve redundancy across multiple endpoints, all the remapped IPs are injected as HostAlias entries pointing at kubernetes / kubernetes.default.svc, so the pod can reach the API server through any of them.                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                         
The behavior is opt-in behind a new flag, defaulting to the existing ClusterIP-based remapping so nothing changes for clusters that are not affected    

## Backwards compatibility                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                         
This mode is off by default. Existing clusters continue to use the ClusterIP remapping exactly as before. To opt in:                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                         
 ```yaml                                                                                                                                                                                                                                                                                                                                                 
   networking:                                                                                                                                                                                                                                                                                                                                           
     apiServerAccessThroughEndpointSlices: true        
```

### DIFF
--- a/deployments/liqo/values.yaml
+++ b/deployments/liqo/values.yaml
@@ -59,6 +59,10 @@ networking:
   # -- Prevents the usage of direct connections by provider clusters.
   # When enabled, the provider cluster will not route traffic directed to another provider through their direct connection.
   denyDirectConnections: false
+  # -- Access remote API server through its Kubernetes service EndpointSlices instead of its ClusterIP.
+  # This is useful when the consumer CNI or VPC security groups are preventing the access to the remote API server through its ClusterIP service; known examples are GKE dataplane v2 and EKS.
+  # When enabled, the consumer cluster will access the remote API server through the endpointslices of the Kubernetes service.
+  apiServerAccessThroughEndpointSlices: false
   # -- The port used by the geneve tunnels.
   genevePort: 6091
   # -- Set the list of resources that implement the GatewayServer
--- a/cmd/liqo-controller-manager/main.go
+++ b/cmd/liqo-controller-manager/main.go
@@ -262,8 +262,26 @@ func run(cmd *cobra.Command, _ []string) error {
 			return fmt.Errorf("unable to start the configuration reconciler: %w", err)
 		}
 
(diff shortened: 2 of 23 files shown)

### CODE REVIEW

**Review by cheina97 — APPROVED** [review]

File: deployments/liqo/values.yaml [hunk 1]
```diff
@@ -59,6 +59,10 @@ networking:
   # -- Prevents the usage of direct connections by provider clusters.
   # When enabled, the provider cluster will not route traffic directed to another provider through their direct connection.
   denyDirectConnections: false
+  # -- Access remote API server through its Kubernetes service EndpointSlices instead of its ClusterIP.
+  # This is useful when the consumer CNI is preventing the access to the remote API server through its ClusterIP service.
```
frisso:
```suggestion
  # This is useful when the consumer CNI or VPC security groups are preventing the access to the remote API server through its ClusterIP service; known examples are GKE dataplane v2 and EKS.
```

aleoli (approved):

Issue #3306, opened 2026-06-11
Allow providers to control the "default" ResourceSlice class behavior globally(cluster-wide) and per tenant


everpeace:
### Describe the problem you are having (if any)

Liqo currently ships with a built-in `ResourceSlice` class controller for the `default` class. This controller accepts incoming `ResourceSlice` requests and fills missing resource values with the configured defaults.

This is useful for the standard Liqo workflow, but it makes it difficult for provider clusters to enforce custom resource approval logic.

Even if a provider deploys a custom `ResourceSlice` class controller, a consumer can still set `spec.class: default`, or omit `spec.class` entirely, and the request will be handled by Liqo's built-in default controller instead of the provider's custom controller.

This creates a _bypass_ path for providers that need strict approval policies, quota checks, tenant-level admission workflows, or organization-specific resource governance.

In addition, providers may want different behavior of the default class for different tenants. For example, the default class may be acceptable for trusted tenants, while other tenants should be required to use a custom class that implements explicit approval logic.

### Describe the solution you would like

I would like Liqo to let provider clusters control whether the built-in `default` `ResourceSlice` class is allowed or denied.

This should be possible at two levels:

- globally, as the provider cluster default behavior
- per tenant, so providers can apply stricter rules only to selected tenants

When the effective policy denies the default class, Liqo should treat both `spec.class: default` and an omitted or empty `spec.class` as default-class requests and deny them.

The denial should be reported through the normal `ResourceSlice` status, for example with `Resources=Denied`, instead of leaving the consumer waiting until timeout.

Custom `ResourceSlice` classes should remain usable when they are allowed by the provider policy, and their resources status should continue to be handled by the corresponding custom class controller.

### Describe the user value of this feature

This feature would help provider operators enforce resource governance policies reliably.

Today, a consumer can bypass a provider's custom `ResourceSlice` class controller by selecting `class: default` or by omitting the class. This makes it difficult to require approval workflows, quota checks, or other provider-defined allocation logic.

Provider-side control over the default class would support use cases such as:

- disabling the `default` class for the whole provider cluster
- applying stricter rules only to selected tenants
- requiring custom approval logic for untrusted or external tenants
- allowing trusted tenants to keep using the default behavior
- avoiding unclear timeout behavior when a request is intentionally denied

This would make custom resource allocation policies enforceable while preserving compatibility for providers that want to keep the current default behavior.

### Describe your proposed solution

Introduce two complementary mechanisms.

[...]

PR #3319, opened 2026-07-08
feat(authentication): allow disabling the built-in default ResourceSlice class

thisiskazem:
# Description

Today the built-in `ResourceSlice` class controller on the provider accepts **every** request of the `default` class (and of the empty class, which Liqo treats as `default`). As a result a provider cannot enforce its own allocation policy: even with a custom `ResourceSlice` class controller, a consumer can bypass it by simply requesting `class: default` or omitting the class.

This PR adds a **provider-side switch to disable the built-in `default` class**:

- New Helm value **`authentication.defaultResourceSliceClassEnabled`** (default `true`), wired to a new controller-manager flag **`--default-resource-slice-class-enabled`**.
- When set to `false`, the `RemoteResourceSliceReconciler` **denies** `ResourceSlice`s of the `default` and empty (`""`) class, setting the `Resources` condition to `Denied` with a clear message (`the default ResourceSlice class has been administratively disabled on the provider`) instead of leaving the consumer to time out. It reuses the existing deny path, so there is no new status surface.
- **Custom classes are unaffected** and continue to be handled by their own controllers. No change to `Quota` creation or runtime enforcement.
- **Opt-in and fully backward-compatible** — the default preserves today's behavior.

This is the first, minimal step of the direction agreed in #3306 (make the default class disable-able); the `ResourceSliceClass` registry and per-tenant `Accept`/`Deny` policy are intended as follow-ups.

Part of #3306

cc @everpeace @claudiolor @frisso 

# How Has This Been Tested?

Tested end-to-end on **two k3s clusters (consumer + provider) running Liqo v1.2.0**, with the provider running the controller-manager image built from this branch. Verified via **both** the controller-manager flag and the Helm value (`helm upgrade … --set authentication.defaultResourceSliceClassEnabled=false`).

- [x] **Default enabled (baseline):** `default` and empty-class `ResourceSlice`s are `Accepted` — behavior unchanged.
- [x] **Disabled:** a `default`-class slice transitions to `Resources=Denied` with the clear message.
- [x] **Empty class:** a slice with `class: ""` is also `Denied` (cannot bypass by omitting the class).
- [x] **Custom class:** a slice with a custom class is left untouched (no accept/deny by the built-in controller).
- [x] **Re-enable:** toggling back re-accepts the default class (backward-compatible round-trip).

### DIFF
--- a/deployments/liqo/README.md
+++ b/deployments/liqo/README.md
@@ -10,6 +10,7 @@
 | authentication.awsConfig.region | string | `""` | AWS region where the clsuter is runnnig. |
 | authentication.awsConfig.secretAccessKey | string | `""` | SecretAccessKey for the Liqo user. |
 | authentication.awsConfig.useExistingSecret | bool | `false` | Use an existing secret to configure the AWS credentials. |
+| authentication.defaultResourceSliceClassEnabled | bool | `true` | Enable the built-in default ResourceSlice class. When set to false, the provider denies ResourceSlices that use the "default" class, so consumers must use an explicit, provider-approved ResourceSlice class. |
 | authentication.enabled | bool | `true` | Enable/Disable the authentication module. |
 | authentication.tlsCompatibilityMode | bool | `false` | Enable TLS compatibility mode for client certificates and keys. If set to true, Liqo will use widely supported algorithm (RSA) instead of Ed25519 (default) for generating private keys and CSRs. Enable this option to ensure compatibility with systems that do not yet support Ed25519 as signature algorithm. |
 | common.affinity | object | `{}` | Affinity for all liqo pods, excluding virtual kubelet, gateway and fabric pods. |
--- a/cmd/liqo-controller-manager/modules/authentication.go
+++ b/cmd/liqo-controller-manager/modules/authentication.go
@@ -64,12 +64,13 @@ func NewAuthOption(identityProvider identitymanager.IdentityProvider, namespaceM
 		TrustedCA:                opts.TrustedCA,
 		TLSCompatibilityMode:     opts.TLSCompatibilityMode,
 		SliceStatusOptions: &remoteresourceslicecontroller.SliceStatusOptions{
-			EnableStorage:             opts.EnableStorage,
-			LocalRealStorageClassName: opts.RealStorageClassName,
-			IngressClasses:            opts.IngressClasses,
-			LoadBalancerClasses:       opts.LoadBalancerClasses,
-			ClusterLabels:             opts.ClusterLabels.StringMap,
-			DefaultResourceQuantity:   opts.DefaultNodeResources.ToResourceList(),
+			EnableStorage:                    opts.EnableStorage,
+			LocalRealStorageClassName:        opts.RealStorageClassName,
+			IngressClasses:                   opts.IngressClasses,
+			LoadBalancerClasses:              opts.LoadBalancerClasses,
+			ClusterLabels:                    opts.ClusterLabels.StringMap,
+			DefaultResourceQuantity:          opts.DefaultNodeResources.ToResourceList(),
+			DefaultResourceSliceClassEnabled: opts.DefaultResourceSliceClassEnabled,
 		},
 	}
 }
--- a/deployments/liqo/templates/liqo-controller-manager-deployment.yaml
+++ b/deployments/liqo/templates/liqo-controller-manager-deployment.yaml
@@ -53,6 +53,7 @@ spec:
           - --networking-enabled={{ .Values.networking.enabled }}
           - --authentication-enabled={{ .Values.authentication.enabled }}
           - --tls-compatibility-mode={{ .Values.authentication.tlsCompatibilityMode }}
+          - --default-resource-slice-class-enabled={{ .Values.authentication.defaultResourceSliceClassEnabled }}
           - --offloading-enabled={{ .Values.offloading.enabled }}
           - --default-limits-enforcement={{ .Values.controllerManager.config.defaultLimitsEnforcement }}
           {{- $d := dict "commandName" "--default-node-resources" "dictionary" .Values.offloading.defaultNodeResources -}}
--- a/deployments/liqo/values.yaml
+++ b/deployments/liqo/values.yaml
@@ -245,6 +245,10 @@ authentication:
   # Enable this option to ensure compatibility with systems that do not yet
   # support Ed25519 as signature algorithm.
   tlsCompatibilityMode: false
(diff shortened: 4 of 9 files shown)


thisiskazem:
@claudiolor The docs linkcheck failed on a transient timeout to krew.sigs.k8s.io, unrelated to this PR
Could you re-run it, please?

deployments/liqo/README.md
```diff
@@ -10,6 +10,7 @@
 | authentication.awsConfig.region | string | `""` | AWS region where the clsuter is runnnig. |
 | authentication.awsConfig.secretAccessKey | string | `""` | SecretAccessKey for the Liqo user. |
 | authentication.awsConfig.useExistingSecret | bool | `false` | Use an existing secret to configure the AWS credentials. |
+| authentication.defaultResourceSliceClassEnabled | bool | `true` | Enable the built-in default ResourceSlice class. When set to false, the provider denies ResourceSlices that use the "default" class (or leave the class empty), so consumers must use an explicit, provider-approved ResourceSlice class. |
```
claudiolor: I believe we could define a cluster global default for the ResourceSliceClass. So when the consumer does not specify a ResourceSliceClass, then the provider fills with its default. I believe this would simplify a lot the user experience.  We might even do it in a separate PR
```suggestion
| authentication.defaultResourceSliceClassEnabled | bool | `true` | Enable the built-in default ResourceSlice class. When set to false, the provider denies ResourceSlices that use the "default" class , so consumers must use an explicit, provider-approved ResourceSlice class. |
```

claudiolor (approved):

deployments/liqo/README.md
frisso: This is a great idea, @claudiolor !
So, if I understand correctly your idea, let's define a sort of global config on the provider side such as "defaultResourceSliceClass=<string>", and if the consumer leaves the requested class it empty, this is the class that will be used on the provider side.
Is it correct?

frisso:
@thisiskazem I would also adjust the documentation to reflect the new policy (e.g., definitely here https://docs.liqo.io/en/v1.2.0/usage/resource-reservation.html#custom-resource-allocation-policies, but we may need to update also in other points)

thisiskazem:
Thanks @claudiolor and @frisso! 🙏

@claudiolor — applied your suggestion ✅ (updated `values.yaml` + regenerated the helm-docs `README.md`). Your `defaultResourceSliceClass=<string>` idea (empty class → provider's chosen default) is interesting — I'll open a follow-up PR for it.

@frisso — also refined the docs in `usage/resource-reservation.md`: the **Custom resource allocation policies**, **Limitations**, and **What Liqo does not** sections now describe disabling the default class natively, instead of using an external webhook.

deployments/liqo/README.md
everpeace: For the customizable provider class for the empty("") class, I think it's not a bad idea. But, in such a case, I think we should consider to define what happens if _the customized provider class for the empty class_ is updated by the provider admin.

- All the change impacts to _existing_ and _future_ ResoruceClass??
  - It will cause some disturbance on both the provider and the consumer sides
  - But, no need to record activated class for "" (it does not follow the descriptive principle in Kubernetes)
- No impact to _existing_, but, only affect to _future_ ResourceClass?? (I think it's preferred)
  - We should record which class was activated for the input "" class (admission webhook or similar would work for this)

aleoli (approved):

everpeace:
> Your defaultResourceSliceClass=<string> idea (empty class → provider's chosen default) is interesting — I'll open a follow-up PR for it.

I commented my idea on https://github.com/liqotech/liqo/pull/3319#discussion_r3570728051. PTAL.

deployments/liqo/README.md
frisso: I would prefer the second option.

frisso:
So, to recap, we agreed to implement two options here:
* provider can *disable* the default class (this PR)
* provider can *remap* the default class to a specific resource provider (another PR)

Do we agree?

frisso (approved):
It looks very good to me, including the updated docs.

deployments/liqo/README.md
claudiolor: Agree that the second option is preferable: we don't have to reconsider all the ResourceSlices once the provider changes the default and everything stays declarative

thisiskazem:
@claudiolor @frisso — I'd like your decision to finalize the design for the follow-up PR (remap an empty `""` class → the provider's configured `defaultResourceSliceClass`), building on the point @everpeace raised about recording which class was activated.

We're aligned on the semantics: the resolved class is recorded **once** at creation, so a later change to the provider default affects **future** ResourceSlices only — existing ones are untouched (declarative).

The open decision is **where** we record it:

- **Option A — mutating webhook on `resourceslices`:** on create, if `spec.class` is empty, set it to the configured class (@everpeace's admission-webhook suggestion). Concern: the `ResourceSlice` is created on the consumer and replicated to the provider by the crd-replicator, which re-syncs `spec` — so a provider-side `spec.class` mutation could get reverted. We'd likely need the crd-replicator to treat the resolved class as authoritative.
- **Option B — record in provider-owned `status` (or an annotation), set once by the reconciler:** avoids the spec re-sync conflict and stays provider-owned, at the cost of a small API addition (a resolved-class field) that downstream then reads.

Which option do you prefer (or a better third one)?

everpeace:
> Option A — mutating webhook on resourceslices: on create

`ResourceClass.spec.class` is an immutable field, isn't it?? Then, I'd prefer Option A. because I think the ResourceSlice class should be recognized/handled of its _spec_. The empty class is adjusted/completed by the webhook into `.spec.class`.

thisiskazem:
Good question @everpeace

I checked, and `ResourceSlice.spec.class` isn't immutable today: there's no CRD `x-kubernetes-validations` rule on it, and the validating webhook only guards the `remoteClusterID` label and the cordon annotation, not `class`.

I'm on board with Option A (spec-side handling). To get the "future-only" guarantee we agreed on, I'd pair the mutating webhook (`"" → spec.class` on create) with **making `spec.class` immutable** (`self == oldSelf`, like `Tenant.spec.clusterID` already does). One thing I'll verify: the slice is created on the consumer with `class: ""` and replicated to the provider by the crd-replicator, so we need to ensure its spec re-sync doesn't conflict with the immutability rule.

@claudiolor — ok to go with Option A on this basis?

claudiolor:
> @claudiolor — ok to go with Option A on this basis?

Agree, I'd proceed with option A.

fra98 (approved):

PR #3324, opened 2026-07-09
feat: set affinity and nodeSelector to wg gateway templates

fra98:
# Description

This PR adds helm value to set affinity and nodeSelector to wg gateway templates from Helm


anjmao (approved):

Gabbe64:
Great addition! I really like the extra scheduling control from these values.

NITPICK: could it be better to move these flags under a template-agnostic path like `networking.gateway.pod` ? 
These settings target the gateway Pod itself rather than a specific template, so keeping them in a common location would make them reusable across different templates as well.

fra98:
> Great addition! I really like the extra scheduling control from these values.
> 
> NITPICK: could it be better to move these flags under a template-agnostic path like `networking.gateway.pod` ? These settings target the gateway Pod itself rather than a specific template, so keeping them in a common location would make them reusable across different templates as well.

@Gabbe64 given the current liqo architecture, all gateway settings control the default WireGuard template only, since it is the only one supported through Helm. If you are using your own custom template, you can only create it externally from the Liqo Helm chart, so having a "shared" Helm setting right now doesn't help and could be confusing for users. Instead if we nest it under `wireguardTemplate` it's obvious that this is used only by the default template

claudiolor (approved):

PR #3325, opened 2026-07-09
[VK] Add type=virtual-kubelet label to Liqo virtual nodes

claudiolor:
# Description

The K8s autoscaler relies on the type=virtual-kubelet label to skip the validation of the provider ID. When this label is not present, a Liqo virtual node might block nodes upscaling.
This commit fixes this issue by always adding that label by default.

### DIFF
--- a/pkg/virtualKubelet/liqoNodeProvider/setup.go
+++ b/pkg/virtualKubelet/liqoNodeProvider/setup.go
@@ -39,6 +39,8 @@ const (
 	labelNodeExcludeBalancersAlpha = "alpha.service-controller.kubernetes.io/exclude-balancer"
 	roleLabelKey                   = "kubernetes.io/role"
 	roleLabelValue                 = "agent"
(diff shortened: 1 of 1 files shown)


pkg/virtualKubelet/liqoNodeProvider/setup.go
```diff
@@ -39,6 +39,7 @@ const (
 	labelNodeExcludeBalancersAlpha = "alpha.service-controller.kubernetes.io/exclude-balancer"
 	roleLabelKey                   = "kubernetes.io/role"
 	roleLabelValue                 = "agent"
+	virtualKubeletType             = "virtual-kubelet"
```
fra98, line 43: NIT: can we have a short comment on why we need `type=virtual-kubelet` label? I would also define the label key as constant since it's hardcoded

fra98 (approved):

PR #3329, opened 2026-07-13
feat: add watch for network config in shadowendpointslice controller

### LINKED ISSUE #3328: ShadowEndpointSlice improvement for direct-connections
# Problem

In direct-connection scenarios, endpoint remapping depends on Network Configuration data. If that configuration is not ready when the controller first reconciles a ShadowEndpointSlice, the controller may complete without applying the correct remapping. Before this change, the controller could miss later Configuration updates and therefore never re-trigger reconciliation for the affected ShadowEndpointSlices. The result is stale or incorrect EndpointSlice replication.

# Proposed Fix

With PR #3329, I propose updating the ShadowEndpointSlice controller to explicitly respond to Configuration lifecycle events that affect endpoint remapping.

The fix introduces a watch on network Configuration resources and enqueues the related ShadowEndpointSlices on:

* Configuration create, when network CIDRs are configured.
* Configuration update, when network CIDRs change.

To avoid unnecessary reconciliations, dedicated predicates filter events so only meaningful Configuration changes trigger processing. The controller maps each Configuration to the impacted ShadowEndpointSlices (via remote-cluster metadata/indexed relationships) and requeues them for reconciliation.

This ensures that if Configuration was missing or incomplete during the first ShadowEndpointSlice reconcile, the controller is woken up again when Configuration becomes available or changes, preventing stale EndpointSlice replication in direct-connection paths.

Franc-Zar:
# Description

This PR strengthens ShadowEndpointSlice reconciliation triggers by adding explicit handling for Network Configuration create and update events.

In direct-connection scenarios, endpoint remapping depends on Network Configuration data. If that configuration is not ready when the controller first reconciles a ShadowEndpointSlice, the controller may complete without applying the correct remapping. Before this change, the controller could miss later Configuration updates and therefore never re-trigger reconciliation for the affected ShadowEndpointSlices. The result is stale or incorrect EndpointSlice replication.

With this PR, Configuration create/update events are explicitly watched and used to enqueue the involved ShadowEndpointSlices, so the controller wakes up again as soon as a Configuration becomes available or changes. This ensures remapping is eventually applied and prevents silent drift after startup timing races or delayed network configuration.

Main changes:

* The controller now watches Configuration resources and enqueues all involved ShadowEndpointSlices when relevant events occur;
* On Configuration create, reconciliation is triggered only when network CIDRs are configured;
* On Configuration update, reconciliation is triggered only when network CIDRs actually change;
* A dedicated mapping helper retrieves impacted ShadowEndpointSlices from the RemoteClusterID and indexed direct-connection metadata, with safe handling when labels are missing.

This reduces the risk of missed events and stale EndpointSlice state after network configuration changes, improving the reliability of endpoint replication/remapping.

# How Has This Been Tested?

Added Unit tests for shadowendpointslice-controller package:
* Predicate-focused tests:
    * validateNetworkConfigOnCreate
    * validateNetworkConfigOnUpdate
    * missing RemoteClusterID handling for network config mapping
* Reconciliation behavior test:
    * verifies endpoint remapping after network configuration changes for existing ShadowEndpointSlices

Fixes #3328.

### DIFF
--- a/pkg/utils/indexer/indexer.go
+++ b/pkg/utils/indexer/indexer.go
@@ -20,11 +20,19 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	ctrlruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	offloadingv1beta1 "github.com/liqotech/liqo/apis/offloading/v1beta1"
+	"github.com/liqotech/liqo/pkg/consts"
+	"github.com/liqotech/liqo/pkg/utils/directconnection"
 )
 
 const (
 	// FieldNodeNameFromPod is the field name of the node name of a pod.
 	FieldNodeNameFromPod = "spec.nodeName"
+
+	// FieldDirectConnectionClusterIDs is the field index key used to index ShadowEndpointSlices by the
+	// clusterIDs present as keys in the direct-connections-data annotation (i.e., the clusterAddresses map).
+	FieldDirectConnectionClusterIDs = "metadata.annotations.direct-connections-data.clusterIDs"
 )
 
 // ExtractNodeName returns the node name of the given object.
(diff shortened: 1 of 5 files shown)


pkg/utils/indexer/indexer.go
```diff
@@ -37,6 +45,30 @@ func ExtractNodeName(rawObj client.Object) []string {
...
+	if err := ca.FromJSON([]byte(val)); err != nil {
+		return nil
+	}
+	clusterIDs := make([]string, 0, len(ca.Clusters))
+	for id := range ca.Clusters {
+		clusterIDs = append(clusterIDs, id)
+	}
+	return clusterIDs
+}
+
```
Gabbe64, line 64: Hello!
Everything looks good to me, I only have a nitpick change to suggest:

I would move the annotation parsing and cluster ID extraction into the `directconnection` package:
```go
func FromAnnotations(annotations map[string]string) (*ClusterAddresses, error) {...}
func (c *ClusterAddresses) ClusterIDs() []string {...}
```
This way the annotation handling stays inside its own package: the `indexer` becomes a simple adapter on top of it, and `Reconcile()` can rely on `FromAnnotations()` instead of unmarshalling the annotation itself, so the format is defined in a single place.

Gabbe64 (approved):

pkg/utils/indexer/indexer.go
Franc-Zar: Hello @Gabbe64 , thanks for the review and your suggestion.
I've implemented it in the latest commit, and it’s indeed a cleaner approach. 
If there’s anything else I can do for this PR, just let me know.

aleoli (approved):

PR #3332, opened 2026-07-15
[VIrtualKubelet] virtualkubelet net configuration as flag

cheina97:
# Virtual-Kubelet network configuration via controller-injected flags

## Summary

Stop the virtual-kubelet from querying the Kubernetes API server for `networkingv1beta1.Configuration`. Instead, the `liqo-controller-manager` fetches the network CIDR remappings and injects them as virtual-kubelet container arguments. When the `Configuration` changes, the VirtualNode controller updates the VK Deployment so a new virtual-kubelet pod is rolled out with the correct CIDRs.

## Motivation

Today, every virtual-kubelet pod needs direct read access to the `networking.liqo.io/configurations` resource to discover how remote pod and external CIDRs are remapped. Centralising this lookup in the VirtualNode controller has two benefits:

1. It reduces the RBAC surface of the virtual-kubelet.
2. It guarantees that a change in the network `Configuration` is propagated to running virtual-kubelets by re-forging the Deployment.

## Scope

- Limited to the `networkingv1beta1.Configuration` lookup in the virtual-kubelet.
- `ForeignCluster` and `VirtualNode` lookups are left unchanged.
- No new API types or CRD changes.
- Network CIDR flags are injected by the controller, **not** by the VirtualNode mutating webhook.

### Virtual-kubelet flags and internal representation

- Added four new VK flags:
  - `--remote-pod-cidr`
  - `--remote-pod-cidr-remap`
  - `--remote-external-cidr`
  - `--remote-external-cidr-remap`
- Added `buildRemoteCIDR` to assemble a new internal `*networkconfig.RemoteCIDR` value from the flags.
- Updated the provider and pod reflectors to use the internal `RemoteCIDR` struct instead of `*networkingv1beta1.Configuration`.

### Controller-side arg forging

- Added `vkforge.SetNetworkConfigurationArgs(..., cfg)` to append the four network CIDR flag groups to a VK container.
- The VirtualNode controller first fetches the `ForeignCluster` for the VirtualNode's cluster ID.
  - If the `ForeignCluster` cannot be found, the reconcile returns an error (the resource is expected to exist).
  - If `Status.Modules.Networking.Enabled` is `false`, the controller skips injecting network args.
  - Only when networking is enabled does it fetch the `Configuration` by cluster ID and pass it to the forge helper.
- If the `Configuration` is not found while networking is enabled, the reconcile returns an error.
- Added a watch on `Configuration` and `ForeignCluster` to re-enqueue affected `VirtualNodes` when the network configuration changes.

### RBAC cleanup

- Removed the `configurations` RBAC permission from the virtual-kubelet `ClusterRole` (`pkg/virtualKubelet/roles/local/role.go`).
- Removed the `configurations` RBAC permission from the VirtualNode webhook `ClusterRole`.
- Regenerated RBAC manifests (`make rbacs`).

### Tests

- Updated `cmd/virtual-kubelet/root/networkconfig_test.go` for the new flag builder.
- Added `pkg/vkMachinery/forge/networkconfig_test.go` covering arg forging.
- Updated the VirtualNode controller integration tests to create `ForeignClusters` with networking enabled and ready `Configuration` objects where needed.
- Updated `pkg/virtualKubelet/reflection/workload/pod_test.go` and `podns_test.go` to use `*networkconfig.RemoteCIDR` and the new `func() []string` Kubernetes service IP getter.

## Verification

All relevant packages build and pass tests:

```bash
go test ./cmd/virtual-kubelet/root/... ./pkg/vkMachinery/forge/... ./pkg/utils/ipam/mapping/...
KUBEBUILDER_ASSETS=/tmp/envtest-binaries/k8s/1.36.2-darwin-arm64 go test ./pkg/liqo-controller-manager/offloading/virtualnode-controller/...
go test ./pkg/virtualKubelet/reflection/workload/...
go build ./cmd/virtual-kubelet
GOOS=linux go build ./cmd/webhook
GOOS=linux go build ./cmd/liqo-controller-manager
```

## Checklist

- [x] Code compiles and tests pass.
- [x] Manifests/RBACs regenerated (`make manifests`, `make rbacs`).
- [x] Scope limited to the requested change.
- [x] No new API types or CRD changes.


pkg/vkMachinery/forge/networkconfig.go
```diff
@@ -0,0 +1,88 @@
...
+		if !strings.Contains(args[i], "=") {
+			out = append(out, args[i])
+			continue
+		}
+		key, _ := DestringifyArgument(args[i])
+		if !containsString(networkConfigurationArgKeys, key) {
+			out = append(out, args[i])
+		}
+	}
+	return out
```
fra98: NIT: we can probably use slices.DeleteFunc()

pkg/vkMachinery/forge/networkconfig.go
```diff
@@ -0,0 +1,88 @@
...
+	args = appendCIDRArgs(args, RemotePodCIDRRemap, cfg.Status.Remote.CIDR.Pod)
+	args = appendCIDRArgs(args, RemoteExternalCIDR, cfg.Spec.Remote.CIDR.External)
+	args = appendCIDRArgs(args, RemoteExternalCIDRRemap, cfg.Status.Remote.CIDR.External)
+	return args
+}
+
+// RemoveNetworkConfigurationArgs removes all network configuration args from the slice.
+func RemoveNetworkConfigurationArgs(args []string) []string {
+	return removeNetworkConfigurationArgs(args)
+}
```
fra98: this seems unused

cmd/virtual-kubelet/root/networkconfig.go
```diff
@@ -0,0 +1,62 @@
...
+type CIDRPair struct {
+	Original []string
+	Remapped []string
+}
+
+// RemoteCIDR holds the remote pod and external CIDRs, with their local remappings.
+type RemoteCIDR struct {
+	Pod      CIDRPair
+	External CIDRPair
+}
```
fra98: These are duplicates of the struct we already have of in `pkg/virtualKubelet/networkconfig`.

cmd/virtual-kubelet/root/networkconfig.go
```diff
@@ -0,0 +1,62 @@
...
+		Pod: networkconfig.CIDRPair{
+			Original: c.RemotePodCIDR,
+			Remapped: c.RemotePodCIDRRemap,
+		},
+		External: networkconfig.CIDRPair{
+			Original: c.RemoteExternalCIDR,
+			Remapped: c.RemoteExternalCIDRRemap,
+		},
+	}, nil
+}
```
fra98: NIT: if we delete the above struct only this function is left. Could we move it to opts.go? As this would be the third file called networkconfig.go in the codebase and it might be confusing why we have so many

pkg/utils/ipam/mapping/ips.go
```diff
@@ -159,6 +160,61 @@ func MapAddressWithConfiguration(cfg *networkingv1beta1.Configuration, address s
...
+	}
+
+	podAddr, ok, err := remapWithinStrings(ip, address, cidr.Pod.Original, cidr.Pod.Remapped)
+	if err != nil {
+		return "", fmt.Errorf("remapping the address within the pod CIDRs: %w", err)
+	} else if ok {
+		return podAddr, nil
+	}
+
+	extAddr, ok, err := remapWithinStrings(ip, address, cidr.External.Original, cidr.External.Remapped)
```
fra98: Can we cast Oiriginal and Reamapped to `networkingv1beta1.CIDR` and reuse remapWithin? Code is nearly duplicated apart from the type.
Alternatively we can also refactor the whole networkconfig.RemoteCIDR to use networkingv1beta1.CIDR instead of string directly

pkg/virtualKubelet/reflection/workload/pod.go
```diff
@@ -245,27 +245,28 @@ func (pr *PodReflector) Stats(ctx context.Context) (*statsv1alpha1.Summary, erro
...
 
 		// If the addresses have already been saved in cache, then return them directly.
 		if len(addresses) > 0 {
-			return addresses, nil
+			return addresses
 		}
 
 		var err error
-		addresses, err = pr.config.KubernetesServiceIPMapper(ctx)
+		addresses, err = pr.config.KubernetesServiceIPMapper(context.Background())
```
fra98: I don't get this refactoring, why removing the context and start a context.Background()?

pkg/virtualKubelet/reflection/workload/pod.go
```diff
@@ -245,27 +245,28 @@ func (pr *PodReflector) Stats(ctx context.Context) (*statsv1alpha1.Summary, erro
...
 		if err != nil {
-			return nil, err
+			klog.Warningf("Failed to retrieve kubernetes service IPs: %v", err)
+			return nil
 		}
```
fra98, line 266: Also why error handling was removed?

.kimchi/docs/pr-description.md
fra98, line 1: Just a comment to remind to remove file before merging

pkg/virtualKubelet/reflection/workload/podns.go
```diff
@@ -327,21 +327,20 @@ func (npr *NamespacedPodReflector) ForgeShadowPod(ctx context.Context, local *co
...
+	// Wrap the kubernetes service remapped IP retrieval, so that we do not have to handle errors in the forge logic.
+	ipGetter := func() []string {
+		if npr.config.RemoteCIDR == nil {
 			return nil
 		}
 
-		ips, kserr = npr.kubernetesServiceIPGetter(ctx)
-		return ips
+		return npr.kubernetesServiceIPGetter()
 	}
```
fra98: Also here let's restore error and ctx as this can fail and we should retry in such cases

pkg/virtualKubelet/reflection/workload/podns.go
```diff
@@ -327,21 +327,20 @@ func (npr *NamespacedPodReflector) ForgeShadowPod(ctx context.Context, local *co
 		// The function will never be invoked
 	}
 
-	// Wrap the kubernetes service/endpointslices remapped IP retrieval, so that we do not have to handle errors in the forge logic.
-	ipsGetter := func() (ips []string) {
-		if npr.config.NetConfiguration == nil {
+	// Wrap the kubernetes service remapped IP retrieval, so that we do not have to handle errors in the forge logic.
```
fra98: I;d bring back endpointslices as this is possible
```suggestion
	// Wrap the kubernetes service/endpointslices remapped IP retrieval, so that we do not have to handle errors in the forge logic.
```

fra98 (approved):

PR #3333, opened 2026-07-16
feat: add force uninstall

claudiolor:
# Description

This pull request introduces a "force uninstall" feature to the Liqo uninstaller, allowing users to forcibly remove all Liqo resources even if there are active peerings or leftover namespaces that would normally block a standard uninstall. This is intended for use in situations where the standard uninstall is stuck.

These changes make the Liqo uninstall process more robust in failure scenarios but introduce the risk of data loss when the force option is used. Use the force uninstall feature only when the standard uninstall process cannot succeed and you are certain that remaining resources are no longer needed.

### DIFF
--- a/cmd/uninstaller/main.go
+++ b/cmd/uninstaller/main.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/dynamic"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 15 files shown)


frisso:
This is really great, @claudiolor .
I would just add that, in our experience (which is mainly in environments dedicated to experimentations, where things are not so stable) the most critical feature here would be to have a "force unpeer" command, which removes a peering (and all the associated resources) against a cluster that is no longer reachable.
Also "force uninstall" is helpful, but the "force unpeer" would be really a key feature for us at Polito.

claudiolor:
> Also "force uninstall" is helpful, but the "force unpeer" would be really a key feature for us at Polito.

Hey @frisso, force unpeer is already supported via `liqoctl` and by annotating the ForeignCluster resource: https://docs.liqo.io/en/stable/usage/liqoctl/liqoctl_force.html#liqoctl-force-unpeer

It was introduced in #3122 and #3126.
This PR uses the same internal mechanism to support the forced Liqo uninstall.

PR #3338, opened 2026-07-24
feat: support dynamic RouteConfiguration expansion for multi-tunnel gateways

MircoBarone:
Part of the multi-tunnel WireGuard implementation (8th PR related to #3225).

When managing return traffic from WireGuard gateway interfaces to cluster pods, routing rules are defined in RouteConfiguration resources (e.g., <local-cluster-id>-<node-name>-gw-node and <node>-extcidr). These resources are created during the initial Liqo installation, before peering occurs, and are shared across all gateways.

Previously, these resources used a hardcoded incoming interface (`iif: liqo-tunnel`).

In a multi-tunnel setup:
- Each gateway instance may be configured with a different number of WireGuard interfaces.
- Statically expanding the rules in the shared RouteConfiguration resources is impractical, as it would bloat the resources and require continuous synchronization based on the maximum number of interfaces across all active gateways.

To address this, this PR introduces a dynamic placeholder mechanism (`liqo-tunnel*`) in the RouteConfiguration resources. Each gateway's RouteConfigurationBinding controller expands the placeholder locally according to its configured number of WireGuard interfaces.

### Changes
- Added the `--num-interfaces` flag (default: `1`) to the gateway container. The value is populated in both gateway client and server templates based on the number of ports.
- Added validation during gateway startup to ensure `--num-interfaces >= 1`.
- Updated `internalnode_k8s.go` and `pod_k8s.go` to generate RouteConfiguration resources using `iif: liqo-tunnel*` instead of the static `iif: liqo-tunnel`.
- Extended `RouteConfigurationBindingReconciler` to store both the tunnel interface prefix and the list of concrete tunnel interface names for the local gateway instance (e.g., `liqo-tunnel0`, `liqo-tunnel1`).
- Introduced `expandRules()` in `RouteConfigurationBindingReconciler.Reconcile()` to expand `iif: liqo-tunnel*` placeholder rules into one rule per local WireGuard interface before applying kernel routing rules and routes.

### DIFF
--- a/cmd/fabric/main.go
+++ b/cmd/fabric/main.go
@@ -216,6 +216,8 @@ func run(cmd *cobra.Command, _ []string) error {
 		mgr.GetClient(),
 		mgr.GetScheme(),
 		mgr.GetEventRecorder("route-binding-controller"),
+		"",
+		nil,
 	)
 
 	if err := rcr.SetupWithManager(cmd.Context(), mgr, routeTargetRef,
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -39,6 +39,7 @@ import (
 	"github.com/liqotech/liqo/pkg/gateway/concurrent"
 	"github.com/liqotech/liqo/pkg/gateway/connection"
 	"github.com/liqotech/liqo/pkg/gateway/connection/conncheck"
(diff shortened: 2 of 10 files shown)


MircoBarone:
Closed in favor of #3363

PR #3341, opened 2026-07-26
feat(virtual-kubelet): add opt-in custom resource reflection

### LINKED ISSUE #3330: Custom resource reflection for namespace offloading (OwnershipShared)

Liqo namespace offloading extends a namespace across clusters and reflects a fixed set of native Kubernetes resources (ConfigMaps, Secrets, Services, PVCs, etc.) through the Virtual Kubelet. It does not support user-defined Custom Resources (CRs).

This is a blocker for multi-cluster setups where:

- A consumer offloads a namespace to a provider (twin namespace).
- CRs are created or updated on the consumer.
- A controller on the provider reconciles the remote CR and writes its status.
- The consumer needs the desired `spec` propagated remotely and the resulting `status` reflected locally.

Historically, Liqo supported peering-level CR replication through the CRD replicator, but that mechanism was removed and was not integrated with namespace offloading. There is currently no supported, generic way to reflect selected namespaced GVRs across an offloaded namespace pair.

Add generic, opt-in Custom Resource reflection to the Virtual Kubelet.

Operators configure arbitrary namespaced Group/Version/Resources (GVRs) through Helm:

```yaml
offloading:
  reflection:
    customResources:
      - group: example.io
        version: v1
        resource: widgets
        workers: 2
        type: AllowList
```

The configuration is propagated to `VkOptionsTemplate`. For every configured GVR, the Virtual Kubelet starts a dynamic, unstructured reflector when a namespace is activated through the existing `NamespaceMap` lifecycle.

Reflection follows OwnershipShared semantics:

- `spec`: consumer/local → provider/remote
- `status`: provider/remote → consumer/local

Custom Resource reflection defaults to `AllowList`. Individual CR instances opt in with:

```yaml
metadata:
  annotations:
    liqo.io/allow-reflection: "true"
```

An empty `customResources` list means that Custom Resource reflection is disabled. No application-specific CRDs are hardcoded in Liqo.

- Enables operators based on Custom Resources to work across offloaded namespaces.
- Preserves clear ownership: the consumer owns desired state, while the provider reconciles it and owns status.
- Reuses the existing Virtual Kubelet reflection model and namespace lifecycle.
- Supports arbitrary namespaced CRDs without coupling Liqo to a specific application or operator.
- Keeps the feature disabled by default and requires explicit GVR and per-object opt-in.

### Describe your proposed solution

The implementation is centered on the Virtual Kubelet reflection subsystem rather than changes to `NamespaceOffloading` or new `liqoctl` flags.

#### Configuration and API

- Add a `customResources` list to `VkOptionsTemplate`.
- Expose it through the Helm value `offloading.reflection.customResources`.
- Each entry contains `group`, `version`, `resource`, `workers`, and reflection `type`.
- Do not add `customResources` to `NamespaceOffloading`.
- Do not add a `liqoctl --replicate-cr` flag.

#### Reflection

- Implement a generic reflector using dynamic Kubernetes clients and unstructured objects.
- Register one reflector for each configured GVR.
- Start and stop reflectors through the existing `NamespaceMap` / Virtual Kubelet namespace lifecycle.
- Create or update the remote object from the local desired state.
- Propagate remote status back through the local status subresource.
- Preserve the normal Liqo reflection annotations, labels, policies, and cleanup behavior.

#### RBAC

- Generate local and remote Virtual Kubelet RBAC rules from the configured GVR list.
- Grant only the verbs required for reflection and status synchronization.
- Keep the default RBAC unchanged when the list is empty.

### MVP scope and limitations

- OwnershipShared semantics only: `spec` local → remote and `status` remote → local.
- Namespaced Custom Resources only; cluster-scoped resources are not supported.
- Compatible CRDs must already exist on both clusters; Liqo does not install CRDs.
- Default policy is `AllowList` using `liqo.io/allow-reflection: "true"`.
- Existing VirtualNodes may need to be recreated or re-peered after changing the configured GVR list so that configuration and RBAC are refreshed.
- The controller for a reflected type should run on one side only to prevent competing reconciliations.
- Remote changes to `spec` are overwritten by the local desired state.
- Cross-resource references and hardcoded namespaces inside a CR are copied as-is and may require application-level handling.

### Validation

The design has been validated with bidirectional peering and selected-name namespace mappings:

- Annotated CR instances were reflected from the consumer namespace to its remote twin.
- Provider-side status updates were reflected back to the local CR.
- Unannotated CR instances were not reflected under the default AllowList policy.
- No application-specific CRD types were required in the implementation.

### Code of Conduct

sinadarbouy:
# Description

Adds optional reflection of arbitrary namespaced custom resources for offloaded namespaces.

Liqo already reflects a fixed set of built-in resources (Pods, ConfigMaps, Secrets, Services, …). Operators that run CR controllers on the provider need the same pattern for their own GVRs: **spec local → remote**, **status remote → local**, without hardcoding application CRDs in Liqo.

This PR implements that as an opt-in list on the Virtual Kubelet path:

* Configure GVRs via Helm `offloading.reflection.customResources` (rendered into `VkOptionsTemplate` and VK flags).
* Empty list keeps the feature off (default).
* Default reflection policy is **AllowList** (`liqo.io/allow-reflection: "true"`).
* CRDs must already exist on both clusters; Liqo does not install them.
* Local/remote VK RBAC is generated from the same list.
* Dynamic clients/informers are created only when at least one custom GVR is configured.

**Notable design choices (intentional):**

* Cluster-wide GVR list on the consumer install (not per-`NamespaceOffloading`).
* Spec ownership is local-only (remote spec edits are overwritten).
* Status is synced only when the CRD exposes a status subresource.
* Controllers for a reflected type should run on one side only to avoid double reconcile.

Docs: `docs/usage/reflection.md`.

Fixes #3330

# How Has This Been Tested?

- [x] Unit tests for the custom GVR reflector (`pkg/virtualKubelet/reflection/custom`)
- [x] Unit tests for `--custom-resource-reflection` flag parsing (`cmd/virtual-kubelet/root`)
- [x] Unit tests for reflection manager / provider packages touched by this change
- [x] New e2e cruise suite `test/e2e/cruise/customreflection` (AllowList spec sync, status sync, skip without annotation, deletion), with Widget CRD + `customResources` wired in the e2e installer
- [x] Manual validation on peered clusters (neo ↔ docker-desktop): annotated CR reflected with remote status written back to the consumer
- [ ] Full upstream `make e2e` / CI pipeline (to be run by GitHub Actions after images are built)

**Test notes:** e2e installer applies `widgets.example.io` before Liqo install and sets `offloading.reflection.customResources[0]` to that GVR so Virtual Kubelets register the reflector at startup.


sinadarbouy:
Hi @fra98 @cheina97, just a gentle ping on this PR when you get a chance.

I’ve rebased it onto the latest master, including the recent VK deployment changes, and would really appreciate a first look.

If you’re not the right reviewers for this, no worries; happy to take a pointer to who would be best to review it.

Thanks!

fra98:
Hi @sinadarbouy, thanks for your contribution!

CRD reflection is a hot topic and one of the most wanted features. Some design choices are not trivial, like:
- spec and status ownership
- RBACs
- CRDs installation prerequisites

Looking at the PR description alone, the default choices you made regarding the above items make sense in my opinion, but of course they will be up for debate. I'll take a further look at the code, hopefully soon

PR #3343, opened 2026-07-29
fix: rework webhook certificates

claudiolor:
# Description

- liqo-webhook makes sure that certificates are properly configured for the webhooks. It gets a secret name as input and generates the required certificates and inject them. However, the watch on the secret was cluster-scoped and based on the labels, while the main function used only the secret name. This caused some issues if a stale secret was present in a different namespace. This patch fixes this inconsistency just relying on the provided secret name.
- Additionally it fixes certificates lifecycle, making sure that the certificate bundle is regenerated before its expiration


pkg/utils/certificate/certificate.go
```diff
@@ -0,0 +1,66 @@
...
+	if time.Now().Before(twoThirdsPoint) {
+		// Calculate requeue time as the remaining time until the 2/3 point of the certificate expiration time + 10%
+		timeUntilTwoThirds := time.Until(twoThirdsPoint)
+		requeueIn = timeUntilTwoThirds * 11 / 10
+
+		klog.V(4).Infof("Certificate not ready for renewal, will check again in %v", requeueIn)
+		return false, requeueIn, nil
+	}
+
+	return true, requeueIn, nil
```
aleoli, line 57: Probably we can use the `NextRenewalCheck` function to avoid redefinition of the two-thirds time rule

aleoli (approved):

pkg/utils/certificate/certificate.go
claudiolor: fixed!

fra98 (approved):

PR #3346, opened 2026-08-03
[Networking] ping in internalnode source-detector to check if the CNI is ready

cheina97:
## Problem

When a new node is spun up, the CNI may not have finished programming the host routes yet. The `source-detector` controller runs with `hostNetwork: true` and computes the source IP used to reach a gateway pod via `netlink.RouteGet`. During the CNI setup window `RouteGet` can return a transient or wrong route, and the controller currently writes that result directly into `InternalNode.Status.NodeIP.Local/Remote`. Once written, downstream consumers treat the value as authoritative.

## Solution

Add an ICMP reachability check before trusting the route. The controller now sends a single ICMP echo request to the gateway pod IP and waits for an echo reply. If the pod is not reachable, the reconciliation is requeued after 5 seconds instead of recording a wrong source IP.

This approach is CNI-agnostic: it does not rely on CNI-specific files, annotations, or behaviours. It only requires that the destination is reachable through the routes installed by the CNI, which is true for any correctly configured CNI.

## Changes

- Added `pkg/fabric/source-detector/ping.go` with a `PingIP` helper supporting IPv4 and IPv6.
- Updated `pkg/fabric/source-detector/gateway_controller.go` to call `PingIP` before `GetSrcIPFromDstIP`.
- Added a basic unit test for invalid IP handling in `pkg/fabric/source-detector/ping_test.go`.
- Promoted `golang.org/x/net` to a direct dependency in `go.mod`.

## Verification

```bash
GOOS=linux go build ./pkg/fabric/source-detector/...
GOOS=linux go vet ./pkg/fabric/source-detector/...
GOOS=linux go test -c ./pkg/fabric/source-detector/...
```

## Notes / Limitations

- The fabric DaemonSet is `hostNetwork: true`, `privileged: true`, and has `NET_RAW`, so raw ICMP sockets are available.
- If ICMP is firewalled between nodes, the controller will keep requeueing. In such environments this check should be combined with additional route/interface readiness validation.

Special thanks to @Gabbe64 for finding this bug

### DIFF
--- a/pkg/fabric/source-detector/gateway_controller.go
+++ b/pkg/fabric/source-detector/gateway_controller.go
@@ -17,6 +17,7 @@ package sourcedetector
 import (
 	"context"
 	"fmt"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -87,6 +88,15 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	klog.V(4).Infof("Reconciling gateway pod %s", req.String())
+
+	// Verify that the gateway pod is actually reachable via ICMP before
+	// trusting the route selected by the kernel. This avoids recording a
+	// transient/wrong source IP when the CNI has not finished setting up
+	// the node routes yet.
+	if err := PingIP(pod.Status.PodIP, PingTimeout); err != nil {
+		klog.V(4).Infof("Gateway pod %s is not reachable yet: %v. Checking later", req.String(), err)
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
 
 	src, err := GetSrcIPFromDstIP(pod.Status.PodIP)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/fabric/source-detector/gateway_controller.go
```diff
@@ -88,6 +89,15 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	klog.V(4).Infof("Reconciling gateway pod %s", req.String())
 
+	// Verify that the gateway pod is actually reachable via ICMP before
+	// trusting the route selected by the kernel. This avoids recording a
+	// transient/wrong source IP when the CNI has not finished setting up
+	// the node routes yet.
+	if err := PingIP(pod.Status.PodIP, PingTimeout); err != nil {
+		klog.V(4).Infof("Gateway pod %s is not reachable yet: %v. Checking later", req.String(), err)
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
fra98, line 98: Maybe we can log with V(0), this seems quite important to check if there is any error

fra98 (approved):

PR #3347, opened 2026-08-04
fix: report latency metrics directly from the origin container

varnastadeus:
Latency metric collection improvements

 ### What changed

 Latency metrics are now recorded at measurement time (on each PONG) instead of scrape time (reading from CR status). CR status updates use a poll-and-requeue pattern instead of per-PONG callbacks.

 ### Key changes

 - pkg/conncheck/: Replaced UpdateFunc (K8s-writing callback) with Observer (metrics-only, no error). AddSender takes a per-sender observer. Merged GetLatency/GetConnected into GetStatus.
 - pkg/gateway/tunnel/metrics.go: Converted MetricsPeerLatency, MetricsPeerIsConnected, MetricsGeneveLatency, MetricsGeneveIsConnected from *prometheus.Desc to *prometheus.GaugeVec. Removed dead PrometheusMetrics/MetricsErrorHandler.
 - pkg/gateway/connection/: Removed ForgeUpdateConnectionCallback. Reconciler polls GetStatus and requeues at PingUpdateStatusInterval. ObserveLatency sets gauge + connected + histogram at PONG time. Deleted custom collector — metrics registered directly in main.go.
 - pkg/gateway/tunnel/wireguard/metrics.go: Stripped latency/connected metrics. Only byte counters remain. NewPrometheusCollector no longer takes client.Client.
 - pkg/fabric/genevetunnel_controller.go: Same poll-and-requeue pattern. observeGeneveLatency reads remote_cluster_id from InternalFabric labels.
 - pkg/gateway/tunnel/geneve/metrics.go: Deleted (163 lines). Geneve latency/connected metrics now live in the fabric process where PONGs arrive. Only GeneveTrafficCollector (netlink byte counters) remains as a custom collector.
 - Added --pprof-bind-address flag to the gateway.


fra98 (approved):

cheina97:
@varnastadeus can you add a podmonitor on the liqo-fabric from helm and enable/disable it through values?

cmd/gateway/main.go
```diff
@@ -191,6 +194,13 @@ func run(cmd *cobra.Command, _ []string) error {
 		if err = connr.SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to setup connections reconciler: %w", err)
 		}
+
+		// Register peer latency and connection metrics.
+		metrics.Registry.MustRegister(
+			tunnel.MetricsPeerLatency,
+			tunnel.MetricsPeerIsConnected,
+			tunnel.MetricsPeerLatencyHistogram,
+		)
```
cheina97, line 203: CAn you wrap it ina. function or call them in an init function the metrics package?

pkg/conncheck/receiver.go
```diff
@@ -35,7 +35,8 @@ type Peer struct {
 	lastPingTimestamp time.Time
 	// lastPongTimestamp is the time when the last PONG was received.
 	lastPongTimestamp time.Time
-	updateCallback    UpdateFunc
+	// observer is called on PONG and disconnect for this peer.
+	observer Observer
```
cheina97: CAn you explain better the role of the observer in the comment?

cmd/gateway/main.go
```diff
@@ -151,6 +153,7 @@ func run(cmd *cobra.Command, _ []string) error {
 			BindAddress: connoptions.GwOptions.MetricsAddress,
 		},
 		HealthProbeBindAddress: connoptions.GwOptions.ProbeAddr,
+		PprofBindAddress:       connoptions.GwOptions.PprofAddr,
```
cheina97, line 156: Why are you adding these here and in this PR? Personally I didn't know about this endpoint in controller runtime.
We can add it in a separate PR to every liqo components and add a value in the helm chart to configure it

cmd/gateway/main.go
varnastadeus: I have put registration explicitly as now metrics.go holds both gateway and fabric metrics, having registration in `init` would be confusing as different containers expose different metrics

PR #3348, opened 2026-08-04
feat(authentication): allow the provider to resolve the ResourceSlice class when unspecified

thisiskazem:
# Description

Today the class of a `ResourceSlice` is entirely up to the consumer, and a consumer that does not care about class names has no option other than the built-in default one. This means a provider that wants its own allocation policy has to ask every consumer to spell out a class name it defines.

This PR lets the provider declare which class to use when the consumer does not request one, as suggested by @claudiolor while reviewing #3319:

- New Helm value **`authentication.defaultResourceSliceClass`** (default empty, hence no behavior change), wired to a new controller-manager flag `--default-resource-slice-class`.
- The provider now reports the class every `ResourceSlice` is handled with in the new **`status.class`** field: the class requested by the consumer, or, when none is requested, the configured default (the built-in `default` class if nothing is configured). The consumer's `spec.class` is left untouched, so the spec keeps describing what the consumer asked for while the provider decides how it is served. As discussed on #3319, keeping the resolution out of the spec also avoids fighting the crd-replicator, which owns the spec in the consumer → provider direction.
- Class controllers, including the built-in one, select the `ResourceSlice`s they are responsible for through `status.class` alone.
- The class is resolved **once**. Configuring or changing the default therefore only affects the `ResourceSlice`s created afterwards, which is the "future-only" semantics agreed with @everpeace and @frisso.
- **`spec.class` is now immutable**, as requested by @claudiolor: changing the class of an existing `ResourceSlice` is rejected rather than silently ignored.
- **liqoctl** no longer defaults `--class` / `--resource-slice-class` to `"default"`, so that the provider-side resolution is actually reachable. It also preserves a class already set on a `ResourceSlice` when none is requested, so that re-running `liqoctl create resourceslice` to adjust the resources keeps working under the new immutability rule.

Composes with `defaultResourceSliceClassEnabled` from #3319: a `ResourceSlice` resolved to a custom class is no longer a default-class request, so it is routed to that class' controller instead of being denied. This gives consumers a working path even while the lenient default class is disabled.

Docs updated in `usage/resource-reservation.md` (a new *Choosing the class on the provider* section) and in `advanced/peering/offloading-in-depth.md` (the selection contract for custom class controllers, and what happens to a request for a class no controller handles).

Note this changes the `ResourceSlice` CRD: it adds the optional `status.class` field, and a validation rule that makes `spec.class` immutable, so updates that previously succeeded on that field are now rejected.

Follow-up to #3319. Part of #3306.

cc @claudiolor @frisso @everpeace

## Two notes for reviewers

**The immutability rule is intentionally defined on the spec, not on the `class` field.** `Class` is optional, and CRD validation rules on an optional field are not evaluated when the field is unset, so a field-level `self == oldSelf` would still allow setting a class on a `ResourceSlice` that had none. The spec-level rule covers that case too, and it is verified by one of the tests below.

**`ResourceSlice`s that predate this change keep the built-in default class.** A `ResourceSlice` created before this version has no `status.class`. If the provider is upgraded and `defaultResourceSliceClass` is configured in the same step, resolving such a slice to the new default would silently move an already-accepted slice away from the controller that accepted it. So when the `Resources` condition is already set, an empty class resolves to the built-in `default`. This scenario was caught while testing an earlier revision on real clusters, and it is covered by a unit test.

# How Has This Been Tested?

Unit tests were added for the resolution rules (`classresolution_test.go`) and for the liqoctl class preservation (`forge/resourceslice_test.go`).

End to end on **two k3s clusters** (consumer + provider) running this branch, with the provider configured through the Helm value.

- [x] **Not configured:** an empty class is still accepted by the built-in controller, and reported as `default` in `status.class`.
- [x] **Enabling the default:** `ResourceSlice`s that already existed keep their class.
- [x] **Remap:** a new `ResourceSlice` with no class gets `status.class` set to the configured class, and is left to that class' controller.
- [x] **Future-only:** after changing the configured default, existing `ResourceSlice`s keep their class and only the new ones get the new value, verified across three successive values.
- [x] **Explicit class:** a class requested by the consumer is reported as is in `status.class`, and never overridden.
- [x] **Pre-existing slices:** a `ResourceSlice` whose resources were handled before any class was resolved resolves to `default`, even when a custom default is configured.
- [x] **Immutability:** changing, unsetting, and adding `spec.class` on an existing `ResourceSlice` are all rejected.
- [x] **liqoctl re-run:** re-running `liqoctl create resourceslice` without `--class` preserves the class and updates the resources.
- [x] **Status propagation:** the resolved class is visible on the consumer, replicated with no change to the crd-replicator.
- [x] **Composition with #3319:** with the default class disabled, `ResourceSlice`s resolved to `default` are still denied, while those resolved to a custom class are routed to their class controller.

### DIFF
--- a/apis/authentication/v1beta1/resourceslice_types.go
+++ b/apis/authentication/v1beta1/resourceslice_types.go
@@ -45,6 +45,9 @@ const (
 )
 
 // ResourceSliceSpec defines the desired state of ResourceSlice.
+// +kubebuilder:validation:XValidation:rule="has(self.class) == has(oldSelf.class) && (!has(self.class) || self.class == oldSelf.class)",message="Class is immutable"
+//
+//nolint:lll // ignore long lines given by Kubebuilder marker annotations
 type ResourceSliceSpec struct {
 	// ConsumerClusterID is the id of the consumer cluster.
 	ConsumerClusterID *liqov1beta1.ClusterID `json:"consumerClusterID,omitempty"`
@@ -52,10 +55,13 @@ type ResourceSliceSpec struct {
 	ProviderClusterID *liqov1beta1.ClusterID `json:"providerClusterID,omitempty"`
 	// Resources contains the slice of resources requested.
 	Resources corev1.ResourceList `json:"resources,omitempty"`
-	// Class contains the class of the ResourceSlice.
+	// Class contains the class of the ResourceSlice. It is immutable once the ResourceSlice is created.
 	Class ResourceSliceClass `json:"class,omitempty"`
 	// CSR is the Certificate Signing Request of the consumer cluster.
 	CSR []byte `json:"csr,omitempty"`
+
+	// NOTE: the immutability rule for Class is enforced at the spec level, and not on the field itself,
+	// as validation rules defined on optional fields are not evaluated when the field is unset.
 }
 
 // ResourceSliceConditionType represents different types of conditions that a ResourceSlice could assume.
@@ -99,6 +105,11 @@ type ResourceSliceCondition struct {
 type ResourceSliceStatus struct {
 	// Conditions contains the conditions of the ResourceSlice.
 	Conditions []ResourceSliceCondition `json:"conditions,omitempty"`
+	// Class contains the class resolved by the provider cluster, when the consumer did not request
+	// an explicit one. It is set at most once, when the resources of the ResourceSlice are first
+	// handled, so that configuring or changing the default on the provider only affects the
+	// ResourceSlices created afterwards.
+	Class ResourceSliceClass `json:"class,omitempty"`
 	// Resources contains the slice of resources accepted.
 	Resources corev1.ResourceList `json:"resources,omitempty"`
 	// AuthParams contains the authentication parameters for the resources given by the provider cluster.
@@ -129,6 +140,17 @@ type ResourceSlice struct {
 
 	Spec   ResourceSliceSpec   `json:"spec,omitempty"`
 	Status ResourceSliceStatus `json:"status,omitempty"`
+}
+
+// EffectiveClass returns the class the ResourceSlice is handled with: the one resolved by the
+// provider cluster, if any, or the one requested by the consumer otherwise.
+// ResourceSlice class controllers must rely on this method, and not on the spec alone, to select
+// the ResourceSlices they are responsible for.
+func (rs *ResourceSlice) EffectiveClass() ResourceSliceClass {
+	if rs.Status.Class != ResourceSliceClassUnknown {
+		return rs.Status.Class
+	}
+	return rs.Spec.Class
 }
 
 // +kubebuilder:object:root=true
--- a/cmd/liqoctl/cmd/peer.go
+++ b/cmd/liqoctl/cmd/peer.go
@@ -121,7 +121,8 @@ func newPeerCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
 
 	// Authentication flags
 	cmd.Flags().BoolVar(&options.CreateResourceSlice, "create-resource-slice", true, "Create a ResourceSlice for the peering")
-	cmd.Flags().StringVar(&options.ResourceSliceClass, "resource-slice-class", "default", "The class of the ResourceSlice")
+	cmd.Flags().StringVar(&options.ResourceSliceClass, "resource-slice-class", "", "The class of the ResourceSlice. "+
+		"When left empty, the class configured on the provider cluster is used")
 	cmd.Flags().BoolVar(&options.InBand, "in-band", false, "Use in-band authentication. Use it only if required and if you know what you are doing")
 	cmd.Flags().StringVar(&options.ProxyURL, "proxy-url", "", "The URL of the proxy to use for the communication with the remote cluster")
 
--- a/docs/advanced/peering/offloading-in-depth.md
+++ b/docs/advanced/peering/offloading-in-depth.md
@@ -153,6 +153,9 @@ Once a custom class is defined in the `ResourceSlice` spec, the custom ResourceS
 The custom controller might deny the request, fully accept it, or partially accept it by providing only a portion of the requested resources.
 The `VirtualNode` in the consumer cluster and the `Quota` in the provider cluster will be created based on the resources granted by the custom controller.
 
+If the class is left unspecified, the provider cluster may assign one of its own, configured through the `authentication.defaultResourceSliceClass` Helm value, and record it in `status.class` (see [Choosing the class on the provider](/usage/resource-reservation.md#choosing-the-class-on-the-provider)).
+The class a `ResourceSlice` is handled with is therefore the one in `status.class` when set, and the one in `spec.class` otherwise: a custom controller must select the `ResourceSlice`s it is responsible for through this pair, as implemented by the `EffectiveClass()` method of the `ResourceSlice` type.
+
 ### Delete ResourceSlice
 
 You can revert the process by deleting the `ResourceSlice` in the consumer cluster.
--- a/docs/usage/liqoctl/liqoctl_create.md
+++ b/docs/usage/liqoctl/liqoctl_create.md
@@ -508,7 +508,7 @@ liqoctl create resourceslice [flags]
 ### Options
 `--class` _string_:
 
->The class of the ResourceSlice **(default "default")**
+>The class of the ResourceSlice. When left empty, the class configured on the provider cluster is used
 
 `--cpu` _string_:
 
--- a/docs/usage/resource-reservation.md
+++ b/docs/usage/resource-reservation.md
@@ -251,6 +251,27 @@ authentication:
 
 When the default class is disabled, any `ResourceSlice` that uses the `default` class is denied: its `Resources` condition is set to `Denied` with an explanatory message, instead of leaving the consumer waiting until timeout. Consumers must then request an explicit class handled by a custom controller. The default value is `true`, which preserves the previous behavior.
 
+(ResourceReservationDefaultClass)=
+
+### Choosing the class on the provider
+
+Requiring an explicit class means that every consumer has to know the class names used by the provider. To avoid that, the **provider** can declare which class to use for the `ResourceSlice`s that do not request one:
+
+```{code-block} yaml
+:caption: "values.yaml (provider)"
+authentication:
+  defaultResourceSliceClass: "my-class"
+```
+
+When a consumer creates a `ResourceSlice` without specifying a class, the provider assigns the configured one and records it in `status.class`. The consumer's `spec.class` is left untouched, so the request keeps reflecting what was asked for, while the provider decides how it is served.
+
+The class is resolved **once**, when the `ResourceSlice` is first reconciled. Changing `defaultResourceSliceClass` later therefore affects only the `ResourceSlice`s created afterwards, and never reshuffles the existing ones. Consistently, `spec.class` is immutable: attempting to change it on an existing `ResourceSlice` is rejected.
+
+```{admonition} Writing a class controller
+:class: note
+The class actually in use is the one in `status.class` when set, and `spec.class` otherwise. Custom class controllers must select the `ResourceSlice`s they are responsible for through this pair, and not through `spec.class` alone, otherwise they will not see the requests remapped by the provider. The `EffectiveClass()` method on the `ResourceSlice` type implements this.
+```
+
 (ResourceReservationSuspendReclaim)=
 
 ## Suspend and reclaim a reservation
--- a/pkg/liqo-controller-manager/authentication/remoteresourceslice-controller/remoteresourceslice_controller.go
+++ b/pkg/liqo-controller-manager/authentication/remoteresourceslice-controller/remoteresourceslice_controller.go
@@ -17,6 +17,7 @@ package remoteresourceslicecontroller
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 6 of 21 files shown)


cmd/liqoctl/cmd/peer.go
```diff
@@ -121,7 +121,8 @@ func newPeerCommand(ctx context.Context, f *factory.Factory) *cobra.Command {
 
 	// Authentication flags
 	cmd.Flags().BoolVar(&options.CreateResourceSlice, "create-resource-slice", true, "Create a ResourceSlice for the peering")
-	cmd.Flags().StringVar(&options.ResourceSliceClass, "resource-slice-class", "default", "The class of the ResourceSlice")
+	cmd.Flags().StringVar(&options.ResourceSliceClass, "resource-slice-class", "", "The class of the ResourceSlice. "+
+		"When left empty, the class configured on the provider cluster is used")
```
frisso: The "class configured on the provider cluster is used" or "default class configured on the provider cluster is used"?

docs/advanced/peering/offloading-in-depth.md
```diff
@@ -153,6 +153,9 @@ Once a custom class is defined in the `ResourceSlice` spec, the custom ResourceS
 The custom controller might deny the request, fully accept it, or partially accept it by providing only a portion of the requested resources.
 The `VirtualNode` in the consumer cluster and the `Quota` in the provider cluster will be created based on the resources granted by the custom controller.
 
+If the class is left unspecified, the provider cluster may assign one of its own, configured through the `authentication.defaultResourceSliceClass` Helm value, and record it in `status.class` (see [Choosing the class on the provider](/usage/resource-reservation.md#choosing-the-class-on-the-provider)).
+The class a `ResourceSlice` is handled with is therefore the one in `status.class` when set, and the one in `spec.class` otherwise: a custom controller must select the `ResourceSlice`s it is responsible for through this pair, as implemented by the `EffectiveClass()` method of the `ResourceSlice` type.
```
frisso: I would remove from "as implemented by...". This does not look appropriate for user-level documentation.

docs/advanced/peering/offloading-in-depth.md
```diff
@@ -153,6 +153,9 @@ Once a custom class is defined in the `ResourceSlice` spec, the custom ResourceS
 The custom controller might deny the request, fully accept it, or partially accept it by providing only a portion of the requested resources.
 The `VirtualNode` in the consumer cluster and the `Quota` in the provider cluster will be created based on the resources granted by the custom controller.
 
+If the class is left unspecified, the provider cluster may assign one of its own, configured through the `authentication.defaultResourceSliceClass` Helm value, and record it in `status.class` (see [Choosing the class on the provider](/usage/resource-reservation.md#choosing-the-class-on-the-provider)).
```
frisso:
```suggestion
If the class is left unspecified, the provider cluster may assign one of its own, configured through the `authentication.defaultResourceSliceClass` Helm value, and report it in `status.class` (see [Choosing the class on the provider](/usage/resource-reservation.md#choosing-the-class-on-the-provider)).
```

docs/advanced/peering/offloading-in-depth.md
frisso: It is not clear to me what happens if the user requests a specific class, and the provider denies it (because it does not exist).
I would document how this behavior would be handled.

docs/usage/liqoctl/liqoctl_create.md
```diff
@@ -508,7 +508,7 @@ liqoctl create resourceslice [flags]
 ### Options
 `--class` _string_:
 
->The class of the ResourceSlice **(default "default")**
+>The class of the ResourceSlice. When left empty, the class configured on the provider cluster is used
```
frisso: the class configured on the provider cluster is used
or
the default class configured on the provider cluster is used
?

docs/usage/resource-reservation.md
```diff
@@ -251,6 +251,27 @@ authentication:
 
 When the default class is disabled, any `ResourceSlice` that uses the `default` class is denied: its `Resources` condition is set to `Denied` with an explanatory message, instead of leaving the consumer waiting until timeout. Consumers must then request an explicit class handled by a custom controller. The default value is `true`, which preserves the previous behavior.
 
+(ResourceReservationDefaultClass)=
+
+### Choosing the class on the provider
+
+Requiring an explicit class means that every consumer has to know the class names used by the provider. To avoid that, the **provider** can declare which class to use for the `ResourceSlice`s that do not request one:
```
frisso:
```suggestion
Requiring an explicit class means that every consumer has to know the class names used by the provider. To avoid that, the **provider** can declare which class to use for the `ResourceSlice`s that do not request one explicitely:
```

docs/usage/resource-reservation.md
```diff
@@ -251,6 +251,27 @@ authentication:
...
+
+```{code-block} yaml
+:caption: "values.yaml (provider)"
+authentication:
+  defaultResourceSliceClass: "my-class"
+```
+
+When a consumer creates a `ResourceSlice` without specifying a class, the provider assigns the configured one and records it in `status.class`. The consumer's `spec.class` is left untouched, so the request keeps reflecting what was asked for, while the provider decides how it is served.
```
frisso:
```suggestion
When a consumer creates a `ResourceSlice` without specifying a class, the provider assigns the predefined one (configured through its Helm parameters) one and records it in `status.class`.
The consumer's `spec.class` is left untouched, so the request keeps reflecting what was asked for, while the provider decides how it is served.
```

docs/usage/resource-reservation.md
```diff
@@ -251,6 +251,27 @@ authentication:
...
+
+The class is resolved **once**, when the `ResourceSlice` is first reconciled. Changing `defaultResourceSliceClass` later therefore affects only the `ResourceSlice`s created afterwards, and never reshuffles the existing ones. Consistently, `spec.class` is immutable: attempting to change it on an existing `ResourceSlice` is rejected.
```
frisso:
```suggestion
The class is resolved **once**, when the `ResourceSlice` is first reconciled.
Changing `defaultResourceSliceClass` later in the provider's cluster configuration therefore affects only the `ResourceSlice`s created afterwards, and never reshuffles the existing ones.
Consistently, `spec.class` is immutable: attempting to change it on an existing `ResourceSlice` is rejected.
```

docs/usage/resource-reservation.md
```diff
@@ -251,6 +251,27 @@ authentication:
...
+
+```{admonition} Writing a class controller
```
frisso:
```suggestion
```{admonition} Writing a custom resource class controller
```

frisso (approved):
LGTM. Proposed minor changes in the documentation.

cmd/liqoctl/cmd/peer.go
thisiskazem: Good catch, "default class" is more precise since it maps to the authentication.defaultResourceSliceClass value.
Applied here and to liqoctl create resourceslice --class for consistency, and regenerated the CLI reference.

docs/advanced/peering/offloading-in-depth.md
thisiskazem: Agreed, that's implementation detail rather than user documentation.
Removed it here, and applied the same to the equivalent sentence in usage/resource-reservation.md.
The remaining text still states the rule a custom controller needs, that the class in use is status.class when set and spec.class otherwise, without naming the Go helper.
I'll document EffectiveClass() in the resource-slice-class-controller-template repo instead, which is where controller authors will look for it.

docs/advanced/peering/offloading-in-depth.md
thisiskazem: Applied. "Report" is also the better fit for a status field, so I used the same wording in the equivalent sentence in usage/resource-reservation.md.

docs/advanced/peering/offloading-in-depth.md
thisiskazem: Good point, that was undocumented. Added a short paragraph after this one.

To summarise the actual behaviour: Liqo keeps no registry of the available classes, so the provider cannot distinguish an unknown class from one whose controller is not running yet. A request for a class nobody handles is therefore not denied, it just stays pending: the Resources condition is left unset, no Quota and no VirtualNode are created, and liqoctl reports that the resources were not accepted. If a controller for that class is started later, the request is served then. An explicit rejection (Resources=Denied) only comes from a controller that does handle the class.

This is pre-existing behaviour, unchanged by this PR, but it is worth writing down.

docs/usage/liqoctl/liqoctl_create.md
thisiskazem: Same as your comment on peer.go, agreed and already applied.
This page is generated from the flag's help text, so I changed it in pkg/liqoctl/rest/resourceslice/create.go and regenerated with make docs.
Both liqoctl create resourceslice --class and liqoctl peer --resource-slice-class now read "the default class configured on the provider cluster is used".

docs/usage/resource-reservation.md
thisiskazem: Done, added "explicitly" (kept the spelling already used elsewhere in the page).

docs/usage/resource-reservation.md
thisiskazem: Applied, thanks.
Two small adjustments: dropped a duplicated "one" from the snippet, and kept "reports" rather than "records" to match your other comment about status reporting rather than recording.
Happy to switch it back if you prefer "records" here.

docs/usage/resource-reservation.md
thisiskazem: Applied, thanks.

docs/usage/resource-reservation.md
thisiskazem: Done.

apis/authentication/v1beta1/resourceslice_types.go
```diff
@@ -45,17 +45,23 @@ const (
...
 	// Resources contains the slice of resources requested.
 	Resources corev1.ResourceList `json:"resources,omitempty"`
-	// Class contains the class of the ResourceSlice.
+	// Class contains the class of the ResourceSlice. It is immutable once the ResourceSlice is created.
 	Class ResourceSliceClass `json:"class,omitempty"`
 	// CSR is the Certificate Signing Request of the consumer cluster.
 	CSR []byte `json:"csr,omitempty"`
+
+	// NOTE: the immutability rule for Class is enforced at the spec level, and not on the field itself,
+	// as validation rules defined on optional fields are not evaluated when the field is unset.
```
claudiolor, line 64: I think we can skip this comment
```suggestion
```

apis/authentication/v1beta1/resourceslice_types.go
```diff
@@ -99,6 +105,11 @@ type ResourceSliceCondition struct {
 type ResourceSliceStatus struct {
 	// Conditions contains the conditions of the ResourceSlice.
 	Conditions []ResourceSliceCondition `json:"conditions,omitempty"`
+	// Class contains the class resolved by the provider cluster, when the consumer did not request
+	// an explicit one. It is set at most once, when the resources of the ResourceSlice are first
+	// handled, so that configuring or changing the default on the provider only affects the
+	// ResourceSlices created afterwards.
```
claudiolor, line 111:
```suggestion
	// Class contains the class resolved by the provider cluster. When the consumer does not specify
	// a class, the provider can provide its own default.
```

pkg/liqo-controller-manager/authentication/remoteresourceslice-controller/remoteresourceslice_controller.go
```diff
@@ -242,10 +243,40 @@ func (r *RemoteResourceSliceReconciler) handleAuthenticationStatus(ctx context.C
 	return requeueIn, nil
 }
 
+// resolveClass records the class the provider assigns to a ResourceSlice that did not request an explicit
+// one. It is recorded in the status, which is owned by the provider, so that the spec keeps reflecting what
+// the consumer asked for. The class is resolved at most once, and never for the ResourceSlices whose
+// resources have already been handled, so that configuring or changing the default on the provider leaves
+// the existing ResourceSlices untouched, and only affects the ones created later on.
+func (r *RemoteResourceSliceReconciler) resolveClass(resourceSlice *authv1beta1.ResourceSlice) {
+	resourcesCondition := authentication.GetCondition(resourceSlice, authv1beta1.ResourceSliceConditionTypeResources)
```
claudiolor, line 252: I think this function can be really simple for now:
- If .Status.Class != "" -> skip
- if .Spec.Class == "" -> Set to default
- if Spec.Class != "" -> Status.Class = Spec.Class 

This allow us to always look at .Status.Class without any need to use the EffectiveClass() function.
If one day we will have a registry of the available classes, this controller might just reject the ResourceSlice.

apis/authentication/v1beta1/resourceslice_types.go
thisiskazem: Done, removed.

apis/authentication/v1beta1/resourceslice_types.go
thisiskazem: Done, thanks.

pkg/liqo-controller-manager/authentication/remoteresourceslice-controller/remoteresourceslice_controller.go
thisiskazem: Good idea, done. resolveClass() now always fills in status.class (the requested class, or the default when none is requested), and everything reads status.class directly, so EffectiveClass() is gone  and the docs describe a single rule.

I kept one extra case: if the Resources condition is already set but status.class is empty, the class resolves to the built-in default. That covers slices created before this version: if the provider is upgraded and defaultResourceSliceClass is set in the same step, an already-accepted slice would otherwise be relabelled to the new class and silently move away from the controller that accepted it. 

Happy to drop it if you think that scenario isn't worth handling.

thisiskazem:
@claudiolor thanks for the review, all your comments are addressed and the description is updated.
resolveClass() now always fills status.class, and EffectiveClass() is gone.
I re-tested it end to end on two k3s clusters. Could you take another look when you have time?

pkg/liqo-controller-manager/authentication/remoteresourceslice-controller/remoteresourceslice_controller.go
```diff
@@ -242,10 +243,41 @@ func (r *RemoteResourceSliceReconciler) handleAuthenticationStatus(ctx context.C
...
+		case resourcesCondition != nil && resourcesCondition.Status != "":
+			class = authv1beta1.ResourceSliceClassDefault
+		case r.sliceStatusOptions.DefaultResourceSliceClass != authv1beta1.ResourceSliceClassUnknown:
+			class = r.sliceStatusOptions.DefaultResourceSliceClass
+		default:
+			class = authv1beta1.ResourceSliceClassDefault
+		}
+	}
+
+	resourceSlice.Status.Class = class
```
claudiolor: Not sure why we need to look at the condition, if we eventually always assign `class = authv1beta1.ResourceSliceClassDefault`. I think we can simplify it like the following:
```suggestion
	resourceSlice.Status.Class = cmp.Or(
	    resourceSlice.Spec.Class,
	    r.sliceStatusOptions.DefaultResourceSliceClass,
	    authv1beta1.ResourceSliceClassDefault,
	)
```

thisiskazem: @claudiolor `cmp.Or` is much nicer, thanks. The condition isn't redundant though: the three branches don't all end in `default`. They only differ for a slice whose resources were already accepted while a provider default is configured.

For example, upgrading Liqo and setting `defaultResourceSliceClass=custom-x` in the same `helm upgrade`: a slice from an existing peering was already accepted by the built-in class and has no `status.class` yet. With `cmp.Or` alone it would resolve to `custom-x` and move to another controller, breaking the "only new slices are affected" behaviour we agreed on.

So I'd keep `cmp.Or` and reduce that case to a single `if` that only overrides the provider default:

```go
defaultClass := r.sliceStatusOptions.DefaultResourceSliceClass
if cond := authentication.GetCondition(resourceSlice, authv1beta1.ResourceSliceConditionTypeResources); cond != nil && cond.Status != "" {
	// Resources already handled before any class was resolved: keep the built-in default class.
	defaultClass = authv1beta1.ResourceSliceClassDefault
}
resourceSlice.Status.Class = cmp.Or(resourceSlice.Spec.Class, defaultClass, authv1beta1.ResourceSliceClassDefault)
```

This case is covered by a unit test and was verified on two k3s clusters. If you still prefer to drop it, I'll remove it.

claudiolor: Ah ok, now I see your point. Thanks!

thisiskazem:
@claudiolor Thanks! Pushed with `cmp.Or` and the guard reduced to a single `if`.

PR #3353, opened 2026-08-11
[Networking] RouteConfigurationBinding

cheina97:
## Summary

This PR introduces the `RouteConfigurationBinding` CRD and reworks route configuration application to follow the same per-target binding pattern already established by `FirewallConfigurationBinding`.

Instead of having a single `RouteConfiguration` controller per node/pod that reconciles route rules directly, a new `RouteConfigurationBinding` resource now binds a `RouteConfiguration` to a specific target entity (e.g. a fabric `Node` or a gateway `Pod`). Each target runs its own `RouteConfigurationBinding` reconciler that applies only the bindings matching its identity, and a central garbage-collector ensures orphaned bindings are removed when their target disappears.

## Motivation

- **Target-scoped reconciliation**: route rules are now applied by the entity that actually owns the network namespace (fabric node or gateway pod), scoped to its own identity.
- **Cleanup on target disappearance**: if a node or gateway pod is removed without running its normal shutdown path, the corresponding `RouteConfigurationBinding` resources would be stuck forever. A new GC controller detects missing targets and force-removes the binding finalizers.
- **Scalability**: per-host status is moved from the `RouteConfiguration` status to individual `RouteConfigurationBinding` resources, avoiding scalability issues on large clusters.

### API (`apis/networking/v1beta1`)

- Added the `RouteConfigurationBinding` CRD with:
  - `Spec.RouteConfigurationRef`: reference to the `RouteConfiguration` to apply.
  - `Spec.TargetRef`: typed object reference (`APIVersion`, `Kind`, `Name`, `Namespace`) identifying the entity responsible for applying the binding.
  - `Status.Conditions`: per-binding applied condition.
- Added `RouteBindingTargetReference`, a target reference helper type with a `String()` method.
- Deprecated the `Conditions` field in `RouteConfigurationStatus`; per-host status is now reported on `RouteConfigurationBinding` resources.
- Regenerated `zz_generated.deepcopy.go`.

### Controllers

- **`pkg/route/routeconfigurationbinding_controller.go`**: new reconciler that applies netlink routes/rules/tables from the referenced `RouteConfiguration` for bindings whose `TargetRef` matches the controller's own target.
- **`pkg/route/binding_creator_base.go`**: shared base logic used by both fabric and gateway binding creators to ensure a `RouteConfigurationBinding` exists for each target.
- **`pkg/fabric/route_binding_creator_controller.go`** and **`pkg/gateway/route_binding_creator_controller.go`**: fabric and gateway-specific creators that produce bindings for their respective targets (`Node` / `Pod`).
- **`pkg/liqo-controller-manager/networking/route-binding-gc/bindinggc_controller.go`**: new garbage-collector controller running in `liqo-controller-manager` that:
  - Periodically re-checks all `RouteConfigurationBinding` resources.
  - Watches `Node` and `Pod` deletions.
  - Force-removes the binding finalizer when the referenced target no longer exists.
- **`pkg/route/watch.go`**: added event handlers and watch sources for the binding reconciler.
- Removed the legacy `RouteConfiguration` finalizer logic from `pkg/route/finalizer.go` (replaced by per-binding finalizers).

### Main binaries

- **`cmd/fabric/main.go`**: starts the `RouteConfigurationBindingReconciler` scoped to the local fabric `Node`.
- **`cmd/gateway/main.go`**: starts the `RouteConfigurationBindingReconciler` scoped to the local gateway `Pod` and the gateway route binding creator.
- **`cmd/liqo-controller-manager/modules/networking.go`**: starts the new route binding GC controller.

### Uninstaller

- **`pkg/uninstaller/deletion.go`**: updated to clean up `RouteConfigurationBinding` resources during uninstall.

### RBAC and CRDs

- Generated CRD manifests for `RouteConfigurationBinding`.
- Updated `RouteConfiguration` CRD.
- Updated ClusterRole manifests for `liqo-controller-manager`, `liqo-fabric`, and `liqo-gateway`.
- Updated Helm templates (`liqo-gateway-rbac`, WireGuard gateway client/server templates).

### Tooling

- **`hack/rt-stats.sh`**: added helper script for runtime statistics.

## Deprecations

- `RouteConfiguration.Status.Conditions` is deprecated and no longer populated by route controllers. Consumers should read the conditions from the corresponding `RouteConfigurationBinding` resources instead.

## Testing

- Added unit tests for the new route binding reconciler and the route binding GC controller.
- Existing route tests were updated or removed to reflect the new architecture.

## Migration notes

Existing `RouteConfiguration` resources are automatically migrated:
- The legacy `routeconfiguration-controller.liqo.io/finalizer` finalizer is stripped by the new binding creators on first reconcile.
- New `RouteConfigurationBinding` resources are created for each target.


apis/networking/v1beta1/routeconfiguration_types.go
```diff
@@ -133,8 +133,12 @@ type RouteConfigurationStatusCondition struct {
 }
 
 // RouteConfigurationStatus defines the observed state of RouteConfiguration.
+// The per-host status has been moved to RouteConfigurationBinding resources to avoid
+// scalability issues on large clusters. The Conditions field is kept for API compatibility
+// but is no longer populated by the route controllers.
 type RouteConfigurationStatus struct {
 	// Conditions is the list of conditions of the RouteConfiguration.
+	// Deprecated: per-host status is now reported on RouteConfigurationBinding resources.
```
fra98, line 142: Nice one! This deprecation warning is not present on firewall equivalent

cmd/fabric/main.go
```diff
@@ -242,6 +255,49 @@ func run(cmd *cobra.Command, _ []string) error {
...
+			APIVersion: firewall.TargetAPIVersionV1,
+			Kind:       firewall.TargetKindNode,
+			Name:       options.NodeName,
+		}, true)
+	route.CleanupRouteConfigurationBindings(cleanupCtx, cl,
+		networkingv1beta1.RouteBindingTargetReference{
+			APIVersion: "v1",
+			Kind:       "Node",
+			Name:       options.NodeName,
+		})
```
fra98: Let's use constants for fields like Firewall equivalent.

cmd/fabric/main.go
```diff
@@ -242,6 +255,49 @@ func run(cmd *cobra.Command, _ []string) error {
...
+	cleanupCtx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+	firewall.CleanupFirewallConfigurationBindings(cleanupCtx, cl,
+		networkingv1beta1.TargetReference{
+			APIVersion: firewall.TargetAPIVersionV1,
+			Kind:       firewall.TargetKindNode,
+			Name:       options.NodeName,
+		}, true)
+	route.CleanupRouteConfigurationBindings(cleanupCtx, cl,
+		networkingv1beta1.RouteBindingTargetReference{
```
fra98: Since this is used in two places maybw we can declare TargetReference consts also for easier reading

apis/networking/v1beta1/routeconfigurationbinding_types.go
```diff
@@ -0,0 +1,142 @@
...
+	APIVersion string `json:"apiVersion"`
+	// Kind of the referenced target (e.g. "Pod" or "InternalNode").
+	Kind string `json:"kind"`
+	// Name of the referenced target.
+	Name string `json:"name"`
+	// Namespace of the referenced target. It must be empty for cluster-scoped targets
+	// (e.g. an InternalNode) and set for namespaced targets (e.g. a Pod).
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+}
```
fra98, line 53: NIT (totally optional): I see there is some code duplication between similar (Firewall|Route)TargetReference and its methods. I wonder if we should have common struct so we can re-use some functions which can be struct function like:
```go
func (t TargetReference) MatchesTargetRef(other TargetReference) bool {
```

fra98 (approved):

PR #3367, opened 2026-09-02
feat: allow configuring hostNetwork, dnsPolicy and nodeSelector for the virtual kubelet pod

Awadabang:
# Description

Extend the VkOptionsTemplate CRD with hostNetwork, dnsPolicy and nodeSelector so the virtual kubelet pod can be scheduled on specific nodes and use the host network namespace when required.

The VkOptionsTemplateSpec gains three fields, wired through to the forged VirtualKubelet deployment pod spec. When hostNetwork is enabled and no dnsPolicy is set, it defaults to ClusterFirstWithHostNet.

Adds the API fields (with generated deepcopy), the matching Helm chart CRD, values, templates and README documentation, and a unit test covering the forged pod spec.

# How Has This Been Tested?

- [x] go build ./pkg/vkMachinery/forge/...
- [x] go test ./pkg/vkMachinery/forge/... (new TestVirtualKubeletDeploymentHostNetworkOptions)


claudiolor:
Hey @Awadabang. thanks for working on this! The nodeSelector addition makes sense to me, as someone might want to control where Virtual Kubelet pod gets scheduled.

Not sure about the use case for hostNetwork and dnsPolicy, though. 
I think enabling hostNetwork could be potentially dangerous when multiple Virtual Kubelet instances are scheduled on the same node.
Additionally, port 10250, used by default by the VK, is already bound by the node's virtual kubelet.

Could you provide a bit more context on the scenarios that require hostNetwork? And, relatedly, what is the intended use case for exposing dnsPolicy as an option?

PR #3372, opened 2026-09-07
fix(metric-agent): set APIResourceList.APIVersion to v1

broshell:
`metric-agent`'s discovery handler sets `APIResourceList.APIVersion` to `v1beta1` (the group's version). Per Kubernetes API conventions, `APIResourceList` is a `meta.k8s.io/v1` type, so `APIVersion` must always be the literal `v1` — the group/version being described belongs in `GroupVersion`, which is already set correctly.

`kubectl`/client-go decode this leniently and never noticed, but strict discovery clients (e.g. kube-rs via k8s-openapi) validate `APIVersion` against the fixed `v1` constant and reject the response, which breaks API discovery for any tool using such a client on a cluster with Liqo installed.

The fix aligns `APIResourceList.APIVersion` with the convention, matching every other API group's discovery response.


fra98 (approved):

fra98:
@broshell thank you for the contribution!

PR #3374, opened 2026-09-08
feat(offloading): dynamically update virtualkubelet deployments

fra98:
# Dynamic VirtualKubelet Deployment Updates

## Summary

VirtualNodes now automatically re-render their virtual-kubelet deployment whenever the referenced `VkOptionsTemplate` changes — no more unpeer/repeer cycles to pick up configuration updates. The virtual-kubelet exposes readiness endpoints, so a broken virtual-kubelet never replaces a working one.

### Automatic template propagation

Updating a `VkOptionsTemplate` (e.g., a new virtual-kubelet image, changed reflector settings) now propagates to **all existing VirtualNodes** referencing it. Previously, changes only applied to newly created virtual nodes.

### New VirtualNode status fields

The resolved (effective) configuration is published in the VirtualNode status, providing full observability of what the virtual-kubelet is actually running:

| Field | Description |
|---|---|
| `status.effectiveOffloadingPatch` | The offloading patch enforced by the virtual-kubelet: the spec's patch with the labels/annotations-not-reflected merged from the `VkOptionsTemplate` |
| `status.effectiveCreateNode` | Whether the node is being created (resolved from spec or template default) |
| `status.effectiveDisableNetworkCheck` | Whether the network check is disabled (resolved from spec or template default) |

### Skip the virtual-kubelet deployment

Annotate a VirtualNode with `liqo.io/skip-vk-deployment` to prevent the controller from creating the virtual-kubelet deployment and its supporting resources (ServiceAccount, ClusterRoleBinding). Useful when the deployment is managed externally (e.g., by an operator or a GitOps tool).

The check follows the standard pattern: the annotation must be present, non-empty, and not `"false"` (case-insensitive) to take effect.

### Virtual-kubelet readiness

The virtual-kubelet now exposes `/healthz` and `/readyz` endpoints (on port 8081), and the Deployment enforces a readiness probe against them. This guarantees that a virtual-kubelet failing to start (e.g., due to invalid arguments) never replaces a working one — the rollout simply stalls, keeping the old pod running.

### ServiceAccount collision with wireguard gateways

The virtual-kubelet ServiceAccount was named after the VirtualNode (e.g., `cluster2`), which could collide with wireguard gateway ServiceAccounts (also named after the cluster ID) in the same tenant namespace. Two controllers fighting over the same ServiceAccount caused reconcile errors and unpredictable behavior.

The virtual-kubelet ServiceAccount is now renamed with a `vk-` prefix (e.g., `vk-cluster2`), eliminating the collision. The ClusterRoleBinding subjects are updated additively, so virtual-kubelets from older versions keep their permissions during the rollout.

**Migration**: any external RBAC bound to the old ServiceAccount name must be updated to `vk-<virtual-node-name>`.

### Stale virtual-kubelet deployment on template reversion

When a `VkOptionsTemplate` was updated with a bad image and then fixed back, the virtual-kubelet deployment did not recover correctly due to a race between the controller's reconcile and the Deployment controller's rollout. The controller now preserves operator-managed pod template annotations (`kubectl.kubernetes.io/restartedAt`) and the `paused` flag, ensuring that `kubectl rollout restart` and `kubectl rollout pause` work correctly on virtual-kubelet deployments.

## Upgrade notes

- **CRDs are updated first** — existing VirtualNodes are unaffected until the new controller-manager rolls out.
- The virtual-kubelet deployment is rolled (pod template changes), with the old pod kept running until the new one is Ready.
- The controller-manager now receives `--podcidr` and `--vk-options-default-template` arguments (previously on the webhook deployment).

## Breaking changes

- The virtual-kubelet ServiceAccount name changed from `<virtual-node-name>` to `vk-<virtual-node-name>`. Any external RBAC bound to the old ServiceAccount must be updated to `vk-<virtual-node-name>`.

## Testing

- Envtest: deployment forging, template propagation, effective values in status, ownership, skip annotation, legacy template cleanup, shared ServiceAccount, operator actions preservation
- Unit: forge functions (base args, effective values, offloading-patch merge, update strategy)


cmd/virtual-kubelet/root/root.go
```diff
@@ -168,15 +149,39 @@ func runRootCommand(ctx context.Context, c *Opts) error {
...
 		if err != nil {
 			klog.Errorf("Unable to get network configuration: %v", err)
 			return err
 		}
 	}
 
+	// Start the health server after the startup inputs have been retrieved and before the
+	// leader election: the server being up certifies that the startup succeeded, so that
+	// also virtual-kubelets waiting to acquire the leadership report themselves as ready.
+	setupHealthServer()
```
anjmao: Better Go practice is to avoid starting goroutine inside and allow caller to do it. This is then more clear and have better control for tests etc.
```
go setupHealthServer()
```

cmd/virtual-kubelet/root/root.go
```diff
@@ -168,15 +149,39 @@ func runRootCommand(ctx context.Context, c *Opts) error {
...
+			PodName:   c.PodName,
+			Namespace: c.TenantNamespace,
+		},
+		LeaderElectorName: c.NodeName,
+		LeaseDuration:     c.VirtualKubeletLeaseLeaseDuration,
+		RenewDeadline:     c.VirtualKubeletLeaseRenewDeadline,
+		RetryPeriod:       c.VirtualKubeletLeaseRetryPeriod,
+	}); err != nil {
+		return err
+	} else if !leader {
```
anjmao: Nit: Can be simpler without else if

```go
if !leader {
    klog.Error("This virtual-kubelet is not the leader")
	os.Exit(1)
}
```

pkg/webhooks/virtualnode/validate.go
```diff
@@ -0,0 +1,101 @@
...
+func (w *vnwhv) Handle(ctx context.Context, req admission.Request) admission.Response {
+	switch req.Operation {
+	case admissionv1.Create:
+		return w.handleCreate(ctx, &req)
+	default:
+		return admission.Allowed("")
+	}
+}
+
+func (w *vnwhv) handleCreate(ctx context.Context, req *admission.Request) admission.Response {
```
anjmao, line 63: It was great to remove virtualnode webhook. 
But do we really need validating webhook? Why can't this validation be done inside liqo controller manager which now fully controls creation and updates?

pkg/consts/annotations.go
```diff
@@ -53,4 +53,9 @@ const (
 	// ForeignClusterPermanentlyUnreachableAnnotationKey is the annotation used to signal that the foreign cluster is not reachable and it will
 	// never come up.
 	ForeignClusterPermanentlyUnreachableAnnotationKey = "liqo.io/foreign-cluster-permanently-unreachable"
+
+	// SkipVkDeploymentAnnotation is the annotation used to signal that the virtual-kubelet
+	// deployment (and its supporting resources) must not be created by the controller,
+	// as it is managed externally.
+	SkipVkDeploymentAnnotation = "liqo.io/skip-vk-deployment"
```
anjmao, line 60: What is the use case for this? Do we have any "managed externally" virtual kubelets?

fra98: This is a possibility for power users who might want to control the VK deployment directly instead of it being created automatically by Liqo.

The ResourceSlice has the same pattern (i.e., disabling automatic virtual node creation) which is widely used by builders on top of liqo to control virtual nodes directly

pkg/webhooks/virtualnode/validate.go
fra98: Yes, it can be done, but it's less robust compared to a validating webhook. 
With the controller-manager approach, the window from the check to the actual resource creation in the API server is wider and can become outdated in case of high concurrency.

Keep in mind that users can still add `liqo.io/webhook-skip` to the VirtualNode to bypass this webhook check if they prefer to handle this in autonomy: https://github.com/fra98/liqo/blob/6da49f6f8af74530399a4fa9dfabb894fbc46ea7/deployments/liqo/templates/webhooks/liqo-validating-webhook.yaml#L150-L154

pkg/consts/annotations.go
claudiolor: Do we want to add a line somewhere in the docs?

claudiolor (approved):

pkg/consts/annotations.go
fra98: Adding in new PR

